PR #81, opened 2025-09-19
added glossary document with some definitions to get us started

felipeblazing:
Just wanted to get us started on making the glossary


bwyogatama:
Thanks for making the glossary!
Quick comment is about the term GPU thread, who people might misunderstood it to be an actual GPU thread running on GPU

PR #85, opened 2025-09-26
added first basic implementation of memory reservation system and tests

felipeblazing:
Creates a MemoryReservationManager with reservations and some basic tests to verify functionality

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,7 @@ set(EXTENSION_SOURCES
   src/cpu_cache.cpp
   src/gpu_query_result.cpp
   src/config.cpp
+  src/memory/memory_reservation.cpp
 )
 add_subdirectory(src/operator)
 add_subdirectory(src/plan)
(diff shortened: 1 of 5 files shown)


felipeblazing:
Options to Consider

We have different memory spaces to think about. 

There are 1 or more NVMEs we can use for storage
There are 1 or more GPUs whose memory we care about
There is probably only 1 Host Memory tier we currently care about (lets handle numaness later)

Tier currently describes the kind of memory but doesn't tell us which nvme, or which gpu.

For the nvme's the handle is basically a file path
For GPUs it is a device_id 

We want  a way to make something like a MemoryResource that can map to each one of these, while understanding which kind of memory they are referring to. 

One option might be something like

MemoryResource{
  Tier (either gpu, host, or disk)
   std::string handle; <== if its device id conver it to string and back, kinda gross but simple
}

MemoryResource{
  Tier tier
  std::variant<string,int> handle

}

MemoryResource{
  Tier tier
  std::vector<Device> devices
}


How these look to the memory reservation manager

(Will edit this later we have a call in a few minutes so going to post so it doesn't get lost to the ether!)

PR #86, opened 2025-09-30
Add interfaces for base task, task queue, and task executor.

Yifei-yang7:
Basic interfaces for scheduler and task executor, which can be customized with concrete task types, states, scheduling policies, etc.

### DIFF
--- a/src/include/parallel/task.hpp
+++ b/src/include/parallel/task.hpp
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2025, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "duckdb/common/shared_ptr.hpp"
+#include "duckdb/common/unique_ptr.hpp"
+
+namespace sirius {
+namespace parallel {
+
+/**
+ * Interface for concrete task local states.
+ */
+class ITaskLocalState {
+public:
+  virtual ~ITaskLocalState() = default;
+};
+
+/**
+ * Interface for concrete task global states.
+ */
+class ITaskGlobalState {
+public:
+  virtual ~ITaskGlobalState() = default;
+};
+
+/**
+ * Interface for concrete executor tasks.
+ */
+class ITask {
+public:
+  ITask(duckdb::unique_ptr<ITaskLocalState> local_state, duckdb::shared_ptr<ITaskGlobalState> global_state)
+    : local_state_(std::move(local_state)), global_state_(global_state) {}
+
+  virtual ~ITask() = default;
+
+  // Non-copyable and movable.
+  ITask(const ITask&) = delete;
+  ITask& operator=(const ITask&) = delete;
+  ITask(ITask&&) = default;
+  ITask& operator=(ITask&&) = default;
+
+  // Execution function.
+  virtual void Execute() = 0;
+
+protected:
+  duckdb::unique_ptr<ITaskLocalState> local_state_;
+  duckdb::shared_ptr<ITaskGlobalState> global_state_;
+};
+
+} // namespace parallel
+} // namespace sirius
--- a/src/include/parallel/task.hpp
+++ b/src/include/parallel/task.hpp
@@ -16,8 +16,7 @@
 
 #pragma once
 
-#include "duckdb/common/shared_ptr.hpp"
-#include "duckdb/common/unique_ptr.hpp"
+#include <memory>
 
 namespace sirius {
 namespace parallel {
@@ -43,7 +42,7 @@ public:
  */
 class ITask {
 public:
-  ITask(duckdb::unique_ptr<ITaskLocalState> local_state, duckdb::shared_ptr<ITaskGlobalState> global_state)
+  ITask(std::unique_ptr<ITaskLocalState> local_state, std::shared_ptr<ITaskGlobalState> global_state)
     : local_state_(std::move(local_state)), global_state_(global_state) {}
 
   virtual ~ITask() = default;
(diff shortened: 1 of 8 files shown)


sarda-devesh:
@Yifei-yang7 Where is the thread local state (such as the threads memory resource) going to be stored? Is it in the `WorkerThread` struct or thoughts on having something like a `IThreadState`?

src/include/parallel/task.hpp
```diff
@@ -0,0 +1,66 @@
...
+ */
+class ITaskLocalState {
+public:
+  virtual ~ITaskLocalState() = default;
+};
+
+/**
+ * Interface for concrete task global states.
+ */
+class ITaskGlobalState {
```
felipeblazing, line 36: What is the global state meant to capture?

src/include/parallel/task.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include "duckdb/common/shared_ptr.hpp"
+#include "duckdb/common/unique_ptr.hpp"
+
+namespace sirius {
+namespace parallel {
+
+/**
+ * Interface for concrete task local states.
+ */
+class ITaskLocalState {
```
felipeblazing, line 28: What will we be storing in a tasks local state?

src/include/parallel/task.hpp
```diff
@@ -0,0 +1,66 @@
...
+public:
+  virtual ~ITaskGlobalState() = default;
+};
+
+/**
+ * Interface for concrete executor tasks.
+ */
+class ITask {
+public:
+  ITask(duckdb::unique_ptr<ITaskLocalState> local_state, duckdb::shared_ptr<ITaskGlobalState> global_state)
```
felipeblazing, line 46: if we want this engine to be applicable to multiple runtimes sometime in the future wouldn't it be better to rely on the std implementations?

src/include/parallel/task_executor.hpp
```diff
@@ -0,0 +1,86 @@
...
+public:
+  struct WorkerThread {
+    explicit WorkerThread(duckdb::unique_ptr<std::thread> thread)
+      : internal_thread_(std::move(thread)) {}
+
+	  duckdb::unique_ptr<std::thread> internal_thread_;
+  };
+
+  struct Config {
+    int num_threads = std::thread::hardware_concurrency();
```
felipeblazing, line 43: the number of threads for task execution should correspond to the number of cudastreams we want scheduling work against the gpu. The default should be something more like 2-5 probably.

src/include/parallel/task_executor.hpp
```diff
@@ -0,0 +1,86 @@
...
+  virtual void OnTaskError(int worker_id, duckdb::unique_ptr<ITask> task, const std::exception& e);
+
+  // Main thread loop.
+  virtual void WorkerLoop(int worker_id);
+
+private:
+  duckdb::unique_ptr<ITaskScheduler> scheduler_;
+  Config config_;
+  std::atomic<bool> running_;
+  duckdb::vector<duckdb::unique_ptr<WorkerThread>> threads_;
```
felipeblazing, line 82: instead of storing a vector of threads why not use soemthing like https://github.com/vit-vit/CTPL?

src/include/parallel/task_executor.hpp
```diff
@@ -0,0 +1,86 @@
...
+private:
+  // Helper functions.
+  virtual void OnStart();
+  virtual void OnStop();
+  virtual void OnTaskError(int worker_id, duckdb::unique_ptr<ITask> task, const std::exception& e);
+
+  // Main thread loop.
+  virtual void WorkerLoop(int worker_id);
+
+private:
```
felipeblazing, line 78: it feels like we need a queue inside of the executor to hold tasks that have been created but that  we don't currently have threads to service yet. eventually we want to be able to spill things from that queue as well. And we presumably want to have the queue be a priority queue where the priority becomes the policy mechanism for deciding what gets processed when.

src/include/parallel/task_scheduler.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "task.hpp"
+
+namespace sirius {
+namespace parallel {
+
+/**
+ * Interface for concrete scheduling policies.
+ */
+class ITaskScheduler {
```
felipeblazing, line 27: Is this scheduler acting as some kind of queue?

felipeblazing (commented):
I would love to discuss some of the questions I raised here to make sure we are all aligned on how these interfaces whould be implemented.

src/parallel/task_executor.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+void ITaskExecutor::Stop() {
+  bool expected = true;
+  if (!running_.compare_exchange_strong(expected, false)) {
+    return;
+  }
+  OnStop();
+  for (auto& thread : threads_) {
+    thread->internal_thread_->join();
```
felipeblazing, line 44: stopping can usually be forced or non forced. I think a threadpool has a cleaner api where  instead of joining threads we focus on something thats more like a futures api where you schedule work against a threadpool and get back a future rather than controlling the threads directly

src/include/parallel/task.hpp
Yifei-yang7: duckdb scan requires all tasks on the same table share a `TableScanGlobalSourceState` to make tasks won't scan same ranges. Usually when processing a pipeline in parallel with multiple tasks we will want to put shared states among tasks here.

src/include/parallel/task_executor.hpp
Yifei-yang7: I'm not quite familiar with this library. But by checking its intro it seems we cannot change behavior of its internal queue (FIFO) to support various scheduling policies. Here we have interface `ITaskScheduler` which can be derived to support this.

src/include/parallel/task_scheduler.hpp
Yifei-yang7: Yes as mentioned in the previous comment, `ITaskScheduler` is simply an interface, and the concrete queues, how scheduling is performed will be defined in its derived classes later on.

src/include/parallel/task.hpp
felipeblazing: I heard at some point we want to support something like Doris. The checks don't seem incredibly helpfpul to me particularly if you can ensure that the value is not null before hand.

src/parallel/task_executor.cpp
felipeblazing: we can change the scheduling policies by not oversbuscribing to it. at that point I do get a weird feeling about not using it so much as a thread pool and it starts to look more like a vector of threads as you proposed.

PR #91, opened 2025-10-23
Data Representation, Data Batch, and Data Repository

bwyogatama:
As per @felipeblazing's request, breaking down the large PR to smaller ones, starting from Data Batch/Data Repository.


src/include/data/data_batch_view.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include "helper/helper.hpp"
+#include "data/common.hpp"
+
+namespace sirius {
+
+using sirius::memory::Tier;
+
+class DataBatchView : public cudf::table_view {
+public:
+    DataBatchView(DataBatch* batch, std::vector<cudf::column_view> const& cols)
```
felipeblazing: I think the constructor looks more like DataBatchView(DataBatch* batch)

We should do something which is as follows

{
 acquire some kind of lock
ensure the data is currently in gpu if its not we did something wrong
increment the reference count
releasee the lock
}
make the view from the data in the databatch. now come to think of it. Our databatch could have a method which creates a DataBatchView and it can do the updating of the reference count when it creates this view. But we want the movement of this data to the gpu to have happened previously and asynchronously, if this data batch is not in gpu it should be an error

}

src/include/data/data_repository.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include "data_batch.hpp"
+#include "helper/helper.hpp"
+
+namespace sirius {
+
+class IDataRepository {
+public:
+    virtual void AddNewDataBatch(sirius::unique_ptr<DataBatchView> data_batch) = 0;
+
+    virtual sirius::unique_ptr<DataBatchView> EvictDataBatch() = 0;
```
felipeblazing: I am not sure we have defined how eviction happens in our current execution design. So for example it is possible that we know upfront every view that will be made from a databatch, we create them all, when refcount = 0 then we evict, or it might be the case that we don't know how many tasks will need access to that batch. If the later is the case then we need some way of knowing which batch to evict. and eviction should probably not return a view. If we are using status it should return a status, if we are using exceptions it should return void or throw

felipeblazing: also now that i think about it there is no such thing as eviction from the DataRepository right? The data repository has a view that will only be used once unless I am mistaken, the Data Batch will resides in the Data Repository Manager Class

src/include/data/data_repository.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#pragma once
+#include "data_batch.hpp"
+#include "helper/helper.hpp"
+
+namespace sirius {
+
+class IDataRepository {
+public:
+    virtual void AddNewDataBatch(sirius::unique_ptr<DataBatchView> data_batch) = 0;
```
felipeblazing: lets call it data_batch_view to not confuse it with an owning container

felipeblazing (changes requested):

src/include/data/data_repository.hpp
bwyogatama: What i mean by eviction is when a task pulling the data batch out of that repository

src/include/data/data_batch_view.hpp
bwyogatama: Since it's a derived class from cudf::table_view, it has to pass the constructor to the cudf::table_view right? Or are you thinking about something that is not a derived class of cudf::table_view?

src/include/data/data_repository.hpp
bwyogatama: sounds good!

src/include/data/data_repository.hpp
bwyogatama: If I understand correctly, I think you are talking about evicting from GPU memory

src/include/data/data_batch_view.hpp
felipeblazing: so do something like batch->CreateView()

felipeblazing: Ok I see where some of the confusion is coming from. So my main point is that cols comes from the data that resides inside of batch,so if batch has a cudf_table call the table->view() method and thats your view

felipeblazing (changes requested):

src/include/data/cpu_data_representation.hpp
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * @brief Convert this CPU table representation to a different memory tier
+     * 
+     * @param target_tier The target memory tier to convert to
+     * @return sirius::unique_ptr<IDataRepresentation> A new data representation in the target tier
+     */
+    sirius::unique_ptr<IDataRepresentation> ConvertToTier(Tier target_tier) override;
+
+public:
+    sirius::unique_ptr<MultipleBlocksAllocation> allocation_; ///< The allocation where the actual data resides
```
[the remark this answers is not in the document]
bwyogatama: Converted them to private

src/include/data/data_repository.hpp
```diff
@@ -0,0 +1,102 @@
...
+     * 
+     * @param num_data_batches Maximum number of candidates to return
+     * @return std::vector<uint64_t> Vector of batch IDs suitable for downgrading
+     * 
+     * @note Thread-safe operation protected by internal mutex
+     */
+    virtual std::vector<uint64_t> GetDowngradableDataBatches(size_t size_in_bytes) = 0;
+private:
+    sirius::mutex mutex_;                                      ///< Mutex for thread-safe access to repository operations
+    sirius::vector<sirius::unique_ptr<DataBatchView>> data_batches_;  ///< Map of pipeline source to DataBatchView
```
Yifei-yang7: I recall to make this composable, we probably want to use the base type `cudf::table_view` instead of `sirius::DataBatchView` since other systems will be agnostic of `sirius::DataBatchView`.

src/include/data/data_repository_manager.hpp
```diff
@@ -0,0 +1,212 @@
...
+     * Delegates to the appropriate pipeline's repository to remove a data batch
+     * according to that repository's eviction policy.
+     * 
+     * @param pipeline_id ID of the pipeline whose repository should evict a batch
+     * @return sirius::unique_ptr<DataBatchView> The evicted batch, or nullptr if repository is empty
+     * 
+     * @throws std::out_of_range If no repository exists for the specified pipeline
+     * @note Thread-safe operation
+     */
+    sirius::unique_ptr<DataBatchView> EvictDataBatchView(size_t pipeline_id) {
```
felipeblazing: I feel like no one should be calling this method on this class. You call this on the DataRepositories themselves. We pull data from the datarepositories. For removing from this method every time someone decrements the ref count we should check if the refcount = 0, if it does then we should remove this DataBatch because no one else is going to use it unless we think of scenarios where we might want to use it again. Now that I think about it it feels like we need TWO properties. A refcount on the view AND a way of saying this view is currently in an executing task or not (pinning). So it could be two refcounts I guess. One that tracks that the DataBatch needs to keep existing, and another which tracks the number of tasks currently operating on it in gpu. In order to spill a DataBatch that second refcount will need to be 0 and we need a way of locking that batch so it doesn't get scheduled while it is being spilled.

src/include/data/data_repository_manager.hpp
```diff
@@ -0,0 +1,212 @@
...
+     * This method removes a DataBatch from the holder and destroy it. It should only be called
+     * when the reference counter is zero (i.e., no DataBatchViews are referencing it).
+     * 
+     * @param batch_id The unique ID of the batch to destroy
+     * 
+     * @throws std::out_of_range If no batch exists with the given ID
+     * @note Thread-safe operation
+     * @warning Caller must ensure the reference count is zero before calling this method
+     */
+    void DestroyDataBatch(uint64_t batch_id) {
```
felipeblazing: this should happen automatically when no views exist to this data

felipeblazing (changes requested):

src/include/data/data_repository.hpp
felipeblazing: its ok for now but yes at some point we want this to be less opinionated and have a unique ptr to some interface

felipeblazing (approved):

Yifei-yang7 (approved):

PR #92, opened 2025-10-25
Memory Spaces and Memory Reservation multi device

felipeblazing:


src/include/memory/memory_reservation.hpp
```diff
@@ -16,79 +16,413 @@
...
+    
+    /**
+     * Configuration for a single MemorySpace.
+     * Contains all parameters needed to create a MemorySpace instance.
+     */
+    struct MemorySpaceConfig {
+        Tier tier;
+        size_t device_id;
+        size_t memory_limit;
+        std::vector<std::unique_ptr<rmm::mr::device_memory_resource>> allocators;
```
bwyogatama: When do we use the allocator in Memory Reservation Manager? I'm guessing we use the allocator in Memory Space to move data arounds?

felipeblazing: Just some stuff I have to clean up

felipeblazing: ok sorry I misread where this was. This is the allocators that are being set up for the MemorySpace.

PR #96, opened 2025-11-03
Pipeline Task, Pipeline Queue, and Pipeline Executor

bwyogatama:


src/include/pipeline/gpu_pipeline_task.hpp
```diff
@@ -0,0 +1,137 @@
...
+    public:
+        /**
+         * @brief Construct a new gpu_pipeline_task_local_state object
+         * 
+         * @param task_id Unique identifier for this task
+         * @param pipeline_id Identifier for the pipeline this task belongs to
+         * @param pipeline Shared pointer to the GPU pipeline to execute
+         * @param batch_views Vector of data batch views serving as input to the pipeline
+         */
+        explicit gpu_pipeline_task_local_state(uint64_t task_id, uint64_t pipeline_id, duckdb::shared_ptr<duckdb::GPUPipeline> pipeline,
```
felipeblazing: Is this local state where we put the reservation and the allocator?

src/include/pipeline/gpu_pipeline_queue.hpp
```diff
@@ -0,0 +1,101 @@
...
+     */
+    void close() override;
+
+    /**
+     * @brief Push a new task to be scheduled.
+     * 
+     * @param task The task to be scheduled
+     * @throws sirius::runtime_error If the scheduler is not currently accepting requests
+     */
+    void push(sirius::unique_ptr<itask> task) override;
```
felipeblazing: why is there a push and enque task?

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -0,0 +1,55 @@
...
+uint64_t gpu_pipeline_task::get_task_id() const {
+    return _task_id;
+}
+
+const duckdb::GPUPipeline* gpu_pipeline_task::get_pipeline() const {
+    return _global_state->cast<gpu_pipeline_task_local_state>()._pipeline.get();
+}
+
+void gpu_pipeline_task::execute() {
+   // Execute the task
```
felipeblazing: is this where we are going to matertialize the data into gpu memory?

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -0,0 +1,96 @@
...
+    : itask_executor(sirius::make_unique<gpu_pipeline_task_queue>(), config) {}
+
+void gpu_pipeline_executor::schedule_gpu_task(sirius::unique_ptr<gpu_pipeline_task> gpu_task) {
+    // Convert to itask and use parent's schedule method
+    schedule(std::move(gpu_task));
+}
+
+void gpu_pipeline_executor::schedule(sirius::unique_ptr<itask> task) {
+     // GPU-specific scheduling logic
+     auto gpu_task = cast_to_gpu_pipeline_task(task.get());
```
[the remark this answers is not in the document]
bwyogatama: I will fix this

src/pipeline/gpu_pipeline_task.cpp
bwyogatama: that's the plan yes

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -0,0 +1,96 @@
...
+     _threads.clear();
+ }
+ 
+ void gpu_pipeline_executor::worker_loop(int worker_id) {
+     while (true) {
+         if (!_running.load()) {
+             // Executor is stopped.
+             break;
+         }
+         auto task = _task_queue->pull();
```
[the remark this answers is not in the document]
bwyogatama: Yes the `task_queue` use a counting semaphore, and it should block until there is something on the queue (semaphore is incremented when tasks are added into the queue).
Technically, it will return `nullptr` when we call `Stop` when shutting down the database. In that scenario, we will increase the counting semaphore when there is nothing on the queue, which will return nullptr in the `WorkerLoop`

src/include/pipeline/gpu_pipeline_task.hpp
```diff
@@ -0,0 +1,153 @@
...
+public:
+    /**
+     * @brief Construct a new gpu_pipeline_task_local_state object
+     * 
+     * @param task_id Unique identifier for this task
+     * @param batch_views Vector of data batch views serving as input to the pipeline
+     */
+    explicit gpu_pipeline_task_local_state(uint64_t task_id, 
+        sirius::vector<sirius::unique_ptr<data_batch_view>> batch_views,
+        sirius::unique_ptr<sirius::memory::reservation> res = nullptr) :
```
felipeblazing: I don't think  the   reservation should  be allowed to be nullptr

src/include/creator/task_completion.hpp
```diff
@@ -0,0 +1,108 @@
...
+        return dequeue_message();
+    }
+
+private:
+    sirius::mutex _mutex;                                                        ///< Mutex for thread-safe queue access
+    sirius::queue<sirius::unique_ptr<task_completion_message>> _message_queue;  ///< Underlying message queue
+    std::counting_semaphore<> _sem{0};                                   ///< Semaphore for blocking/signaling (starts with 0 permits)
+};
+
+} // namespace sirius
```
aminaramoon, line 108: I see formatting issues here! what do we need semaphore and muex, also I think gcc's semaphore implementaton has a bug that causes deadlock. I had to reimplement it for theseus. 

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100806

it looks like it is fixed, but I remeber we had a test that failed forced to reimplement it. but since ou have a mutex, ou can probably avoid another sync primitive here

bwyogatama: Maybe I'm missing something here but how do we implement `enqueue` and `dequeue` without semaphore and mutex?

The semaphore is incremented when an item is being pushed into the queue. So when another thread calls `dequeue`, it will block until the counting semaphore is incremented.

The mutex is to ensure that the pushing/pulling an item into the queue is in the critical section.

bwyogatama: Resolved with moodycamel queue

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -0,0 +1,129 @@
...
+    _pipeline_exec->submit_task_request(std::move(request));
+}
+
+void gpu_pipeline_executor::manager_loop() {
+    while (true) {
+        if (!_running.load()) {
+            // Executor is stopped.
+            break;
+        }
+        auto task = _local_task_buffer->consume();
```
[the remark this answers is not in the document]
bwyogatama: Any suggestion on how to implement this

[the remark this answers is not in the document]
aminaramoon: in general, in asynchronous programming, the shutdown is managed by the highest level code starting a cascading call of shutdown methods on the lower level APIs. 
here, whoever calls shutdown and sets _running to false, must also call `shutdown` method of local_task_buffer, the local task buffer then listens to that signal, and returns to avoid deadlock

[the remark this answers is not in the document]
bwyogatama: Thanks, I got it down already

PR #105, opened 2025-11-12
Scan Task, Scan Queue, Scan Executor

kevkrist:
This PR implements the scan executor, scan queue, and scan task classes. A few things are left as future work in the `duckdb_scan_task::execute()` method:
1. Pushing data to the data repository.
2. Sending a message to the task creator of task completion.
3. If a DataChunk overflows the task's local data buffers (because of VARCHAR columns), the DataChunk should be pushed to the next task that is scheduled, with information about how much memory is needed. Currently, an exception is thrown. I will spawn a PR for this once this PR is approved.


src/include/operator/scan/physical_table_scan_adapter.hpp
```diff
@@ -0,0 +1,64 @@
...
+   * @brief Construct a new physical_table_scan_adapter object (has the same signature as DuckDB's
+   * PhysicalTableScan).
+   *
+   * @param[in] physical_table_scan The DuckDB PhysicalTableScan operator to be wrapped.
+   */
+  physical_table_scan_adapter(std::unique_ptr<PhysicalTableScan> physical_table_scan)
+    : GPUPhysicalOperator(PhysicalOperatorType::TABLE_SCAN,
+                          physical_table_scan->types,
+                          physical_table_scan->estimated_cardinality),
+      physical_table_scan(std::move(physical_table_scan)) {};
```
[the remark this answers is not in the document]
kevkrist: It is already correct and clear.

src/include/operator/scan/duckdb_scan_task.hpp
```diff
@@ -0,0 +1,550 @@
...
+   *
+   * @param[in,out] l_state The local state of the scan task.
+   * @return true if the chunk fits, false otherwise.
+   */
+  static bool chunk_fits(duckdb_scan_task_local_state& l_state);
+
+  /**
+   * @brief Processes the current data chunk and copies its data into the column builders' buffers.
+   */
+  void process_chunk(duckdb_scan_task_local_state& l_state);
```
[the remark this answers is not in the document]
kevkrist: This is an error.

src/include/operator/scan/duckdb_scan_task.hpp
```diff
@@ -0,0 +1,550 @@
...
+//===----------------------------------------------------------------------===//
+// DuckDB Scan Task
+//===----------------------------------------------------------------------===//
+
+/**
+ * @brief A scan task for the scan_executor using the DuckDB table function interface.
+ *
+ * The duckdb_scan_task represents a unit of work for scanning data from a DuckDB table function.
+ * It accumulates approximately the target batch size specified in the local state before pushing
+ * the batch to the data repository and notifying the task creater. If the table scan is
```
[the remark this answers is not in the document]
kevkrist: Fixed

src/operator/scan/duckdb_scan_task.cpp
```diff
@@ -0,0 +1,468 @@
...
+  // Release reservations. Memory is deallocated automatically.
+  auto& mem_res_mgr = memory::memory_reservation_manager::get_instance();
+  if (data_reservation) { mem_res_mgr.release_reservation(std::move(data_reservation)); }
+  if (mask_reservation) { mem_res_mgr.release_reservation(std::move(mask_reservation)); }
+  if (offset_reservation) { mem_res_mgr.release_reservation(std::move(offset_reservation)); }
+}
+
+void duckdb_scan_task_local_state::column_builder::reserve_memory(size_t estimated_num_rows)
+{
+  assert(size > 0);
```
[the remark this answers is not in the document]
kevkrist: Fixed.

src/include/operator/scan/duckdb_scan_task.hpp
```diff
@@ -0,0 +1,550 @@
...
+   *
+   * @param[in,out] l_state The local state of the scan task.
+   * @param[in,out] g_state The global state of the scan task.
+   * @return true if a new chunk was fetched, false if the source is drained.
+   */
+  static bool get_next_chunk(duckdb_scan_task_local_state& l_state,
+                             duckdb_scan_task_global_state& g_state);
+
+  /**
+   * @brief Checks if the current data chunk fits in the allocated buffers of the column buiilders.
```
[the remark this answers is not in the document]
kevkrist: Fixed.

src/include/operator/scan/duckdb_scan_task.hpp
```diff
@@ -0,0 +1,550 @@
...
+   * @param[in] op The physical table scan operator being executed.
+   * @param[in] exec_ctx The duckdb execution context.
+   * @param[in] global_tf_state The duckdb table function global state.
+   */
+  void initialize_local_table_function_state(duckdb::PhysicalTableScan const& op,
+                                             duckdb::ExecutionContext& exec_ctx,
+                                             duckdb::GlobalTableFunctionState* global_tf_state);
+
+  /**
+   * @brief Initilizes the column builders. Does not reserve or allocate memory for the builders.
```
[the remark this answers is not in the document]
kevkrist: Fixed.

src/include/operator/scan/duckdb_scan_task.hpp
```diff
@@ -0,0 +1,550 @@
...
+ *
+ * This class manages the state specific to a single scan task, most importantly the memory buffers
+ * into which to accumulate data from a DuckDB table scan and the logic for processing DuckDB data
+ * chunks into those buffers.
+ *
+ */
+class duckdb_scan_task_local_state : public itask_local_state {
+  static constexpr size_t DEFAULT_APPROXIMATE_BATCH_SIZE = 2ULL * 1024 * 1024 * 1024;  ///< 2 GB
+
+  // The following VARCHAR size is selected to accomodate all CHAR/VARCHAR TPC-H columns.
```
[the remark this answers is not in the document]
kevkrist: Fixed.

src/include/operator/scan/physical_table_scan_adapter.hpp
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * @brief Get the underlying PhysicalTableScan operator.
+   *
+   * @return PhysicalTableScan& Reference to the underlying PhysicalTableScan operator.
+   */
+  PhysicalTableScan const& get_physical_table_scan() const { return *physical_table_scan; }
+
+  //===----------Fields----------===//
+  std::unique_ptr<PhysicalTableScan> physical_table_scan;
+};
```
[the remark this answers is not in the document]
kevkrist: Fixed.

src/include/operator/scan/duckdb_scan_task.hpp
```diff
@@ -0,0 +1,550 @@
...
+  size_t estimated_rows_per_batch;         ///< Estimated number of rows per batch
+  vector<column_builder> column_builders;  ///< Column builders for each column
+  vector<size_t> varchar_indices;          ///< Indices of VARCHAR columns
+
+  duckdb::DataChunk chunk;  ///< DataChunk buffer
+  size_t row_offset = 0;    ///< Current row offset in buffers
+
+  unique_ptr<duckdb::LocalTableFunctionState>
+    local_tf_state;                    ///< Local state for the table function.
+  duckdb::ExecutionContext& exec_ctx;  ///< The duckdb execution context, needed for initializings
```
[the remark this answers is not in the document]
kevkrist: Fixed.

src/operator/scan/duckdb_scan_task.cpp
```diff
@@ -0,0 +1,468 @@
...
+                                       lower_bits);
+      mask_blocks_accessor.advance();
+
+      // There may be leftover bits (the upper bits from src_byte) to propagate to the next byte
+      auto const num_upper_bits = cur_bits - num_lower_bits;
+      if (num_upper_bits > 0) {
+        // The mask for the bits in the next destination byte
+        auto const upper_mask = utils::make_mask<uint8_t>(num_upper_bits);
+        // The upper bits from the source byte to copy into the next destination byte, shifted into
+        // postion for copying into the destination byte
```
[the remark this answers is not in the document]
kevkrist: Fixed.

src/operator/scan/duckdb_scan_task.cpp
```diff
@@ -0,0 +1,468 @@
...
+  }
+}
+
+void duckdb_scan_task_local_state::initialize_builders(const duckdb::PhysicalTableScan& op)
+{
+  estimated_row_size = 0;
+  // Initialize projected types and column sizes
+  column_builders.reserve(num_columns);
+  for (size_t i = 0; i < num_columns; ++i) {
+    duckdb::LogicalType const col_type = op.returned_types[op.column_ids[i].GetPrimaryIndex()];
```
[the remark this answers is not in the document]
kevkrist: Assertion added.

src/operator/scan/duckdb_scan_executor.cpp
```diff
@@ -0,0 +1,68 @@
...
+// standard library
+#include <mutex>
+
+namespace sirius::parallel {
+
+void duckdb_scan_executor::schedule(std::unique_ptr<itask> task)
+{
+  {
+    std::unique_lock<std::mutex> lock(_finish_mutex);
+    _total_tasks.fetch_add(1, std::memory_order_release);
```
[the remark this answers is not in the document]
kevkrist: Fixed with suggestion.

src/operator/scan/duckdb_scan_task.cpp
```diff
@@ -0,0 +1,468 @@
...
+                                                        l_state.approximate_batch_size,
+                                                        l_state.default_varchar_size,
+                                                        std::move(l_state.local_tf_state));
+
+    // Create a new reference to the global state
+    auto shared_global_state =
+      std::static_pointer_cast<duckdb_scan_task_global_state>(this->_global_state);
+    auto next_task = sirius::make_unique<duckdb_scan_task>(
+      new_task_id, dr_mgr, std::move(new_local_state), shared_global_state);
+    g_state.scan_executor.schedule(std::move(next_task));
```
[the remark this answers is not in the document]
kevkrist: This is a valid question. Seeking others' input.
What happens now is the next task has its state initialized (reserving and allocating memory, etc.), even though it will be a no-op (there could be multiple). The alternative is a lock around the scheduling of subsequent tasks.

src/include/operator/scan/duckdb_scan_task.hpp
```diff
@@ -0,0 +1,550 @@
...
+  void set_source_drained() { source_drained.store(true, std::memory_order_release); }
+
+  //===----------Fields----------===//
+  std::atomic<bool> source_drained{false};  ///< Whether the table scan source is fully drained
+  uint64_t pipeline_id;                     ///< The pipeline id to which this table scan belongs
+  uint64_t max_threads;                     ///< Maximum number of threads for this scan task
+
+  unique_ptr<duckdb::GlobalTableFunctionState>
+    global_tf_state;                    ///< Global state for the table function
+  unique_ptr<duckdb::LocalTableFunctionState>
```
Yifei-yang7: This `LocalTableFunctionState` also appears in `duckdb_scan_task_local_state`, is this one a duplicate?

src/operator/scan/duckdb_scan_task.cpp
```diff
@@ -0,0 +1,470 @@
...
+    auto const full_bytes = utils::div_8(num_bits);
+    auto const tail_bits  = utils::mod_8(num_bits);
+    mask_blocks_accessor.memcpy_from(src_valid, full_bytes);
+    if (tail_bits > 0) {
+      auto const tail_mask = utils::make_mask<uint8_t>(tail_bits);
+      auto const tail      = src_valid[full_bytes] & tail_mask;
+      mask_blocks_accessor.set_current(tail);
+    }
+  } else {
+    // Byte unaligned case
```
Yifei-yang7, line 188: Just put here as a thought: so far we use a unified buffer to handle both aligned and unaligned bits, essentially if we got one unaligned byte, the unalignment can be propagated to every subsequent mask byte assignment. The potential solution is to split aligned and unaligned portion separately, such that for each duckdb::Datachunk the unalignment only happens on at most 1 byte.

That said, the current solution is still good enough, unless later on we observe performance issues on mask bytes assignment.

src/operator/scan/duckdb_scan_task.cpp
```diff
@@ -0,0 +1,470 @@
...
+
+void duckdb_scan_task_local_state::column_builder::process_mask_for_column(
+  duckdb::ValidityMask const& validity, size_t num_rows, size_t row_offset)
+{
+  auto const* src_valid = reinterpret_cast<uint8_t const*>(validity.GetData());
+  auto const cur_bit    = utils::mod_8(row_offset);  //< bit offset in current byte
+  auto const num_bits   = num_rows;
+
+  if (src_valid == nullptr) {
+    // All valid case
```
Yifei-yang7, line 130: Will we just produce a null mask if we finally figure out `src_valid` is null for all duckdb::DataChunk in this batch? Or we still assign full bytes for the entire batch?

src/operator/scan/duckdb_scan_task.cpp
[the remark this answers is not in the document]
Yifei-yang7: I feel even with lock guarding the scheduling, it can still not prevent next tasks with no-op, because it can happen that after we atomically "check (know not drained) + schedule", before this new task is put to execute, some other tasks finish the last part of scan and drain the source.

src/include/operator/scan/duckdb_scan_task.hpp
```diff
@@ -0,0 +1,550 @@
...
+ * This class manages the state specific to a single scan task, most importantly the memory buffers
+ * into which to accumulate data from a DuckDB table scan and the logic for processing DuckDB data
+ * chunks into those buffers.
+ *
+ */
+class duckdb_scan_task_local_state : public itask_local_state {
+  static constexpr size_t DEFAULT_APPROXIMATE_BATCH_SIZE = 2ULL * 1024 * 1024 * 1024;  ///< 2 GB
+
+  // The following VARCHAR size is selected to accommodate all CHAR/VARCHAR TPC-H columns.
+  static constexpr size_t DEFAULT_VARCHAR_SIZE = 256ULL;  ///< 256 bytes
```
Yifei-yang7: Consider making these as configurable parameters like the ones in `config.hpp`.

Yifei-yang7:
Thanks @kevkrist ! This is really a large complex component and requires a lot of effort. I left a few small comments for potential discussions.

src/operator/scan/duckdb_scan_task.cpp
kevkrist: I don't believe that one nullptr null mask means that the entire column is NOT NULL, which I know is not what you're suggesting. What we could do is add a flag to indicate that all null masks were nullptrs. And then, when we migrate the data batch (future work), we know that we don't need to materialize the null mask for cudf. Is that what you had in mind?

src/operator/scan/duckdb_scan_task.cpp
kevkrist: Yes, I definitely thought of this.

I can see 2 alternatives.
Alternative 1: _every_ column would need 2 buffers for each of data, mask, offset (for VARCHAR) so that we can keep column data aligned. Then we concatenate at the end. We wouldn't know how to size them in advance, however, and there would be lots of wasted memory.

Alternative 2: we could store the row id with the mask-aligned and mask-unaligned data and interleave them together at the end. This still increases the memory requirement though (2 mask buffers per column) and adds the overhead of interleaving the data together at the end. 

Lastly, since we aren't pushing filters into the table scan, in the test cases at least all but the last data chunk are vector-sized and thus aligned, so I think this is a rare situation.

Let me know if you disagree or see other solutions.

src/include/operator/scan/duckdb_scan_task.hpp
kevkrist: Yes this is a duplicate. Thanks! Removed.

src/include/operator/scan/duckdb_scan_task.hpp
kevkrist: Done.

src/operator/scan/duckdb_scan_task.cpp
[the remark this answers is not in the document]
kevkrist: Yeah I'm inclined to agree, so I'll leave it as-is.

src/operator/scan/duckdb_scan_task.cpp
Yifei-yang7: yes!

src/operator/scan/duckdb_scan_task.cpp
Yifei-yang7: I guess the Alternative 1 is the easy way to go. For sizing them, since the major of the data would be aligned (for each duckdb::Datachunk there will be at most 7 unaligned rows), we can still use the existing sizing approach for the aligned portion; for the unaligned portion, the size can be much smaller (e.g., 1/100 size of the aligned portion, which will be enough if each duckdb DataChunk has >700 rows)

src/operator/scan/duckdb_scan_task.cpp
kevkrist: Ok I have added a flag in the `column_builder` that gets set whenever a non-`nullptr` validity mask is encountered for a data chunk and I added some unit tests.

src/operator/scan/duckdb_scan_task.cpp
kevkrist: I am actually trying to argue that we should leave as-is, since things are very likely only to become unaligned for the last chunk of a table, so the unaligned-ness won't get propagated to the processing of subsequent chunks most of the time.

Yifei-yang7: I see. So I also agree we can leave it as now, the reason is that it probably won't have performance issue compared to the string format conversion.
But for the number of unaligned data chunks, I just checked on clickbench and there are 716 unaligned chunks in the table scan. The reason is that in duckdb the data is likely to store in the row group granularity, and every last scanned chunk of a row group can be unaligned. And the first time when finishing scan a row group it can trigger and then propagate unaligned-ness.

kevkrist: Ah I see that makes sense. If it becomes a problem, we can address it later. For now, I'd rather not prematurely optimize.

Yifei-yang7 (approved):

src/operator/scan/duckdb_scan_task.cpp
```diff
@@ -0,0 +1,469 @@
...
+}
+
+void duckdb_scan_task_local_state::column_builder::reserve_memory(size_t estimated_num_rows)
+{
+  assert(estimated_num_rows > 0);
+
+  auto& mem_res_mgr = memory::memory_reservation_manager::get_instance();
+  data_reservation  = mem_res_mgr.request_reservation(res_request, type_size * estimated_num_rows);
+  mask_reservation =
+    mem_res_mgr.request_reservation(res_request, utils::ceil_div_8(estimated_num_rows));
```
felipeblazing: you would normally make one reservation here for everything instead of a reservation for each part. So add up all the sizes and make one reservation. With reservations you want the whole thing to fail or all of it to succeed. Doing it piecemeal ytou end up in situations where some of them work others don't and now you can have several tasks with partially fulfilled reservations.

src/operator/scan/duckdb_scan_task.cpp
```diff
@@ -0,0 +1,469 @@
...
+      mem_res_mgr.request_reservation(res_request, sizeof(int64_t) * (estimated_num_rows + 1));
+  }
+}
+
+void duckdb_scan_task_local_state::column_builder::allocate_memory()
+{
+  assert(data_reservation);
+  assert(mask_reservation);
+
+  data_blocks_accessor.initialize(sirius::make_unique<multiple_blocks_allocation>(
```
felipeblazing: This is not how you want to use the multiple bllocks allocation. With this approach you have wasted space for each column. Instead you want to make one large allocation and lets say you have fixed size buffers of 1MB and your first column is a 1.5MB buffer and your second one is a 1.25MB buffer
In the code below this would translate to having to allocate 4 chunks when actually 3 would have sufficed.

So you write out 1MB of the first buffer in the first chunk, .5 MB of the first buffer and .5MB of the second buffer into the second chunk, and then the remaining .75MB into the third chunk with only .25MB of wasted space. Your limit on wasted space in this case is 1 chunk whereas in the other its 1 chunk per allocate_multiple_blocks invocation. So similar to reservations. figure out how much you need at once and allocate it all at once.

src/operator/scan/duckdb_scan_task.cpp
```diff
@@ -0,0 +1,469 @@
...
+  }
+}
+
+// This method should usually be called only on variable-length data
+bool duckdb_scan_task_local_state::column_builder::sufficient_space_for_column(
+  duckdb::Vector& vec, duckdb::ValidityMask const& validity, size_t num_rows)
+{
+  size_t data_bytes = 0;
+  if (type.InternalType() == duckdb::PhysicalType::VARCHAR) {
+    auto const* str_data = reinterpret_cast<duckdb::string_t const*>(vec.GetData());
```
felipeblazing, line 108: shouldn't we be checking the validity mask before adding the size? If we aren't chekcing that max isn't there an api that can get this size without having to do a full pass over the data? There might not be I am asking

felipeblazing (changes requested):

src/operator/scan/duckdb_scan_task.cpp
kevkrist: EDIT: I see. My question would then be how are the data that belong to a column tracked? 

I suppose one solution would be to keep a vector of (block_idx, offset, length) triples for each column? Or is there a better alternative?

felipeblazing: correct the allocation needs to go out of scope after or with the columns. And to be clear if you have 3 buffers that fit into one chunk they should all be in the same chunk. You can look at https://arxiv.org/html/2508.05029v1#S3.SS4

src/operator/scan/duckdb_scan_task.cpp
kevkrist: It seems that there are some indirect ways to get the size of the string data allocated on the heap (according to Claude). The issue is that you still need to go through each entry to see if the string data is inlined (<= 12 B). In that case the characters aren't on the heap, but are directly in the `string_t` representation. Data format conversion is such a pain and makes some of this code painful I think inherently.

Yes you are definitely right about checking the validity mask I just forgot it here -- I even put it in the function signature! Fixed.

src/operator/scan/duckdb_scan_task.cpp
kevkrist: So then each column would need to maintain something like a vector of (block_idx, offset, length) triples? Because each column's data would be dispersed in ~vector-sized chunks across the multiple blocks allocation. So for 2 columns A, B, you would have A [vecsize] | B[vecsize] | A[vecsize] | B[vecsize] ... across the allocation.

src/operator/scan/duckdb_scan_task.cpp
kevkrist: Fixed.

src/operator/scan/duckdb_scan_task.cpp
kevkrist: This was my misunderstanding. There is a way to do this without data fragmentation across the blocks.
Fixed.

PR #107, opened 2025-11-20
Add Pixi manifest

mbrobbel:
I want to propose to add a [Pixi](https://pixi.sh) manifest with an environment to build and test Sirius.
The goal is to simplify setting up a reproducible development environment that is also usable for CI.

To test this locally:
```
curl -fsSL https://pixi.sh/install.sh | sh
git clone --branch pixi --recurse-submodules https://github.com/mbrobbel/sirius.git
cd sirius
pixi shell
make
```

I added `sccache` to this environment (it gets picked up by DuckDB via https://github.com/duckdb/duckdb/blob/7c039464e452ddc3330e2691d3fa6d305521d09b/CMakeLists.txt#L44-L47 - note we should handle https://github.com/duckdb/duckdb/pull/19326 when we update DuckDB i.e. add CUDA) to help speed up rebuilds locally and in CI.

A few notes about the other changes:

- Adds an example workflow (in `.github/workflows/check.yml`) that uses the Pixi environment to build (`linux-64` and `linux-aarch64`) using free public runners. It also uses `sccache` to reduce build times.
- Also adds a `.github/dependabot.yml` to update the used actions.
- Moved the `substrait` extension out of the `duckdb` submodule (it's now a normal submodule). This avoids (the removed) `git clone` in the `duckdb` submodule.
- Explicitly list the supported CUDA archs for building, because `native` doesn't work when the build system has no GPU.
- Set `EXTENSION_VERSION` to `dev` to avoid `sccache` cache misses due to the added `-DEXT_VERSION_SIRIUS=<git hash>` definition (if not set).
- Fixed a typo in the spdlog cmake file.
- Added `setup_test_datasets.sh` with the steps required to get test data sources. This script is used as activation for the environment so users don't have to manually run it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# SCM syntax highlighting & preventing 3-way merges
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,27 @@
+name: Check
+
+on:
+  pull_request:
+
+env:
+  SCCACHE_GHA_ENABLED: true
+
+jobs:
+  build:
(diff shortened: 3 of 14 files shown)


bwyogatama:
This is awesome! Thanks so much Matthijs!

PR #108, opened 2025-11-21
Read parquet using DuckDB parquet reader

ducndh:
Fixing metadata mismatch when calling table function with GPU to support read_parquet from inside gpu_processing.

DuckDB only loads the necessary columns from the parquet file when doing a read_parquet. Therefore, a different column projection is treated as a different table in the DuckDB metadata. As a result, we have to use the file path + column names necessary for that projection for the table cache.


bwyogatama:
This is great! Thanks for doing this. 
1. Can you add scripts or instructions on how to generate the parquet file so that people can try right away?
2. Can you add some basic test suites for the Parquet support? Here is some example: https://github.com/sirius-db/sirius/blob/dev/test/sql/tpch-sirius.test

ducndh:
I just added the test files and the script to generate Parquet file that is consistent with the way we generated the TPCH dataset. Running "make test" matches all the solutions provided in the tpch-dbgen.

bwyogatama:
Hi Duke, I would like to merge this PR, can we rebase to the dev branch before merging?

src/gpu_buffer_manager.cpp
```diff
@@ -423,36 +430,45 @@ GPUBufferManager::createTableAndColumnInGPU(Catalog& catalog, ClientContext& con
...
-            }
+			}
 		}
 	}
-    
-    //finding column_name in column_names
-    //convert column_name to uppercase
+
+    // Normalize names
+    string up_table_name = table_name;
```
felipeblazing, line 439: is there a reason why this case handling isn't done before it gets to the Sirius part of the code? Duckdb treats these as all case insensitive  can we rely on duckdb handling this?

src/operator/gpu_physical_table_scan.cpp
```diff
@@ -73,8 +73,76 @@ GPUPhysicalTableScan::GPUPhysicalTableScan(vector<LogicalType> types, TableFunct
...
+    }
+
+    // Add projection signature to differentiate column sets
+    // Extract ONLY the projected column names (not all table columns)
+    vector<string> projected_names;
+    for (const auto& col_id : column_ids) {
+        projected_names.push_back(names[col_id.GetPrimaryIndex()]);
+    }
+    // Sort for consistent hashing across different query orders
+    sort(projected_names.begin(), projected_names.end());
```
felipeblazing, line 105: if the columns aren't in the same order this isn't really the same. What is the situation this is trying to handle?

felipeblazing (changes requested):

Issue #55, opened 2025-07-07
sum on NULL columns should return NULL


bwyogatama:
The following returns incorrect results:
```
CREATE TABLE S (C BIGINT, D BIGINT);
INSERT INTO S VALUES (NULL, 1), (NULL, 2), (NULL, 3), (NULL, 4), (NULL, 5);
call gpu_processing("select sum(C) from S");
```
It returns 0 instead of NULL

bwyogatama:
Fixed by #109

PR #109, opened 2025-11-23
Fix sum on NULL columns should return NULL

BarryWu0812:
Fix sum on NULL columns should return NULL #55

Test result:

image


bwyogatama:
Thanks for the contributions!

PR #112, opened 2025-11-25
Topology aware memory manager initialization 

nirandaperera:
This PR adds topology-aware memory manager initialization based on the [rapidsmpf-topology discovery](https://github.com/rapidsai/rapidsmpf/blob/main/cpp/include/rapidsmpf/topology_discovery.hpp) 

- Topology discovery source files have been copied to sirius-db (at least for the moment) 
- There are several questions, 
    - how to handle `CUDA_VISIBLE_DEVICES` env var?
    - Memory spaces currently only hold rmm device allocators. This would be problematic for host allocators. So, this PR only adds GPU memory spaces 
    - what is the GPU allocation policy? Single process managing many devices? One-device-per-process?

PR is targeted to https://github.com/sirius-db/sirius/pull/97 branch

The PR might be touching many lines due to a clang-format run on the files. Maybe this needs to happen prior to this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,7 @@ set(EXTENSION_SOURCES
   src/memory/memory_reservation.cpp
   src/memory/fixed_size_host_memory_resource.cpp
   src/memory/null_device_memory_resource.cpp
+  src/memory/topology_discovery.cpp
   src/parallel/task_executor.cpp
 )
 add_subdirectory(src/operator)
--- a/pixi.lock
+++ b/pixi.lock
@@ -3,190 +3,179 @@ environments:
   default:
     channels:
     - url: https://conda.anaconda.org/rapidsai/
+    - url: https://conda.anaconda.org/nvidia/
     - url: https://conda.anaconda.org/conda-forge/
     packages:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/attr-2.5.2-h39aace5_0.conda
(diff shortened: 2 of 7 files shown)


src/memory/memory_reservation.cpp
```diff
@@ -125,272 +189,398 @@ std::unique_ptr<memory_reservation_manager> memory_reservation_manager::_instanc
...
-            // Release the wait lock before returning the reservation
-            lock.unlock();
-            return std::move(res.value());
-        }
-        // Wait until notified that memory may be available again
-        _wait_cv.wait(lock);
-    }
+    // Create allocators for this GPU
+    std::vector<std::unique_ptr<rmm::mr::device_memory_resource>> allocators;
+    allocators.push_back(std::make_unique<rmm::mr::cuda_async_memory_resource>());
```
felipeblazing: for device the allocator should be  https://github.com/sirius-db/sirius/blob/dev/src/include/memory/per_stream_tracking_resource_adaptor.hpp instead of just the cuda async memory resource

src/memory/memory_reservation.cpp
```diff
@@ -125,272 +189,398 @@ std::unique_ptr<memory_reservation_manager> memory_reservation_manager::_instanc
...
+  // TODO: How would we handle CUDA_VISIBLE_DEVICES env var?
+  const auto& topology = discovery.get_topology();
+
+  // Save and restore CUDA device context using RAII
+  scoped_device_context device_guard;
+
+  // Create a vector of memory_space_config for all GPUs
+  std::vector<memory_space_config> configs;
+  configs.reserve(topology.num_gpus);
+
```
felipeblazing: We also need to make memory spaces for the different numa nodes. Given that all memory spaces are connected to numa nodes we should probably add another value to memory space which is the numa ID of the node. The memory spaces for host should use https://github.com/sirius-db/sirius/blob/dev/src/include/memory/fixed_size_host_memory_resource.hpp as their allocator

felipeblazing (changes requested):

felipeblazing (approved):
we can merge as is and updatelater

PR #113, opened 2025-11-26
DuckDB Data Representation

sarda-devesh:
This PR introduces some of the changes needed to make to the result collector to work with new architecture: 
- Introduces a `duckdb_table_representation` to represent a table as a collection of `duckdb::DataChunk`
- Introduces a `result_table_allocation` which actually stores the raw data referenced by the `duckdb::DataChunk` spread across multiple blocks of a `fixed_size_host_memory_resource::multiple_blocks_allocation`
- Introduces a static `data_representation_converter::convert_gpu_table_to_result_format` to perform the conversion from cudf's representation to Duckdb's representation

Future work for a followup PRs:
- - Extend the Downgrade Executor to introduce a downgrade task that downgrades to the `duckdb_table_representation`. This is currently blocked on #97
- Update the task creator to create downgrade tasks for Data Batches in the query output Data Repository. This is currently blocked on #98
- Update our extension interface to pass the Data Chunks to DuckDB by reading from the Data Repository
- Extend support to other types than string/integer + support type casting

### DIFF
--- a/src/data/data_representation_converter.cpp
+++ b/src/data/data_representation_converter.cpp
@@ -1,0 +1,328 @@
+/*
+ * Copyright 2025, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#include "data/data_representation_converter.hpp"
+#include "memory/result_table.hpp"
+#include "memory/fixed_size_host_memory_resource.hpp"
+
+#include "duckdb/common/types/string_type.hpp"
+#include "duckdb/common/types.hpp"
+
+#include <cuda_runtime.h>
+#include <cudf/column/column_view.hpp>
+#include <cudf/strings/strings_column_view.hpp>
+#include <cudf/types.hpp>
+#include <cudf/utilities/bit.hpp>
+
+#include <cmath>
+#include <unordered_map>
+
+namespace sirius {
+
+using sirius::memory::fixed_size_host_memory_resource;
+
+constexpr size_t BYTE_TO_BITS = 8;
+
+// Helper map to map CUDF types to DuckDB types. TODO(Devesh): Make sure that all of the mapping make sense because not all of them are 1:1 mappings
+const std::unordered_map<cudf::type_id, duckdb::LogicalTypeId> CUDF_TO_DUCKDB_TYPES_MAP = {
+    // Signed Integers
+    {cudf::type_id::INT8,    duckdb::LogicalTypeId::TINYINT},
+    {cudf::type_id::INT16,   duckdb::LogicalTypeId::SMALLINT},
+    {cudf::type_id::INT32,   duckdb::LogicalTypeId::INTEGER},
+    {cudf::type_id::INT64,   duckdb::LogicalTypeId::BIGINT},
+
+    // Unsigned Integers
+    {cudf::type_id::UINT8,   duckdb::LogicalTypeId::UTINYINT},
+    {cudf::type_id::UINT16,  duckdb::LogicalTypeId::USMALLINT},
+    {cudf::type_id::UINT32,  duckdb::LogicalTypeId::UINTEGER},
+    {cudf::type_id::UINT64,  duckdb::LogicalTypeId::UBIGINT},
+
+    // Floating Point
+    {cudf::type_id::FLOAT32, duckdb::LogicalTypeId::FLOAT},
+    {cudf::type_id::FLOAT64, duckdb::LogicalTypeId::DOUBLE},
+
+    // Boolean & String
+    {cudf::type_id::BOOL8,   duckdb::LogicalTypeId::BOOLEAN},
+    {cudf::type_id::STRING,  duckdb::LogicalTypeId::VARCHAR},
+
+    // Timestamps
+    // cudf::TIMESTAMP_DAYS -> DuckDB DATE (days since epoch)
+    {cudf::type_id::TIMESTAMP_DAYS,         duckdb::LogicalTypeId::DATE},
+    {cudf::type_id::TIMESTAMP_SECONDS,      duckdb::LogicalTypeId::TIMESTAMP_SEC},
+    {cudf::type_id::TIMESTAMP_MILLISECONDS, duckdb::LogicalTypeId::TIMESTAMP_MS},
+    {cudf::type_id::TIMESTAMP_MICROSECONDS, duckdb::LogicalTypeId::TIMESTAMP}, // Default TIMESTAMP is us
+    {cudf::type_id::TIMESTAMP_NANOSECONDS,  duckdb::LogicalTypeId::TIMESTAMP_NS},
+};
+
+inline duckdb::LogicalTypeId cudf_type_to_duckdb_type(cudf::type_id id) {
+    auto it = CUDF_TO_DUCKDB_TYPES_MAP.find(id);
+    if (it != CUDF_TO_DUCKDB_TYPES_MAP.end()) {
+        return it->second;
+    }
+
+    return duckdb::LogicalTypeId::INVALID;
+}
+
+// Helper function used to determine the number of bytes needed to store a particular column
+size_t determine_column_bytes_needed(cudf::column_view& column_view, rmm::cuda_stream_view& stream) { 
+    size_t num_rows = column_view.size();
+
+    // First determine the bytes needed to hold the null mask
+    size_t null_mask_bytes = cudf::bitmask_allocation_size_bytes(num_rows); // By default aligns to 64 byte boundary
+
+    // Now determine the bytes needed to hold the actual data
+    size_t data_bytes = 0;
+    cudf::data_type col_type = column_view.type();
+    if(col_type.id() == cudf::type_id::STRING) { 
+        // For string get the number bytes for the offset and chars column
+        cudf::strings_column_view strings_view(column_view);
+        size_t offset_bytes = num_rows * cudf::size_of(strings_view.offsets().type());
+        size_t chars_bytes = strings_view.chars_size(stream);
+        size_t duckdb_string_bytes = num_rows * sizeof(duckdb::string_t);
+        data_bytes = offset_bytes + chars_bytes;
+    } else {
+        int32_t width = cudf::size_of(col_type);
+        data_bytes = width * num_rows;
+    }
+
+    return null_mask_bytes + data_bytes;
+}
+
+// Ensures that the location is aligned to the specificed size within the block
+inline void align_location(sirius::memory::result_table_location& location, size_t target_alignment, fixed_size_host_memory_resource::multiple_blocks_allocation& allocated_blocks) { 
+    if(location.block_offset % target_alignment != 0) { 
+        location.block_offset += target_alignment - (location.block_offset % target_alignment);
+        
+        // Deal with this shift overflowing into the next block
+        if(location.block_offset >= allocated_blocks.block_size) { 
+            location.allocation_index += 1;
+            location.block_offset = 0;
+            allocated_blocks.ensure_capacity(location.allocation_index + 1);
+        }
+    }
+} 
+
+// Helper function to copy data from GPU to CPU in a way that ensures that records are aligned to their size and no record
+// is spread across pages
+inline void copy_buffer_to_cpu(
+    const char* gpu_data_buffer, size_t num_records, size_t record_size, 
+    fixed_size_host_memory_resource::multiple_blocks_allocation& allocated_blocks,
+    sirius::memory::result_table_location& write_location,
+    rmm::cuda_stream_view& stream) { 
+    
+    size_t records_left = num_records;
+    while(records_left > 0) {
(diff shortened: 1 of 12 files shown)


src/data/data_representation_converter.cpp
```diff
@@ -0,0 +1,436 @@
...
+#include <unordered_map>
+
+namespace sirius {
+
+using sirius::memory::fixed_size_host_memory_resource;
+
+constexpr size_t BYTE_TO_BITS = 8;
+
+// Helper map to map CUDF types to DuckDB types. TODO(Devesh): Make sure that all of the mapping make sense because not all of them are 1:1 mappings
+const std::unordered_map<cudf::type_id, duckdb::LogicalTypeId> CUDF_TO_DUCKDB_TYPES_MAP = {
```
Yifei-yang7, line 41: Can you also include `DATE`, `DECIMAL` and `BOOLEAN` into this and put it to util sources like `cudf_utils` so that future development can access?

src/data/data_representation_converter.cpp
```diff
@@ -0,0 +1,436 @@
...
+        if(records_left > 0 || curr_location.block_offset >= allocated_blocks.block_size) { 
+            curr_location.allocation_index += 1;
+            curr_location.block_offset = 0;
+            allocated_blocks.ensure_capacity(curr_location.allocation_index + 1);
+        }
+    }
+}
+
+// Helper method to convert arrow style string into German style string
+void create_duckdb_strings(sirius::memory::result_table_column& results_column, fixed_size_host_memory_resource::multiple_blocks_allocation& allocated_blocks,
```
Yifei-yang7, line 173: It seems we used to use a cuda kernel `create_cpu_strings` to create duckdb strings in parallel since we need to handle both inlined and non-inlined cases. So here do we create all duckdb strings as non-inlined?

src/data/duckdb_data_representation.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "data/duckdb_data_representation.hpp"
+
+namespace sirius {
+
+duckdb_table_representation::duckdb_table_representation(sirius::unique_ptr<sirius::memory::result_table_allocation> result_table, std::vector<sirius::unique_ptr<duckdb::DataChunk>> output_vectors, sirius::memory::memory_space& memory_space)
+    : idata_representation(memory_space), _result_table(std::move(result_table)) {
+
+    // Create the output chunks to reference the data in the result table but using duckdb types
+    for(int i = 0; i < output_vectors.size(); i++) {
```
Yifei-yang7, line 25: What is the reason to create a new vector of duckdb::Datachunks (`_output_chunks`) from the input one (`output_vectors`)?

src/include/memory/fixed_size_host_memory_resource.hpp
```diff
@@ -158,6 +158,26 @@ class fixed_size_host_memory_resource : public rmm::mr::device_memory_resource {
...
+         */
+        void ensure_capacity(size_t num_blocks) { 
+            if(num_blocks <= blocks.size()) { // We already have the requested number of blocks
+                return;
+            }
+
+            int num_additional_blocks = num_blocks - blocks.size();
+            multiple_blocks_allocation additional_allocation = mr->allocate_multiple_blocks(num_additional_blocks * block_size);
+            
+            // Copy over the block pointers from that allocation to this allocation
```
Yifei-yang7, line 174: Does it make more sense to implement an `allocate_more_blocks()` for multiple_blocks_allocation directly? The function internal will transfer the ownership of the additional blocks to the one we want to expand.

src/include/memory/fixed_size_host_memory_resource.hpp
```diff
@@ -158,6 +158,26 @@ class fixed_size_host_memory_resource : public rmm::mr::device_memory_resource {
             return *this;
         }
 
+        /**
+         * @brief Ensures that the allocation has at least the specified number of blocks
+         * 
+         * Checks if the allocation has at least the specified number of blocks and if not requests more blocks
+         */
+        void ensure_capacity(size_t num_blocks) {
```
felipeblazing, line 166: When would we call ensure_capacity? Normally when allocating multiple blocks you know ahead of time how much space you are going to need. Why do you need to call this function that increases this amount? This is not typically how this memory resource is intended to be used

src/data/data_representation_converter.cpp
felipeblazing, line 41: This mapping should be handled somewhere else and not in the duckdb data representation. This understanding of how one type systems maps to another should be handled in one place so that when it needs to expand to include some other type system that becomes easier to do so.  Some of these mappings don't have valid conversions and so we need to be able to decide how we will deal with types that don't have conversions

src/data/data_representation_converter.cpp
```diff
@@ -0,0 +1,436 @@
...
+    } else {
+        int32_t width = cudf::size_of(col_type);
+        data_bytes = width * num_rows;
+    }
+
+    return null_mask_bytes + data_bytes;
+}
+
+// Ensures that the location is aligned to the specificed size within the block
+inline void align_location(sirius::memory::result_table_location& location, size_t target_alignment, fixed_size_host_memory_resource::multiple_blocks_allocation& allocated_blocks) {
```
felipeblazing, line 107: What is the purpose of this alignment? The multiple block allocations should not be aligned or have any constraints of this sort. The point is to densely pack things into them and leave alignment constraints to encodings that need to read and manipulate data. These representations are meant to hold bytes until they are ready to be converted to an encoding that is for processing.

src/data/data_representation_converter.cpp
```diff
@@ -0,0 +1,436 @@
...
+        // Deal with this shift overflowing into the next block
+        if(location.block_offset >= allocated_blocks.block_size || location.block_offset >= allocated_blocks.block_size) { 
+            location.allocation_index += 1;
+            location.block_offset = 0;
+            allocated_blocks.ensure_capacity(location.allocation_index + 1);
+        }
+    }
+} 
+
+// Helper function to copy data from GPU to CPU in a way that ensures that records are aligned to their size and no record
```
felipeblazing, line 120: records are supposed to span pages. Take a look at https://arxiv.org/abs/2508.05029 and look at the section on host encoding and fixed sized buffers

felipeblazing (changes requested):

src/data/data_representation_converter.cpp
sarda-devesh: I can definitely add in any additional types that are currently missing but I am not sure if placing it `cudf_utils` makes the most sense. Felipe brought up something similar in https://github.com/sirius-db/sirius/pull/113#discussion_r2648813902 so I will continue the discussion with some of my additional thoughts there

src/data/data_representation_converter.cpp
sarda-devesh: So, in fact the string is initialized using the `duckdb::string_t` [constructor](https://github.com/duckdb/duckdb/blob/9fffe278189182bc4538dfe99e4c885d3d45db27/src/include/duckdb/common/types/string_type.hpp#L44) which will automatically make it inlined if the length is less than 12 characters. 

The main goal of the cuda kernel was to optimize the conversion of the arrow style string representations into the german style string representation. However, that kernel assumed that we were storing all of the characters of the strings in a single contiguous buffer versus here we are distributing the strings across different block so therefore we can't generate the `duckdb::string_t` on GPU as we can't predict what address they will be stored in CPU memory

src/data/data_representation_converter.cpp
sarda-devesh: I do agree that ideally we need some sort of `type_converter` and associated `type_converter_registry` similar to what you have started implementing in https://github.com/NVIDIA/cuCascade/pull/4. 

However, what the interface for what is unclear to me because currently we don't have a concrete representation of consists of a type in Sirius? I think we do need to discuss how we represent type systems and types in a way universal way so that we can support say Doris's type system in the future but I am not sure if it makes sense in this PR because these seem orthogonal issues to me

src/data/data_representation_converter.cpp
sarda-devesh: The data that is stored in these multi block allocations are going to be utilized to initialize the `duckdb::DataChunk` that we pass to duckdb to output the result to the user which means that I need to allocate the data in the multiple block allocations such that it meets the alignment constraints of DuckDB. 

See my response in https://github.com/sirius-db/sirius/pull/113/files/95d6bb165c91e18b33d2b9cc22da2fe66c53dc09#r2648815012 for more details on why this is neccessary

src/data/data_representation_converter.cpp
sarda-devesh: We are currently using the page based memory layout on CPU (the multi byte allocations) whereas duckdb expects that all of the values for a column are stored contiguously in memory (more of a arrow style format) so we need to do these tricks so that can present a view to DuckDB as if data is stored in this arrow style contiguous format even though it is spread across different pages.

sarda-devesh: Note that this implementation will spread out column data across different pages in the multiple blocks allocation but does in an aligned way to ensure that if we are copying over a 16 byte integer then 8 bytes don't end up on one page and another 8 bytes end up on another page

src/data/duckdb_data_representation.cpp
sarda-devesh: The thing is that duckdb's expects the DataWrapper to be wrapped around `duckdb::unique_ptr` when it passed to it but wanted to avoid using duckdb::unique_ptr in the data representation as the previous code was also using `sirius::unique_ptr` in the representation convert. 

Now that we are seperating out a lot of logic into cuCascades this is not an issue and I will update the code to directly copy the vectors

src/include/memory/fixed_size_host_memory_resource.hpp
sarda-devesh: So the thing is that in this implementation we don't know how much space we are going to need in order to respect the alignment and contiguous records logic that we need to store the data in a format that can be processed by DuckDB. 

For example - Let us say that we have copied string characters up to offset 950/1000 (say a block is 1KB) and now we need to copy a 128 byte string. In this case, because we need to ensure that the characters of the string are contiguous we will need to copy the string to the next block so that the 128 characters are stored contiguously. 

However, the current logic for how many blocks we need is just based on `total data size/block size` and doesn't account for the fact that we can't just tightly pack everything (as the amount of padding we are going to need to add is data dependent) which means that we can run into a scenario where we will have to spill to using more blocks than what we anticipated

src/include/memory/fixed_size_host_memory_resource.hpp
sarda-devesh: In a sense the `allocate_more_blocks` logic is what is implemented by this function. In my head the public API should be ensure that we have X blocks/Y bytes allocated and the method takes care of that rather than the user needing to explicitly call `allocate_more_blocks`

src/include/memory/fixed_size_host_memory_resource.hpp
sarda-devesh: The other thing is that since this is only needed by the `result_table_allocation` is that instead of adding the ability to grow an already allocated `multiple_blocks_allocation`, I can update the `result_table_allocation` to contain a vector of "spill blocks" that can be used to store any additional data blocks needed

src/data/data_representation_converter.cpp
Yifei-yang7: I see, makes sense

src/include/memory/fixed_size_host_memory_resource.hpp
Yifei-yang7: I understand the logic is here. It seems to be better if we don't need an extra variable of `additional_allocation` but directly grow the existing blocks, that's why a member function of `allocate_more_blocks` can be used, and with this you can avoid the extra handling of "ensures that they don't get deallocated when additional_allocation goes out of scope"

sarda-devesh: See my comment [here](https://github.com/sirius-db/sirius/pull/113#discussion_r2652097760) - I am going to update `result_table_allocation` to store potentially multiple `multiple_blocks_allocation` in which case we don't need to implement a `allocate_more_blocks` function in the first place

Yifei-yang7:
@sarda-devesh As discussed today in the meeting, for a result batch in gpu representation, it may better to first convert to `cpu_data_representation` (fixed-size pages and pinned mem), and then copy it to a duckdb Datachunk with memory managed duckdb (continuous buffer and pageable mem).

@felipeblazing I think what we discussed is a good way to go, just wonder compared to this, would it be same if we use another separate rmm device_memory_resource on the host purely for result memory allocation? In a sense it's just a replacement of using duckdb's managed memory, but we can avoid this double copying.

bwyogatama:
> @sarda-devesh As discussed today in the meeting, for a result batch in gpu representation, it may better to first convert to `cpu_data_representation` (fixed-size pages and pinned mem), and then copy it to a duckdb Datachunk with memory managed duckdb (continuous buffer and pageable mem).
> 
> @felipeblazing I think what we discussed is a good way to go, just wonder compared to this, would it be same if we use another separate rmm device_memory_resource on the host purely for result memory allocation? In a sense it's just a replacement of using duckdb's managed memory, but we can avoid this double copying.

Let's just do this optimization later since we are on express timeline

bwyogatama:
Two things we need to do here:
1. Rebase with main, a lot of things has changed and we are using cuCascade now, so we can use the representation converter in cuCascade that @felipeblazing introduced for this purpose
2. Integrate with the Result Collector. I think the easiest way of getting this working is to do the conversion in the Result Collector operator and execute it in the Pipeline Executor as one of the operators in a pipeline, no need to create task for the downgrade executor.

kevkrist:
One thing I worry about here is the lifetime of the `duckdb_table_representation`. Can we guarantee that the `duckdb_table_representation` doesn't release its memory or get downgraded to disk before the duckdb data chunk finishes its own lifetime?

PR #115, opened 2025-11-28
Avoid invalid CUDA launches when converting empty row-ids

### LINKED ISSUE #114: Sirius crashes when a join query returns no rows
### What happened?

When complex joins that include string columns produce zero rows, Sirius silently exits. 

### What did you expect to happen?

That Sirius does not silently exit and instead returns 0 rows as expected.

### Steps to Reproduce:

Run a string join whose predicates eliminate every row through the `gpu_processing`.
Observe that the engine logs a GPU assert and exits.

Image

Aisha630:
### Summary

Added explicit `N == 0` guards to `convertUInt64ToInt32` and `convertInt32ToUInt64` (the helpers used when translating Sirius row-id/offset buffers to cuDF-compatible ones). When the input length is zero, the functions now immediately return `nullptr` instead of launching `<<<0, BLOCK_THREADS>>>`, which CUDA treats as an invalid configuration.
This prevents the crash encountered when GPU pipelines produce empty result sets but still run through the row-id conversion path (e.g., joins that ultimately yield zero rows). 

Closes: #114 

fixed

### DIFF
--- a/src/cuda/cudf/cudf_utils.cu
+++ b/src/cuda/cudf/cudf_utils.cu
@@ -62,6 +62,9 @@ __global__ void convert_int32_to_uint64(int32_t* data, uint64_t* output, size_t 
 }
 
 int32_t* convertUInt64ToInt32(uint64_t* data, size_t N) {
+    if (N == 0) {
+        return nullptr;
+    }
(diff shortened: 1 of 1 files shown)


bwyogatama:
This is great, thanks for the contributions!

PR #116, opened 2025-12-02
Read Parquet using cuDF parquet reader

BarryWu0812:
This PR is based on #108 to develop, and add the feature that supporting gpu parquet reading by using `cudf::io::read_parquet`.

This PR also passes the parquet test added in a1e9f16e02c6f41956bd7d9bc815a813aecc05df.


bwyogatama:
Thanks Barry for the contributions, can you rebase to the dev branch so that we can merge these changes?

BarryWu0812:
Sure! I think I can rebase it after duke's PR is merged.

bwyogatama:
No worries, will reach out to you again once Duke PR's is merged

felipeblazing:
When we think about how to make a CUDF reader we should think about how we are going to be able to separate out the I/O from the decoding and decompression that is happening on the GPU. If we block the decoding for I/O the way read_parquet does when invoked in the way in this PR we probably will get worse performance than using the duckdb scanner. 

There are a few reasons for this. 

1. Look at the pre-fetching executor here [Theseus paper](https://arxiv.org/abs/2508.05029)
2. Reading this way will consume GPU memory for longer than necessary. We want to make sure that we find ways to minimize the consumption of memory over time. Moving bytes from data source to Pinned host memory first as a staging area before decoding / decompressing. We can do this by making some kind of custom data source that can fetch those bytes seperately from actually invoking read_parquet

kevkrist:
@felipeblazing 
I've read the paper closely and I'm interested in working on this. I think I'm stuck on the fact that the cudf APIs seem to need a full parquet file image (not just a byte range + the required metadata). Does the pre-loader rebuild a parquet file image from the byte range, or are there custom decoding/decompression kernels? Or am i just missing something?

BarryWu0812:
Hi @bwyogatama , I saw @kevkrist have a new commit to solve the parquet issue, so does this PR still need to be merge?

bwyogatama:
closed in favor of #231

PR #121, opened 2025-12-13
Add DECIMAL support for comparison operations

woodyycchang:
Problem:
Comparison operations (=, <, >, etc.) on DECIMAL columns with constant values threw 'Unsupported constant type for comparison: 27', causing queries with HAVING clauses on DECIMAL aggregates to fall back to CPU.

Root cause:
The comparison dispatcher only handled INT, FLOAT, BOOL, and STRING types, but not DECIMAL32, DECIMAL64, or DECIMAL128.

Solution:
Add cases for DECIMAL types in the comparison switch statement, creating fixed_point_scalar objects for the comparison operation.

Testing:
- TPC-DS SF1 query with HAVING clause on DECIMAL SUM
- Before fix: Falls back to CPU
- After fix: Runs on GPU with correct results

Tested on:
- Tesla V100-SXM2-32GB (Compute 7.0)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,12 @@ add_compile_definitions(ENABLE_CUDF)
 # Set extension name here
 set(TARGET_NAME sirius)
 
+# Set CUDA architecture
+if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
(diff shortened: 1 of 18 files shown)


bwyogatama:
Thanks for the contributions Woody! Can we rebase and resolve the conflict before I merge the changes?

PR #122, opened 2025-12-14
Fix COUNT(column) to exclude NULL values in ungrouped aggregates

woodyycchang:
## Problem
COUNT(column) returned the same result as COUNT(*), incorrectly counting NULL values. This violates SQL standard semantics.

## Root Cause
The ungrouped aggregate COUNT implementation returned total row count instead of counting non-NULL values.

## Solution
Use cudf column's `null_count()` to properly exclude NULLs:
```cpp
cudf::size_type null_count = cudf_column.null_count();
uint64_t non_null_count = cudf_column.size() - null_count;
```

## Testing
TPC-DS SF1 store_sales (2.88M rows, 129K NULLs):

| Query | CPU | GPU Before | GPU After |
|-------|-----|------------|-----------|
| COUNT(*) | 2,880,404 | 2,880,404 | 2,880,404 ✅ |
| COUNT(ss_customer_sk) | 2,750,652 | 2,880,404 ❌ | 2,750,652 ✅ |

Note: This PR also includes commits from #118 and #120 due to branch history.

Tested on Tesla V100-SXM2-32GB

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,7 @@ add_subdirectory(src/operator)
 add_subdirectory(src/plan)
 add_subdirectory(src/data)
 add_subdirectory(src/pipeline)
+add_subdirectory(src/downgrade)
 
 set(CUDA_SOURCES
   src/cuda/communication.cu
--- a/src/creator/downgrade_task_creator.cpp
+++ b/src/creator/downgrade_task_creator.cpp
@@ -14,19 +14,18 @@
  * limitations under the License.
  */
 
-#include "memory/common.hpp"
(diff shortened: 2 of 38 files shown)


bwyogatama:
Thanks for the contributions, can we rebase to the dev branch so that we can merge this commit?

woodyycchang:
Sure, rebased and resolved the conflict, ready for review!

bwyogatama:
There is an error in the build that we would need to fix (see the failed checks below)

bwyogatama:
Since this PR also includes commits from #118 and #120, we will try to merge in those two first before this, then we can rebase and merge this one

PR #123, opened 2025-12-14
Fix sign extension in integer to int128 conversions

woodyycchang:
## Problem
Converting INT64/INT32/INT16 to INT128 corrupted negative values. For example:
- Input: `-155262`
- GPU output: `18446744073709396354` (which is 2^64 - 155262)

## Root Cause
The `convert_intXX_to_int128` CUDA kernels always set upper 64 bits to zero:
```cpp
output_ptr[1] = 0;  // Wrong for negative numbers!
```

For proper two's complement sign extension:
- Negative values need upper bits = `0xFFFFFFFFFFFFFFFF` (all 1s)
- Positive values need upper bits = `0x0000000000000000` (all 0s)

## Solution
Fixed all three conversion kernels to properly sign-extend:
```cpp
output_ptr[1] = (value < 0) ? -1 : 0;
```

Also fixed `gpu_columns.cpp` to use `int64_t` instead of `uint64_t` for INT64 scalar casting.

## Testing
```sql
SELECT SUM(ss_quantity - 100) FROM store_sales WHERE ss_sold_date_sk = 2452260;
```
| | Before | After | CPU |
|--|--------|-------|-----|
| Result | 18446744073709396354 ❌ | -155262 ✅ | -155262 |

Positive values regression test passed.

Tested on Tesla V100-SXM2-32GB

### DIFF
--- a/src/cuda/utils.cu
+++ b/src/cuda/utils.cu
@@ -52,15 +52,13 @@ __global__ void convert_int64_to_int128(uint8_t* input, uint8_t* output, size_t 
 {
   size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
   if (idx < count) {
-    // Store the converted value as 16 bytes in output
-    uint64_t* input_ptr  = reinterpret_cast<uint64_t*>(input + idx * 8);
-    uint64_t* output_ptr = reinterpret_cast<uint64_t*>(output + idx * 16);
-    // for (int i = 0; i < 8; ++i) {
-    // output[idx * 16 + i] = input[idx * 8 + i];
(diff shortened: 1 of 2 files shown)


bwyogatama:
Thanks for the contributions, can we rebase to the dev branch so that we can merge this commit?

woodyycchang:
Sure, rebased and resolved the conflicts, ready for review!

bwyogatama:
Thanks again Woody! We can merge this in but we need to pass the pre-commit hooks first (check the linter error before)

woodyycchang:
Fixed the formatting with pre-commit

PR #124, opened 2025-12-14
Enable string scalar materialization for GPU aggregates

Aisha630:
This pull request includes adding support for string scalars in GPU columns. This support for string aggregation is already provided by cuDF but was not being utilized.

* Added support for handling `STRING` type scalars in `GPUColumn::setFromCudfScalar`, including allocation and copying of string data and offsets, and setting string-specific metadata.
* Replaced `memcpy` with a byte-by-byte copy for inlined strings in the CUDA materialization kernel to avoid unaligned memory access issues on the GPU.
* Modified aggregation checks in `HandleAggregateExpressionCuDF` to allow `avg`, `min`, and `max` to run on empty inputs instead of being rejected, returning 0 rows downstream.

This PR was tested on the entire JOB benchmark which was not proceeding earlier due to unsupported string aggregation. Results were also compared with DuckDB. `make test` was also verified.

Previously:

image

Now:
job_sirius_vs_duckdb_run5

### DIFF
--- a/src/cuda/operator/materialize.cu
+++ b/src/cuda/operator/materialize.cu
@@ -501,7 +501,13 @@ __global__ void create_cpu_strings(duckdb_string_type* gpu_strings,
     if (str_length <= inline_threshold) {
       curr_string.value.inlined.length = str_length;
       char* gpu_data_ptr               = gpu_chars + str_offset;
-      memcpy(curr_string.value.inlined.inlined, gpu_data_ptr, str_length);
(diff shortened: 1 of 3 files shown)


bwyogatama:
Thanks for the contribution! Can we resolve the conflict and rebase to dev branch so that we can merge these changes?

Aisha630:
> Thanks for the contribution! Can we resolve the conflict and rebase to dev branch so that we can merge these changes?

Done

bwyogatama:
This is great! I will merge this commit

PR #131, opened 2025-12-22
Extend expression executor to handle `data_batch_view` inputs

kevkrist:
This commit adds new APIs to the expression executor for `execute()` and `select()` to work with `data_batch` inputs and outputs. The existing code is not affected in any way.


src/include/data/data_repository.hpp
```diff
@@ -16,7 +16,7 @@
 
 #pragma once
 
-#include "data_batch.hpp"
+#include <data/data_batch_view.hpp>
```
bwyogatama: This data_repository file should no longer exist as it has been moved to cuCascade

src/expression_executor/gpu_expression_executor.cpp
```diff
@@ -14,22 +14,39 @@
...
+// rmm
+#include <cuda_stream_view.hpp>
+#include <resource_ref.hpp>
 
 namespace duckdb {
 namespace sirius {
 
-GpuExpressionExecutor::GpuExpressionExecutor(const Expression& expr) { AddExpression(expr); }
+GpuExpressionExecutor::GpuExpressionExecutor(const Expression& expr,
+                                             rmm::device_async_resource_ref resource_ref)
```
bwyogatama, line 41: I am pretty sure the one that you should have used is the cucascade::reservation_aware_resource_adapter instead of rmm::device_async_resource_ref

bwyogatama: and you can get cucascade::reservation_aware_resource_adapter from the memory_space

kevkrist: I think this should be the responsibility of the caller. The expression executor needs the `rmm::device_async_resource_ref` for cudf APIs, and using the async resource ref requires the fewest changes to the existing code. The caller can get the async resource ref easily:
```
rmm::device_async_resource_ref get_resource_ref(memory_space& space)
{
  return rmm::to_device_async_resource_ref_checked(space.get_default_allocator());
}
```

src/include/data/data_repository.hpp
kevkrist: Rebasing now.

bwyogatama:
@kevkrist So I talked with @aminaramoon to make sure that we get this right. 

Here is the correct workflow:
1. Before we launched the task, we will call
`auto res = manager->request_reservation(specific_memory_space{memory tier, device id}, reservation_size);`
And then we pass the `res` to the task.

2. In the beginning of the task, we call
`auto* mr = res->get_memory_resource_as<reservation_aware_resource_adaptor>();`
This will give us our `reservation_aware_resource_adapter`.

3. After that, we need to attach the resource to the reservation for tracking:
`mr->attach_reservation_to_tracker(stream, std::move(res));`
This will attach the tracker into the calling thread, so now all the cudf operator executed by this thread can be tracked.

4. We then set cudf default allocator to use our resource
`cudf::set_current_device_resource(mr)`

5. Inside our operator implementation (`GPUPhysicalOperator::Execute()`), we can call a bunch of cudf operator (WE DON'T NEED TOP PASS MEMORY RESOURCE TO THE cuDF OPERATOR)

6. Then after all the operators in the pipeline are done, we call: 
`mr->get_peak_total_allocated_bytes()` 
This will update the total allocated bytes in the tracker.

7. At the end of the task, we call 
`mr->reset_peak_allocated_bytes()`
This will reset our tracker

kevkrist:
Ok, so it seems to me that the least invasive option, and the one that preserves existing code function most simply, is to store the current device resource ref (via `cudf::get_current_device_resource_ref()`  [[link](https://docs.rapids.ai/api/libcudf/stable/group__memory__resource)]) as a field in the executor, and pass that to the cudf functions (the old code still needs to explicitly pass its own resource reference as far as I can tell). So either
1) The caller can pass `cudf::get_current_device_resource_ref()` (link) for the `resource_ref` argument in the constructor, or
2) The underlying `resource_ref` member is set when the `data_batch` APIs are invoked to the current device resource ref.
Does this sound reasonable? Preference for 1 or 2?

bwyogatama:
I understand (1) but can you elaborate (2)? Which data batch API you were talking about?

kevkrist:
I mean the `select` and `execute` methods that use `data_batch` as inputs and outputs. So, whenever one of these methods is called, the resource ref field of the executor is set to the current device resource ref before executing the projections/filters.
Option (1) requires no changes to current PR, option (2) very minor changes.

bwyogatama:
Let's just do option (1)

PR #133, opened 2025-12-24
Feature/sirius context and config

aminaramoon:
Features addded:

- added a confiration manager using libconfig that allows configuration to be set from different libraries without injecting dependency to the third-party libraries : i.e. you can configure cucascade without adding the configuration manager to the cucascade.
- added sirius_context, a global singleton store for all engine resource
- added sirius_context_lifetime_manager that manages lifetime of the states, begin/end queries, configuring the engine using a config file.


src/sirius_context.cpp
```diff
@@ -0,0 +1,226 @@
...
+  return *task_creator_;
+}
+
+const sirius::creator::task_creator& SiriusContext::get_task_creator() const
+{
+  thorw_if_not_initialized();
+  return *task_creator_;
+}
+
+void SiriusContext::thorw_if_not_initialized() const
```
joosthooz, line 193: typo - throw

src/include/pipeline/config.hpp
```diff
@@ -0,0 +1,39 @@
...
+namespace sirius {
+namespace pipeline {
+
+struct pipeline_executor_config {
+  pipeline_executor_config() noexcept
+  {
+    pipeline_task_executor_config.num_threads    = 4;
+    pipeline_task_executor_config.retry_on_error = false;
+
+    gpu_task_executor_config.num_threads    = 2;
```
joosthooz: What is the difference between the pipeline and gpu task executors? Why does one have 2 threads and the other 4?

src/creator/task_creator.cpp
```diff
@@ -100,23 +114,23 @@ void task_creator::process_next_task(op::sirius_physical_operator* node)
     auto pipeline = std::get<duckdb::shared_ptr<pipeline::sirius_pipeline>>(hint);
     process_next_task(&pipeline->get_inner_operators()[0].get());
   } else {
-    if (!priority_scans.empty()) {
-      duckdb::shared_ptr<pipeline::sirius_pipeline> pipeline = priority_scans.front();
+    if (!_priority_scans.empty()) {
```
joosthooz, line 117: Not about this PR specifically, but if we are doing `while (!_priority_scans.empty()) {` here https://github.com/sirius-db/sirius/pull/133/changes#diff-daca317e191e57036793b6c21fb12b69ca29699cfe5609bf34b8bf78f3cd54a3R129, shouldn't `_priority_scans` be unneeded here? Or will there be additional tasks pushed into it?

joosthooz: Also, if this function is to be called by an executor thread, we need to make these queue accesses thread-safe

PR #136, opened 2026-01-02
Push duckdb scan data to data repository via `host_table_allocation`

kevkrist:
This PR will
1. Add utilities for constructing metadata for `host_table_allocation` from host data produced by duckdb scan tasks.
2. Uses the `host_table_allocation` to push data batches to the data repository when a duckdb scan task completes.

It also fixes some scan task concurrency issues, which are now well-tested.


felipeblazing (changes requested):
This makes sense. We probably want to consider moving the metadata info to somewhere more generic because I think that could be useful for all kinds of things like spilling or storing on disk. The only change I am requesting is that this gets updated to the latest version. There were some changes recently in cucascade and in Sirius that is conflicting with this now.

test/cpp/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
 add_subdirectory(utils)
+add_subdirectory(memory)
```
joosthooz, line 20: Consider merging test subdirs `memory` and `memory_management`

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -478,16 +523,16 @@ void duckdb_scan_task::execute()
...
     auto next_task = std::make_unique<duckdb_scan_task>(
       new_task_id, _data_repo, std::move(new_local_state), shared_global_state);
     g_state.scan_executor.schedule(std::move(next_task));
   }
 
-  /// FUTURE WORK: Create the data batch and push it to the data repository.
-
-  /// FUTURE WORK: Notify task_creator of completion by sending a message to the message queue
+  // Make data batch and push to repository
+  if (l_state.row_offset > 0) { _data_repo->add_data_batch(l_state.make_data_batch()); }
```
joosthooz: Why only if row offset is 0?

src/include/sirius_context.hpp
```diff
@@ -62,33 +62,34 @@ class SiriusContext : public ClientContextState {
...
-  void initialize(const sirius::sirius_config& config);
+  void initialize(const ::sirius::sirius_config& config);
 
   /// \brief Terminate the Sirius context, releasing all resources.
   void terminate();
 
   /// \brief Get the memory reservation manager.
-  [[nodiscard]] sirius::memory::sirius_memory_reservation_manager& get_memory_manager();
-  [[nodiscard]] const sirius::memory::sirius_memory_reservation_manager& get_memory_manager() const;
+  [[nodiscard]] ::sirius::memory::sirius_memory_reservation_manager& get_memory_manager();
```
joosthooz: Most of the codebase currently doesn't use the `::` prefix for the sirius namespace, let's remove them for consistency

src/include/op/scan/duckdb_scan_task.hpp
```diff
@@ -433,7 +460,15 @@ class duckdb_scan_task_local_state : public sirius::parallel::itask_local_state
     duckdb::ExecutionContext& exec_ctx,
     size_t approximate_batch_size = duckdb::Config::DEFAULT_SCAN_TASK_BATCH_SIZE,
     size_t default_varchar_size   = duckdb::Config::DEFAULT_SCAN_TASK_VARCHAR_SIZE,
-    duckdb::unique_ptr<duckdb::LocalTableFunctionState> existing_local_tf_state = nullptr);
+    std::unique_ptr<duckdb::LocalTableFunctionState> existing_local_tf_state = nullptr);
```
joosthooz, line 483: Are we moving away from `duckdb::unique_ptr`? (for consistency)

src/include/op/scan/duckdb_scan_task.hpp
```diff
@@ -88,18 +92,32 @@ class duckdb_scan_task_global_state : public sirius::parallel::itask_global_stat
...
+    auto const remaining = active_local_states.fetch_sub(1, std::memory_order_acq_rel) - 1;
+    if (remaining == 0) { set_source_drained(); }
+  }
+
   //===----------Fields----------===//
-  std::atomic<bool> source_drained{false};  ///< Whether the table scan source is fully drained
-  duckdb::shared_ptr<pipeline::sirius_pipeline>
-    pipeline;            ///< The pipeline to which this table scan belongs
-  uint64_t max_threads;  ///< Maximum number of threads for this scan task
+  std::atomic<bool> source_drained{false};      ///< Whether the table scan source is fully drained
```
joosthooz: Not sure what kind of conventions we want to work towards in this codebase, but do we want to make (some of) these fields private and create getters/setters for them?

src/include/memory/host_table_utils.hpp
```diff
@@ -0,0 +1,139 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+// sirius
+#include <expression_executor/gpu_expression_executor_state.hpp>  /// FIXME: remove type conversion from expression executor
```
joosthooz: Do you wish to address this in a follow-up? Consider adding an issue for it

src/include/op/scan/duckdb_scan_task.hpp
```diff
@@ -88,18 +92,32 @@ class duckdb_scan_task_global_state : public sirius::parallel::itask_global_stat
    */
   void set_source_drained() { source_drained.store(true, std::memory_order_release); }
 
+  /**
+   * @brief Register a new local table function state.
+   */
+  void register_local_state() { active_local_states.fetch_add(1, std::memory_order_relaxed); }
```
joosthooz: This function and its documentation confuse me a bit; it doesn't really register something, just increments a counter. Does it use the counter to keep track of how many instances of something are working on this object?

src/include/op/scan/duckdb_scan_task.hpp
joosthooz: This should be prefixed with a `_` for consistency (not specific to this PR)

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -266,6 +272,26 @@ void duckdb_scan_task_local_state::column_builder::process_column(
   process_mask_for_column(validity, num_rows, row_offset, allocation);
 }
 
+metadata_node duckdb_scan_task_local_state::column_builder::make_metadata_node(
+  size_t num_rows) const
+{
+  if (type.InternalType() == duckdb::PhysicalType::VARCHAR) {
```
joosthooz, line 281: Are there any other types that are currently not supported by this function yet?

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -275,34 +301,31 @@ duckdb_scan_task_local_state::duckdb_scan_task_local_state(
...
     default_varchar_size(default_varchar_size),
-    exec_ctx(exec_ctx),
-    local_tf_state(std::move(existing_local_tf_state))  // Move the existing state if provided
+    exec_ctx(exec_ctx)
 {
   auto const& op = g_state.op;
   num_columns    = op.projection_ids.size();
 
-  auto& mem_res_mgr = g_state.mem_res_mgr;
+  bool const has_existing_state = static_cast<bool>(existing_local_tf_state);
```
joosthooz: This explicit cast to bool isn't needed, you can just do `if (existing_local_tf_state)`.

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -275,34 +301,31 @@ duckdb_scan_task_local_state::duckdb_scan_task_local_state(
...
+  local_tf_state                = std::move(existing_local_tf_state);
```
joosthooz: What if the pointer is empty, shouldn't `local_tf_state` be initialized with something in that case?

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -382,6 +405,32 @@ void duckdb_scan_task_local_state::initialize_local_table_function_state(
...
+  // Create metadata nodes for each column and assemble metadata buffer
+  std::vector<metadata_node> column_metadata;
+  column_metadata.reserve(num_columns);
+  for (auto const& builder : column_builders) {
+    column_metadata.push_back(builder.make_metadata_node(row_offset));
+  }
+  auto metadata = std::make_unique<std::vector<uint8_t>>(pack_metadata_from_nodes(column_metadata));
+
+  // Make the host table allocation
+  auto const sz = allocation->size_bytes();  /// TODO: change to actual used tail byte
```
joosthooz: Can we already do this in this PR or in a follow-up?

test/cpp/scan/test_scan_executor.cpp
```diff
@@ -38,265 +43,50 @@
...
-
-      app.EndRow();
-    }
+static std::filesystem::path get_test_config_path()
+{
+  return std::filesystem::path(__FILE__).parent_path() / "memory.cfg";
+}
 
-    app.Close();
+static duckdb::shared_ptr<duckdb::SiriusContext> get_sirius_context(duckdb::Connection& con)
```
joosthooz: This function occurs 3x, can it be moved into a utilities file for deduplication?

joosthooz (commented):
Thank you for this, I've posted some questions and comments, I hope it is useful!

src/op/scan/duckdb_scan_task.cpp
kevkrist: `row_offset` is the row offset in the local allocation, so if and only if `row_offset > 0`, we know that this task has accumulated data that needs to be pushed to the repository.

src/op/scan/duckdb_scan_task.cpp
kevkrist: Yes, it's initialized by the final call in the constructor, to `initialize_local_table_function_state()`, which omits re-initializing if `local_tf_state` is not nullptr.

src/include/op/scan/duckdb_scan_task.hpp
kevkrist: Yeah I think we want to use std::unique/shared_ptr whenever we can, but when interfacing with duckdb code (e.g. `ClientContext`), a lot of things only work with the duckdb variant.

test/cpp/scan/test_scan_executor.cpp
kevkrist: Refactored and moved to utils/utils/hpp in the test/cpp directory.

src/op/scan/duckdb_scan_task.cpp
kevkrist: Fixed

src/op/scan/duckdb_scan_task.cpp
kevkrist: Done.

src/include/memory/host_table_utils.hpp
kevkrist: Moved to `cudf_utils.hpp`.

src/include/sirius_context.hpp
kevkrist: The reason I did this is because the TU with the changes included a `duckdb::sirius` namespace, which was causing namespace collisions here. After moving `GetCudfType()` out of the expression executor code (which lives in `duckdb::sirius`, everything now compiles cleanly without the leading `::`. Fixed.

src/include/op/scan/duckdb_scan_task.hpp
kevkrist: Fixed.

test/cpp/CMakeLists.txt
kevkrist: I feel that they have somewhat separate purposes. The `memory` directory is for things belonging under the `memory` subdirectory (e.g., utilities for interacting with objects in `cucascade::memory` namespace), but `memory_management` deals with code that manages those objects (local to Sirius). In any case, I'm happy to merge one way or the other if you feel strongly about it.

src/include/op/scan/duckdb_scan_task.hpp
kevkrist: yes exactly. I've updated the function names and added some documentation. Please let me know if it's still not clear or if you'd like more explanation.

src/include/op/scan/duckdb_scan_task.hpp
kevkrist: I've made the fields private and declared `duckdb_scan_task` and the states as friends as a kind of compromise. I originally followed the duckdb pattern here for the local and global states.

src/op/scan/duckdb_scan_task.cpp
kevkrist: Yeah, any nested types that aren't VARCHAR. The question is where to enforce this constraint. Probably at scan task construction, or even better, when the task creator receives the types from duckdb. I'm just not sure that enforcing type safety is the responsiblity of the code in this PR. What do you think?

felipeblazing (approved):

src/op/scan/duckdb_scan_task.cpp
joosthooz: If that's a more general limitation of sirius then it's not the responsibility of this code, the reason I asked was in case there were things you'd like to add later, then you'd might like to leave a TODO or make an issue

src/include/op/scan/duckdb_scan_task.hpp
```diff
@@ -470,6 +490,7 @@ class duckdb_scan_task_local_state : public sirius::parallel::itask_local_state
    */
   std::shared_ptr<data_batch> make_data_batch();
 
+ private:
```
joosthooz, line 493: If we make all the fields below private, we should add a `_` prefix for consistency.

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -339,6 +341,16 @@ duckdb_scan_task_local_state::duckdb_scan_task_local_state(
   initialize_local_table_function_state(op, exec_ctx, g_state.global_tf_state.get());
 }
 
+size_t duckdb_scan_task_local_state::get_tail_byte_offset() const
+{
+  auto const& last_builder = column_builders.back();
+  auto last_byte_offset    = last_builder.mask_blocks_accessor.get_current_global_byte_offset();
+  if (utils::mod_8(last_byte_offset) != 0) {
+    last_byte_offset++;  // Round up to next byte if partially filled
```
joosthooz, line 352: `get_current_global_byte_offset` already returns entire bytes, doesn't it? Currently this code adds 1 byte if the offset is not 8-byte aligned, not sure if that is your intention

src/include/op/scan/duckdb_scan_task.hpp
```diff
@@ -78,46 +81,51 @@ class duckdb_scan_task_global_state : public sirius::parallel::itask_global_stat
...
    */
-  void release_local_state()
+  void decrement_local_states()
   {
-    auto const remaining = active_local_states.fetch_sub(1, std::memory_order_acq_rel) - 1;
+    auto const remaining = _active_local_states.fetch_sub(1, std::memory_order_acq_rel) - 1;
     if (remaining == 0) { set_source_drained(); }
   }
 
+ private:
```
joosthooz, line 117: Some of the fields don't have a  `_` yet

joosthooz (changes requested):
Thanks for addressing all of my comments, I just want to ask to have a look at `get_tail_byte_offset` because I'm not sure it does what you want. Besides that just tiny cosmetic things, trying to improve the consistency in the codebase a bit, hope you don't mind

src/include/op/scan/duckdb_scan_task.hpp
kevkrist: Fixed.

src/include/op/scan/duckdb_scan_task.hpp
kevkrist: Fixed.

src/op/scan/duckdb_scan_task.cpp
kevkrist: It's not -- definitely a typo, thank you. It should be row_offset to check whether the current mask cursor is pointing to a byte that has some but not all of its bits filled.

joosthooz (approved):
Great, thank you!

PR #139, opened 2026-01-03
Task Creator

bwyogatama:
This PR implements the task creator in Sirius following @aminaramoon design and is split from #132

### DIFF
--- a/src/include/creator/task_creator.hpp
+++ b/src/include/creator/task_creator.hpp
@@ -50,6 +50,9 @@ class task_creation_info {
                      ::duckdb::shared_ptr<::duckdb::GPUPipeline> pipeline)
     : _node(node), _pipeline(std::move(pipeline))
   {
+    if (!_pipeline) {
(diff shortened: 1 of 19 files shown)


src/include/creator/task_creator.hpp
```diff
@@ -16,54 +16,250 @@
...
+   * is closed and empty, it returns nullptr to signal that no more tasks will arrive.
+   *
+   * @return A unique pointer to the task creation info if available, nullptr if the
+   *         queue is closed and empty.
+   */
+  std::unique_ptr<task_creation_info> pull();
+
+ private:
+  size_t _num_threads;
+  duckdb_moodycamel::BlockingConcurrentQueue<std::unique_ptr<task_creation_info>> _queue;
```
felipeblazing: At some point we should endeavour to move away from the duckdb containers and classes and use ones that do not depend on it if we some day want to support other runtimes like Doris. Lets leave this for now just a comment

PR #141, opened 2026-01-08
Fix: Prevent silent data corruption in BIGINT SUM on GPU

woodyycchang:
## Summary
Fixes critical silent data corruption bug where GPU SUM operations on BIGINT columns returned incorrect results due to integer overflow.

- **Bug # 6**: BIGINT SUM aggregation  
- **Bug # 7**: BIGINT multiplication
- **Bug # 8**: BIGINT addition
- **Bug # 9**: BIGINT subtraction

## Problem
- GPU SUM used INT64 accumulator for BIGINT columns
- When sum exceeded 2^63-1, silent overflow occurred
- Examples: returned `0` instead of `461168601842738790400`, or `-80` instead of `92233720368547758000`
- **No error thrown** - silent data corruption

## Root Cause
- cuDF's `reduce` operation uses INT64 accumulator for BIGINT SUM
- GPU pipeline doesn't support INT128/DECIMAL128 accumulator
- Overflow wraps around silently

## Solution
Modified `src/cuda/cudf/cudf_aggregate.cu` to detect BIGINT SUM operations and throw `NotImplementedException`, triggering automatic CPU fallback which correctly handles overflow with INT128 accumulator.

## Testing
All BIGINT SUM operations now fall back to CPU with correct results  
Bug case: Returns `461168601842738790400` (was `0` before)  
No more silent data corruption

## Trade-offs
- Performance: BIGINT SUM now executes on CPU instead of GPU
- Correctness: 100% accurate results guaranteed
- Future: Can be optimized when GPU pipeline adds INT128 support

## Files Changed
- `src/cuda/cudf/cudf_aggregate.cu`: Added INT64 overflow detection (+4 lines)

Screenshot 2026-01-07 at 9 54 14 PM

### DIFF
--- a/src/cuda/cudf/cudf_aggregate.cu
+++ b/src/cuda/cudf/cudf_aggregate.cu
@@ -127,6 +127,10 @@ void cudf_aggregate(vector<shared_ptr<GPUColumn>>& column,
         } else if (num_rows > 1) {
           to_cudf_type = cudf::data_type(cudf::type_id::INT32);
         }
+      } else if (to_cudf_type.id() == cudf::type_id::INT64) {
+        // INT64 SUM can overflow - GPU doesn't support INT128 accumulator
+        // Throw exception to trigger CPU fallback which handles overflow correctly
+        throw NotImplementedException("GPU SUM of BIGINT may overflow - falling back to CPU");
       }
 
       // CuDF reduce will return an invalid scalar if all values are NULL
--- a/src/expression_executor/specializations/gpu_execute_function.cpp
+++ b/src/expression_executor/specializations/gpu_execute_function.cpp
@@ -463,13 +463,13 @@ struct NumericBinaryFunctionDispatcher {
     auto left  = executor.Execute(*expr.children[0], state->child_states[0].get());
     auto right = executor.Execute(*expr.children[1], state->child_states[1].get());
 
-
     // Check for BIGINT multiplication overflow potential
     if constexpr (BinOp == cudf::binary_operator::MUL) {
       if (left->view().type().id() == cudf::type_id::INT64 &&
           right->view().type().id() == cudf::type_id::INT64) {
         // BIGINT multiplication can overflow - GPU doesn't check for overflow
-        throw NotImplementedException("GPU BIGINT multiplication may overflow - falling back to CPU");
+        throw NotImplementedException(
+          "GPU BIGINT multiplication may overflow - falling back to CPU");
       }
     }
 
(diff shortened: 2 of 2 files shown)


bwyogatama:
Thanks for the contribution @woodyycchang . Seems like some of the test still doesn't work here, can you make sure it works before we start the review?

woodyycchang:
Hi @bwyogatama, all 4 CI checks are passing now, rebased onto latest dev and applied clang-format. Ready for review whenever you get a chance. Thanks!

bwyogatama (approved):

src/sirius_extension.cpp
```diff
@@ -311,16 +311,16 @@ void SiriusExtension::GPUProcessingFunction(ClientContext& context,
         "DuckDB\n=============================================\n");
       data.res = data.conn->Query(data.query);
     } else if (data.plan_error) {
-      printf(
-        "=============================================\nError in GPUExecuteQuery, fallback to "
-        "DuckDB\n=============================================\n");
+      SIRIUS_LOG_DEBUG("Plan error in GPUExecuteQuery, fallback to DuckDB");
```
bwyogatama: What is the purpose of eliminating the printf? I would say put both

src/sirius_extension.cpp
```diff
@@ -311,16 +311,16 @@ void SiriusExtension::GPUProcessingFunction(ClientContext& context,
...
       data.res = data.conn->Query(data.query);
     } else {
       data.res = data.gpu_context->GPUExecuteQuery(context, data.query, data.gpu_prepared, {});
       if (data.res->HasError()) {
-        printf(
-          "=============================================\nError in GPUExecuteQuery, fallback to "
-          "DuckDB\n=============================================\n");
+        if (data.res->GetErrorObject().Type() != ExceptionType::NOT_IMPLEMENTED) {
```
bwyogatama: For this as well, we still want the original error message

bwyogatama (changes requested):

src/sirius_extension.cpp
bwyogatama: Cause i want us to know that we fallback to duckdb when there is error, not just error specific to `ExceptionType::NOT_IMPLEMENTED`

woodyycchang:
Reverted all sirius_extension.cpp changes. PR now only modifies cudf_aggregate.cu and gpu_execute_function.cpp.

bwyogatama (approved):

PR #146, opened 2026-01-13
Result collector from `host_table_representation`

kevkrist:
This PR adapts the `GPUPhysicalResultCollector` to `data_batch` inputs. If the `data_batch` is in GPU memory, it is first converted to pinned memory via `host_table_representation` using the default converter. Then the data is copied into pageable DuckDB `DataChunk`s for consumption by the DuckDB endpoints.


src/include/memory/host_table_utils.hpp
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * @brief Metadata node structure for serializing cudf column metadata.
+ *
+ * This structure is used to create host_table metadata, similar to cudf::pack(), but without a
+ * cudf::table_view. This is important in contexts where data has not been migrated to CPU memory
+ * from GPU memory, e.g., after table scans, into a `host_table_allocation`.
+ *
+ */
+struct metadata_node {
```
felipeblazing, line 53: don't we need scale here too for decimal columns?

src/include/memory/host_table_utils.hpp
```diff
@@ -0,0 +1,244 @@
...
+};
+
+namespace detail {
+
+/**
+ * @brief Local copy of cudf::detail::serialized_column for metadata parsing.
+ *
+ * This struct mirrors the layout of the serialized metadata used by cudf::pack().
+ */
+struct serialized_column {
```
felipeblazing, line 155: is this a copy out of some struct found in libcudf that doesn't normally get exposed?

src/include/memory/host_table_utils.hpp
```diff
@@ -0,0 +1,244 @@
...
+  cudf::data_type type;
+  cudf::size_type size;
+  cudf::size_type null_count;
+  int64_t data_offset;
+  int64_t null_mask_offset;
+  cudf::size_type num_children;
+  int pad{};
```
felipeblazing, line 164: lets add some c docs so we can know what this padding will be used for.

src/include/memory/multiple_blocks_allocation_accessor.hpp
```diff
@@ -0,0 +1,224 @@
...
+ * size of the allocation.
+ */
+template <typename T>
+struct multiple_blocks_allocation_accessor {
+  using underlying_type = T;
+  using multiple_blocks_allocation =
+    cucascade::memory::fixed_size_host_memory_resource::multiple_blocks_allocation;
+
+  //===----------Fields----------===//
+  size_t block_size          = 0;  ///< The size of each block in bytes
```
felipeblazing: do you need block_size and num_blocks? I feel you can just get this by accessing the allocation which doesn't change.

src/include/memory/multiple_blocks_allocation_accessor.hpp
```diff
@@ -0,0 +1,224 @@
...
+   */
+  void reset_cursor() { set_cursor(initial_byte_offset); }
+
+  /**
+   * @brief Set value at the current position in the allocation.
+   *
+   * @param[in] value The value to set.
+   * @param[in] allocation The allocation.
+   */
+  void set_current(T value, std::unique_ptr<multiple_blocks_allocation>& allocation)
```
felipeblazing, line 116: Do we every use this api? It seems inadvisiable to have apis that update one value at a time. We want to normally discourage from writing and readnig single values.

src/include/memory/multiple_blocks_allocation_accessor.hpp
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * @brief Get the value at the current position in the allocation as a different type.
+   *
+   * @tparam S The type to which to cast the value.
+   * @param[in] allocation The allocation.
+   */
+  template <typename S>
+  [[nodiscard]] S get_current_as(
```
felipeblazing, line 133: Similar comment to the one above. Do we want to have single accessors?

src/include/memory/multiple_blocks_allocation_accessor.hpp
```diff
@@ -0,0 +1,224 @@
...
+    return get_current_as<underlying_type>(allocation);
+  }
+
+  /**
+   * @brief Advance the cursor into the allocation to the next position as type S.
+   *
+   * @tparam S The type size to use for advancing the cursor.
+   */
+  template <typename S>
+  void advance_as()
```
felipeblazing, line 179: Similar to above do we want to have apis that allow for this kind of traversal.

src/include/memory/multiple_blocks_allocation_accessor.hpp
```diff
@@ -0,0 +1,224 @@
...
+ * position within the allocation and provides methods to set/get values,
+ * advance the cursor, and perform memcpy operations.
+ * NOTE: the caller is responsible for ensuring the cursor does not go out of bounds. Otherwise,
+ * behavior is undefined.
+ *
+ * @tparam T The underlying data type to be accessed. It is assumed that T is aligned with the block
+ * size of the allocation.
+ */
+template <typename T>
+struct multiple_blocks_allocation_accessor {
```
felipeblazing, line 47: what does this class make simpler? It feels like you just need the copy from and to methods to me but maybe I am mistaken.

src/include/operator/result/host_table_chunk_reader.hpp
```diff
@@ -0,0 +1,163 @@
...
+  // column_reader
+  //===----------------------------------------------------------------------===//
+  /**
+   * @brief Reads a single column from the host table representation and produces duckdb vectors and
+   * validity masks
+   */
+  struct column_reader {
+    size_t size{0};        ///< The number of rows in the column
+    size_t null_count{0};  ///< The number of null values in the column
+    ::sirius::memory::multiple_blocks_allocation_accessor<uint8_t>
```
felipeblazing: is the idea that the accessors here are all sharing a single multi block allocation?

src/operator/gpu_physical_result_collector.cpp
```diff
@@ -516,4 +527,83 @@ unique_ptr<QueryResult> GPUPhysicalMaterializedCollector::GetResult(GlobalSinkSt
...
+    auto reservation = memory_mgr.request_reservation(
+      cucascade::memory::any_memory_space_in_tier{cucascade::memory::Tier::HOST},
+      data->get_size_in_bytes());
+    if (!reservation) {
+      throw InternalException(
+        "[GPUPhysicalMaterializedCollector] Failed to reserve host memory for result collection");
+    }
+
+    // Get the memory space for host tier
+    auto mem_space = memory_mgr.get_memory_space(reservation->tier(), reservation->device_id());
```
felipeblazing: reservation has a get_memory_space() method

src/operator/gpu_physical_result_collector.cpp
```diff
@@ -516,4 +527,83 @@ unique_ptr<QueryResult> GPUPhysicalMaterializedCollector::GetResult(GlobalSinkSt
...
+  }
+
+  auto* data = input_batch->get_data();
+  if (!data) {
+    throw InvalidInputException(
+      "[GPUPhysicalMaterializedCollector] data_batch has no data representation");
+  }
+
+  // If data is in GPU tier, convert to HOST tier first
+  if (data->get_current_tier() == cucascade::memory::Tier::GPU) {
```
felipeblazing: I think this can be ok for now but we should make an issue where in the future we try to register converters to the representations we want. And if a converter doesn't exist between these two types direclty we should find the shortest path to conversion. For now we should not add that feature yet since it would take a minute to get right and I dont want to block merging this pr

src/operator/gpu_physical_result_collector.cpp
```diff
@@ -516,4 +527,83 @@ unique_ptr<QueryResult> GPUPhysicalMaterializedCollector::GetResult(GlobalSinkSt
...
+        "[GPUPhysicalMaterializedCollector] memory_manager is not initialized");
+    }
+    if (::sirius::converter_registry::is_initialized()) {
+      ::sirius::converter_registry::initialize();
+    }
+
+    // Make the memory reservation for host memory
+    auto& memory_mgr = ::sirius::memory_manager::get();
+    auto reservation = memory_mgr.request_reservation(
+      cucascade::memory::any_memory_space_in_tier{cucascade::memory::Tier::HOST},
```
felipeblazing: add a todo here for finding teh closest memory space in the future. In an ideal world we would use the one that is closest to the gpu we are  moving from

src/operator/result/host_table_chunk_reader.cpp
```diff
@@ -0,0 +1,261 @@
...
+    if (leftover_bits > 0) {
+      // Should only be for the last vector in a chunk
+      validity_ptr[bytes_to_copy] =
+        mask_accessor.get_current(allocation) & utils::make_mask<uint8_t>(leftover_bits);
+    }
+    return;
+  }
+
+  // Unaligned case (shouldn't happen)
+  SIRIUS_LOG_DEBUG(
```
felipeblazing: cant we use some kind of padding to prevent this from ever happening? in what case would this happen?

src/operator/result/host_table_chunk_reader.cpp
```diff
@@ -0,0 +1,261 @@
...
+  // NULL case
+  if (null_count != 0) {
+    auto& validity = duckdb::FlatVector::Validity(vector);
+    copy_mask_to_validity(validity, row_offset, count, allocation);
+
+    // Get the current offset
+    auto start = offset_accessor.get_current(allocation);
+
+    // Copy each string individually
+    for (size_t i = 0; i < count; ++i) {
```
felipeblazing: Why are we copying over strings one at a time this feels VERY slow.

felipeblazing (changes requested):

src/operator/result/host_table_chunk_reader.cpp
bwyogatama: @kevkrist We should be able to avoid this here. This copy is inevitable from DuckDB format to cuDF format, but can be avoided in the other direction.

I think even in the legacy Sirius we already didn't do this because it's too slow to copy one string at a time.

We just need to copy the entire strings buffer, and construct DuckDB strings one by one.

src/include/memory/host_table_utils.hpp
kevkrist: Scale is a field of `cudf::data_type` and so is captured by `serialized_column` here [libcudf types.hpp](https://github.com/rapidsai/cudf/blob/d63d978bf949e278a650a829b85c0744e52d60b0/cpp/include/cudf/types.hpp#L320).

But I do implicitly *expect* in this PR that the scale of the cudf type and the corresponding duckdb type are the same, since scale is preserved in the original conversion from duckdb to cudf and via casts in the expression executor. Should I add checks to the constructor of `host_table_chunk_reader` that confirm that the cudf and duckdb types match? I think this is OK as-is for now and can leave as TODO for when we have type converter registry as you mention below, but I am happy to add it now.

src/include/memory/host_table_utils.hpp
kevkrist: Yes, the `serialized_column` struct is in a hidden namespace :( [here](https://github.com/rapidsai/cudf/blob/d63d978bf949e278a650a829b85c0744e52d60b0/cpp/src/copying/pack.cpp#L167).

```cpp
struct serialized_column {
  serialized_column() = default;

  serialized_column(data_type _type,
                    size_type _size,
                    size_type _null_count,
                    int64_t _data_offset,
                    int64_t _null_mask_offset,
                    size_type _num_children)
    : type(_type),
      size(_size),
      null_count(_null_count),
      data_offset(_data_offset),
      null_mask_offset(_null_mask_offset),
      num_children(_num_children)

  {
  }

  data_type type;
  size_type size{};
  size_type null_count{};
  int64_t data_offset{};       // offset into contiguous data buffer, or -1 if column data is null
  int64_t null_mask_offset{};  // offset into contiguous data buffer, or -1 if column data is null
  size_type num_children{};
  // Explicitly pad to avoid uninitialized padding bits, allowing `serialized_column` to be bit-wise
  // comparable
  int pad{};
};
```

src/include/memory/multiple_blocks_allocation_accessor.hpp
kevkrist: They're used for `set_cursor()` and `advance()` methods. I could, wherever those are used, pass a const reference to the allocation. Let me know if you'd like me to do that and then we can erase these fields.

src/include/memory/host_table_utils.hpp
kevkrist: Yeah -- I've copied the c docs from the `cudf` source (see above).

src/operator/result/host_table_chunk_reader.cpp
kevkrist: This will never happen, since the chunk reader produces vectors of 2048 tuples until it hits the end of the data chunk, in which case it may produce a vector of fewer tuples of course. I'll delete this code and add an assertion to state the invariant on the starting row offset.

src/operator/gpu_physical_result_collector.cpp
kevkrist: Done.

src/include/operator/result/host_table_chunk_reader.hpp
kevkrist: Yes correct.

src/operator/result/host_table_chunk_reader.cpp
kevkrist: Yeah, I don't like this either, and the only reason I wasn't worried about it for now is because all TPC-H result sets are small, and I am going for the quickest/easiest working solution. As @felipeblazing mentioned above, ideally we would have a representation converter registry and a custom representation. [PR 113](https://github.com/sirius-db/sirius/pull/113) develops a custom representation along these lines, but the approach in this PR felt much simpler to me.

@felipeblazing The issue results from the duckdb string representation, which varies based on the string length, making bulk copies difficult:
```cpp
	union {
		struct {
			uint32_t length;
			char prefix[4];
			char *ptr;
		} pointer;
		struct {
			uint32_t length;
			char inlined[12];
		} inlined;
	} value;
```

@bwyogatama I see that legacy Sirius did that, but I can't use pointers to the `multiple_blocks_allocation` because the host representation, from what I can tell, can go out of scope, and the underlying memory can thus be released, after `sink()` is invoked and potentially before duckdb is done using the data chunks. Am I wrong about that? I think that this issue may also apply to PR 113. Also, a duckdb `string_t` needs to point to a contiguous buffer, which a `multiple_blocks_allocation` in a `host_table_representation` will not in general satisfy. PR 113 took some care to avoid this issue at the expense of some fragmentation in the `multiple_blocks_allocation` and a restriction to strings fewer than 2MB.  

We could alternatively copy all the string data into a buffer owned by the data chunk, and then construct the DuckDB strings one by one, which may actually be what @bwyogatama suggested in the first place. Another alternative is to abandon this PR and finish PR 113, assuming the data representation is guaranteed to continue to exist for the lifetime of the data chunk, about which I am doubtful, since in the arrow -> duckdb converter the memory ownership needs to be registered in the vector [[see here](https://github.com/duckdb/duckdb/blob/9612b5bea5a6df924daf5ce696d6992df2483bfe/src/function/table/arrow_conversion.cpp#L840)].

I'll wait to gather your thoughts before I make any changes.

src/operator/gpu_physical_result_collector.cpp
kevkrist: I added a note to the method definition in the .cpp file.

src/operator/gpu_physical_result_collector.cpp
kevkrist: Changed.

src/include/memory/multiple_blocks_allocation_accessor.hpp
kevkrist: I agree that single value access should be discouraged, but there are a couple of places where I think it's essentially unavoidable. Here are some examples.
1) Converting duckdb strings to arrow-style strings. We have to perform a running prefix-sum over the duckdb string lengths to get the offsets ([here](https://github.com/sirius-db/sirius/blob/dev/src/op/scan/duckdb_scan_task.cpp#L238)) and do the reverse when converting back to duckdb [[here](https://github.com/sirius-db/sirius/blob/kevin/result_collector/src/operator/result/host_table_chunk_reader.cpp#L183)]. We could alternatively build the offsets in a local buffer and then bulk copy to the `multiple_blocks_allocation` if you'd prefer, but I don't know that that would be any faster.
2) There are places where I could replace a loop with `set_current()` and `advance()` with a `memset` function [[example](https://github.com/sirius-db/sirius/blob/kevin/result_collector/src/operator/scan/duckdb_scan_task.cpp#L136)], which I am happy to do.
3) When there are trailing bits to set in the null mask (if the row count is not divisible by 8), I use `set_current()` to only set the desired number of bits in the mask and not advance the cursor (because I will need to set more bits in that byte).
3) If the incoming vectors from the table scan have size not divisible by 8 (the 'unaligned' case), then when building the null mask for a data batch column, we have to stitch together the trailing bits from one vector (say 3) with the leading bits from the next (5 bits from the first byte), which is a REAL headache, since every succeeding byte of the null mask for that vector then needs to be stitched together individually [[here](https://github.com/sirius-db/sirius/blob/kevin/result_collector/src/operator/scan/duckdb_scan_task.cpp#L189)]. I WISH there were a way around this, and if you have any thoughts, I'm open to them.

src/include/memory/multiple_blocks_allocation_accessor.hpp
kevkrist: The idea was to simplify looping over elements of the multiple blocks allocation by maintaining a cursor into the block index and the byte offset in the block without the user having to worry about e.g. changing the block index, resetting the byte offset, etc.

src/include/memory/multiple_blocks_allocation_accessor.hpp
kevkrist: I have done 2.

src/operator/result/host_table_chunk_reader.cpp
kevkrist: I have changed this to follow more closely the string conversion duckdb->arrow logic [here](https://github.com/duckdb/duckdb/blob/9612b5bea5a6df924daf5ce696d6992df2483bfe/src/function/table/arrow_conversion.cpp#L332): 

I use duckdb to allocate a buffer to bulk copy the string data into. I still then need to loop over the offsets to collect successive differences to determine the string length for each string.

felipeblazing (commented):
looks good to go. Update it again and we can merge it in!

PR #162, opened 2026-01-15
migrated to using new version of cucascade

felipeblazing:


src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -16,13 +16,84 @@
...
+  const std::shared_ptr<cucascade::data_batch>& batch,
+  const cucascade::memory::memory_space* requested_memory_space)
+{
+  const auto* target_space =
+    requested_memory_space != nullptr ? requested_memory_space : batch->get_memory_space();
+  if (target_space == nullptr) { return std::nullopt; }
+
+  auto lock_result = batch->try_to_lock_for_processing(target_space->get_id());
+
+  auto cancel_task_if_needed = []() {};
```
joosthooz, line 44: What is this for, is it a placeholder?

src/downgrade/downgrade_task.cpp
```diff
@@ -43,19 +43,14 @@ void downgrade_task::execute()
...
   }
 
-  // Try to acquire downgrade lock - if batch is being processed, we can't downgrade
-  if (!batch->try_to_lock_for_downgrade()) {
-    // Batch is currently being processed, skip downgrade for now
+  // Try to acquire an in-transit lock - if batch is being processed, we can't downgrade
+  if (!batch->try_to_lock_for_in_transit()) {
+    // Batch is currently being processed or moving, skip downgrade for now
     // The scheduler can retry later
     mark_task_completion();
```
joosthooz, line 50: Wouldn't we want to signal something else to show that the downgrade did not proceed?

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -16,13 +16,84 @@
...
+
+  const bool needs_conversion =
+    requested_memory_space != nullptr &&
+    lock_result.status == cucascade::lock_for_processing_status::memory_space_mismatch;
+
+  if (!lock_result.success && needs_conversion) {
```
joosthooz, line 50: Is the combination `lock.success == true` and `needs_conversion == true` not possible? Maybe create an outer `if (needs_conversion)`

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -45,14 +116,19 @@ void gpu_pipeline_task::execute()
...
+  const auto* requested_memory_space =
+    reservation != nullptr ? &reservation->get_memory_space() : nullptr;
+  std::vector<cucascade::data_batch_processing_handle> processing_handles;
+  processing_handles.reserve(local_state._batches.size());
+
+  for (const auto& batch : local_state._batches) {
+    auto handle = lock_or_prepare_batch(batch, requested_memory_space);
+    if (!handle) {
+      // Failed to lock (or convert) one of the batches. Caller can retry later.
+      return;
```
joosthooz, line 129: How does the caller know the task was not executed?

src/pipeline/gpu_pipeline_task.cpp
felipeblazing: yes i dont think we will need it but i thought we might need a callback for this someday for systems like rapidsmpf.

src/pipeline/gpu_pipeline_task.cpp
felipeblazing: it is not possible if it needed conversion lock would not succeed

src/downgrade/downgrade_task.cpp
felipeblazing: it can just queue up something else this wasn't moved out of its container

src/pipeline/gpu_pipeline_task.cpp
felipeblazing: this method is not fully implemented yet lock or prepare batch is step 1 of the instructions below. this whole method will be fixed in another pr

src/gpu_physical_operator.cpp
```diff
@@ -262,9 +262,8 @@ std::vector<::std::shared_ptr<::cucascade::data_batch>> GPUPhysicalOperator::get
   std::vector<::std::shared_ptr<::cucascade::data_batch>> input_batch;
   for (auto& [port_name, port_ptr] : ports) {
     // For Pipeline barrier: need at least one data batch in the port's repository
-    // TODO: later on we will adjust to the new data repository interface in cuCascade
-    auto batch_and_handle = port_ptr->repo->pull_data_batch();
-    input_batch.push_back(std::move(batch_and_handle.first));
+    auto batch = port_ptr->repo->pull_data_batch(::cucascade::batch_state::task_created);
+    if (batch) { input_batch.push_back(std::move(batch)); }
```
joosthooz, line 266: Not a problem specific to this PR, but if one of these ports is empty or encounters a batche that is being downgraded, it will not be included in the returned vector here, so the task that is being [created in the task_creator](https://github.com/sirius-db/sirius/blob/eb91d7c7954b65eb7e77af18fe7de3a727015669/src/creator/task_creator.cpp#L193) will not have all of its inputs. How to deal with that?

bwyogatama:
Instead of modifying `gpu_physical_operator.cpp`, you should know modify `sirius_physical_operator.cpp` instead

PR #167, opened 2026-01-16
Feature/adding new physical operators

felipeblazing:
Started adding execute methods to the physical operators in the op directory. 

Implemented:
Limit
Filter
ungrouped aggregation
topn
projection

topn and ungrouped aggregation both require state. I need to make sure that it is available to all the gpus scheduling tasks so that is going to need some work

things left to do:
1. need to make sure we lock the global state in topn and ungrouped aggregation to prevent race conditions
2. need to make topn and ungrouped aggregation available on multiple gpus, we might do something like let each gpu accumulate its own state and then reconcile them all

### DIFF
--- a/src/include/op/sirius_physical_operator.hpp
+++ b/src/include/op/sirius_physical_operator.hpp
@@ -21,6 +21,7 @@
 #include "duckdb/common/enums/operator_result_type.hpp"
 #include "duckdb/common/enums/order_preservation_type.hpp"
 #include "duckdb/common/enums/physical_operator_type.hpp"
+#include "duckdb/common/exception.hpp"
 #include "duckdb/common/optional_idx.hpp"
 #include "duckdb/common/types/data_chunk.hpp"
 #include "duckdb/execution/execution_context.hpp"
@@ -31,6 +32,7 @@
 
 #include <cucascade/data/data_batch.hpp>
 #include <cucascade/data/data_repository.hpp>
+#include <cucascade/data/data_repository_manager.hpp>
 
 namespace duckdb {
 class GPUExecutor;
@@ -67,9 +69,17 @@ class sirius_physical_operator {
  public:
   sirius_physical_operator(duckdb::PhysicalOperatorType type,
                            duckdb::vector<duckdb::LogicalType> types,
-                           duckdb::idx_t estimated_cardinality)
-    : type(type), types(std::move(types)), estimated_cardinality(estimated_cardinality)
+                           duckdb::idx_t estimated_cardinality,
+                           ::cucascade::shared_data_repository_manager* data_repo_mgr = nullptr)
+    : type(type),
+      types(std::move(types)),
+      estimated_cardinality(estimated_cardinality),
+      data_repo_mgr(data_repo_mgr)
   {
+    if (data_repo_mgr == nullptr) {
+      throw duckdb::InternalException(
+        "sirius_physical_operator requires a non-null data_repository_manager");
+    }
   }
   sirius_physical_operator() = default;
 
@@ -225,6 +235,16 @@ class sirius_physical_operator {
   std::vector<std::pair<sirius_physical_operator*, std::string_view>> next_port_after_sink;
   //! The creator of the task
   creator::task_creator* creator;
+
+  //! Shared data repository manager (owned upstream, propagated to operators)
+  ::cucascade::shared_data_repository_manager* data_repo_mgr = nullptr;
+
+ public:
+  ::cucascade::shared_data_repository_manager* get_data_repository_manager() const
+  {
+    return data_repo_mgr;
+  }
+
 };
 
 }  // namespace op
--- a/src/include/op/sirius_physical_operator.hpp
+++ b/src/include/op/sirius_physical_operator.hpp
@@ -70,7 +70,7 @@ class sirius_physical_operator {
   sirius_physical_operator(duckdb::PhysicalOperatorType type,
                            duckdb::vector<duckdb::LogicalType> types,
                            duckdb::idx_t estimated_cardinality,
-                           ::cucascade::shared_data_repository_manager* data_repo_mgr = nullptr)
+                           cucascade::shared_data_repository_manager* data_repo_mgr = nullptr)
     : type(type),
       types(std::move(types)),
       estimated_cardinality(estimated_cardinality),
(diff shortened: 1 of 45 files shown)


src/include/op/sirius_physical_operator.hpp
```diff
@@ -223,8 +233,18 @@ class sirius_physical_operator {
   std::unordered_map<std::string, std::unique_ptr<port>> ports;
   //! The next operators to be executed after this operator when it is used as a sink
   std::vector<std::pair<sirius_physical_operator*, std::string_view>> next_port_after_sink;
-  //! The creator of the task
+  //! The creator of the task<<<<<<< HEAD
```
wmalpica: there was a commit merge issue here

src/op/sirius_physical_limit.cpp
```diff
@@ -28,14 +32,74 @@ sirius_physical_streaming_limit::sirius_physical_streaming_limit(
...
+    if (slices.empty()) { continue; }
+
+    // cudf::slice returns a vector of table_views; materialize into a table
+    auto sliced_table = std::make_unique<cudf::table>(slices.front());
+
+    std::unique_ptr<cucascade::idata_representation> output_data =
+      std::make_unique<cucascade::gpu_table_representation>(std::move(*sliced_table),
+                                                            *batch->get_memory_space());
+
+    auto const batch_id = data_repo_mgr->get_next_data_batch_id();
```
wmalpica: I think we want to use this: https://github.com/sirius-db/sirius/blob/dev/src/include/data/data_batch_utils.hpp#L43 to get next batch id. Which means we dont need data_repo_mgr to be passed into all the physical operators.
ALSO, we need to remote `get_next_data_batch_id()` function from data repo manager. We cant have two separate counters, otherwise the batch_ids will not be unique. We have to pick just one

wmalpica: This would also mean that `GpuExpressionExecutor` does not need the data_repository_manager

wmalpica (changes requested):

src/op/sirius_physical_top_n.cpp
```diff
@@ -34,21 +45,149 @@ sirius_physical_top_n::sirius_physical_top_n(
...
+      batch->get_data()->cast<cucascade::gpu_table_representation>().get_table());
+    concat_views.push_back(table->view());
+    owned_tables.push_back(std::move(table));
+  }
+
+  if (concat_views.empty()) { return {}; }
+
+  std::unique_ptr<cudf::table> combined;
+  if (concat_views.size() == 1) {
+    combined = std::make_unique<cudf::table>(concat_views.front());
```
wmalpica: Here if you only have a single input batch, you will end up making a copy. 
You should instead check if you have a single input_batch and handle that case before going into the view collection and concatenation path

src/op/sirius_physical_top_n.cpp
```diff
@@ -34,21 +45,149 @@ sirius_physical_top_n::sirius_physical_top_n(
...
+    combined = cudf::concatenate(concat_views);
+  }
+
+  auto combined_view = combined->view();
+
+  std::unique_ptr<cudf::table> kept;
+  auto const keep_rows = std::min<cudf::size_type>(combined_view.num_rows(),
+                                                   static_cast<cudf::size_type>(offset + limit));
+
+  if (orders.size() == 1) {
```
wmalpica, line 66: I think for a lot, or all of this logic, we want to use what we already have in `gpu_order_impl::local_top_n` and if that function is missing something we should fix it there. that it why we have it,

src/op/sirius_physical_ungrouped_aggregate.cpp
```diff
@@ -38,5 +58,458 @@ sirius_physical_ungrouped_aggregate::sirius_physical_ungrouped_aggregate(
...
+    auto view  = table.view();
+
+    for (idx_t i = 0; i < aggregates.size(); ++i) {
+      auto& agg = aggregates[i]->Cast<duckdb::BoundAggregateExpression>();
+      if (agg.IsDistinct()) {
+        throw duckdb::NotImplementedException("Distinct aggregates not supported in GPU path yet");
+      }
+      auto fname = agg.function.name;
+
+      if (fname == "count_star") {
```
wmalpica: I think we want to use `gpu_aggregate_impl::local_ungrouped_aggregate`.  In general this code is very repetitive. You probably want a block of code that figures out the index, output type and aggregation type and only have one cudf::reduce

test/cpp/operator/operator_test_utils.hpp
```diff
@@ -0,0 +1,400 @@
...
+
+template <typename TFirst, typename TSecond>
+inline std::shared_ptr<cucascade::data_batch> make_two_column_batch(
+  data_repository_mgr& repo_mgr,
+  cucascade::memory::memory_space& space,
+  const std::vector<TFirst>& col0_values,
+  const std::vector<TSecond>& col1_values,
+  cudf::type_id col1_type_id,
+  std::optional<int32_t> decimal_scale      = std::nullopt,
+  std::optional<cudf::type_id> col0_type_id = std::nullopt)
```
wmalpica, line 299: This API is funky. You have two template types, but then you are making the col0 type id optional and as the last parameter? If we are making testing utilities, we should make them with the intention of good reusability, otherwise, just put this in the test file where you are using it

wmalpica (changes requested):

test/cpp/operator/operator_test_utils.hpp
felipeblazing: What is werid about this being optional. If you don't set this it will assume an int64. I don't see why its not ok to all them to make a two column batch without having to specify both column types. The templated types are the C++ primitive types that store the data in the vector for initialization.

src/op/sirius_physical_top_n.cpp
felipeblazing: this api takes in a const std vector so we would have to copy. Technically I guess I could change ALL of the execute methods but that is out of scope for this pr

src/op/sirius_physical_top_n.cpp
felipeblazing: I can move that method into here but it should not exist in a separate file if it is not being used elsewhere.

src/op/sirius_physical_top_n.cpp
wmalpica: What you can do to avoid making a copy is that if `input_batches.size() == 1` then `combined_view = input_batches[0]->get_data()->cast<cucascade::gpu_table_representation>().get_table().view()`

wmalpica (approved):

Issue #56, opened 2025-07-07
A IS NOT DISTINCT FROM B  vs A = B


bwyogatama:
`A IS NOT DISTINCT FROM B`  is currently handled the same way as `A = B`, while it should be handled differently. 
`A IS NOT DISTINCT FROM B` returns `True` when both A and B are `NULL`, while `A = B` returns False when both A and B are `NULL`.

mbrobbel:
Closed via #170

PR #170, opened 2026-01-17
Add support for IS NOT DISTINCT FROM

Darmin-Geek:
Resolves Issue #56
* Adds support for IS NOT DISTINCT FROM comparisons
* Adds sql tests for IS NOT DISTINCT FROM


src/expression_executor/specializations/gpu_execute_comparison.cpp
```diff
@@ -222,63 +227,107 @@ struct ComparisonDispatcher {
...
     // If the right side is a constant, do not materialize in a column
-    if (expr.right->GetExpressionClass() == ExpressionClass::BOUND_CONSTANT) {
-      auto right_value = expr.right->Cast<BoundConstantExpression>().value;
+    if (right_expr->GetExpressionClass() == ExpressionClass::BOUND_CONSTANT) {
+      auto right_value = right_expr->Cast<BoundConstantExpression>().value;
 
-      switch (GetCudfType(expr.right->return_type).id()) {
+      switch (GetCudfType(right_expr->return_type).id()) {
         case cudf::type_id::INT16:
           return DoScalarComparison<int16_t>(
```
bwyogatama, line 265: This is great but I believe there is a bug here.

What if the expression is something like `0 IS NOT DISTINCT FROM NULL`

If you treat NULL as 0 in this case it will return True, where it should return False

bwyogatama:
Once you fix that, and the CI is all green, we can merge this

src/expression_executor/specializations/gpu_execute_comparison.cpp
Darmin-Geek: Hi, through testing and my own static analysis, it seems to me that there is no bug there. In my testing, as shown below, ```0 IS NOT DISTINCT FROM NULL``` returns false.

The reason why the code works is that in the call to the constructor of numeric_scalar for the right hand side, I set the is_valid parameter equal to false if and only if the scalar is NULL. The cudf NULL_EQUALS operator must check the validity of the data before accessing it. This is because using the value of a null/invalid scalar causes undefined behavior. (https://docs.rapids.ai/api/libcudf/stable/classcudf_1_1scalar.html#a1a1fd922cacc88449792ed329ed3ebdb)

Cudf will see that the left hand side (0) is valid/not null and the right hand side is invalid/null and return false. As a result, the value of the scalar has no effect if the scalar is null/invalid. 


Since the call to the constructor of numeric_scalar requires a value, and doing GetValue causes issues when reading from NULL, I pass a default value if and only if the scalar is NULL. Thankfully, this default value is ignored because is_valid is set to false. In all, there is no bug. Does this answer your question?

image

Thank you for your help.

Darmin-Geek:
Commit [3334fd6](https://github.com/sirius-db/sirius/pull/170/commits/3334fd61a8b309f6e37ee2403883c0c86b20587b) failed CI on the step where duckdb is linked (only fails on arm+release). The reason it fails appears to be that the operation times out (or is manually canceled). I merged the outstanding commits in the hope of decreasing cache misses. _If you have any advice, please let me know._

Darmin-Geek:
Hi @bwyogatama, thank you for running the GitHub action.  As I explained, the "0 to True" bug that appeared to exist does not actually exist. Is there anything else I should do before you merge the pull request?

EDIT: I am currently testing with gpu_execution instead of gpu_processing.

bwyogatama (approved):

PR #171, opened 2026-01-19
Add JIT-based REGEX_REPLACE via regex interpreter and cuDF JIT Transform

madmtang:
## Summary
This PR adds a new implementation of `REGEX_REPLACE` in Sirius that uses a custom regex interpreter to dynamically generate cuDF JIT Transform UDFs. The goal is to eliminate the overhead of libcudf’s general-purpose GPU regex engine for a large, structured subset of regex patterns and significantly improve performance on regex-heavy queries (for example, ClickBench Q28).

## Motivation
Sirius currently implements `REGEX_REPLACE` using libcudf regex functions, which exhibit high overhead due to interpreter-style execution, warp divergence, and instruction pressure on the GPU. This causes Sirius to perform substantially worse than DuckDB on queries involving regex replacement, despite Sirius’ otherwise strong GPU performance.

Prior proof-of-concept work showed that a hand-written JIT UDF for ClickBench Q28 could outperform DuckDB by a large margin, but that approach was not general. This PR generalizes that idea by automatically generating specialized JIT kernels from regex patterns at runtime.

## Design and Implementation
This PR introduces three main components:

1. **Regex Interpreter**
   - Parses a regex pattern into an AST (Sequence, Group, Quantified, Literal).
   - Transpiles the AST into CUDA C++ device code compatible with the libcudf JIT Transform API.
   - Generates pattern-specific kernels that avoid general regex interpreter loops, reducing branching and warp divergence.

2. **Supported Regex Subset**
   - Anchors: `^` at the beginning and `$` at the end (both required).
   - Literals, escaped characters, and `.`.
   - Negated character classes with restrictions under `+` and `*`.
   - Capturing and non-capturing groups (no nested groups).
   - Quantifiers `?`, `+`, `*` with constrained atom types.
   - Replacement supports extracting the first capture group (`\1`); unsupported replacements fall back to returning the input unchanged.

3. **UDF Caching**
   - A thread-safe cache keyed by `(pattern, replacement)` avoids regenerating and recompiling identical JIT UDFs.
   - Amortizes NVRTC compilation cost across repeated queries.

The rest of the query plan and execution pipeline is unchanged; only the execution of `REGEX_REPLACE` is modified.

## Performance Impact
Using a custom benchmark (RegexBench, based on ClickBench data), this implementation:
- Achieves up to **6× speedup** over the Sirius baseline for supported regex patterns.
- Reduces total query runtime almost entirely by reducing `REGEX_REPLACE` execution time.
- Outperforms both the Sirius baseline and DuckDB on all RegexBench queries.
- Improves geometric mean runtime by approximately **4.6×** relative to the Sirius baseline.

## Limitations
- Only a subset of regex syntax is supported.
- Replacement is limited to the first capture group (`\1`).
- Complex patterns requiring heavy looping or branching are not suitable for this approach and should fall back to existing implementations.


bwyogatama:
Thanks for the contribution @madmtang. Seems like some of the test still doesn't work here, can you make sure it works before we start the revirew?

mbrobbel:
@madmtang marking this as draft because there are some conflicts.

mtervo:
@bwyogatama @mbrobbel @madmtang, we brought the branch up-to-date with the latest version of dev. It's building, and all the unit tests pass. I believe we are ready for review. Thanks!

wmalpica:
@madmtang Thank you for your effort here, but we are planning on deprecating the legacy version of Sirius soon.

PR #191, opened 2026-01-26
build error for missing noexcept

awitten1:
Hi! Thanks for building this, it's really cool.

I was trying to build Sirius today and got the following compile error:
```
/home/awitt/code/sirius/src/memory/sirius_memory_reservation_manager.cpp:69:8: error: looser exception specification on overriding virtual function ‘virtual void sirius::memory::sirius_device_memory_resource::do_deallocate(void*, std::size_t, rmm::cuda_stream_view)’
   69 |   void do_deallocate(void* ptr, std::size_t bytes, rmm::cuda_stream_view stream) override
```

It looks like do_deallocate needs to be marked noexcept, so this PR adds noexcept to that function.  After this change everything built for me!

### DIFF
--- a/src/memory/sirius_memory_reservation_manager.cpp
+++ b/src/memory/sirius_memory_reservation_manager.cpp
@@ -66,14 +66,11 @@ class sirius_device_memory_resource : public rmm::mr::device_memory_resource {
     return it->second->allocate(bytes, stream);
   }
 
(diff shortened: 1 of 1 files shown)


bwyogatama:
@aminaramoon Are you good with these changes? I faced the same build issue and apparently it's because in librmm 25.12, they add `noexcept` to `rmm::device_memory_resource`

bwyogatama (approved):

awitten1:
Great thank you!

PR #194, opened 2026-01-27
WIP: operator_data concept

wmalpica:
This PR is to demonstrate the `operator_data` class concept, which is a wrapper for a `std::vector<std::shared_ptr<cucascade::data_batch>` that allows for adding operator dependent metadata to a batch that is type-erasable.


wmalpica:
This is getting implemented elsewhere

PR #195, opened 2026-01-28
Feature: GpuPipeline Executor, DuckDBScanExecutor and more

aminaramoon:
In this PR:

1. new utilties for building executors: 
 - kiosk: a booking manager that can be used for controlling number of concurrent ops, as well as waiting for all of them ot finish
 2. thread-pool: exactly what it sounds like
 3. channel: a thread-safe queue with producer/consumer model that can be useful for breaking down dependcies
 4. interruptible mpmc queue
 5. gpu pipeline executor 
 6. duckdb scan executor
 7. managing the end of like of pipelines, when they are closed
 8 utilties for starting adn ending queries


src/include/exec/kiosk.hpp
```diff
@@ -0,0 +1,109 @@
...
+#include <atomic>
+#include <chrono>
+#include <condition_variable>
+#include <memory>
+#include <mutex>
+#include <optional>
+
+namespace booking {
+
+class kiosk;
```
wmalpica, line 27: where is this class used?

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -129,7 +90,13 @@ void gpu_pipeline_executor::worker_loop(int worker_id)
       // Executor is stopped.
       break;
     }
-    auto task = _task_queue->pull();
+    auto task    = _task_queue->pull();
+    auto cleanup = absl::MakeCleanup([this]() {
+      std::unique_lock<std::mutex> lock(_task_count_mutex);
+      _num_active_requests--;
```
wmalpica: Is it guaranteed that you will have an active request on each thread? 
Does it make sense to just zero out _num_active_requests instead of decrementing it once per thread?

bwyogatama (approved):

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -473,6 +474,12 @@ void duckdb_scan_task::process_chunk(duckdb_scan_task_local_state& l_state)
 }
 
 void duckdb_scan_task::execute()
+{
+  auto output_batches = compute_task();
+  publish_output(output_batches);
+}
```
joosthooz, line 480: This implementation is the same as in parent sirius_pipeline_itask::execute() so it doesn't need to be overridden (although it is more clear navigating through the code). If you want to keep it, I think it is missing in the header, it should be there with override specifier.

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -518,11 +525,23 @@ void duckdb_scan_task::execute()
...
+
+  return std::vector<std::shared_ptr<cucascade::data_batch>>{};
+}
+
+void duckdb_scan_task::publish_output(
+  std::vector<std::shared_ptr<cucascade::data_batch>> output_batches)
+{
+  std::for_each(std::make_move_iterator(output_batches.begin()),
+                std::make_move_iterator(output_batches.end()),
+                [this](auto batch) { this->_data_repo->add_data_batch(std::move(batch)); });
```
joosthooz, line 544: Is this suitable as a default implementation in `gpu_pipeline_task`? Is this class going to be a subclass of that or will it stay a subclass of `sirius_pipeline_itask`?

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -16,158 +16,95 @@
...
+    auto output_consumers = gpu_task->get_output_consumers();
+    _thread_pool->schedule([this,
+                            task      = std::move(pipeline_task),
+                            ticket    = std::move(ticket),
+                            consumers = std::move(output_consumers)]() mutable {
+      task->execute();
+      task.reset();
+      if (_schedule_callback) {
+        for (auto* consumer : consumers) {
+          _schedule_callback(consumer);
```
joosthooz: Is this callback going to be responsible to schedule tasks for consumers?

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -518,11 +525,23 @@ void duckdb_scan_task::execute()
       std::static_pointer_cast<duckdb_scan_task_global_state>(this->_global_state);
     auto next_task = std::make_unique<duckdb_scan_task>(
       new_task_id, _data_repo, std::move(new_local_state), shared_global_state);
-    g_state._scan_executor.schedule(std::move(next_task));
+    g_state._pipeline_executor.schedule(std::move(next_task));
```
joosthooz, line 528: Do we want to keep this follow-up task creation here for now?

joosthooz (approved):
Small comments and questions, but if you're happy I'm ok with merging

src/pipeline/pipeline_executor.cpp
```diff
@@ -17,131 +17,166 @@
...
-  on_stop();
-  for (auto& thread : _threads) {
-    if (thread.joinable()) { thread.join(); }
-  }
-  _threads.clear();
+  if (_management_thread.joinable()) { _management_thread.join(); }
 }
 
-void pipeline_executor::worker_loop(int worker_id)
+void pipeline_executor::set_task_creator(sirius::creator::task_creator& task_creator)
```
joosthooz, line 104: Is this still relevant? The task creator code is commented out in this PR, are we planning to remove it?

bwyogatama (approved):

PR #197, opened 2026-01-28
Define Sirius Pipeline Task APIS for processing scan and pipeline tasks

aminaramoon:
removes task downgrade creator
removes implementaiton of task creator 
 - defines apis for receivign a request 
- defines task apis for setting, releaseing reservation
- reporting peak consumption
- getting conumer operators for task creator
-


src/pipeline/sirius_pipeline.cpp
```diff
@@ -165,6 +165,17 @@ sirius_pipeline::get_inner_operators()
   return result;
 }
 
+std::vector<sirius_pipeline*> sirius_pipeline::get_parents()
```
bwyogatama, line 168: So unfortunately right now I don't think the parents are accurate after I modified the DuckDB plan. So this wouldn't work until I do something about it. Can you mark it as a TODO for me?

src/op/sirius_physical_operator.cpp
```diff
@@ -223,29 +226,60 @@ sirius_physical_operator::get_next_port_after_sink()
...
-      if (!port_ptr->src_pipeline->is_pipeline_finished()) {
-        // Src pipeline not finished, return it to continue processing
-        return creator::task_creation_hint(port_ptr->src_pipeline);
+    } else {
+      auto* producer = &(unfinished_barrier->second->src_pipeline->get_inner_operators()[0].get());
+      return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, producer};
+    }
+  }
+
+  // if they are complete, create data if you can
```
bwyogatama, line 253: I am not sure I am following this part, so please correct me if I misunderstood something here. To my understanding this code is saying,

_If at least one of the ports has at least one input batch and they are not `MemoryBarrierType::FULL`, then we can schedule the task for this pipeline_

A couple questions:
1. What if all the ports have `MemoryBarrierType::FULL`? Then we can never schedule the task for this pipeline?
2. Shouldn't we schedule the task for this pipeline when all the ports have inputs? Using `all_of` instead of `any_of`

wmalpica: I spoke with @bwyogatama and he was right, the fix is implemented in another branch

wmalpica:
Closing this PR because its contents are included in https://github.com/sirius-db/sirius/pull/195

PR #200, opened 2026-02-02
Implement execute() and some fixes to scans

joosthooz:
This PR adds some missing links after recent changes, and it attempts to schedule gpu_pipeline_tasks instead of individual operators. But that is still WIP for scans.
Scan tasks are already being scheduled, but do not finish yet (I'm troubleshooting this).
Then, another missing link is to schedule a pipeline task that follows a scan pipeline (this should be a pipeline with a single DUCKDB_SCAN operator, otherwise we will also need to add handling multiple operators in a scan pipeline).


bwyogatama:
Closing this branch as the commits are already included in #206

PR #202, opened 2026-02-02
Parquet Scan Task

kevkrist:
This PR will implement the functionality to perform table scans on parquet files in the new execution model. It specifically tries to perform the role of the `byte-range preloader` described in the [Theseus](https://arxiv.org/html/2508.05029v1#S3.SS4) paper. It uses the cudf `hybrid_scan_reader` ([docs](https://docs.rapids.ai/api/libcudf/nightly/classcudf_1_1io_1_1parquet_1_1experimental_1_1hybrid__scan__reader)), which essentially provides all the needed functionality. This reader is great because it also supports filter pushdown and row-group pruning (zone maps, Bloom filters), etc., which are left as future work. Moreover, this reader doesn't require footer reconstruction for each batch, as the metadata is cached in the reader itself. 

Here are the main components:
- `host_parquet_representation`: This represents compressed data from row groups and associated column chunks whose decompressed size is approximately the target batch size. Included are a cross-HOST converter and a HOST->GPU converter, which are added to the converter registry. The HOST->GPU converter converts to a `gpu_table_representation` by decompressing the column chunks.
- `parquet_scan_task`: The `hybrid_scan_reader` is used to determine the byte ranges for the reserved row groups + column chunks and reads them into the local allocation, which is sized to the compressed size of the reserved row groups. It then pushes the `parquet_data_representation` as a data batch to the data repository. **Tasks are not rescheduled by completed tasks**.

Some advantages:
- One of the main advantages of this scan route is that there are no painful duckdb -> sirius data conversion headaches.
- The number of tasks that needs to be scheduled to exhaust the scan can be fetched from the global state after its construction by `get_num_row_group_partitions()`. A simple atomic counter is used to reserve row group ranges for tasks.
- Reservations and allocations are perfectly sized for scan tasks.
- It avoids the following issue with scan tasks scheduling subsequent tasks themselves: the reservation for the newly created task is made before the reservation for the current, nearly-complete task is released, which might lead to reservation contention.

Limitations:
- For scans with projections, nested types (e.g., STRUCT, MAP, LIST) are not currently supported


src/op/scan/parquet_scan_task.cpp
```diff
@@ -0,0 +1,473 @@
...
+/**
+ * @brief Read parquet file row groups and column chunks into local state memory.
+ *
+ * @tparam USE_FILE_CACHE Whether to use the file cache when reading column chunk data.
+ * @param g_state The global state for the parquet scan task.
+ * @param l_state The local state for the parquet scan task.
+ * @param new_meta The new file metadata to populate.I
+ */
+template <bool USE_FILE_CACHE>
+void read_parquet_file(parquet_scan_task_global_state& g_state,
```
felipeblazing: if you are just getting the bytes you will need to read cant we use 
https://github.com/apache/arrow/blob/main/cpp/src/parquet/file_reader.h#L226
GetReadRanges

kevkrist: Yeah I think that would work and would simplify some of the logic here, namely `get_chunk_min_offset` and the global state constructor. I think my goal here was to just use the native duckdb Parquet reader for simplicity.

EDIT: It might help a lot with nested data types, actually.

felipeblazing: i definitely recommend this approach since it will remove the need to understand where the bytes are

kevkrist: Ok I'll work on that.

kevkrist:
Switching back to draft as I start to incorporate cudf APIs.

src/op/scan/parquet_scan_task.cpp
kevkrist: Switched to using the CuDF `hybrid_scan_reader` which provides the same functionality.

src/include/op/scan/parquet_scan_task.hpp
```diff
@@ -0,0 +1,341 @@
...
+//===----------------------------------------------------------------------===//
+class parquet_scan_task_global_state : public parallel::itask_global_state {
+  using hybrid_scan_reader = cudf::io::parquet::experimental::hybrid_scan_reader;
+
+ public:
+  /**
+   * @brief Struct representing a range of row groups assigned to a scan task.
+   */
+  struct row_group_range {
+    size_t start_row_group;
```
aminaramoon, line 72: initialize them to something reasonable. or remove default constructor and replace it with a consturctor that initlizes all the members

src/data/host_parquet_representation_converters.cpp
```diff
@@ -0,0 +1,229 @@
...
+                                        counts.data(),
+                                        counts.size(),
+                                        &attr,
+                                        &attrs_idxs,
+                                        num_attrs,
+                                        &fail_idx,
+                                        copy_stream));
+#endif
+  };
+  if (stream.value() == nullptr) {
```
aminaramoon: default initialize the stream to `cudf::get_default_stream()`
creating streams are not cheap, also allocation on a local stream can result in segfault when you deallocate the buffers on that stream if the stream goes out of scope, which in case of a local stream, it will always go out of scope immediately

src/data/host_parquet_representation.cpp
```diff
@@ -0,0 +1,59 @@
...
+  // Allocate new blocks for the copy
+  auto allocation_copy = host_mr->allocate_multiple_blocks(_size_in_bytes);
+
+  // Copy data block by block
+  const auto& src_blocks = _column_chunks->get_blocks();
+  auto dst_blocks        = allocation_copy->get_blocks();
+  std::size_t remaining  = _size_in_bytes;
+  std::size_t block_size = _column_chunks->block_size();
+
+  for (std::size_t i = 0; i < src_blocks.size() && remaining > 0; ++i) {
```
aminaramoon, line 40: since the blocks are fixed size, we could just count the number of blocks populated with data and copy them without caring about remianing and stuff

src/data/host_parquet_representation_converters.cpp
```diff
@@ -0,0 +1,237 @@
...
+                    });
+
+  // Copy HOST data to GPU with a single async batch copy.
+  size_t bytes_copied = 0;
+  std::vector<void*> dst_ptrs;
+  std::vector<void*> src_ptrs;
+  std::vector<size_t> counts;
+  cudaMemcpyAttributes attr{};
+  attr.srcAccessOrder = cudaMemcpySrcAccessOrderStream;
+  attr.srcLocHint     = {cudaMemLocationTypeHost, host_src.get_device_id()};
```
aminaramoon: please add a comment here that host id here is `numa_id`
which means people shouldn;t bind host ids to gpu ids.

src/data/host_parquet_representation_converters.cpp
kevkrist: Per offline discussion, using per-thread default stream if the stream is null or legacy default and falling back to individual async copies.

src/include/op/scan/parquet_scan_task.hpp
kevkrist: Followed your second recommendation.

src/data/host_parquet_representation_converters.cpp
aminaramoon: I am sorruy for the confusion, you don;t need to fallback to async copies, you can do batch copy on per-thread stream

aminaramoon (approved):
LGTM! thanks @kevkrist

PR #208, opened 2026-02-06
updated to rapids 26.04

felipeblazing:
Updated to rapids 26.04
removed deprecated memory resources
changed memory resource apis

removed "using std" which conflicted with the cuda::std name space cause CUB to not be able to build

### DIFF
--- a/src/op/scan/duckdb_scan_task.cpp
+++ b/src/op/scan/duckdb_scan_task.cpp
@@ -50,8 +50,11 @@ duckdb_scan_task_global_state::duckdb_scan_task_global_state(
 {
   // Initialize global table function state
   if (_op.function.init_global) {
-    duckdb::TableFunctionInitInput tf_input(
-      _op.bind_data.get(), _op.column_ids, _op.scanned_ids, nullptr, _op.extra_info.sample_options);
+    duckdb::TableFunctionInitInput tf_input(_op.bind_data.get(),
+                                            _op.column_ids,
+                                            _op.scanned_ids,
+                                            _op.table_filters.get(),
+                                            _op.extra_info.sample_options);
     _global_tf_state = _op.function.init_global(client_ctx, tf_input);
   }
 
@@ -388,8 +391,11 @@ void duckdb_scan_task_local_state::initialize_local_table_function_state(
   // Note: local_tf_state might already be set if it was moved from a previous task
   // Only create a new one if it doesn't exist
   if (!_local_tf_state && op.function.init_local) {
-    duckdb::TableFunctionInitInput tf_input(
-      op.bind_data.get(), op.column_ids, op.projection_ids, nullptr, op.extra_info.sample_options);
+    duckdb::TableFunctionInitInput tf_input(op.bind_data.get(),
+                                            op.column_ids,
+                                            op.projection_ids,
+                                            op.table_filters.get(),
+                                            op.extra_info.sample_options);
     _local_tf_state = op.function.init_local(exec_ctx, tf_input, global_tf_state);
   }
 }
--- a/src/op/scan/duckdb_scan_task.cpp
+++ b/src/op/scan/duckdb_scan_task.cpp
@@ -49,11 +49,14 @@ duckdb_scan_task_global_state::duckdb_scan_task_global_state(
     _op(*scan_op)
 {
   // Initialize global table function state
+  // Note: We pass nullptr for table_filters because filters are applied by the Sirius physical
+  // table scan operator, not by the DuckDB table function. This ensures consistent filtering
+  // behavior and allows GPU-accelerated filter execution.
   if (_op.function.init_global) {
     duckdb::TableFunctionInitInput tf_input(_op.bind_data.get(),
                                             _op.column_ids,
                                             _op.scanned_ids,
-                                            _op.table_filters.get(),
+                                            nullptr,  // Don't pass filters to DuckDB
                                             _op.extra_info.sample_options);
     _global_tf_state = _op.function.init_global(client_ctx, tf_input);
   }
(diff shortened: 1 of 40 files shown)


bwyogatama (approved):

src/op/sirius_physical_operator.cpp
```diff
@@ -224,9 +224,6 @@ std::optional<task_creation_hint> sirius_physical_operator::get_next_task_hint()
     auto* producer = &(unfinished_barrier->second->src_pipeline->get_operators()[0].get());
     return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, producer};
   }
-  // WSM TODO: we may want to implement the status of where it says that all tasks are scheduled but
```
wmalpica, line 227: BTW this TODO that was deleted, I put into a ticket
https://github.com/sirius-db/sirius/issues/211
cc @aminaramoon

wmalpica (approved):

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -207,10 +207,12 @@ std::vector<op::sirius_physical_operator*> gpu_pipeline_task::get_output_consume
       _global_state->cast<gpu_pipeline_task_global_state>()._pipeline.get() == nullptr) {
     return output_consumers;
   }
-  auto parents =
-    _global_state->cast<gpu_pipeline_task_global_state>()._pipeline.get()->get_parents();
-  for (auto& parent : parents) {
-    output_consumers.push_back(&parent->get_operators()[0].get());
+  auto* pipeline = _global_state->cast<gpu_pipeline_task_global_state>()._pipeline.get();
+  auto sink      = pipeline->get_sink();
+  if (!sink) { return output_consumers; }
```
bwyogatama: I don't think there is a pipeline without sink ever so you can throw an error here

src/sirius_extension.cpp
```diff
@@ -89,6 +89,9 @@ struct GPUTableFunctionData : public TableFunctionData {
       DBConfig::GetConfig(context).options.disabled_optimizers;
     disabled_optimizers.insert(OptimizerType::IN_CLAUSE);
     disabled_optimizers.insert(OptimizerType::COMPRESSED_MATERIALIZATION);
+    // The Sirius scan does not apply table_filters (for caching), so keep filters
+    // as explicit LogicalFilter operators rather than pushing them into the scan.
+    disabled_optimizers.insert(OptimizerType::FILTER_PUSHDOWN);
```
bwyogatama: As discussed, let's enable the filter pushdown again

src/sirius_extension.cpp
```diff
@@ -356,6 +362,7 @@ unique_ptr<FunctionData> SiriusExtension::GPUExecutionBind(ClientContext& contex
                                                            vector<string>& names)
 {
   auto result              = make_uniq<SiriusTableFunctionData>();
+  result->conn             = make_uniq<Connection>(*context.db);
```
bwyogatama, line 359: I don't think this is ever used right?

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -50,8 +50,11 @@ duckdb_scan_task_global_state::duckdb_scan_task_global_state(
 {
   // Initialize global table function state
   if (_op.function.init_global) {
-    duckdb::TableFunctionInitInput tf_input(
-      _op.bind_data.get(), _op.column_ids, _op.scanned_ids, nullptr, _op.extra_info.sample_options);
+    duckdb::TableFunctionInitInput tf_input(_op.bind_data.get(),
+                                            _op.column_ids,
+                                            _op.scanned_ids,
+                                            _op.table_filters.get(),
```
bwyogatama: okay here the table filters should be set to nullptr on purpose

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -388,8 +391,11 @@ void duckdb_scan_task_local_state::initialize_local_table_function_state(
   // Note: local_tf_state might already be set if it was moved from a previous task
   // Only create a new one if it doesn't exist
   if (!_local_tf_state && op.function.init_local) {
-    duckdb::TableFunctionInitInput tf_input(
-      op.bind_data.get(), op.column_ids, op.projection_ids, nullptr, op.extra_info.sample_options);
+    duckdb::TableFunctionInitInput tf_input(op.bind_data.get(),
+                                            op.column_ids,
+                                            op.projection_ids,
+                                            op.table_filters.get(),
```
bwyogatama: same here, set to nullptr on purpose

bwyogatama (changes requested):
Requested changes from Bobbi

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -116,19 +116,25 @@ void gpu_pipeline_executor::manager_loop()
...
+      bool query_complete = false;
       if (_completion_handler && pipeline) {
         auto sink = pipeline->get_sink();
         if (sink && sink->type == op::SiriusPhysicalOperatorType::RESULT_COLLECTOR) {
-          if (pipeline->is_pipeline_finished()) { _completion_handler->mark_completed(); }
+          query_complete = pipeline->is_pipeline_finished();
         }
       }
+
+      if (!query_complete && _task_creator) {
```
aminaramoon, line 131: why is this needed ? RESULT_COLLECTOR doesn';t have consumers ?

src/sirius_extension.cpp
felipeblazing: its used in the GPUExecutionFunction 
```
auto gpu_physical_plan =
        GPUGeneratePhysicalPlan(context, *result->gpu_context, query_plan, *result->conn);
      auto gpu_prepared    = make_shared_ptr<GPUPreparedStatementData>(std::move(prepared),
      
```

src/pipeline/gpu_pipeline_executor.cpp
felipeblazing: you are right ican remove that

felipeblazing: actually dont see the problem, this applies to all nodes, result_collector wont have consumers anyway

felipeblazing: the if statement checking for result collector is already closed

src/sirius_extension.cpp
bwyogatama: yeah but we never go to that portion of the code

bwyogatama: in the GPUExecutionFunction, it goes to SiriusGeneratePhysicalPlan

PR #209, opened 2026-02-06
fixed how a pipeline parents are set

wmalpica:
Quick fix for how pipelines are set

### DIFF
--- a/src/sirius_engine.cpp
+++ b/src/sirius_engine.cpp
@@ -747,8 +747,10 @@ void sirius_engine::initialize_internal(op::sirius_physical_operator& plan)
       // iterate through ports at first_op
       if (new_scheduled[i]->sink.get()) {
         for (auto& [next_op, port_id] : new_scheduled[i]->sink.get()->get_next_port_after_sink()) {
-          new_scheduled[i]->parents.push_back(
-            duckdb::weak_ptr<sirius::pipeline::sirius_pipeline>(next_op->get_port(port_id)->dest_pipeline));
+          if (next_op->get_port(port_id)->dest_pipeline) {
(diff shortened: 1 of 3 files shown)


src/sirius_engine.cpp
```diff
@@ -745,8 +745,11 @@ void sirius_engine::initialize_internal(op::sirius_physical_operator& plan)
       new_scheduled[i]->parents.clear();
       auto& first_op = new_scheduled[i]->operators[0].get();
       // iterate through ports at first_op
-      for (auto port_id : first_op.get_port_ids()) {
-        new_scheduled[i]->parents.push_back(first_op.get_port(port_id)->src_pipeline);
+      if (new_scheduled[i]->sink.get()) {
+        for (auto& [next_op, port_id] : new_scheduled[i]->sink.get()->get_next_port_after_sink()) {
+          new_scheduled[i]->parents.push_back(
+            duckdb::weak_ptr<sirius::pipeline::sirius_pipeline>(next_op->get_pipeline()));
```
bwyogatama: actually using `get_pipeline()` will not work here, i think we haven't linked the operator to each pipeline yet here. 

the correct way is to use `next_op->get_port[port_id]->dst_pipeline`

bwyogatama (changes requested):

src/sirius_engine.cpp
wmalpica: ok. I will set it that way. Although, the way I had it did work

bwyogatama: That's a mistery because we for sure haven't set the pipeline yet by then. We set the pipeline after that

bwyogatama (approved):

PR #210, opened 2026-02-06
fix CustomTopNString bugs

fanruan-tuco:
```shell
select l_comment from lineitem order by l_comment limit 10;
```
the sql is extremly slower than duckdb,  then I found out:
in the custom string topN prefix sort stage, the reduction op CustomTopNMin may be not right, it causes more rows are passed to the second full key sort stage

and,  fixed typo: 
```cpp
CustomTopNStringPrefixLessThan prefix_greater_than_compartor;  
```

local test:
after this change, on tpch10 datasets with RTX3070, the hotrun time decreased from 1.5s to 50ms

### DIFF
--- a/src/cuda/cudf/cudf_orderby.cu
+++ b/src/cuda/cudf/cudf_orderby.cu
@@ -138,7 +138,7 @@ struct CustomTopNMin {
   template <typename T>
   __device__ __forceinline__ T operator()(const T& a, const T& b) const
   {
-    return a < b;
+    return a < b ? a : b;
   }
 };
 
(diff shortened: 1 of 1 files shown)


fanruan-tuco:
hi, can someone review this?

bwyogatama:
@HermesCui can you review this?

HermesCui:
Hi Fanruan @fanruan-tuco,

Thanks for fixing this bug! Before we merge your PR, could you please complete the following three steps?

Provide a before-and-after comparison showing the effect of your typo fix.

Verify that this change doesn't affect the results of other queries on TPC-H and ClickBench.

Run the unit test file to confirm correctness.

mbrobbel:
Looks like this was fixed in d78a4f3.

PR #214, opened 2026-02-06
Feature/add stream to tasks

aminaramoon:
CHANGES:
1. acquire stream in the executor
2. pass the stream to `itask::execute` and `compute` methods
3. hold extension lock in `$HOME/.sirius`
4. fix some issues in config, allowing predicate for variants and optional and iterable to check correctness of configs

### DIFF
--- a/src/include/config_option.hpp
+++ b/src/include/config_option.hpp
@@ -146,8 +146,9 @@ struct config_to_type_traits<std::string> {
 
 template <IsBackInsertableWithValue T>
 struct config_to_type_traits<T> {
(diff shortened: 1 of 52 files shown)


src/include/config_option.hpp
```diff
@@ -465,6 +453,118 @@ struct registered_config : config_base {
   bool is_required_{false};
 };
 
+// Specialized registered_config for variant types with validation
+template <typename T, typename... Args>
```
felipeblazing, line 493: lets document this method I have trouble understanding its intent

Issue #213, opened 2026-02-06
Allow using 64-bit string offsets (libcudf large string support)


joosthooz:
Currently we use 32 bit string offsets, because that seems to be the default in libcudf (it has support for both 32 and 64 bit string offsets). It also has some performance advantages especially on GPU, and reduces memory utilization for string columns.
However, it could be desirable for certain users of workload to use 64-bit string offsets instead, so we'd like to add support for this at some point.

bwyogatama:
duplicate issue #484

PR #215, opened 2026-02-09
Interpret string offsets as int32

joosthooz:
As in https://github.com/felipeblazing/sirius/pull/2.
We'll use int32 for string offsets for now, as it seems to be the default in libcudf. Here's an issue for tracking future efforts to add support for int64: #213


joosthooz:
Closed in favor of #217

PR #217, opened 2026-02-09
Adds sort, fixes large strings

felipeblazing:
This PR

Implements the sort operators.
Added two new operators. 
Sample for Sort (samples data to notify downstream of data distribution)
Sort Partition (partitions data specifically for sort)

Large string fixes
Had to deal with 32bit offsets for cudf and 64bit offsets for duckdb. This fix is temporary and we need to revisit this later to handle both types of offsets.

### DIFF
--- a/src/include/op/result/host_table_chunk_reader.hpp
+++ b/src/include/op/result/host_table_chunk_reader.hpp
@@ -66,8 +66,13 @@ class host_table_chunk_reader {
       data_accessor;  ///< Accessor to the column data in the multiple blocks allocation
     memory::multiple_blocks_allocation_accessor<uint8_t>
       mask_accessor;  ///< Accessor to the null mask data in the multiple blocks allocation
+    memory::multiple_blocks_allocation_accessor<int32_t>
+      offset_accessor_32;  ///< Accessor to the STRING offsets (INT32) in the multiple blocks
+                           ///< allocation
     memory::multiple_blocks_allocation_accessor<int64_t>
-      offset_accessor;  ///< Accessor to the STRING offsets in the multiple blocks allocation
+      offset_accessor_64;  ///< Accessor to the STRING offsets (INT64) in the multiple blocks
+                           ///< allocation
+    bool use_int64_offsets{false};  ///< Whether the offset column uses INT64 (from cudf::pack)
 
     /**
      * @brief Construct a new column reader object
--- a/src/include/op/sirius_physical_sort_partition.hpp
+++ b/src/include/op/sirius_physical_sort_partition.hpp
@@ -23,26 +23,24 @@
 namespace sirius {
 namespace op {
 
-class sirius_physical_merge_sort : public sirius_physical_operator {
+class sirius_physical_sort_sample;
+
+class sirius_physical_sort_partition : public sirius_physical_operator {
  public:
-  static constexpr const SiriusPhysicalOperatorType TYPE = SiriusPhysicalOperatorType::MERGE_SORT;
+  static constexpr const SiriusPhysicalOperatorType TYPE =
+    SiriusPhysicalOperatorType::SORT_PARTITION;
 
  public:
-  sirius_physical_merge_sort(sirius_physical_order* order_by);
+  sirius_physical_sort_partition(sirius_physical_order* order_by);
 
-  sirius_physical_merge_sort(duckdb::vector<duckdb::LogicalType> types,
-                             duckdb::vector<duckdb::BoundOrderByNode> orders,
-                             duckdb::vector<duckdb::idx_t> projections_p,
-                             duckdb::idx_t estimated_cardinality,
-                             bool is_index_sort_p = false);
+  sirius_physical_sort_partition(duckdb::vector<duckdb::LogicalType> types,
+                                 duckdb::vector<duckdb::BoundOrderByNode> orders,
+                                 duckdb::vector<duckdb::idx_t> projections_p,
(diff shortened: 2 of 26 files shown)


src/op/result/host_table_chunk_reader.cpp
```diff
@@ -164,27 +169,39 @@ void host_table_chunk_reader::column_reader::copy_string(
...
+    auto const total_data_bytes = end_offset - start_offset;
+    auto str_buffer             = duckdb::make_buffer<duckdb::VectorBuffer>(total_data_bytes);
+    auto str_buffer_ptr         = str_buffer->GetData();
+    data_accessor.memcpy_to(allocation, str_buffer_ptr, total_data_bytes);
+
+    if (null_count != 0) {
+      auto& validity = duckdb::FlatVector::Validity(vector);
+      copy_mask_to_validity(validity, row_offset, count, allocation);
+    }
     detail::make_duckdb_strings<false>(
```
joosthooz, line 185: I think that 1 of these `<false>` is a a typo and should be true for the `null_count != 0` branch.
So if that is the case there should be 4 calls to make_duckdb_strings here. And they should have the new 2nd template argument `typename OffsetType`

src/include/op/result/host_table_chunk_reader.hpp
```diff
@@ -66,8 +66,13 @@ class host_table_chunk_reader {
       data_accessor;  ///< Accessor to the column data in the multiple blocks allocation
     memory::multiple_blocks_allocation_accessor<uint8_t>
       mask_accessor;  ///< Accessor to the null mask data in the multiple blocks allocation
+    memory::multiple_blocks_allocation_accessor<int32_t>
+      offset_accessor_32;  ///< Accessor to the STRING offsets (INT32) in the multiple blocks
+                           ///< allocation
     memory::multiple_blocks_allocation_accessor<int64_t>
-      offset_accessor;  ///< Accessor to the STRING offsets in the multiple blocks allocation
+      offset_accessor_64;  ///< Accessor to the STRING offsets (INT64) in the multiple blocks
```
joosthooz, line 73: I don't mind this implementation, but be aware we can also use a variant for this (an agent implemented this very quickly for me and it deduplicates some code)

src/include/op/sirius_physical_sort_partition.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+ public:
+  sirius_physical_sort_partition(sirius_physical_order* order_by);
+
+  sirius_physical_sort_partition(duckdb::vector<duckdb::LogicalType> types,
+                                 duckdb::vector<duckdb::BoundOrderByNode> orders,
+                                 duckdb::vector<duckdb::idx_t> projections_p,
+                                 duckdb::idx_t estimated_cardinality);
+
+  //! Order specification (copied from ORDER_BY)
```
joosthooz, line 41: What does the exclamation mark `//!` mean?

src/include/op/sirius_physical_sort_partition.hpp
```diff
@@ -0,0 +1,75 @@
...
+  {
+    return duckdb::OrderPreservationType::FIXED_ORDER;
+  }
+
+ public:
+  // Sink interface
+  bool is_sink() const override { return true; }
+  bool sink_order_dependent() const override { return false; }
+
+ public:
```
joosthooz, line 59: `public` is repeated 3x

src/include/op/sirius_physical_sort_partition.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+  duckdb::OrderPreservationType source_order() const override
+  {
+    return duckdb::OrderPreservationType::FIXED_ORDER;
+  }
+
+ public:
+  // Sink interface
+  bool is_sink() const override { return true; }
+  bool sink_order_dependent() const override { return false; }
```
joosthooz, line 57: These can be grouped with `is_source()` above, the comments aren't needed

src/include/op/sirius_physical_sort_sample.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+ public:
+  std::vector<std::shared_ptr<cucascade::data_batch>> execute(
+    const std::vector<std::shared_ptr<cucascade::data_batch>>& input_batches,
+    rmm::cuda_stream_view stream = cudf::get_default_stream()) override;
+
+  //! Override to wait for N batches before returning READY
+  std::optional<task_creation_hint> get_next_task_hint() override;
+
+ public:
```
joosthooz, line 75: `public` is repeated 6x

src/include/op/sirius_physical_sort_sample.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+  duckdb::OrderPreservationType source_order() const override
+  {
+    return duckdb::OrderPreservationType::FIXED_ORDER;
+  }
+
+ public:
+  // Sink interface
+  bool is_sink() const override { return true; }
+  bool sink_order_dependent() const override { return false; }
```
joosthooz, line 65: Same as above, these can be grouped with `is_source()`

joosthooz (changes requested):
Have a look at the calls to `make_duckdb_strings<false>()`, there may be an issue there, besides that looks good to me

joosthooz (approved):
On second thought I'm going to merge this and fix the comments in my own PR.

PR #222, opened 2026-02-10
added grouped aggregation integration tests, fixed count output type bug

wmalpica:
added grouped aggregation integration tests, fixed count output type bug
Also added a print function which can do a printf on a cudf::table or a data_batch


wmalpica:
the contents of this PR were already merged elsewhere

PR #226, opened 2026-02-10
[do not merge] Joins implementation V0

wmalpica:
WIP:
Currently Inner join, left, right full outer and left semi join seem to be working.
Right semi join is failing.
Anti joins and mark joins have not been started


wmalpica:
this work was effectively merged in another PR

PR #228, opened 2026-02-10
added avg to grouped by aggregations we can do

felipeblazing:
Can now do group by average. 

Should handle all types we support

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -36,7 +36,7 @@ bool Config::ENABLE_REGEX_JIT_IMPL = true;
 
 bool Config::MODIFIED_PIPELINE = false;
 
-uint64_t Config::DEFAULT_SCAN_TASK_BATCH_SIZE   = 2ULL * 1024 * 1024 * 1024;  ///< 2 GB
+uint64_t Config::DEFAULT_SCAN_TASK_BATCH_SIZE   = 512ULL * 1024 * 1024;  ///< 50 MB
 uint64_t Config::DEFAULT_SCAN_TASK_VARCHAR_SIZE = 256ULL;
 
 uint64_t Config::MAX_SORT_PARTITION_BYTES = 0;  ///< 0 = auto (33% of available GPU memory)
--- a/src/extension_lock.cpp
+++ b/src/extension_lock.cpp
@@ -21,6 +21,7 @@
 #include <unistd.h>    // for close
 
 #include <cerrno>
+#include <csignal>
 #include <cstring>
 #include <filesystem>
 #include <iostream>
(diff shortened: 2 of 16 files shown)


joosthooz (approved):
I ran a small test on this branch using TPCH Q1 sf1 and the averages are correct.

src/include/op/aggregate/aggregate_op_util.hpp
```diff
@@ -26,13 +26,30 @@
 namespace sirius {
 namespace op {
 
+/**
+ * @brief Mapping from one original DuckDB aggregate expression to its position(s) in the expanded
+ * cudf_aggregates vector. AVG is decomposed into SUM + COUNT_VALID (two slots), all others use one.
+ */
+struct AggregateSlot {
+  bool is_avg = false;
+  size_t cudf_idx;  ///< Index in cudf_aggregates. For AVG, this is the SUM slot; cudf_idx+1 is
```
aminaramoon, line 35: this will be randomly initialized if not set by user, is that okay ?

src/config.cpp
```diff
@@ -36,7 +36,7 @@ bool Config::ENABLE_REGEX_JIT_IMPL = true;
 
 bool Config::MODIFIED_PIPELINE = false;
 
-uint64_t Config::DEFAULT_SCAN_TASK_BATCH_SIZE   = 2ULL * 1024 * 1024 * 1024;  ///< 2 GB
+uint64_t Config::DEFAULT_SCAN_TASK_BATCH_SIZE   = 512ULL * 1024 * 1024;  ///< 50 MB
```
aminaramoon, line 39: do we still want to keep this ?

src/extension_lock.cpp
```diff
@@ -30,6 +31,28 @@
 
 namespace sirius {
 
+namespace {
+
+// Stored globally so the signal handler can clean up the lock file on abnormal termination.
+std::string g_lock_path;
+
+void cleanup_lock_file(int signum)
```
aminaramoon: why do we need this here ?
the lock is realeased once the process dies. doesn;t matter whether it is with signal or segfault or anything.


we can just add `O_CLOEXEC` to the flags of the file, so when the process is dead for whatever reason, it closes the file and releases the lock

src/extension_lock.cpp
aminaramoon: open(lock_path_.c_str(), O_CREAT | O_RDWR | O_CLOEXEC, 0666);

src/include/op/aggregate/aggregate_op_util.hpp
felipeblazing: if its not set purposefully it has no meaning

src/config.cpp
felipeblazing: yes 2GB is too big

src/extension_lock.cpp
felipeblazing: adding this made it so that i didn;t have to add rm lock file every time i ran a test than i had to ctrl + c
this should also work both are reasonable

aminaramoon (approved):

wmalpica (approved):

PR #240, opened 2026-02-12
fixed issue with limit and multile batches returning too many rows

felipeblazing:


src/op/sirius_physical_limit.cpp
```diff
@@ -38,8 +38,30 @@ sirius_physical_streaming_limit::sirius_physical_streaming_limit(
...
+  }
+  if (offset_val.Type() == duckdb::LimitNodeType::CONSTANT_VALUE) {
+    _remaining_offset.store(static_cast<int64_t>(offset_val.GetConstantValue()),
+                            std::memory_order_relaxed);
+  }
+}
+
+int64_t sirius_physical_streaming_limit::claim(std::atomic<int64_t>& counter, int64_t max_claim)
+{
+  int64_t current = counter.load(std::memory_order_acquire);
```
joosthooz, line 57: Don't you need to reload current inside the loop? Now it looks like if the compare_exchange fails the first time it will always keep failing.

joosthooz (changes requested):
Have a look at the `claim()` function

PR #242, opened 2026-02-12
Use new get_shared_db() in some newly added tests to fix the build

joosthooz:


mbrobbel (approved):

joosthooz:
Merging this because the test worker seems down and the changes do fix the build.

PR #243, opened 2026-02-12
Allow the use of `clang`

dhruv9vats:
Allow the use of `clang` compiler tool chain to build `sirius`.

The approach taken to version control build configurations is documented in the `CONTRIBUTING.md` file in this PR.

Additionally, also updates the `.clangd` config file to work better on the code base.


CONTRIBUTING.md
```diff
@@ -0,0 +1,27 @@
...
+  -DDUCKDB_EXTENSION_CONFIGS=../extension_config.cmake \
+  -DCMAKE_C_COMPILER=${CONDA_PREFIX}/bin/clang \
+  -DCMAKE_CXX_COMPILER=${CONDA_PREFIX}/bin/clang++ \
+  -DCMAKE_C_FLAGS_DEBUG="-g -O0" \
+  -DCMAKE_CXX_FLAGS_DEBUG="-g -O0" \
+  -DCMAKE_CUDA_FLAGS_DEBUG="-g -G -O0" \
+  -DCMAKE_C_COMPILER_LAUNCHER=sccache \
+  -DCMAKE_CXX_COMPILER_LAUNCHER=sccache \
+  -DCMAKE_CUDA_COMPILER_LAUNCHER=sccache \
+  -DBUILD_TESTS=ON
```
dhruv9vats: I can successfully build this.

aminaramoon (approved):

9prady9:
Which version of clang are you evaluating it with ?

pixi.toml
```diff
@@ -25,7 +28,36 @@ libconfig = "*"
...
+[tasks.configure-release]
+cmd = "cmake -S duckdb -B build/release -C CMake/release.cmake"
+description = "Configure an optimized release build"
+
+[tasks.configure-relwithdebinfo]
+cmd = "cmake -S duckdb -B build/relwithdebinfo -C CMake/relwithdebinfo.cmake"
+description = "Configure a release build with debug symbols"
+
+[tasks.build-debug]
+cmd = "cmake --build build/debug --parallel"
```
9prady9: I have suggestion w.r.t to `--parallel` usage, it defaults to the build tool's native default value which is nproc + 2 for ninja. For projects like sirius, it will be headache because it cause cudf build to hang or crash

I would strongly recommend that - we inform the developer to add `<vlaue>` (which will be appended as number of jobs to `--parallel`) as arg to pixi task so that it allows user to run as per their machine's capacity and perhaps default it to 4, so that we don't crash the system.


I am trying this out locally right now, will post an update once it works as expected.

9prady9: Changes seems to be working as expected, will push them once my local build is complete.

CMake/common.cmake
mbrobbel, line 1: Ideally, this (and the other {debug,release,..}.cmake) files should be replaced by a [`CMakePresets.json`](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html).

pixi.toml
```diff
@@ -25,7 +28,36 @@ libconfig = "*"
 libabseil = ">=20260107.0"
 duckdb = ">=1.4.4,<2"
 
-[tasks]
+[tasks.configure-debug]
```
mbrobbel, line 31: Then these tasks can just take a preset arg to build the different variants.

CMake/common.cmake
9prady9, line 1: 👍🏾 its better to club all build config flags under cmake files rather than adding them to pixi file.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Then the workflow simply becomes:
+
+```bash
+pixi shell
+real_clion .
+```
+
+### Gotchas
```
9prady9, line 73: 👍🏾 nice collection

CMake/common.cmake
dhruv9vats: @mbrobbel, see the CONTRIBUTING.md file in this PR. Using a presets file is non trivial in this setup.

9prady9: Yeah, that could also work. But when using presets file, what I noticed is that the extensions.cmake file wasn't being picked up by the build preset - perhaps it was mistake in my preset file but if no one noticed anything like that, we can commit a tested preset. And yes, that would make the commands more clean in pixi.toml.

pixi.toml
9prady9: https://github.com/sirius-db/sirius/pull/243/commits/5eb50d93b4a7d87450103a96d0cc9b2f029f8e2d

~Here's the change, we should ideally update the workflow files to pass appropriate job count otherwise, it will result in slower ci runs.~ Not needed, I have addressed it differently.

9prady9:
I am looking into the build failure, @mbrobbel please share if you have a working preset that I can test out.

.clangd
dhruv9vats, line 1: I asked `claude` to make these changes to get clangd to work better on the codebase. With these changes, the C++ side works correctly as far as I have seen (which previously was not the case), but the cuda files still do not have the static analyzer working correctly, due to conflicts between how nvcc and clang work.

If someone is versed in working with these files, a review would be helpful. High level, these were the changes made:
- Split the config into per-language sections (C, C++, CUDA) instead of
  applying -x cuda globally to all files
- Strip nvcc-specific flags (-arch, -ccbin, --expt-extended-lambda, etc.)
  that clang doesn't understand
- Parse .cu/.cuh files as C++ with CUDA attribute stubs to avoid std type
  corruption from libcudacxx chrono ODR conflicts

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,114 @@
...
+this project), which then pulls in SiriusDB as an extension, and because a `CMakePrests.json` must live alongside the
+CMake source of the project (to gets its idiomatic benefits like IDE integrations), it is a bit of a hack (involving
+symlinks) to version control the equivalent of a presets file which uses `duckdb` as its source dir.
+
+One alternative to this while still version controlling the build flags is the use of initial cache files, in
+combination with pixi tasks.
+
+The initial cache files are under the `CMake` directory, for the different build types.
+
+The available pixi tasks, which make use of these initial cache files, can be listed using:
```
dhruv9vats, line 15: @mbrobbel, given this context what would be some alternate approaches?

.github/workflows/sirius_test.yml
```diff
@@ -8,6 +8,19 @@ concurrency:
...
+          - name: gcc-cuda
+            pixi-env: cuda12
+            build-cmd: "make -j$(nproc)"
+            test-cmd: "make test"
+          - name: clang
+            pixi-env: cuda12
+            build-cmd: "pixi run build-release $(($(nproc) - 2))"
+            test-cmd: "pixi run sql-logic-tests-release"
+      fail-fast: false
+    name: test (${{ matrix.name }})
```
dhruv9vats, line 23: this adds a check that runs the tests on clang's release build. I added it for testing. This takes 20 mins to run. Do we want to keep it?

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,114 @@
+# Contributing to SiriusDB
+
+## Building in debug mode (using clang)
+
+Since SiriusDB is an extension to DuckDB, the CMake project source is actually the `duckdb` directory (a submodule of
+this project), which then pulls in SiriusDB as an extension, and because a `CMakePrests.json` must live alongside the
```
joosthooz, line 6: CMakePrests.json looks like a typo

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,114 @@
...
+The available pixi tasks, which make use of these initial cache files, can be listed using:
+
+```bash
+$ pixi task list
+Tasks that can run on this machine:
+-----------------------------------
+build-debug, build-release, build-relwithdebinfo, configure-debug, configure-release, configure-relwithdebinfo, sql-logic-tests-release
+Task                      Description
+build-debug               Build in debug mode
+build-release             Build in release mode
```
joosthooz, line 24: Can we have task just named `build` that is the same as build-release?

.clangd
joosthooz: I don't know about these files but the current state is not working for me at all so any improvement is welcome

.github/workflows/sirius_test.yml
```diff
@@ -8,6 +8,19 @@ concurrency:
 
 jobs:
   test:
+    strategy:
+      matrix:
+        include:
+          - name: gcc-cuda
```
joosthooz, line 14: Can you use a different name instead of `name`? Like compiler or toolchain

dhruv9vats:
superseded by https://github.com/sirius-db/sirius/pull/286, converting to draft until we want to explore alternative approaches.

PR #244, opened 2026-02-12
operator_data polymorphism

joosthooz:
Changed parameter and return types of functions passing around `operator_data` to ptr. This prevents object slicing when trying to pass `partitioned_operator_data` which is a subclass.
This code builds but the engine hangs, so I have to debug this solution.

### DIFF
--- a/src/pipeline/gpu_pipeline_task.cpp
+++ b/src/pipeline/gpu_pipeline_task.cpp
@@ -125,8 +125,18 @@ const sirius_pipeline* gpu_pipeline_task::get_pipeline() const
 
 std::unique_ptr<op::operator_data> gpu_pipeline_task::compute_task(rmm::cuda_stream_view stream)
 {
-  auto& local_state               = _local_state->cast<gpu_pipeline_task_local_state>();
-  auto operator_input_output_data = std::move(local_state._input_data);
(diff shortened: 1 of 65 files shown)


src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -121,18 +123,37 @@ const sirius_pipeline* gpu_pipeline_task::get_pipeline() const
   return _global_state->cast<gpu_pipeline_task_global_state>()._pipeline.get();
 }
 
-op::operator_data gpu_pipeline_task::compute_task(rmm::cuda_stream_view stream)
+std::unique_ptr<op::operator_data> gpu_pipeline_task::compute_task(rmm::cuda_stream_view stream)
 {
-  auto& local_state               = _local_state->cast<gpu_pipeline_task_local_state>();
-  auto operator_input_output_data = local_state._input_data;
+  auto& local_state                       = _local_state->cast<gpu_pipeline_task_local_state>();
+  auto operator_input_output_data_batches = local_state._input_data->get_data_batches();
```
wmalpica, line 129: one reason this approach works is that it leaves the data to still be own and held by the local state. 
In my implementation I am moving into the execute functions. I think that provides an important benefit, which is to allow for releasing memory faster

bwyogatama:
closing this in favor of #245

PR #251, opened 2026-02-13
Provide translation from duckdb expression trees to cudf ASTs

kevkrist:
This PR will provide a translator from duckdb expression trees to cudf ASTs. This will have a number of expected performance benefits that are left as future work.

- Filter pushdown into the parquet scan
- The current expression executor uses binary ops to evaluate expressions a node-at-a-time. Using cudf ASTs can improve performance by reducing the number of GMEM IOs, at the expense of slower kernels. Not an insignificant fraction of runtime in legacy Sirius was spent in expression evaluation, especially for complex expressions (resulting in many GMEM IOs with binary ops). 

It is important to note that not all duckdb expression trees have cudf AST equivalents (e.g., CASE statements).


wmalpica (approved):

src/expression_executor/gpu_expression_translator.cpp
```diff
@@ -0,0 +1,407 @@
...
+        cudf::ast::ast_operator::LESS, *left_expr, *right_expr);
+    case duckdb::ExpressionType::COMPARE_GREATERTHAN:
+      _ast_tree.emplace<cudf::ast::operation>(
+        cudf::ast::ast_operator::GREATER, *left_expr, *right_expr);
+    case duckdb::ExpressionType::COMPARE_LESSTHANOREQUALTO:
+      _ast_tree.emplace<cudf::ast::operation>(
+        cudf::ast::ast_operator::LESS_EQUAL, *left_expr, *right_expr);
+    case duckdb::ExpressionType::COMPARE_GREATERTHANOREQUALTO:
+      _ast_tree.emplace<cudf::ast::operation>(
+        cudf::ast::ast_operator::GREATER_EQUAL, *left_expr, *right_expr);
```
joosthooz, line 77: This switch is missing `break` statements

src/expression_executor/gpu_expression_translator.cpp
```diff
@@ -0,0 +1,407 @@
...
+    }
+    case duckdb::ExpressionType::OPERATOR_NOT: {
+      // Add the child
+      auto child_expr = add_expression(*expr.children[0], table_src);
+
+      // Check for failure in translating child
+      if (!child_expr) { return std::nullopt; }
+
+      // Add the operator expression
+      return _ast_tree.emplace<cudf::ast::operation>(cudf::ast::ast_operator::NOT, *child_expr);
```
joosthooz, line 386: This can do without the comments

joosthooz (changes requested):
This is wonderful, thank you. It just has a problem in a switch statement that needs to be addressed. Do you wish to add tests in a follow-up?

src/expression_executor/gpu_expression_translator.cpp
kevkrist: Thanks! Fixed with added tests for `translate_join_condition`.

wmalpica (approved):

PR #253, opened 2026-02-14
Getting rid of usage of static DuckDB in unit tests

wmalpica:
The usage of static for a DuckDB context in unit tests, although convenient, when running unit tests in debug mode, results in multiple errors of the extension context file being locked. 

This PR also fixes the physical_paritition tests


felipeblazing (changes requested):
This feels like we are going to add a lot of overhead in regenerating the duckdb instance for each test. We don't have to make it static but they could share an instance in almost all of these cases right?

felipeblazing (approved):

PR #254, opened 2026-02-16
Fix limit with multiple batches

joosthooz:
Same as #240 but merged dev with resolved conflicts and applying a fix to the issues I commented on.
Because I can't push to felipe's fork.


bwyogatama:
This seems to have the same bug as #240 that causes hangs in the unittest

felipeblazing (approved):

PR #256, opened 2026-02-16
Ungrouped aggregation type fix

joosthooz:
Targets the error in sirius_physical_ungrouped_aggregate: CUDF failure at: /tmp/conda-bld-output/bld/rattler-build_libcudf/work/cpp/src/reductions/simple.cuh:432: Output type must be same as input column type


bwyogatama (approved):

bwyogatama (approved):

joosthooz:
For completeness; the changes in this PR also crept into one of the commits in the PR that was merged just before this #255

Issue #230, opened 2026-02-11
Issues to address after GTC


bwyogatama:
- Type erase operator input and output type [DONE] 
- implement data repository push for operator
-- Downgrade executor, implement get_next_downgradable_batch_data for respositories  (https://github.com/sirius-db/sirius/issues/188)
- Getting rid of task global and local states (We dont want to do this anymore)
- Shared lock and unique lock for data batch in favor of states (https://github.com/sirius-db/sirius/issues/189) 
- Properly assigned the parents, children and dependencies for each node in the query plan generation
- Moving the CPU-loading phase of parquet scan to the prefetching executor
- Reservation peak consumption estimation (https://github.com/sirius-db/sirius/issues/472)
- Fixing the sparsity issues and OOM issues in duckdb scan executor
- Configuration to use either int64 or int32 for string offsets
- Making outer joins works with multiple partitions
- update task creation logic, to handle cases for WAITING_FOR_TASK_COMPLETION, WAITING_FOR_INPUT
- create metdata scan operators for parquet, parquet scan task get the range groups and do the scan
- get rid of legacy config in favor of new config
- fix build in debug mode [DONE]
- switch CI build to run build in debug mode
- Revisit "first" aggregation to figure out if (batch) ordering is required

aminaramoon:
- update task creation logic, to handle cases for `WAITING_FOR_TASK_COMPLETION`, `WAITING_FOR_INPUT`

aminaramoon:
- create metdata scan operators for parquet, parquet scan task get the range groups and do the scan,

aminaramoon:
- get rid of legacy config in favor of new config

bwyogatama:
Support reading multiple parquet files for one table

bwyogatama:
Supporting NULL expression (used in Column Lifetime optimization in the debug mode)

bwyogatama:
Handling empty results (https://github.com/sirius-db/sirius/issues/475)

bwyogatama:
Supporting virtual columns in parquet scan (used by simple limit queries such as "select l_orderkey from lineitem limit 100")

bwyogatama:
Right semi and right anti mixed join

bwyogatama:
Create operator input output data for delim join so that we can call distinct->execute and partition_join->execute in side the execute function (instead of sink)

bwyogatama:
Support concatenating multiple scan output when scan output is smaller than default batch size

[...]

PR #257, opened 2026-02-16
Added "first" aggregation kind

joosthooz:
This uses libcudf's NTH_ELEMENT aggregation. Note that the ordering of the columns and batches determines which value gets selected.

### DIFF
--- a/src/include/op/merge/gpu_merge_impl.hpp
+++ b/src/include/op/merge/gpu_merge_impl.hpp
@@ -17,11 +17,13 @@
 #pragma once
 
 #include <cudf/cudf_utils.hpp>
+#include <cudf/types.hpp>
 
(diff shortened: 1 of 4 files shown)


bwyogatama:
What I am not sure about this is that if this is an order-preserving operator, in which case the order of the batch we concatenate during the merging actually matters

This won't matter for the TPC-H query we are running cause there is only one element but it matters for general queries

bwyogatama:
I am good with merging this for now, but maybe we should put a todo here: #230

joosthooz:
> This won't matter for the TPC-H query we are running cause there is only one element

Agreed, that's why I ignored ordering for now in this PR

> put a todo here: https://github.com/sirius-db/sirius/issues/230

Done!

PR #260, opened 2026-02-17
Cast join sides if the types do not match

joosthooz:
This PR aims to fix the error `CUDF failure at: /tmp/conda-bld-output/bld/rattler-build_libcudf/work/cpp/src/join/hash_join.cu:856: Mismatch in joining column data types` encountered in TPC-H Q15.


joosthooz:
Not sure if this is the right approach, it seems to be a runtime problem and this solution works at pipeline building time. Closing for now.

PR #264, opened 2026-02-18
Nested loop join implementation 

### LINKED ISSUE #295: Implement nested loop joins

joosthooz:
Nested loop join implementation. Original code done by cursor, but heavily modified by Joost and William.
Closes #295 

A couple things to note:
Nested Loop Join is intended only for inequality joins, or other joins that cannot be handled by hash join. This means that the partition operator will not do any partitioning. 

Also worth noting that this implementation does support expressions and casting inside the join condition aside from inequality joins.


src/planner/sirius_plan_comparison_join.cpp
```diff
@@ -97,49 +107,49 @@ sirius_physical_plan_generator::plan_comparison_join(duckdb::LogicalComparisonJo
...
 
-  D_ASSERT(op.left_projection_map.empty());
-  duckdb::idx_t nested_loop_join_threshold =
-    duckdb::DBConfig::GetSetting<duckdb::NestedLoopJoinThresholdSetting>(context);
-  if (left->estimated_cardinality < nested_loop_join_threshold ||
-      right->estimated_cardinality < nested_loop_join_threshold) {
-    can_iejoin = false;
-    can_merge  = false;
-  }
+  // D_ASSERT(op.left_projection_map.empty());
```
wmalpica, line 110: @bwyogatama is this ok for now?

src/planner/sirius_plan_comparison_join.cpp
```diff
@@ -74,7 +74,17 @@ sirius_physical_plan_generator::plan_comparison_join(duckdb::LogicalComparisonJo
   //	TODO: Extend PWMJ to handle all comparisons and projection maps
   bool prefer_range_joins = duckdb::DBConfig::GetSetting<duckdb::PreferRangeJoinsSetting>(context);
   prefer_range_joins      = prefer_range_joins && can_iejoin;
-  if (has_equality && !prefer_range_joins) {
+  // Hash join only supports equality and NOT_DISTINCT_FROM; use nested loop join if any condition
```
bwyogatama, line 77: Putting a note here that this will cause performance degradation for Q21. Q21 will join lineitem with lineitem on a condition `l_orderkey = l_orderkey AND l_suppkey != l_suppkey`

Technically, if you have at least one equality condition, you can  use hash join, but this portion of the code will route it to nested loop join instead.

Using nested loop join to join lineitem and lineitem will be really bad for performance

bwyogatama: To solve this we need to support hash join with equality and non-equality condition

bwyogatama (approved):

src/planner/sirius_plan_comparison_join.cpp
bwyogatama: can we mark a TODO here? or create an issue?

bwyogatama: Okay i created an issue and put it in project board

PR #266, opened 2026-02-18
Validate output datatypes

joosthooz:
Adds a check after each operator's execute() function that compares the datatypes of the output columns to the vector (that was set during planning). If it doesn't match, it writes a warning in the log.


wmalpica (approved):

joosthooz:
The test problem is actually in `scan_executor`, not column_builder. At first I though I couldn't reproduce locally. Now I know which test has the problem I can reproduce and am investigating.

joosthooz:
No idea why, but the trace log messages and time measurements that were added to the scan executor caused it to hang in tests. Reverting those changes to allow CI to become green.

PR #269, opened 2026-02-19
added first implementation of count distinct by group. 

### LINKED ISSUE #225: implement COUNT DISTINCT in grouped aggregates
There was some old logic that I think was intended to handle COUND distinct 
https://github.com/sirius-db/sirius/blob/08d07c9730baadbb4261b591ef4b162272549624/src/op/sirius_physical_grouped_aggregate.cpp#L113

Here is one way we could do it. 
Suppose we have the following query:
select A, B, SUM(C), COUNT( DISTINCT D) from table GROUP BY A, B
One way we could try to do this is to have the sirius_physical_grouped_aggregate operator
It would do the SUM(C) grouped by A, B the same way
Then it would do a COLLECT_SET(D) GROUP BY A, B.     I know cudf has a function which can do this.
This would put unique values from each group into a list. THen in the grouped_merge_aggregate operator we can combine the collected sets into a single SET (unique values) and then do a count

felipeblazing:
## Summary

Adds `COUNT(DISTINCT col)` support to the grouped aggregation pipeline using cuDF's `collect_set` / `merge_sets` aggregations.

### How it works

The implementation uses a three-phase pipeline:

1. **Local phase** — `COLLECT_SET` groups rows by key and collects distinct values into a LIST column per group per batch
2. **Merge phase** — `MERGE_SETS` unions the per-batch LIST columns and deduplicates across all batches in a partition, producing one deduplicated LIST per group
3. **Post-processing** — `cudf::lists::count_elements` counts the list length per row and casts to INT64 to produce the final count

This follows the same decomposition pattern already used for AVG (SUM+COUNT), extended via the `AggregateSlot` struct.

### Changes

**Core aggregation logic**
- `aggregate_op_util`: Added `is_count_distinct` to `AggregateSlot` and `has_count_distinct` to `CudfAggregateDefinitions`. Detection: `aggr.IsDistinct() && aggr.function.name == "count"` maps to `COLLECT_SET`
- `gpu_aggregate_impl`: `COLLECT_SET` is handled with an explicit branch before the generic `get_local_aggregation<>` template — necessary because that template is also instantiated with `cudf::reduce_aggregation` for ungrouped aggregates, where `make_collect_set_aggregation` does not exist. The INT64 cast guard (COUNT_VALID/COUNT_ALL) also skips `COLLECT_SET` since its output is a LIST
- `gpu_merge_impl`: Added `COLLECT_SET → MERGE_SETS` case in `merge_grouped_aggregate`
- `sirius_physical_grouped_aggregate_merge`: Added `count_elements` + INT64 cast post-processing step. Fast path (`return input unchanged`) is now also bypassed when `has_count_distinct` is set

**Planner**
- Removed the `IsDistinct()` rejection from `can_use_partitioned_aggregate`. It was a no-op: all three planner paths (partitioned, perfect-hash, default) create the same `sirius_physical_grouped_aggregate` operator

**Test infrastructure**
- Replaced `#define PARTITION_SIZE 10000000` in `sirius_physical_partition` with a static member and `set_partition_size()` / `reset_partition_size()` methods, so integration tests can force multiple partitions without changing the production default

### Tests

**Unit tests** (`test_physical_grouped_aggregate_count_distinct.cpp`)
1. Single batch, basic correctness — hand-verified expected counts
2. Multiple batches, randomly striped — `int32` and `int64` value columns (TEMPLATE_TEST_CASE), 30 groups × 3 distinct values split into 5 random batches
3. Cross-batch duplicate deduplication — same `(key, value)` intentionally appears in multiple batches; `MERGE_SETS` must count it once
4. Mixed `count(distinct)` + `min` + `count(*)` in the same grouped aggregate
5. Multiple partitions — two independent groups of keys, each split into 3 random batches; `execute()` called separately per group of batches, exactly as the real engine does

**Integration tests** (`test_gpu_execution_tpch.cpp`) — all compare GPU vs CPU results:
- `count(distinct n_nationkey)` grouped by region key on `nation`
- `count(distinct n_name)` (string column) grouped by region key on `nation`
- Mixed `count(distinct)` + `min` + `count(*)` on `nation`
- Two group keys on `customer` (15k rows)
- Three multi-partition variants using a `PartitionSizeGuard` RAII helper that lowers `PARTITION_SIZE` to force 5 partitions (`nation`) and 15 partitions (`customer`) through the real pipeline

fixes #225

### DIFF
--- a/src/expression_executor/specializations/gpu_execute_function.cpp
+++ b/src/expression_executor/specializations/gpu_execute_function.cpp
@@ -27,6 +27,7 @@
 #include "log/logging.hpp"
 
 #include <cudf/binaryop.hpp>
+#include <cudf/column/column_factories.hpp>
 #include <cudf/datetime.hpp>
 #include <cudf/scalar/scalar.hpp>
 #include <cudf/strings/attributes.hpp>
@@ -747,6 +748,20 @@ std::unique_ptr<cudf::column> GpuExpressionExecutor::Execute(const BoundFunction
   else if (func_str == REGEXP_REPLACE_FUNC_STR) {
     RegexFunctionDispatcher dispatcher(*this);
     return dispatcher(expr, state);
+  }
+
+  //----------Struct Functions----------//
+  // row() and struct_pack() both construct a struct column from their child expressions.
+  // row() is used by DuckDB for tuple constructors like (col1, col2).
+  if (func_str == "row" || func_str == "struct_pack") {
+    D_ASSERT(!expr.children.empty());
+    std::vector<std::unique_ptr<cudf::column>> child_cols;
+    for (size_t i = 0; i < expr.children.size(); ++i) {
+      child_cols.push_back(Execute(*expr.children[i], state->child_states[i].get()));
+    }
+    cudf::size_type num_rows = child_cols[0]->size();
+    return cudf::make_structs_column(
+      num_rows, std::move(child_cols), 0, rmm::device_buffer{}, execution_stream, resource_ref);
   }
 
   // If we've gotten this far, we've encountered a unimplemented function type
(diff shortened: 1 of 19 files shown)


felipeblazing:
fixes #225

wmalpica (approved):

bwyogatama:
Can we put a todo in #230 saying that his count distinct will only work for single column for now?

bwyogatama (approved):

felipeblazing:
> Can we put a todo in #230 saying that his count distinct will only work for single column for now?

This works on multiple columns

PR #270, opened 2026-02-19
Casting aggregate output columns to their expected types

joosthooz:
To fix a type mismatch error downstream, this PR adds a casting step at the end of aggregation that checks the types of the output columns and casts them to the expected types (set during plan creation) if needed.

### DIFF
--- a/src/include/data/data_batch_utils.hpp
+++ b/src/include/data/data_batch_utils.hpp
@@ -16,15 +16,22 @@
 
 #pragma once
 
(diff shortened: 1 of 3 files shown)


joosthooz:
closed in favor of #272

joosthooz:
For more context; this solution casts a far too wide net, and besides that it can cause trouble with our implementation of AVG (creating a sum and count and then performing the avg calculation at the end)

PR #271, opened 2026-02-19
Added rudimentary query generation to dataset generation.

joosthooz:
Keep in mind that queries should actually be generated by dbgen. This code does change the ratio in q11 but there are other variable elements.

### DIFF
--- a/test/tpch_performance/generate_test_data_tpchgen-rs.py
+++ b/test/tpch_performance/generate_test_data_tpchgen-rs.py
@@ -16,6 +16,50 @@ import duckdb
 import os
 import sys
 
+# Import queries from same directory
+_script_dir = os.path.dirname(os.path.abspath(__file__))
+if _script_dir not in sys.path:
+    sys.path.insert(0, _script_dir)
+import queries as _queries_module
+
+QUERIES = _queries_module.QUERIES
+
+
+def _q11_ratio_for_sf(sf):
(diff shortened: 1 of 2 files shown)


joosthooz:
This is my workflow: 
```
# Generate test data and the queries
python test/tpch_performance/generate_test_data_tpchgen-rs.py 1 1 parquet

# set logging settings
export SIRIUS_LOG_LEVEL=debug
export SIRIUS_LOG_DIR=`pwd`/logs

# Generate duckdb results
for q in `seq 1 22`; do ./build/release/duckdb performance_test.duckdb -f test/tpch_performance/tpch_queries/orig/q${q}.sql 2>&1 | tee result_duckdb_q${q}.txt; done
# Generate Sirius results
for q in `seq 1 22`; do ./build/release/duckdb performance_test.duckdb -f test/tpch_performance/tpch_queries/gpu/q${q}.sql 2>&1 | tee result_sirius_q${q}.txt; done

# When the engine hangs, you have to bash ctrl-c a couple of times.
# Try to find an error message in the log file (also if the engine falls back to DuckDB).
# Re-run with gdb --args, type catch throw, start the program. When an exception is encountered type 'bt' for the stacktrace.
# Now you know where the error originates.
# Then continue the for loop above with a higher start number...


# When you have all the result files, check if the output matches
for q in `seq 1 22`; do if diff result_sirius_q${q}.txt result_duckdb_q${q}.txt > /dev/null; then printf "query %d results match\n" $q; fi; done
```

PR #272, opened 2026-02-19
Cast sum result from decimal64 to decimal128

joosthooz:
A more targeted approach than #270. The source of of the type mismatch is in the expected widening of decimal64 to decimal128 which duckdb expects but libcudf does not perform implicitly.


wmalpica (approved):

wmalpica:
Clothing this in favor of https://github.com/sirius-db/sirius/pull/299

PR #290, opened 2026-02-20
BUG FIXES: Getting Integration test work with debug mode after bug fixes

### LINKED ISSUE #280: Fix unit tests that are failing when run in DEBUG mode

aminaramoon:
fixes #280


src/op/sirius_physical_result_collector.cpp
```diff
@@ -176,8 +176,14 @@ void sirius_physical_materialized_collector::sink(const operator_data& input_dat
     host_table_chunk_reader chunk_reader(_client_ctx, host_table, types);
 
     // Push chunks to result collection
-    duckdb::DataChunk chunk;
-    while (chunk_reader.get_next_chunk(chunk)) {
+    while (true) {
+      // TODO(amin): it is fishy that append take a mutable reference to the chunk reader and we are
```
kevkrist, line 181: Do you mean 'chunk' instead of 'chunk reader' here?
Would returning an optional unique ptr to a `DataChunk` from the chunk reader be a better pattern? Thx.

PR #291, opened 2026-02-20
Update `.clangd` config

dhruv9vats:
Updates `.clangd` config and tweaks the cmake presets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ cufile.log
 .pixi/*
 !.pixi/config.toml
 compile_commands.json
+
+# Zed editor settings
+.zed
--- a/pixi.toml
+++ b/pixi.toml
@@ -5,7 +5,7 @@ platforms = ["linux-64", "linux-aarch64"]
 requires-pixi = ">=0.59"
 
 [activation]
-scripts = ["scripts/pixi_activate.sh", "setup_test_datasets.sh"]
+scripts = ["scripts/pixi_activate.sh"]
 
 [dependencies]
 cmake = "4.1.*"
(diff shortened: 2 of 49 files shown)


.gitignore
```diff
@@ -29,3 +29,6 @@ cufile.log
 .pixi/*
 !.pixi/config.toml
 compile_commands.json
+
+# Zed editor settings
+.zed
```
aminaramoon, line 34: you can go to .zed and add .gitignore there instead

aminaramoon (approved):

.gitignore
dhruv9vats: I think we would then need to version control the .gitignore file. We already ignore the settings dir for vscode and IntelliJ.

pixi.toml
```diff
@@ -5,7 +5,7 @@ platforms = ["linux-64", "linux-aarch64"]
 requires-pixi = ">=0.59"
 
 [activation]
-scripts = ["scripts/pixi_activate.sh", "setup_test_datasets.sh"]
+scripts = ["scripts/pixi_activate.sh"]
```
9prady9, line 8: Where did this dataset setup move to now ?

dhruv9vats: The CI already does this step: https://github.com/sirius-db/sirius/blob/a45a24c5c9c2a990df1e0ad633ba75e75a35860a/.github/workflows/sirius_test.yml#L34

and locally, this should be done manually.

9prady9: hm, I think we can do this during cmake configure. I will figure that out, no need to block this.

PR #292, opened 2026-02-20
Add `wget` to pixi env (used in env activation)

mbrobbel:
We're using `wget` to download test data so we should add it to the pixi env for systems that don't have `wget` available.


aminaramoon (approved):
rebased and approved

PR #298, opened 2026-02-20
Fix/sf10 q1

felipeblazing:
Fix scan task scheduling bugs that prevented SF10+ TPC-H queries

Three issues were found and fixed:

1. Redundant reservation replacement in scan executor (duckdb_scan_executor.cpp)

   The scan executor manager_loop was making a host memory reservation and
   calling set_reservation() on scan tasks, replacing the reservation that
   the duckdb_scan_task_local_state constructor already created and allocated
   from. This pattern was copied from the GPU pipeline executor, where tasks
   do NOT create their own reservations. For scan tasks, this caused the
   constructor allocation to reference a freed reservation, leading to
   cannot make allocation with other reservation type errors under memory
   pressure (SF10+). Fix: removed the redundant reservation and
   set_reservation() call from the scan executor.

2. Unbounded scan re-creation via get_operator_for_next_task (duckdb_scan.hpp, parquet_scan.hpp)

   When downstream operators (e.g. MERGE_GROUP_BY) returned
   WAITING_FOR_INPUT_DATA, the task creator recursively walked upstream
   through get_operator_for_next_task() until reaching the scan operator,
   which always reported READY. This created duplicate scan tasks in an
   infinite loop -- the scan would never exhaust because new tasks kept being
   created, and MERGE_GROUP_BY would never transition to READY because
   its upstream pipeline never finished. Fix: added a scan_initiated flag
   to both sirius_physical_duckdb_scan and sirius_physical_parquet_scan.
   Once the first scan task is created, the operator returns nullopt from
   get_next_task_hint(). Continuation scan tasks are handled internally
   by the scan executor, not through the task creator.

### DIFF
--- a/src/op/scan/duckdb_scan_executor.cpp
+++ b/src/op/scan/duckdb_scan_executor.cpp
@@ -207,7 +207,7 @@ void duckdb_scan_executor::manager_loop()
     }
 
     auto* scan_task = dynamic_cast<pipeline::sirius_pipeline_itask*>(task.get());
-    if (scan_task->is<parquet_scan_task>()) {
+    if (scan_task && scan_task->is<parquet_scan_task>()) {
       auto bytes_needed = scan_task->get_estimated_reservation_size();
       auto reservation  = _mem_mgr->request_reservation(
         cucascade::memory::any_memory_space_in_tier{cucascade::memory::Tier::HOST}, bytes_needed);
--- a/src/op/scan/duckdb_scan_executor.cpp
+++ b/src/op/scan/duckdb_scan_executor.cpp
@@ -206,10 +206,27 @@ void duckdb_scan_executor::manager_loop()
       }
     }
 
-    // Scan tasks manage their own host memory reservations and allocations in
-    // duckdb_scan_task_local_state's constructor — do NOT replace them here.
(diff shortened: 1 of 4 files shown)


aminaramoon (changes requested):
I think we need to make the reservation for parquest scan tasks, not duckdb scan tasks. I will make the change and push it to you r branch

aminaramoon (approved):

src/op/scan/duckdb_scan_executor.cpp
```diff
@@ -206,9 +206,8 @@ void duckdb_scan_executor::manager_loop()
       }
     }
 
-    // Make host memory reservation and set it on the local state
     auto* scan_task = dynamic_cast<pipeline::sirius_pipeline_itask*>(task.get());
-    if (scan_task) {
+    if (scan_task->is<parquet_scan_task>()) {
```
joosthooz: This is unsafe after the dynamic_cast

joosthooz (changes requested):

joosthooz (approved):

src/op/scan/duckdb_scan_executor.cpp
joosthooz: Fixed it

PR #303, opened 2026-02-21
FEATURE: enable stream reservations for Sirius

aminaramoon:
- update cucascade with a bug fix in reservation and counting of bytes
- set reservations and release at the end of each gpu task


joosthooz:
The unit tests hang during `gpu_execution - scan single column`

wmalpica (approved):

PR #305, opened 2026-02-21
Adding the sum widening cast also to the local ungrouped aggregate to fix q15

joosthooz:
It was moved to the merge aggregate in #299 to fix a problem in the unit tests. But change caused q15 to not work anymore.
Maybe it needs to happen in both places, but maybe we should look into why the test had a problem because I think doing the cast here should be enough.

### DIFF
--- a/src/op/aggregate/gpu_aggregate_impl.cpp
+++ b/src/op/aggregate/gpu_aggregate_impl.cpp
@@ -73,7 +73,6 @@ std::shared_ptr<cucascade::data_batch> gpu_aggregate_impl::local_ungrouped_aggre
             output_type = cudf::data_type(cudf::type_id::UINT64);
             break;
           }
-          default: break;
           case cudf::type_id::DECIMAL64:
             if (input_col.type().id() == cudf::type_id::DECIMAL64) {
               output_type = cudf::data_type(cudf::type_id::DECIMAL128, output_type.scale());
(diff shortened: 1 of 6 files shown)


joosthooz:
And note that https://github.com/rapidsai/cudf/issues/21512 is already fixed so we can uncomment the test when we're on a libcudf version that includes it.

src/op/aggregate/gpu_aggregate_impl.cpp
```diff
@@ -219,6 +219,23 @@ std::shared_ptr<cucascade::data_batch> gpu_aggregate_impl::local_grouped_aggrega
 
     const auto& output_idx = input_col_to_output_idx[aggregate_col_id];
     for (size_t j = 0; j < output_idx.size(); ++j) {
+      auto result_view = aggregation_result.results[j]->view();
```
wmalpica, line 222: I think this is the place the upcasting needs to be, but in that case, we should remove it from gpu_merge_impl::merge_grouped_aggregate  because then we can end up in a scenario where we upcast it twice

wmalpica (changes requested):

wmalpica (approved):

src/op/aggregate/gpu_aggregate_impl.cpp
joosthooz: Ok I've removed it there

PR #306, opened 2026-02-23
Add Anti Join Support

### LINKED ISSUE #274: Implement ANTI join
Do the same thing we did for SEMI - join

https://github.com/sirius-db/sirius/pull/261/changes

9prady9:
Fixes #274

### DIFF
--- a/src/op/sirius_physical_hash_join.cpp
+++ b/src/op/sirius_physical_hash_join.cpp
@@ -27,9 +27,9 @@
 #include "duckdb/planner/expression/bound_cast_expression.hpp"
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
 #include "log/logging.hpp"
+#include "op/sirius_physical_hash_join.hpp"
 #include "pipeline/sirius_meta_pipeline.hpp"
 #include "pipeline/sirius_pipeline.hpp"
-#include "print.hpp"
 
 #include <cstdio>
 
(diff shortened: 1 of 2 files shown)


src/op/sirius_physical_hash_join.cpp
joosthooz, line 474: Here we need to set `collect_right` to false for ANTI and `collect_left` for RIGHT_ANTI

joosthooz: Add the new join types here

joosthooz (changes requested):
Small bug and the new join type needs to be added to the big if. I think you're already working an adding RIGHT_ANTI too so that's awesome! With these changes it works for me on TPC-H BTW!

src/op/sirius_physical_hash_join.cpp
9prady9: Thank you, I saw those after you pointed out the right anti :) Should be fixed now.

wmalpica (approved):

joosthooz (approved):
Thanks!

9prady9:
there seems to be typo sort of thing in the right anti join queries, checking ...

9prady9:
> there seems to be typo sort of thing in the right anti join queries, checking ...

I figured out the problem, working on correcting the queries ...

wmalpica (approved):

PR #307, opened 2026-02-23
Add Mark Join Support

### LINKED ISSUE #275: Implement MARK join
This is used by TPCH Q16

9prady9:
Fixes #275

- [x] Enabled TPC-H Q16
- [x] Unit Test(s)

### DIFF
--- a/src/op/sirius_physical_hash_join.cpp
+++ b/src/op/sirius_physical_hash_join.cpp
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-#include "op/sirius_physical_hash_join.hpp"
-
 #include "cudf/copying.hpp"
 #include "cudf/join/filtered_join.hpp"
 #include "cudf/join/join.hpp"
@@ -88,10 +86,6 @@ sirius_physical_hash_join::sirius_physical_hash_join(
     conditions(std::move(cond)),
     delim_types(std::move(delim_types))
 {
-  if (join_type == duckdb::JoinType::MARK) {
-    throw std::runtime_error("Unsupported join type: " + duckdb::JoinTypeToString(join_type));
-  }
-
   reorder_join_conditions(conditions);
 
   filter_pushdown = std::move(pushdown_info_p);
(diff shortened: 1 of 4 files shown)


src/op/sirius_physical_hash_join.cpp
```diff
@@ -448,12 +449,58 @@ std::unique_ptr<operator_data> sirius_physical_hash_join::execute(const operator
...
 
+        cudf::column_view scatter_map(cudf::data_type(cudf::type_id::INT32),
+                                      static_cast<cudf::size_type>(semi_indices->size()),
+                                      semi_indices->data(),
+                                      nullptr,
+                                      0,
+                                      0,
+                                      {});
+
+        auto scattered = cudf::scatter(cudf::table_view({true_col->view()}),
```
wmalpica: You probably should use this cudf API instead https://github.com/rapidsai/cudf/blob/26153ca59218aeae1b01c77dd5adf345967c9923/cpp/include/cudf/copying.hpp#L205
That way you dont have to materialize a whole true column

wmalpica (changes requested):

9prady9:
This is ready with Q16 working, but I am working on a unit test nevertheless. Will try to push it before this is merged

src/op/sirius_physical_hash_join.cpp
```diff
@@ -460,6 +456,57 @@ std::unique_ptr<operator_data> sirius_physical_hash_join::execute(const operator
...
+      // All left output columns pass through directly (no gather -- all rows kept)
+      cudf::table_view left_cols_to_output =
+        get_cudf_table_view(*input_batches[0]).select(lhs_output_columns.col_idxs);
+      auto num_left_rows = left_cols_to_output.num_rows();
+
+      std::vector<std::unique_ptr<cudf::column>> mark_out_cols;
+      for (cudf::size_type i = 0; i < left_cols_to_output.num_columns(); i++) {
+        mark_out_cols.push_back(
+          std::make_unique<cudf::column>(left_cols_to_output.column(i), stream));
+      }
```
joosthooz, line 475: Is there a way to do this in a zero-copy fashion?

joosthooz (approved):
Thanks for this! I've got 1 question but not a deal-breaker

src/op/sirius_physical_hash_join.cpp
9prady9: I don't think that is possible because `left_cols_to_output` is a non-owning object that provides a view into the columns (cudf::column_views I believe) for creating the output. The output has to own it's columns, so unless we can modify inputs in place, I don't think we can make it zero-copy here.

That is my understanding of this function alone - but if there is a bigger picture I might have missed, perhaps that could provide more light onto how to achieve zero-copy.

PR #309, opened 2026-02-23
Add TPC-H parquet runner script and SQL query files

felipeblazing:
Adds tooling to run TPC-H benchmark queries directly against parquet files


joosthooz:
I have a small objection with this; q11 will produce 0 results when scale factor is >1. That's why I added query generation into the data generation script. We need a way to deal with that.

PR #319, opened 2026-02-24
Fixed bug in setting partitioning keys for hash join.

joosthooz:
The problem is not triggered in smaller runs (like the tests) because the 10M rows partition threshold is not reached.


joosthooz:
One of the new tests fails, I'm investigating if it is due to a different problem we have not encountered before.

joosthooz:
Ok I didn't mean to merge this without any reviews, I mixed up this PR with another, sorry

PR #320, opened 2026-02-24
Fix bug in hashing when keys have different datatypes

joosthooz:
When a join condition has mixed key types (e.g. INT32 vs INT64), cuDF's murmur3 hash produces different values for the same integer in different representations. That causes elements that are identical (except for their datatype) to end up in different partitions. This PR creates has partitions based on cast datatypes if needed to circumvent this.


bwyogatama (approved):

bwyogatama:
Closed in favor of #342

PR #322, opened 2026-02-24
Q21 optimization: use mixed_join instead of conditional_join if possible

### LINKED ISSUE #316: Supporting mixed conditional join in hash join
Currently if there is at least one inequality condition in the join, it will perform nested loop join.
However, technically, if you have at least one equality condition, you can use hash join, using the `mixed_conditional_join` from cudf. This optimization is needed for Q21 cause otherwise the join between lineitem and lineitem will be painfully slow

joosthooz:
This has a consequential impact on TPC-H Q21 performance and with this branch it no longer hangs at sf10.
Closes #316 

Likely to be superseded by #326

### DIFF
--- a/src/op/sirius_physical_nested_loop_join.cpp
+++ b/src/op/sirius_physical_nested_loop_join.cpp
@@ -41,6 +41,7 @@
 #include <cudf/copying.hpp>
 #include <cudf/join/conditional_join.hpp>
 #include <cudf/join/join.hpp>
+#include <cudf/join/mixed_join.hpp>
 #include <cudf/table/table_view.hpp>
 
 #include <rmm/resource_ref.hpp>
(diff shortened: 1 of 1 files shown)


wmalpica (commented):
I think we actually want the hash_join operator to be able to do mixed_join.
This is because if we do have some equality condition, then we do want to do some partitioning. 
I think this is what @kevkrist was going to do using his new translator https://github.com/sirius-db/sirius/pull/251

Its in this conversation https://sirius-db.slack.com/archives/C09F39V0030/p1771883507300469?thread_ts=1771874038.573699&cid=C09F39V0030 

I dont care if you or Kevin does it, but if we are going to do mixed joins, it needs to be done with partitioning

bwyogatama:
Close this in favor of #316

PR #325, opened 2026-02-24
Multi-file support for `parquet_scan_task`

### LINKED ISSUE #323: Support multiple files in `parquet_scan_task`
The current implementation of the parquet scan restricts a table to consist of a single file. It should be able to support multiple files.

kevkrist:
Fixes #323.


src/include/op/scan/parquet_scan_task.hpp
```diff
@@ -280,6 +297,7 @@ class parquet_scan_task_local_state : public pipeline::sirius_pipeline_task_loca
   [[nodiscard]] std::vector<cudf::size_type> move_rg_indices() { return std::move(_rg_indices); }
```
aminaramoon, line 297: why move ?

aminaramoon (changes requested):

PR #326, opened 2026-02-24
Usage of Mixed Join in Hash Join Operator

### LINKED ISSUE #316: Supporting mixed conditional join in hash join

wmalpica:
Adds support for mixed joins in the hash join operator.
It modified partition operator so that it partitions tables, only on the columns that are part of equality conditions.
It ensures that mixed joins are only allowed if a column is not referenced in both an equality and inequality condition.

The PR refactored a lot of the hash join code so that its not too repetitive between the mixed and non-mixed join case.

PR adds only integration tests. 

Closes #316

### DIFF
--- a/src/expression_executor/gpu_expression_translator.cpp
+++ b/src/expression_executor/gpu_expression_translator.cpp
@@ -35,54 +35,79 @@ gpu_expression_translator::translate_expression(duckdb::Expression const& expr,
   return result;
 }
 
+std::optional<expr_ref> gpu_expression_translator::add_join_condition(
+  duckdb::JoinCondition const& condition, bool swap_sides)
+{
+  auto left_table_ref =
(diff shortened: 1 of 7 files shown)


joosthooz:
At sf100 I'm getting this error: `Exception during task execution: CUDF failure at: /tmp/conda-bld-output/bld/rattler-build_libcudf/work/cpp/src/partitioning/partitioning.cu:476: Requested number of partitions does not fit in shared memory.`

bwyogatama (approved):

PR #327, opened 2026-02-24
Multi-file support for `parquet_scan_task`

### LINKED ISSUE #323: Support multiple files in `parquet_scan_task`

kevkrist:
Fixes #323.

### DIFF
--- a/src/include/op/scan/parquet_scan_task.hpp
+++ b/src/include/op/scan/parquet_scan_task.hpp
@@ -58,17 +58,20 @@ class parquet_scan_task_global_state : public pipeline::sirius_pipeline_task_glo
    * @brief Struct representing a range of row groups assigned to a scan task.
    */
   struct row_group_range {
-    row_group_range(size_t start_row_group_p,
+    row_group_range(size_t file_idx,
+                    size_t start_row_group_p,
                     size_t row_group_count_p,
                     size_t reserved_uncompressed_bytes_p,
                     size_t reserved_compressed_bytes_p)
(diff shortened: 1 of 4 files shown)


src/op/scan/parquet_scan_task.cpp
joosthooz: We might want to throw here, in case there is something wrong with a file

src/include/op/scan/parquet_scan_task.hpp
joosthooz: Now that we're working on this file, can we do something with this function? It doesn't make sense at all right now. `g_state.get_operator().get_operator_id();` misses the return statement and it doesn't return 0 like it says in the comment. 
Maybe it can just be removed? The field in the parent class is default initialized to zero so maybe that will just work.

joosthooz (approved):
Thanks for this!

joosthooz:
I can have a look at those comments in a different PR, they are unrelated to your changes. But I'll hold off merging until Amin has had the change to review.

src/op/scan/parquet_scan_task.cpp
kevkrist: Good idea.

src/include/op/scan/parquet_scan_task.hpp
kevkrist: You're right. Everything just works. Removed this override and made the corresponding `sirius_pipeline_itask` method not virtual.

aminaramoon (approved):

PR #334, opened 2026-02-25
Add iterations, timing, and engine selection to tpch parquet run script

joosthooz:
This PR adds the following functionality to the tpch parquet run script:
- Iterations. Queries will be repeated in the same duckdb instance to make use of scan caching
- Timing. Time measurements are taken between the query instances and these results are written into a separate csv file
- Engine selection. There is now an additional parameter to choose between sirius/duckdb for result validation purposes.
 
Example:
```
$ ./test/tpch_performance/run_tpch_parquet.sh sirius 1 2 2 3 4 5 6 7
...
$ ./test/tpch_performance/run_tpch_parquet.sh duckdb 1 2 2 3 4 5 6 7
...
$ for q in `seq 2 7`; do if diff result_sirius_sf1_q${q}.txt result_duckdb_sf1_q${q}.txt > /dev/null; then printf "query %d results match\n" $q; fi; done
query 2 results match
query 3 results match
query 4 results match
query 5 results match
query 6 results match
query 7 results match
$ cat timings_sirius_sf1_q2.csv 
step,runtime_s
views,0.002852
iter_1,0.766575
iter_2,0.079601
```
So you can see the effects of scan caching here.
I'm considering to add another script that performs all of this.


test/tpch_performance/run_tpch_parquet.sh
```diff
@@ -2,14 +2,22 @@
 # Run TPC-H GPU queries against Parquet files
 #
 # Creates views from parquet files found in the dataset directory,
-# then runs the standard GPU query files unchanged.
+# then runs the specified queries on the specified engine with the specified number of iterations.
+# All iterations run in a single DuckDB session so that scan caches warm on the first run.
+# Timings are recorded using current_timestamp between steps and written to a CSV file.
+# Output results are saved as result_<engine>_sf<scale_factor>_q<query_number>.txt
+# and timing results as timings_<engine>_sf<scale_factor>_q<query_number>.csv.
```
9prady9, line 9: nit: it would be easy to import/export if all queries write to single csv, but things like agents, this isn't much of problem though.

test/tpch_performance/benchmark_and_validate.sh
```diff
@@ -0,0 +1,182 @@
...
+# compare_tpch_engines.sh
+#
+# Runs all 22 TPC-H queries for both sirius and duckdb, compares results,
+# and writes two CSVs:
+#   comparison_sf<SF>.csv  - per-query match/error status
+#   timings_sf<SF>.csv     - long-format iteration runtimes (engine,query,iteration,runtime_s)
+#
+# Usage:
+#   export SIRIUS_CONFIG_FILE=...
+#   ./test/tpch_performance/compare_tpch_engines.sh <scale_factor> <iterations>
```
9prady9:
```suggestion
#   ./test/tpch_performance/benchmark_and_validate.sh <scale_factor> <iterations>
```

?

test/tpch_performance/benchmark_and_validate.sh
```diff
@@ -0,0 +1,182 @@
...
+#
+# Example:
+#   ./test/tpch_performance/compare_tpch_engines.sh 1 3
```
9prady9:
```suggestion
#   ./test/tpch_performance/benchmark_and_validate.sh 1 3
```

9prady9 (approved):
lgtm, minor docs fix and how does the benchmark and validation summary look like ? Can you please share that - to understand output.

PR #337, opened 2026-02-25
fix q5 sf100 inconsistent results

felipeblazing:
fixed issue where we had a race condition where an operator could end early if we pulled a batch and created = completed before we marked created

### DIFF
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -301,10 +301,20 @@ void task_creator::manager_loop()
         } else {
           // need to exhaust input batches until all ports are empty
           while (!node->all_ports_empty()) {
+            // Mark task created BEFORE popping data from ports to prevent a race
+            // condition where update_pipeline_status() sees empty ports and matching
+            // task counters, prematurely marking the pipeline as finished.
+            pipeline->mark_task_created();
+
             auto input_data = node->get_next_task_input_data();
-            if (!input_data) { break; }
-            pipeline->mark_task_created();  // WSM TODO: this needs to be done atomically with the
-                                            // task creation
+            if (!input_data || input_data->get_data_batches().empty()) {
+              // No data was available (e.g., another thread already consumed it).
+              // Balance the counter. mark_task_completed() calls update_pipeline_status()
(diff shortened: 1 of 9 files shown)


bwyogatama (approved):

aminaramoon (approved):

src/creator/task_creator.cpp
```diff
@@ -278,6 +278,12 @@ void task_creator::manager_loop()
           auto const partition_idx = parquet_task_global_state->get_next_rg_partition_idx();
           if (!partition_idx.has_value()) {
             pipeline->mark_task_completed();
+            if (pipeline->is_pipeline_finished()) {
+              auto output_consumers = pipeline->get_output_consumers();
+              for (auto& output_consumer : output_consumers) {
+                schedule(output_consumer);
+              }
```
joosthooz, line 285: We already do this in the duckdb_scan_executor https://github.com/sirius-db/sirius/blob/c295e5e791ace6296381a2929640b9b54bb559d1/src/op/scan/duckdb_scan_executor.cpp#L244-L246

src/creator/task_creator.cpp
```diff
@@ -301,10 +307,26 @@ void task_creator::manager_loop()
...
+              // No data was available (e.g., another thread already consumed it).
+              // Balance the counter. mark_task_completed() calls update_pipeline_status()
+              // which is correct: if all ports are truly empty and all real tasks have
+              // completed, the pipeline should finish.
+              pipeline->mark_task_completed();
+              if (pipeline->is_pipeline_finished()) {
+                auto output_consumers = pipeline->get_output_consumers();
+                for (auto& output_consumer : output_consumers) {
+                  this->schedule(output_consumer);
+                }
```
joosthooz, line 326: We already do this in the gpu_pipeline_executor https://github.com/sirius-db/sirius/blob/c295e5e791ace6296381a2929640b9b54bb559d1/src/pipeline/gpu_pipeline_executor.cpp#L166-L169

joosthooz (commented):
Just wondering why we need to schedule the additional consumer task request, when we are already doing that after each task (scan/pipeline) finishes

PR #345, opened 2026-02-26
Fix/get pop order issue

felipeblazing:
ensures that get and pop are called in the right order in getting data for tasks.


joosthooz (approved):
It took me a while to realize, but now I see that `pop_data_batch_by_id` removes batches that would otherwise be superfluously returned by `get_batch_ids`. So both calls need to happen under the lock. Thanks for sorting this out.

PR #349, opened 2026-02-26
Fix loadable extension build by adding EXTENSION_STATIC_BUILD

mbrobbel:
PR #286 replaced the extension-ci-tools Makefile with CMake presets but dropped the EXTENSION_STATIC_BUILD=1 flag. This flag is required on Linux to link duckdb_static into the loadable extension, without which DuckDB symbols are left unresolved causing missing symbol errors.


dhruv9vats (approved):

joosthooz (approved):
Thanks for looking into this, it caused headaches!

PR #350, opened 2026-02-26
Update duckdb to 1.4.4 to fix mixed right semi/anti joins integration tests

9prady9:
`compare_gpu_cpu` utility uses `duckdb` execution to generate the reference results which had a bug in `v1.4.3` that was later fixed in `v1.4.4`. This confirms that we don't have an actual bug in our mixed join implementation but our validation logic was using a buggy version of duckdb for that specific join logic.

Related duckdb PR that fixed the problem is duckdb/duckdb#20435

### DIFF
--- a/test/cpp/integration/test_gpu_execution_tpch.cpp
+++ b/test/cpp/integration/test_gpu_execution_tpch.cpp
@@ -2400,7 +2400,7 @@ TEST_CASE_METHOD(GPUExecutionDuckDBFixture,
 
 TEST_CASE_METHOD(GPUExecutionDuckDBFixture,
                  "gpu_execution - mixed right semi join one equality and one inequality condition",
-                 "[.][integration_disabled][gpu_execution][mixed_join]")
+                 "[integration][gpu_execution][mixed_join]")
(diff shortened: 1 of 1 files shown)


9prady9:
@mbrobbel This might affect your dorris work as @bwyogatama reminded me. Can you we sync on this tomorrow.

bwyogatama:
@mbrobbel Just as a note, the bug is actually in DuckDB, not in Sirius, so DuckDB returns the wrong result all along, our result has always been correct. Updating duckdb to v1.4.4 fixes their result and therefore it passes the result validation. However, if this breaks the substrait extension, it might not be worth it since Sirius result is correct regardless

mbrobbel (commented):
We should also bump to https://github.com/duckdb/extension-ci-tools/tree/v1.4.4

9prady9:
> We should also bump to https://github.com/duckdb/extension-ci-tools/tree/v1.4.4

Took care of this, 🤞🏾 hopefully

mbrobbel (approved):

PR #354, opened 2026-02-26
Move log configuration from compile-time to runtime engine settings

mbrobbel:
Replace the CMake SIRIUS_DEFAULT_LOG_DIR compile definition and SIRIUS_LOG_LEVEL/SIRIUS_LOG_DIR environment variables with DuckDB extension options that can be changed at runtime via SET:
- sirius_log_level (default: "info")
- sirius_log_dir (default: "log", relative to cwd)
- sirius_log_flush_seconds (default: 3)


9prady9 (approved):
lgtm

PR #356, opened 2026-02-26
Simplify CMake build: target-scoped properties, cleanup

mbrobbel:
- Replace 28 subdirectory CMakeLists.txt with file(GLOB_RECURSE CONFIGURE_DEPENDS)
- Replace global include_directories/add_compile_definitions with target-scoped equivalents
- Replace find_library for numa with pkg-config (PkgConfig::NUMA)
- Move CMAKE_EXPORT_COMPILE_COMMANDS to CMake presets
- Inline cuCascade setup directly in CMakeLists.txt, remove third_party/
- Remove unused: ENABLE_CUDF define, -rdc=true flag, -std=c++20 flag, SKIP_SUBSTRAIT_C_TESTS, CUDA --expt-* flags (default since CUDA 12.2)
- Remove bundled DuckDB extensions (json, tpcds, tpch, parquet, icu) from extension_config.cmake
- Remove orphaned sources: memmove.cu, downgrade_task_creator.{cpp,hpp}
- Remove commented-out dead code (substrait refs, old link lines)
- Consolidate set_target_properties and target_link_libraries per target
- Exclude stale disabled tests (test_modified_pipeline, test_task_creator)


pixi.toml
```diff
@@ -27,6 +27,8 @@ libabseil = ">=20260107.0"
 duckdb = ">=1.4.4,<2"
 clang-tools = ">=21.1.8,<22"
 spdlog = "1.8.*"
+pkg-config = "*"
+mold = "*"
```
joosthooz, line 31: `mold` 😎 nice

src/memory/sirius_memory_reservation_manager.cpp
```diff
@@ -38,19 +36,19 @@ sirius_memory_reservation_manager::sirius_memory_reservation_manager(
   if (gpu_spaces.empty()) {
     throw std::runtime_error("At least one GPU memory space must be configured");
   }
-  std::for_each(gpu_spaces.begin(), gpu_spaces.end(), [](const auto* space) {
+  std::for_each(gpu_spaces.begin(), gpu_spaces.end(), [this](const auto* space) {
     auto* device_mr = space->get_default_allocator();
     rmm::cuda_set_device_raii set_device{rmm::cuda_device_id{space->get_device_id()}};
-    cudf::set_current_device_resource(device_mr);
```
joosthooz, line 44: Isn't this line still needed, in addition to setting `previous_device_resource_`?

joosthooz (approved):
This is great, thanks, have a look at 1 question but if that is what you intended let's merge

src/memory/sirius_memory_reservation_manager.cpp
mbrobbel: *This comment was posted by Claude on my behalf.*

The line is still there — it wasn't removed, it was modified. `cudf::set_current_device_resource(device_mr)` is still being called. The only change is that we now capture its return value into `previous_device_resource_`, so we can properly restore it in the destructor.

Before this PR, the destructor called `cudf::reset_current_device_resource_ref()` which resets to the global default — but that caused a segfault (fixed in `e6441bb`) because the default resource could be invalid if something else had changed it in the meantime. Saving and restoring the previous resource is the correct approach.

cmake/CMakePresets.json
```diff
@@ -36,6 +36,8 @@
         "CMAKE_CUDA_COMPILER_LAUNCHER": "sccache",
         "CMAKE_CXX_COMPILER_LAUNCHER": "sccache",
         "CMAKE_C_COMPILER_LAUNCHER": "sccache",
+        "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
+        "CMAKE_LINKER_TYPE": "MOLD",
```
9prady9, line 40: 👍🏾

CMakeLists.txt
```diff
@@ -14,211 +14,150 @@
...
 if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
   set(CMAKE_CUDA_ARCHITECTURES native)
   message(STATUS "CMAKE_CUDA_ARCHITECTURES not defined. Setting to 'native'.")
 endif()
 
-project(${TARGET_NAME} LANGUAGES CXX CUDA)
+# Language standards
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CUDA_STANDARD 20)
```
9prady9: These almost always better to be set at target level instead of globally but if Sirius will not ever be used within another CMake project, this should be fine.

CMakeLists.txt
```diff
@@ -14,211 +14,150 @@
...
+set(WARNINGS_AS_ERRORS
+    OFF
+    CACHE BOOL "" FORCE)
+add_subdirectory(cucascade "${CMAKE_BINARY_DIR}/cucascade" EXCLUDE_FROM_ALL)
+
+file(GLOB_RECURSE EXTENSION_SOURCES CONFIGURE_DEPENDS
+     "${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp")
+file(GLOB_RECURSE CUDA_SOURCES CONFIGURE_DEPENDS
+     "${CMAKE_CURRENT_SOURCE_DIR}/src/*.cu")
+list(FILTER CUDA_SOURCES EXCLUDE REGEX ".*/unused/.*")
```
9prady9: I strongly object to globbing for source files, I understand `CONFIGURE_DEPENDS` will check for output variation and include any new files but there are other pitfalls for this approach.

- File system's case sensitivity or path handling differs, resulting in different file lists.
- `CONFIGURE_DEPENDS` adds a check to every single build run, which can slow down the incremental build process, especially when there are huge set of files.
- It can unintentionally pick up unwanted files, such as temporary or test files, unless specific exclusions are added, making it harder to ensure only the intended sources are compiled

There's one more related to IDE not working correctly with globbing CMakeLists but I have never encountered that as I haven't had large projects use globbing for the above mentioned reasons.

CMakeLists.txt
```diff
@@ -14,211 +14,150 @@
...
-                                                test/cpp)
-target_link_libraries(${TEST_NAME} ${EXTENSION_NAME})
-set(TEST_EXEC_DIR "${CMAKE_CURRENT_BINARY_DIR}/test/cpp")
-set(TEST_LOG_DIR "${CMAKE_CURRENT_BINARY_DIR}/test/cpp/log")
-set_target_properties(${TEST_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY
-                                              "${TEST_EXEC_DIR}")
+file(GLOB_RECURSE TEST_SOURCES CONFIGURE_DEPENDS
+     "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/*.cpp")
+list(FILTER TEST_SOURCES EXCLUDE REGEX ".*/test_modified_pipeline\\.cpp$")
+list(FILTER TEST_SOURCES EXCLUDE REGEX ".*/test_task_creator\\.cpp$")
```
9prady9: Same reasons as I mentioned for sirius target. I don't have a great alternative to `dev` but let me think it - will get back to you soon.

extension_config.cmake
```diff
@@ -20,12 +20,3 @@
 # Extension from this repo
 duckdb_extension_load(sirius SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR} LOAD_TESTS
                       EXTENSION_VERSION dev)
-
-duckdb_extension_load(json)
-duckdb_extension_load(tpcds)
-duckdb_extension_load(tpch)
-duckdb_extension_load(parquet)
-duckdb_extension_load(icu)
```
9prady9, line 28: are these obsolete or something ?

9prady9 (changes requested):
Nice cleanup but I have an objection with globbing source files. Please have a look at my comments.

extension_config.cmake
mbrobbel: No - we can just load these extensions (if we need them) instead of always building and linking them statically into our `duckdb` build.

9prady9: makes sense 👍🏾

CMakeLists.txt
mbrobbel: https://github.com/sirius-db/sirius/pull/356/changes/2a716a0d37cae565f8f042d10ba0224f5a4fe120

CMakeLists.txt
mbrobbel: https://github.com/sirius-db/sirius/pull/356/changes/94e0db23eade87f9a106729f3c1cfe3139d41a56

CMakeLists.txt
9prady9, line 1: this also works, I wonder how this will work if we ever want to switch to multiple executables. But we can cross that bridge if ever take that route :)

9prady9 (approved):
thank you! awesome cleanup 🎉

PR #359, opened 2026-02-27
fix: clangd configuration for pixi environment

johanpel:
- Pin clang <22 to match clang-tools 21 to prevent missing standard lib header resolution
- Move -x cuda from global to CUDA-only file block
- Fix Compiler version pin in clangd, we're pinning in pixi already
- Strip unsupported -fmodules-ts flags


pixi.toml
```diff
@@ -16,7 +16,7 @@ make = "*"
 ninja = "*"
 sccache = "*"
 unzip = "*"
-clang = ">=21"
+clang = ">=21,<22"
```
mbrobbel, line 19: Maybe we should just let `clang-tools` pin this?
```suggestion
clang = "*"
```

.clangd
```diff
@@ -17,20 +17,20 @@ If:
   PathMatch: .*\.cuh?
 CompileFlags:
   Add:
+    - -x
+    - cuda
+    - -Wno-unknown-cuda-version
+    - --no-cuda-version-check
     # Allow variadic CUDA functions
```
mbrobbel: Seems this is just for headers - is that intended?

pixi.toml
johanpel: With my envs cleaned if I do this it still picks up 22

.clangd
johanpel: This was applied to all files previously so all files were treated as cuda sources, which causes clangd to be unable to find the standard library

pixi.toml
mbrobbel: Ah, let's keep this then.

.clangd
johanpel: I think we can actually skip these entirely, [clang automatically recognizes .cu as cuda sources](https://llvm.org/docs/CompileCudaWithLLVM.html#:~:text=%28Clang%20detects%20that%20you%E2%80%99re%20compiling%20CUDA%20code%20by%20noticing%20that%20your%20filename%20ends%20with%20%2Ecu%2E%20Alternatively%2C%20you%20can%20pass%20%2Dx%20cuda%2E%29)

and we seem to be following this convention

johanpel: https://github.com/sirius-db/sirius/pull/359/commits/2e9393b0e6eda71988f1791d352e62595993f02f

mbrobbel (approved):

PR #360, opened 2026-02-27
Restore build jobs, cache in test workflow, update some actions

mbrobbel:


mike-wendt (approved):
Thanks for adding sccache to the GPU test and improving the check workflow for arm64 and x64.

PR #362, opened 2026-02-27
More improvements to benchmarking script

joosthooz:
This code will create a directory for each run with a timestamp, and write all relevant files into subdirs for each query. This includes the sirius log and the query that was run (which can be used to easily run gdb on something).


test/tpch_performance/run_tpch_parquet.sh
```diff
@@ -142,7 +153,11 @@ COPY (
 EOF
 
     START_TIME=$(date +%s.%N)
-    "$DUCKDB" -f "$TEMP_SQL" 2>&1 | tee "$RESULT_FILE"
+    if [ -n "${OUTPUT_DIR:-}" ]; then
```
9prady9, line 159: wouldn't this turn true once lines 100 & 101 run ? Or is that expected route here ?

9prady9 (commented):
lg, one minor query for my clarification

test/tpch_performance/run_tpch_parquet.sh
joosthooz: I wouldn't have notice if you hadn't seen this, but it turns out that this is an input env var that the `benchmark_and_validate.sh` script sets. If it is unset, these checks make sure the script will still function (it will just place the output files in the project dir instead of a per-query subdir).
I've added a note about this in the docs, will push shortly

9prady9: sounds good 👍🏾

test/tpch_performance/benchmark_and_validate.sh
```diff
@@ -45,6 +46,12 @@ QUERIES=($(seq 1 22))
 RUN_DIR="$PROJECT_DIR/runs/$(date +%Y-%m-%d_%H-%M-%S)_sf${SF}_${ITERATIONS}iter"
 mkdir -p "$RUN_DIR"
 
+if [ -n "${SIRIUS_CONFIG_FILE:-}" ] && [ -f "${SIRIUS_CONFIG_FILE}" ]; then
+    cp "$SIRIUS_CONFIG_FILE" "$RUN_DIR/sirius_config.cfg"
+else
+    cp "$HOME/.sirius/sirius.cfg" "$RUN_DIR/"
```
9prady9, line 52: ah, didn't know we had this kind of setup.

9prady9 (approved):

PR #363, opened 2026-02-27
Row group pruning with filter pushdown

kevkrist:
This PR 1) applies row-group pruning with the table filter in the row-group partitioning phase so that reservations/allocations are correct, and 2) attempts to translate the duckdb expression into a cudf AST for filter pushdown. If translation succeeds, the execution of the table scan operator is bypassed; if it fails, the table scan operator executes as before with the gpu expression executor. The key benefits are straightforward: 1) can reduce file I/O, and 2) reduces the size of cached query results in memory and prunes the table scan filter/projection from warm run execution.  

This PR needs cuDF nightly to get [this](https://github.com/rapidsai/cudf/pull/21447) bug fix for decimal columns in evaluating cuDF AST trees.

On my very modest desktop machine (4070Ti), with `cache-level = table_host`, `sf = 10` dev branch shows:
image

With this PR:
image


src/data/host_parquet_representation_converters.cpp
```diff
@@ -56,9 +54,10 @@ std::unique_ptr<cucascade::idata_representation> convert_host_parquet_to_gpu_wit
 {
   std::cerr << "convert_host_parquet_to_gpu_with_hybrid_scan" << std::endl;
   // Source stuff
-  auto& host_src             = source.cast<host_parquet_representation>();
-  auto const& rg_byte_ranges = host_src.get_column_chunk_byte_ranges();
-  auto reader                = host_src.get_parquet_reader();
+  auto& host_src                         = source.cast<host_parquet_representation>();
+  auto const& rg_byte_ranges             = host_src.get_column_chunk_byte_ranges();
+  auto reader                            = host_src.get_parquet_reader();
+  auto const& post_filter_projection_ids = host_src.get_post_filter_projection_ids();
```
joosthooz: This method has been deleted. If you need it for this improvement to work, feel free to bring it back but then we'll need some config setting to choose whether to use it. Or a setting to enable row group pruning, which will then automatically select it for example.

pixi.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-channels = ["rapidsai", "conda-forge" ]
+channels = ["rapidsai-nightly", "conda-forge" ]
```
joosthooz: I think we're going to wait until a new release is out that includes the fix, because moving to the nightlies could cause unforeseen problems. It should be just a couple of weeks.

src/include/op/scan/scan_utils.hpp
```diff
@@ -0,0 +1,78 @@
...
+#include <duckdb/planner/table_filter.hpp>
+
+namespace sirius::op::scan {
+/**
+ * @brief Convert a DuckDB TableFilterSet into a single bound DuckDB expression (conjunction of
+ * all filters), suitable for passing to gpu_expression_translator::translate_expression().
+ *
+ * Returns nullptr if the filter set is empty or contains only unsupported filter types.
+ */
+inline duckdb::unique_ptr<duckdb::Expression> convert_table_filters_to_expression(
```
joosthooz: Consider moving the implementation into a .cpp file, seeing the function is used from multiple places

src/include/op/scan/parquet_scan_task.hpp
```diff
@@ -283,10 +288,14 @@ class parquet_scan_task_local_state : public pipeline::sirius_pipeline_task_loca
   /**
    * @brief Construct the local state for the parquet scan task.
    *
-   * @param[in] g_state The global state for the parquet scan task
```
joosthooz, line 377: This could stay I think? `@param[in] g_state The global state for the parquet scan task`

src/expression_executor/gpu_expression_translator.cpp
```diff
@@ -120,8 +321,13 @@ std::optional<expr_ref> gpu_expression_translator::add_expression(
...
       return add_expression(expr.Cast<duckdb::BoundBetweenExpression>(), table_src);
-    case duckdb::ExpressionClass::BOUND_CASE:
-      return std::nullopt;  // CASE expressions cannot be translated to cuDF ASTs
+    case duckdb::ExpressionClass::BOUND_CASE: {
+      SIRIUS_LOG_DEBUG(
+        "[expression_translator] CASE expressions cannot be translated to cuDF ASTs: {}",
+        expr.ToString());
+      return std::nullopt;
+    }
+      return std::nullopt;
```
joosthooz: Remove this line, it falls between the cases

src/op/sirius_physical_parquet_scan.cpp
```diff
@@ -73,6 +78,35 @@ sirius_physical_parquet_scan::sirius_physical_parquet_scan(
...
+          "[sirius_physical_parquet_scan] Failed to translate filter expression for pushdown. "
+          "Filter will be applied in the table scan operator.");
+      } else {
+        // Instruct the sirius_physical_table_scan operator to bypass execute()
+        if (table_scan) { table_scan->passthrough = true; }
+      }
+      // Move the duckdb_expression into the table scan
+      if (table_scan) { table_scan->filter_expr = std::move(duckdb_expression); }
+    } else {
+      // throw std::runtime_error(
```
joosthooz: Do you intend to swallow this? If so, do add a debug or warning log message

src/op/scan/parquet_scan_task.cpp
```diff
@@ -219,144 +218,155 @@ parquet_scan_task_global_state::parquet_scan_task_global_state(
...
+                    projected_column_names.push_back(scan_op->names[col_idx]);
+                  });
+    _reader_options.set_column_names(std::move(projected_column_names));
+    // We only prune the pure filter columns from the projected set when the reader performs the
+    // filter. Otherwise, the expression executor will not find the filter columns.
+    if (do_filter) {
+      _post_filter_projection_ids.reserve(scan_op->types.size());
+      for (std::size_t i = 0; i < scan_op->projection_ids.size(); i++) {
+        if (i < scan_op->types.size()) {
+          _post_filter_projection_ids.push_back(scan_op->projection_ids[i]);
```
joosthooz: Claude is saying that we should review this indexing - it might mismatch with the original column_ids

src/op/scan/parquet_scan_task.cpp
```diff
@@ -433,17 +443,35 @@ std::unique_ptr<op::operator_data> parquet_scan_task::compute_task(
...
     bytes_read += range.size();
   }
+#endif
   std::for_each(read_futures.begin(), read_futures.end(), [](auto& future) { future.get(); });
 
   if (bytes_read != l_state.get_reserved_compressed_bytes()) {
     throw std::runtime_error(
-      "[parquet_scan_task] Error in reading byte ranges: total bytes read does not match reserved "
-      "compressed bytes");
+      "[parquet_scan_task] Error in reading byte ranges: total bytes read does not match "
```
joosthooz: Claude is saying in the #else case there will be gaps when projection skips columns (in which case this will always throw), could you have a look?

joosthooz (commented):
Thanks, I left some questions, hope it helps!

src/include/op/scan/parquet_scan_task.hpp
kevkrist: Fixed

src/include/op/scan/scan_utils.hpp
kevkrist: Fixed.

src/data/host_parquet_representation_converters.cpp
kevkrist: Leaving deleted.

src/expression_executor/gpu_expression_translator.cpp
kevkrist: Thanks!

src/op/sirius_physical_parquet_scan.cpp
kevkrist: [Edited] This throw actually shouldn't be here, since the table filter set may be nonempty but `convert_table_filters_to_expression` can still return nullptr (e.g., IS_NOT_NULL or OPTIONAL_FILTER table filter types).

src/op/scan/parquet_scan_task.cpp
kevkrist: The pure_filter_column_indices consist of primary indices of the table, whereas the projection ids are indices into the column_ids. We need the pure_filter_column_ids when parsing the metadata and accumulating byte sizes -- where the indexes refer to the primary indices of the file.

src/op/scan/parquet_scan_task.cpp
kevkrist: I've just discarded the use of the hybrid scan reader entirely.

pixi.toml
kevkrist: Yep. I have disabled translation for now with decimal columns, so we can go ahead and merge this now.

kevkrist (commented):
All issues addressed.

src/include/op/scan/parquet_scan_task.hpp
```diff
@@ -193,7 +194,7 @@ class parquet_scan_task_global_state : public pipeline::sirius_pipeline_task_glo
    * @param[in] idx The row group partition index.
    * @return The row group range metadata associated with the given partition.
    */
-  [[nodiscard]] row_group_range const& get_row_group_partition(size_t idx) const
+  [[nodiscard]] row_group_range const& get_row_group_partition(std::size_t idx) const
   {
     return _row_group_partitions[idx];
   }
```
mbrobbel: This has no callers and returns a reference into `_row_group_partitions`, which `claim_next_rg_partition()` moves from. Should be removed.
```suggestion
```

src/op/sirius_physical_table_scan.cpp
```diff
@@ -203,6 +105,14 @@ std::unique_ptr<operator_data> sirius_physical_table_scan::execute(const operato
   nvtx3::scoped_range nvtx_range{"sirius_physical_table_scan::execute"};
   const auto& raw_input_batches = input_data.get_data_batches();
 
+  // For parquet scan pipelines, filter and projection are already applied in
+  // parquet_scan_task and the host_parquet_representation converters.
+  if (passthrough) {
```
mbrobbel: This skips the concatenation path below. Is that intended?

src/include/op/scan/parquet_scan_task.hpp
```diff
@@ -309,48 +324,39 @@ class parquet_scan_task_global_state : public pipeline::sirius_pipeline_task_glo
...
+   * _file_paths has been populated.
    */
-  void partition_row_groups();
+  void initialize_from_files();
 
   //===----------Fields----------===//
-  size_t _approximate_batch_size;          ///< Target approximate batch size for scan tasks
+  std::size_t _approximate_batch_size;     ///< Target approximate batch size for scan tasks
   sirius_physical_parquet_scan* _scan_op;  ///< The physical parquet scan operator being executed
   bool _is_projected;                      ///< Whether projection is applied
```
mbrobbel: This is no longer used.

```suggestion
```

src/op/scan/parquet_scan_task.cpp
```diff
@@ -15,17 +15,21 @@
  */
 
 // sirius
+#include <data/cached_data_representation.hpp>
```
mbrobbel, line 18: Already included at line 44.
```suggestion
```

src/op/scan/parquet_scan_task.cpp
```diff
@@ -39,6 +43,8 @@
 #include "cudf/cudf_utils.hpp"
 #include "data/cached_data_representation.hpp"
 
+#include <cudf/ast/expressions.hpp>
+#include <cudf/cudf_utils.hpp>
```
mbrobbel, line 42: Already included at line 43.
```suggestion
```

mbrobbel (approved):
Some nits only

src/op/scan/parquet_scan_task.cpp
kevkrist: Thanks!

src/op/sirius_physical_table_scan.cpp
kevkrist: Yes the partitioning logic of the parquet_scan_task guarantees that only the file tail will be smaller than config::DEFAULT_SCAN_TASK_BATCH_SIZE, in which case there will be no other batches for it to merge with.

kevkrist (commented):
Nits fixed. Thanks!

PR #368, opened 2026-02-27
Implement memory-pressure-driven downgrade execution with per-space executors

### LINKED ISSUE #358: Downgrade Executor

felipeblazing:
Overhaul the downgrade subsystem to automatically spill GPU data to HOST
when memory pressure is detected, replacing the previous stub executor.

Key changes:
- Bind each downgrade_executor to a specific memory space (GPU:0, etc.)
  with a monitor thread that polls should_downgrade_memory() and triggers
  downgrade passes automatically
- Implement two-pass candidate selection: prioritize partitioned repos
  over non-partitioned, iterate partitions last-to-first, prefer
  non-active partitions before active ones
- Use shared_ptr<data_batch> instead of unique_ptr in downgrade tasks so
  batches remain accessible in repositories during async downgrade
- Preserve batch state across in-transit conversion to avoid invalidating
  pending pipeline tasks
- Convert from host_data_packed_representation to host_data_representation
  and replace metadata_node packing with structured column_metadata
- Add drain() to flush in-flight downgrade tasks at QueryEnd before
  destroying repositories
- Add cudaDeviceSynchronize() in terminate() to prevent deadlocks between
  async copies and pinned memory pool teardown
- Fix in-transit lock leak in gpu_pipeline_task by adding try/catch
  around convert_to calls
- Add Source::DOWNGRADE to task completion message tracking
- Add shared_test_env with Catch2 listener for pause/resume lifecycle,
  avoiding GPU memory conflicts between shared and isolated tests
- Add comprehensive downgrade_executor unit tests covering lifecycle,
  candidate selection, partition ordering, and amount limiting

Fixes #358.

### DIFF
--- a/src/downgrade/downgrade_executor.cpp
+++ b/src/downgrade/downgrade_executor.cpp
@@ -16,26 +16,30 @@
 
 #include "downgrade/downgrade_executor.hpp"
 
+#include "log/logging.hpp"
+
+#include <rmm/cuda_stream.hpp>
+
+#include <algorithm>
+#include <chrono>
+#include <optional>
+#include <thread>
+
 namespace sirius {
 namespace parallel {
 
 void downgrade_executor::schedule(std::unique_ptr<itask> task)
 {
-  // Downgrade-specific scheduling logic
   auto downgrade_task = cast_to_downgrade_task(task.get());
   if (!downgrade_task) {
-    // If it's not a downgrade_task, use the parent's implementation
     itask_executor::schedule(std::move(task));
     return;
   }
-
-  // Schedule the downgrade task using the parent's method
   itask_executor::schedule(std::move(task));
 }
 
 downgrade_task* downgrade_executor::cast_to_downgrade_task(itask* task)
 {
-  // Safely cast to downgrade_task
   return dynamic_cast<downgrade_task*>(task);
 }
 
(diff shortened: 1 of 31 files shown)


bwyogatama (approved):

src/downgrade/downgrade_executor.cpp
```diff
@@ -63,22 +79,180 @@ void downgrade_executor::stop()
...
+      if (amount > 0) {
+        // Collect all repositories from the manager
+        std::vector<downgrade_repository_info> repos;
+        _data_repo_mgr.for_each_repository(
+          [&repos](cucascade::shared_data_repository* repo) { repos.push_back({repo}); });
+        if (!repos.empty()) { run_downgrade_pass(std::move(repos), amount); }
+      }
+    }
+    // Brief sleep to avoid busy-spinning; the monitor re-checks after each interval
+    std::this_thread::sleep_for(10ms);
```
joosthooz, line 116: This could be a constant or a configuration setting

src/sirius_context.cpp
```diff
@@ -130,8 +138,25 @@ void SiriusContext::initialize(const sirius::sirius_config& config)
     *memory_manager_,
     &config_.get_hw_topology());
 
-  downgrade_executor_ = std::make_unique<sirius::parallel::downgrade_executor>(
-    sirius::parallel::task_executor_config{}, *data_repository_manager_);
+  // Create one downgrade executor per GPU memory space.
+  // HOST→DISK downgrade is not yet implemented, so we skip HOST tier for now.
```
joosthooz, line 142: It says here we are skipping HOST, so in this PR we are only spilling to DISK?

joosthooz (approved):
Thanks, please merge if you are happy with it

PR #376, opened 2026-03-02
Misc task scheduling improvements

aminaramoon:
1. dont eagerly scheudle scan tasks in pipeline executor
2. schedule all the scan tasks from pipelines
3. Ensure that task scheduling runs partition tasks on the build side of a join first to result in better number of partitions


joosthooz:
With these changes there are still tasks from each scan being scheduled via pipeline_executor::start_query()->schedule_next_scan_tasks(), what we need to prevent that is just not added all scan nodes to the `_priority_scans` queue and rely on the task hints: https://github.com/sirius-db/sirius/pull/380

wmalpica (approved):
This does not solve the issue, but is one necessary step towards solving the issue of the pipelines not being executed in the right order

PR #378, opened 2026-03-03
Defragment memory pool when running out of memory

joosthooz:
This PR performs a pooltrim and device sync when a task throws an OOM error. This helps when the memory is highly fragmented. This code solves OOM errors in Q9 at sf1k.

The 2nd commit also adds a retry stage in the OOM handler directly: instead of rescheduling the whole pipeline, it runs the pooltrim and device sync and retries the operator directly.


felipeblazing (approved):
This fixes issues right now which is great. I think we should merge this in and in the future consider making something that monitors this asynchronously and can call trim or even downgrade things

PR #379, opened 2026-03-03
Don't cache pixi envs in ci

mbrobbel:
Because we currently don't run CI on `dev` we should skip this cache because it will generate a cache entry per PR, using cache space that we can better use for `sccache`.

### DIFF
--- a/scripts/pixi_activate.sh
+++ b/scripts/pixi_activate.sh
@@ -14,8 +14,9 @@ fi
 
 project_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 cmake_presets_src="$project_root/cmake/CMakePresets.json"
-cmake_presets_dst="$project_root/duckdb/CMakeUserPresets.json"
+cmake_presets_dst="$project_root/duckdb/CMakePresets.json"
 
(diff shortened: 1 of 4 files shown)


mike-wendt (approved):

scripts/pixi_activate.sh
```diff
@@ -14,8 +14,9 @@ fi
 
 project_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 cmake_presets_src="$project_root/cmake/CMakePresets.json"
-cmake_presets_dst="$project_root/duckdb/CMakeUserPresets.json"
+cmake_presets_dst="$project_root/duckdb/CMakePresets.json"
```
dhruv9vats, line 17: qq: why do we prefer `CMakePresets.json` over `CMakeUserPreset.json` here?

mbrobbel: See [2fa8c55](https://github.com/sirius-db/sirius/pull/379/commits/2fa8c5510c24e3c7b4fe0495d568baef79069d24) commit message:
> Use CMakePresets.json (not CMakeUserPresets.json) because CMake 4.x
requires a CMakePresets.json to exist alongside CMakeUserPresets.json.

PR #384, opened 2026-03-04
Print the stacktrace when encountering a segfault

joosthooz:
We need to set `EXPORT_DYNAMIC_SYMBOLS`  to ON for this in `duckdb/CMakePresets.json` for the symbol printing to work. Not sure how to do that.


mbrobbel:
> We need to set `EXPORT_DYNAMIC_SYMBOLS` to ON for this in `duckdb/CMakePresets.json` for the symbol printing to work. Not sure how to do that.

Do we really need that? Instead of using the statically linked sirius in the `duckdb` executable, you could also use the loadable extension?

felipeblazing (approved):

PR #389, opened 2026-03-05
Bug fixes on streams

bwyogatama:
Currently cudf::default_stream() is still used in several places such as in
`auto table = batch->get_data()->cast<cucascade::gpu_table_representation>().get_table();`

This will trigger a deep copy of `cudf::table` on the `cudf::default_stream`. This causes a bunch of validation error in larger scale factor, because even though Sirius is calling `stream.synchronize()`, it won't `synchronize()` against `cudf::default_stream`

### DIFF
--- a/src/data/host_parquet_representation_converters.cpp
+++ b/src/data/host_parquet_representation_converters.cpp
@@ -141,6 +141,7 @@ std::unique_ptr<cucascade::idata_representation> convert_host_parquet_to_gpu(
   for (auto const& span : column_chunk_spans_d) {
     column_chunk_buffers.emplace_back(span.data(), span.size(), stream, mr_ref);
   }
+  stream.synchronize();
   auto result = reader.materialize_all_columns(
     host_src.get_rg_span(), std::move(column_chunk_buffers), host_src.get_reader_options(), stream);
 #endif
--- a/src/data/host_parquet_representation_converters.cpp
+++ b/src/data/host_parquet_representation_converters.cpp
@@ -141,7 +141,7 @@ std::unique_ptr<cucascade::idata_representation> convert_host_parquet_to_gpu(
   for (auto const& span : column_chunk_spans_d) {
     column_chunk_buffers.emplace_back(span.data(), span.size(), stream, mr_ref);
   }
-  stream.synchronize();
+
   auto result = reader.materialize_all_columns(
     host_src.get_rg_span(), std::move(column_chunk_buffers), host_src.get_reader_options(), stream);
 #endif
--- a/src/op/sirius_physical_delim_join.cpp
+++ b/src/op/sirius_physical_delim_join.cpp
@@ -171,7 +171,7 @@ void sirius_physical_right_delim_join::sink(const operator_data& input_data,
   auto distinct_output = distinct->execute(input_data, stream);
 
   stream.synchronize();
-  
+
   partition_join->sink(*partition_join_output, stream);
   // partition_distinct is external — push distinct output via distinct's next_port_after_sink
   distinct->sink(*distinct_output, stream);
--- a/src/op/sirius_physical_delim_join.cpp
+++ b/src/op/sirius_physical_delim_join.cpp
@@ -167,11 +167,12 @@ void sirius_physical_right_delim_join::sink(const operator_data& input_data,
   nvtx3::scoped_range nvtx_range{"sirius_physical_right_delim_join::sink"};
   // partition_join stays inline (still part of the delim join)
   auto partition_join_output = partition_join->execute(input_data, stream);
-  partition_join->sink(*partition_join_output, stream);
-
   // distinct stays inline (still part of the delim join)
   auto distinct_output = distinct->execute(input_data, stream);
 
+  stream.synchronize();
+  
+  partition_join->sink(*partition_join_output, stream);
   // partition_distinct is external — push distinct output via distinct's next_port_after_sink
   distinct->sink(*distinct_output, stream);
 }
@@ -194,11 +195,12 @@ void sirius_physical_left_delim_join::sink(const operator_data& input_data,
   nvtx3::scoped_range nvtx_range{"sirius_physical_left_delim_join::sink"};
   // column_data_scan stays inline (still part of the delim join)
   auto column_data_scan_output = column_data_scan->execute(input_data, stream);
-  column_data_scan->sink(*column_data_scan_output, stream);
-
   // distinct stays inline (still part of the delim join)
   auto distinct_output = distinct->execute(input_data, stream);
 
+  stream.synchronize();
+
+  column_data_scan->sink(*column_data_scan_output, stream);
   // partition_distinct is external — push distinct output via distinct's next_port_after_sink
   distinct->sink(*distinct_output, stream);
 }
--- a/src/op/sirius_physical_limit.cpp
+++ b/src/op/sirius_physical_limit.cpp
@@ -91,8 +91,8 @@ std::unique_ptr<operator_data> sirius_physical_streaming_limit::execute(
     if (_remaining_limit.load(std::memory_order_acquire) <= 0) { break; }
 
     auto& input_table = batch->get_data()->cast<cucascade::gpu_table_representation>().get_table();
-    auto view        = input_table.view();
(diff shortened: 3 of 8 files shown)


src/op/sirius_physical_top_n.cpp
```diff
@@ -233,16 +233,17 @@ std::unique_ptr<operator_data> sirius_physical_top_n_merge::execute(const operat
...
-    auto table = std::make_unique<cudf::table>(
-      batch->get_data()->cast<cucascade::gpu_table_representation>().get_table(),
-      stream,
-      space->get_default_allocator());
-    concat_views.push_back(table->view());
-    owned_tables.push_back(std::move(table));
+    // auto table = std::make_unique<cudf::table>(
+    //   batch->get_data()->cast<cucascade::gpu_table_representation>().get_table(),
+    //   stream,
+    //   space->get_default_allocator());
```
mbrobbel:
```suggestion
```

src/op/sirius_physical_top_n.cpp
```diff
@@ -233,16 +233,17 @@ std::unique_ptr<operator_data> sirius_physical_top_n_merge::execute(const operat
...
+    concat_views.push_back(
+      batch->get_data()->cast<cucascade::gpu_table_representation>().get_table().view());
+    // owned_tables.push_back(std::move(table));
```
mbrobbel:
```suggestion
```

9prady9 (commented):
Good catch!

I think we can catch such errors using tidy or at least have warnings with the following in clang tidy

```.clang-tidy
       performance-for-range-copy,
       performance-unnecessary-copy-initialization,
       performance-unnecessary-value-param,
```

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -289,6 +290,7 @@ void gpu_pipeline_task::execute(rmm::cuda_stream_view stream)
     }
     processing_handles.emplace_back(std::move(*handle));
   }
+  stream.synchronize();  // ensure all conversions are done before we proceed with execution
```
9prady9: we are passing the same stream to `compute_task` at line 300, why do we need to sync here ? 

Another Q, we are calling `publish_output` under a conditional, which does sync, but we also have another sync before that `if-else` block, do we need that as well ?

src/sirius_context.cpp
```diff
@@ -79,6 +79,9 @@ SiriusContext::~SiriusContext() noexcept
 
 void SiriusContext::QueryBegin(ClientContext& context)
 {
+  // Reset operator ID counter so each query starts from 0
+  sirius::op::sirius_physical_operator::next_operator_id.store(0);
```
9prady9, line 83: easier to understand the operator numbers in logs now 👍🏾

src/op/sirius_physical_top_n.cpp
joosthooz: We should clean this up, but we can do it later

joosthooz (approved):
Great find

src/data/host_parquet_representation_converters.cpp
```diff
@@ -141,6 +141,7 @@ std::unique_ptr<cucascade::idata_representation> convert_host_parquet_to_gpu(
   for (auto const& span : column_chunk_spans_d) {
     column_chunk_buffers.emplace_back(span.data(), span.size(), stream, mr_ref);
   }
+  stream.synchronize();
```
9prady9: we already syncing at 149 right ?L145 is already using `stream` we are providing. Unless there is a bug in `materialize_all_columns` we don't need this one right ?

src/op/sirius_physical_top_n.cpp
9prady9: we don't need the commented code anymore right ?

src/op/sirius_physical_delim_join.cpp
```diff
@@ -167,11 +167,12 @@ void sirius_physical_right_delim_join::sink(const operator_data& input_data,
   nvtx3::scoped_range nvtx_range{"sirius_physical_right_delim_join::sink"};
   // partition_join stays inline (still part of the delim join)
   auto partition_join_output = partition_join->execute(input_data, stream);
-  partition_join->sink(*partition_join_output, stream);
-
   // distinct stays inline (still part of the delim join)
   auto distinct_output = distinct->execute(input_data, stream);
 
+  stream.synchronize();
```
9prady9, line 173: do we even need `sync` here ? we seem to be passing the same stream for the next calls right ? we might end up slowing things down by explicit sycns right ?

src/op/sirius_physical_delim_join.cpp
```diff
@@ -194,11 +195,12 @@ void sirius_physical_left_delim_join::sink(const operator_data& input_data,
   nvtx3::scoped_range nvtx_range{"sirius_physical_left_delim_join::sink"};
   // column_data_scan stays inline (still part of the delim join)
   auto column_data_scan_output = column_data_scan->execute(input_data, stream);
-  column_data_scan->sink(*column_data_scan_output, stream);
-
   // distinct stays inline (still part of the delim join)
   auto distinct_output = distinct->execute(input_data, stream);
 
+  stream.synchronize();
```
9prady9, line 201: same question as the one from Line 173.

src/op/sirius_physical_delim_join.cpp
wmalpica: I think we do need this synchronize because this is right before we call `sink` which will put data into data_repositories, which means that other tasks can get scheduled and executed on a different stream on this data.

src/op/sirius_physical_delim_join.cpp
wmalpica: I think we do need this synchronize because this is right before we call sink which will put data into data_repositories, which means that other tasks can get scheduled and executed on a different stream on this data.

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -257,6 +257,7 @@ void gpu_pipeline_task::publish_output(op::operator_data& output_data, rmm::cuda
     _global_state->cast<gpu_pipeline_task_global_state>().get_pipeline()->get_sink();
   if (sink_operators) {
     sink_operators.get()->sink(output_data, stream);
+    stream.synchronize();
```
wmalpica: I dont think we need this stream.synchronize().
By this point we have already put data into other  data_repositories, so if anything we need to call stream.sync before calling op.sink()
Which we are doing in  anyways right after we call op.execute() inside `gpu_pipeline_task::compute_task`
For that matter, I think we should be able to get rid of the stream.synchronize that is right after compute_task(strea)

src/pipeline/gpu_pipeline_task.cpp
wmalpica: I agree, I dont think we need this one as well

src/op/sirius_physical_limit.cpp
```diff
@@ -90,9 +90,9 @@ std::unique_ptr<operator_data> sirius_physical_streaming_limit::execute(
     // Check if limit is already exhausted
     if (_remaining_limit.load(std::memory_order_acquire) <= 0) { break; }
 
-    auto input_table = batch->get_data()->cast<cucascade::gpu_table_representation>().get_table();
-    auto view        = input_table.view();
-    auto num_rows    = static_cast<int64_t>(view.num_rows());
+    auto& input_table = batch->get_data()->cast<cucascade::gpu_table_representation>().get_table();
```
wmalpica, line 93: I am so happy you found this issue. 
But holy hell, we need to fix the API itself so that its not so easy to cause problems

wmalpica (changes requested):

src/op/sirius_physical_limit.cpp
9prady9: We can use tidy checks for catching accidental additions also.

src/pipeline/gpu_pipeline_task.cpp
9prady9, line 1: While looking through the rest of the code, I saw a sync at L218, would like to get an understanding about that in our sync.

wmalpica (approved):

PR #391, opened 2026-03-05
Improvement/non profiling runs optimazation advisor

felipeblazing:
Makes it so that we also capture non profiled run times when doing comparisons between runs

### DIFF
--- a/test/tpch_performance/rewrite_parquet.py
+++ b/test/tpch_performance/rewrite_parquet.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python3
 """
-Rewrite TPC-H parquet files with GPU-optimized settings using cudf.
+Rewrite TPC-H parquet files with GPU-optimized settings.
 
+Uses cudf (GPU) if available, otherwise falls back to pyarrow (CPU-only).
 Reads each table from the source directory and writes it back with:
   - Snappy compression
   - Parquet V2 page headers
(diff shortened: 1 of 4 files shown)


test/tpch_performance/rewrite_parquet.py
```diff
@@ -100,7 +110,72 @@ def cast_to_schema(arrow_table, target_schema):
...
+    """Create a ParquetWriter with standard settings."""
+    return pq.ParquetWriter(
+        path,
+        schema,
+        compression="snappy",
+        version="2.6",
+        data_page_version="2.0",
+        write_statistics=True,
+        use_dictionary=True,
+        data_page_size=MAX_PAGE_SIZE_BYTES,
```
joosthooz, line 123: You also need to set `max_rows_per_page` otherwise that defaults to 20k resulting in very small pages

test/tpch_performance/rewrite_parquet.py
```diff
@@ -100,7 +110,72 @@ def cast_to_schema(arrow_table, target_schema):
...
+def _make_writer(path, schema):
+    """Create a ParquetWriter with standard settings."""
+    return pq.ParquetWriter(
+        path,
+        schema,
+        compression="snappy",
+        version="2.6",
+        data_page_version="2.0",
+        write_statistics=True,
+        use_dictionary=True,
```
joosthooz, line 122: In the tpchgen-rs generation script we disable dictionary for a bunch of columns https://github.com/sirius-db/tpchgen-rs/blame/bwy/sirius/scripts/generate_tpch.py

PR #401, opened 2026-03-06
Fix deadlock in get_next_task_input_data()

joosthooz:
Prevent sibling nodes deadlocking in get_next_task_input_data().

### DIFF
--- a/src/op/sirius_physical_partition.cpp
+++ b/src/op/sirius_physical_partition.cpp
@@ -26,6 +26,8 @@
 
 #include <nvtx3/nvtx3.hpp>
 
+#include <mutex>
+
 namespace sirius {
 namespace op {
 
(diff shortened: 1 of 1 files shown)


felipeblazing (approved):

src/op/sirius_physical_partition.cpp
```diff
@@ -236,38 +238,36 @@ int sirius_physical_partition::determine_num_partitions()
 
 void sirius_physical_partition::set_num_partitions(int num_partitions)
```
bwyogatama, line 239: Seems like this is not needed anymore, since you set it at get_next_task_hint

bwyogatama (changes requested):

aminaramoon (approved):

bwyogatama (approved):

PR #406, opened 2026-03-07
SKILL: for finding optimized config

aminaramoon:


johanpel:
Nice.  Maybe something for later, but leveraging something like Optuna which now has an MCP server: https://github.com/optuna/optuna-mcp may be an interesting venue to explore.

wmalpica (approved):

PR #411, opened 2026-03-08
ASOF join support

Darmin-Geek:
This is still in early stages.


Darmin-Geek:
Hi, while I originally planned to complete the ASOF joins feature, I see that @ducndh is also working on the feature and has a better plan. As a result, to avoid future confusion, I am closing this pull request so that they can create one when they are ready.

ducndh:
I heavily reused the cudf operator to implement this as a stop gap so building a custom CUDA kernel that faster would be much better option. I am currently not having bandwidth to implement full integration of this operator since my use case does not involve this anymore so your support would be much necessary.

Darmin-Geek:
Ok, thank you for clarifying. I will continue to work on the feature.

PR #417, opened 2026-03-09
Skip file I/O when running from cache

joosthooz:
Creating the kvikio file handles and reading parquet metadata can be slightly expensive, skip it if we are running from cache


joosthooz:
This code sometimes crashes, needs investigation. It does have speedup potential of about 5%

joosthooz:
Creating a new PR for a new branch

PR #418, opened 2026-03-09
CI: Catch all sqllogic test failures and resolve test_datasets permissions

mike-wendt:
## Summary

- **Permission denied on `test_datasets/`**: Self-hosted GPU runners reuse the workspace between runs. Files from previous runs could be owned by a different user, causing `setup_test_datasets.sh` to skip regeneration (files already exist) while the test binary fails to read them. Fixed by adding `chmod -R a+r test_datasets/` after the setup script.

- **Sqllogic test failures silently passing**: The `tee` pipe masked the test binary's exit code — `tee` always exits 0, so even a failing test run would pass the CI step. Also, the existing `grep` only caught `Error in GPUExecuteQuery`, missing other failure modes. Fixed by capturing `PIPESTATUS[0]` (the test binary's exit code) and expanding the grep to also match `FAILURES SUMMARY`.

Reported by @bwyogatama 

## Test plan
- [x] Trigger a CI run and verify the `Run sqllogic tests` step fails correctly when tests fail
- [x] Verify `test_datasets/` permissions no longer cause `IO Error: Permission denied` on subsequent runs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,8 +86,8 @@ jobs:
 
       - name: Run sqllogic tests
         run: |
-          ./build/release/test/unittest "$PWD/test/*" | tee test.log
-          if grep -q "Error in GPUExecuteQuery" test.log; then
-            echo "GPU execution errors detected!"
+          ./build/release/test/unittest "$PWD/test/*" | tee test.log; EXIT=${PIPESTATUS[0]}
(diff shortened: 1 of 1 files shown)


mbrobbel (approved):

mike-wendt:
Tests have run, no longer see the reported errors. Thanks @mbrobbel for the review

PR #433, opened 2026-03-10
Bump cucascade

mbrobbel:


mbrobbel:
The build failure is caused by cuCascade propagating `cuda_std_20` as a PUBLIC compile feature, which reaches DuckDB's core targets that don't have CUDA language enabled. CMake then fails with:

```
CMake Error in src/CMakeLists.txt:
  No known features for CUDA compiler ""
  version .
```

Fix submitted upstream: https://github.com/NVIDIA/cuCascade/pull/83

*This comment was posted by Claude.*

9prady9 (approved):

PR #440, opened 2026-03-11
Add rattler-build conda recipe for Sirius DuckDB extension

mbrobbel:
Packages the loadable DuckDB extension as a conda package with CUDA 12/13 variants. Includes activation scripts that export SIRIUS_EXTENSION path.


mbrobbel:
We don't really have a use-case for this right now.

PR #443, opened 2026-03-11
NVTX Ranges added

felipeblazing:
We want to be able to add NVTX ranges to our operators and tasks to be able to better track what is going on in nsys


dhruv9vats (commented):
The use of `fmt::format` maybe considered.

9prady9:
> The use of `fmt::format` maybe considered.

please use `std::format` we have that in C++20

felipeblazing:
@dhruv9vats @9prady9 done

aminaramoon (approved):

PR #448, opened 2026-03-11
remove redudanct retry in gpu_pipeline_task

aminaramoon:
we dont use default mempool almost ever, the retry logic assumes we use default pool and retries the task.
I am removing it so we dont spend extra cucle retrying and failing a task


joosthooz (changes requested):
This can definitely be improved upon but I don't agree with fully removing it like this. Because the pooltrim helps to resolve memory fragmentation and this handler only retries the current operator instead of rescheduling the whole pipeline

PR #450, opened 2026-03-11
Changed get_ranges for cached data use nullptr instead of exception throwing

wmalpica:
This PR changes `cache_ranges::get_ranges()` function to return a std::nullptr when it does not find the data instead of throwing an exception.  This exception was being used to then trigger a fallback (regular reading of parquet). This happens all the time when it does not find something in the cache. Which means in a way , its using an exception throw as part of regular behaviour business logic, which is an anti-pattern.

I usually run gdb with it set to stop on all caught exceptions. The way this is right now, i cant do that anymore, so I am changing it.

### DIFF
--- a/src/op/scan/cached_ranges.cpp
+++ b/src/op/scan/cached_ranges.cpp
@@ -70,8 +70,8 @@ cache_ranges::cache_ranges(std::vector<range> ranges,
   total_packed_bytes_ = running;
 }
 
-std::vector<std::span<const std::byte>> cache_ranges::get_ranges(std::size_t query_offset,
-                                                                 std::size_t query_size) const
+std::optional<std::vector<std::span<const std::byte>>> cache_ranges::get_ranges(
+  std::size_t query_offset, std::size_t query_size) const
 {
   if (query_size == 0) return {};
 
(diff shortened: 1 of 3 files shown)


src/op/scan/cached_ranges.cpp
```diff
@@ -70,8 +70,8 @@ cache_ranges::cache_ranges(std::vector<range> ranges,
   total_packed_bytes_ = running;
 }
 
-std::vector<std::span<const std::byte>> cache_ranges::get_ranges(std::size_t query_offset,
-                                                                 std::size_t query_size) const
+std::optional<std::vector<std::span<const std::byte>>> cache_ranges::get_ranges(
```
aminaramoon, line 73: how about just empty vector ?

aminaramoon (approved):

src/op/scan/cached_ranges.cpp
wmalpica: I was not 100% sure if empty vector was an acceptable output (meaning, did not throw an exception) in the old code, so i wanted to ensure i preserved the previous logic

PR #456, opened 2026-03-12
Running Kevin's row_group_pruning branch on libcudf nightlies and a patched cucascade

joosthooz:
Shamelessly building on Kevin's PR #363 but using libcudf nightly build and cucascade with this patch  https://github.com/NVIDIA/cuCascade/pull/88


kevkrist:
#363 is finally ready. But I don't have access to a good machine for performance evaluation.

PR #460, opened 2026-03-12
docs: add NVIDIA driver >= 570 requirement and fix mkdir -p for TPC-H setup

### LINKED ISSUE #459: [Docs/Install] NVML not found + gpu_buffer_init silent crash when NVIDIA driver < 570 (CUDA 13.x runtime requirement undocumented)
## Summary

When following the installation instructions on a fresh Ubuntu 22.04 GCP VM with a Tesla T4 GPU, two issues occur in sequence:

1. **`NVML not available: failed to load libnvidia-ml.so.1`** — printed on every DuckDB launch
2. **`call gpu_buffer_init(...)` silently crashes** — process exits with code 35, no terminal output, log truncates mid-initialization

Both are caused by the same root cause: the README states `CUDA >= 13.0` as a requirement but does not specify the **minimum NVIDIA driver version** needed to support CUDA 13.x at runtime. Installing a driver that only supports CUDA 12.2 (e.g. driver 535.x, commonly installed via `apt`) is insufficient.

---

## Environment

| Component | Value |
|---|---|
| OS | Ubuntu 22.04.5 LTS (Jammy), kernel `6.8.0-1048-gcp` |
| GCP instance | `n1-standard-16`, zone `us-central1-b` |
| GPU | NVIDIA Tesla T4 (TU104GL, compute capability 7.5) |
| GPU memory | 15360 MiB |
| CPU | Intel Xeon @ 2.30 GHz, 16 vCPUs |
| RAM | 58 GB |
| Repo commit | `e0a42b0316b4fd36fca3560448ceed1fedbf2b2a` (HEAD -> dev, 2026-03-12) |
| CUDA toolkit (pixi/nvcc) | 13.1.115 |
| libcudf (pixi) | 26.02.01 |

---

## Steps to Reproduce

1. Provision a GCP Ubuntu 22.04 VM with a Tesla T4 GPU
2. Clone the repo and run `pixi shell` — CUDA toolkit 13.1 is installed via pixi
3. Install NVIDIA driver via `sudo apt-get install -y nvidia-driver-535` (or any 535.x driver, which is commonly suggested for Ubuntu 22.04)
4. Build with `CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) make`
5. Launch DuckDB and call `gpu_buffer_init`:

```sql
call gpu_buffer_init('1 GB', '2 GB');
```

---

## Observed Behavior

**On every DuckDB launch:**
```
NVML not available: failed to load libnvidia-ml.so.1
```

**On `call gpu_buffer_init('1 GB', '2 GB')`:**
- Process silently exits, exit code `35`
- No terminal error output
- Debug log (`SET sirius_log_level = 'debug'`) truncates after:
```
[info] [gpu_buffer_manager.cpp:151] Initializing GPU buffer manager with params:
  Use Pin - true, GPU Cache Size - 1073741824, GPU Processing Size - 2147483648
```
- `nvidia-smi` with driver 535 reports: `Driver Version: 535.288.01  CUDA Version: 12.2`
- `dmesg` showed no GPU hardware errors — driver loaded cleanly at kernel level

---

## Root Cause

Driver 535.x supports a maximum of **CUDA 12.2** at runtime. Sirius is compiled with **CUDA 13.1** (via pixi). The mismatch causes CUDA context initialization to fail silently inside `gpu_buffer_manager`, crashing the process before any error can be logged.

| | Version |
|---|---|
| CUDA toolkit used to build (pixi/nvcc) | 13.1.115 |
| Max CUDA supported by driver 535 | 12.2 |
| Required driver for CUDA 13.x runtime | >= 570.x |

---

## Fix

Upgrade the NVIDIA driver to >= 570 (tested with 590):

```bash
sudo apt-get install -y nvidia-driver-590
sudo reboot
```

After upgrade, `nvidia-smi` reports `CUDA Version: 13.1` and `gpu_buffer_init` succeeds:

```sql
D call gpu_buffer_init('1 GB', '2 GB');
┌─────────┐
│ Success │
│ boolean │
├─────────┤
│  0 rows │
└─────────┘
```

---

## Suggested README Change

The README currently states:
> CUDA >= 13.0

We suggest adding an explicit driver version requirement, e.g.:

> CUDA >= 13.0 (requires NVIDIA driver >= 570)

This would prevent other users from hitting the same silent crash after installing a driver that is technically "current" for Ubuntu 22.04 but too old for CUDA 13.x runtime support.

---

Reported by @mandrews

mandrews:
## Summary

Two small documentation fixes found during installation on GCP (Ubuntu 22.04, Tesla T4). See #459 for full context.

- **NVIDIA driver version**: The README states `CUDA >= 13.0` but does not specify the minimum driver version required to support CUDA 13.x at runtime. Added `(requires NVIDIA driver >= 570)` inline. Without this, users installing the commonly suggested driver 535.x will hit a silent crash at `gpu_buffer_init`.
- **`mkdir -p` for TPC-H setup**: Changed `mkdir s1` to `mkdir -p s1` so the dataset generation step does not fail if the directory already exists on re-runs.

## Changes

- `docs/README.md`: CUDA requirement line updated with driver version
- `docs/README.md`: `mkdir` → `mkdir -p` in TPC-H dataset generation step

Closes #459

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,7 +22,7 @@ Experiment Setup:
 ## Supported OS/GPU/CUDA/CMake
 - Ubuntu >= 22.04
 - NVIDIA Volta™ or higher with compute capability 7.0+
-- CUDA >= 13.0
+- CUDA >= 13.0 (requires NVIDIA driver >= 570)
 - CMake >= 3.30.4 (follow this [instruction](https://medium.com/@yulin_li/how-to-update-cmake-on-ubuntu-9602521deecb) to upgrade CMake)
 - libcudf >= 26.04
 - We recommend building Sirius with at least **16 vCPUs** to ensure faster compilation.
@@ -75,7 +75,7 @@ To generate the TPC-H dataset
 cd test_datasets
 unzip tpch-dbgen.zip
 cd tpch-dbgen
-./dbgen -s 1 && mkdir s1 && mv *.tbl s1  # this generates dataset of SF1
+./dbgen -s 1 && mkdir -p s1 && mv *.tbl s1  # this generates dataset of SF1
 cd ../../
 ```
 


mbrobbel (approved):
Thank you @mandrews

PR #462, opened 2026-03-12
Limit on task retry and drain on error

wmalpica:
This PR adds a limit to the number of retries a task can have. It also will clean up the tasks queues if a task is failing. This way the query does not hang and ends appropriatelly.
I also improved some of the logging for each task so that we also capture the number of  bytes in the input and output and the peak memory consumption.

After this, we need should track memory consumption of operators or pipelines so that we can use that to improve our task memory consumption needs which right now are WAY off. 
And then we also need to make requests for the downgrade scheduler as needed

### DIFF
--- a/src/include/exec/kiosk.hpp
+++ b/src/include/exec/kiosk.hpp
@@ -69,6 +69,15 @@ class kiosk {
     cv_.notify_all();
   }
 
+  // Re-enable the kiosk after a stop() so it can accept new tickets.
+  // Must only be called once all outstanding tickets have been released
+  // (i.e. after wait_all() has returned).
+  void reset_stopped()
+  {
+    std::lock_guard lock(mutex_);
+    stopped_ = false;
+  }
+
   // Blocking: acquire a ticket (waits if bounded and at capacity)
   [[nodiscard]] ticket acquire();
 
--- a/src/include/exec/kiosk.hpp
+++ b/src/include/exec/kiosk.hpp
@@ -72,7 +72,7 @@ class kiosk {
   // Re-enable the kiosk after a stop() so it can accept new tickets.
   // Must only be called once all outstanding tickets have been released
   // (i.e. after wait_all() has returned).
-  void reset_stopped()
+  void resume()
   {
     std::lock_guard lock(mutex_);
     stopped_ = false;
--- a/src/include/pipeline/completion_handler.hpp
+++ b/src/include/pipeline/completion_handler.hpp
@@ -53,6 +53,7 @@ class completion_handler {
     bool expected = false;
     if (_completed.compare_exchange_strong(expected, true)) {
       try {
+        _has_error.store(true);
         _promise.set_exception(error);
       } catch (...) {
         // Promise already satisfied or other error - ignore
@@ -73,6 +74,7 @@ class completion_handler {
     bool expected = false;
     if (_completed.compare_exchange_strong(expected, true)) {
       try {
+        _has_error.store(true);
         _promise.set_exception(std::make_exception_ptr(std::runtime_error(error.data())));
       } catch (...) {
         // Promise already satisfied or other error - ignore
@@ -112,9 +114,17 @@ class completion_handler {
    */
   [[nodiscard]] bool is_completed() const noexcept { return _completed.load(); }
 
+  /**
+   * @brief Check if the handler has already been completed with an error.
+   *
+   * @return True if an error has been reported, false otherwise.
+   */
+  [[nodiscard]] bool has_error() const noexcept { return _has_error.load(); }
+
  private:
   std::promise<void> _promise;
   std::atomic<bool> _completed{false};
+  std::atomic<bool> _has_error{false};
 };
 
 }  // namespace sirius::pipeline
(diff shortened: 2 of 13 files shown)


src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -152,6 +149,42 @@ void gpu_pipeline_executor::manager_loop()
...
+        static constexpr uint32_t MAX_OOM_RETRIES = 10;
+        if (next_retry_count > MAX_OOM_RETRIES) {
+          SIRIUS_LOG_ERROR(
+            "GPU Pipeline Executor: task {} (original task {}) exceeded {} OOM retries at "
+            "operator index {} — terminating query",
+            gpu_task->get_task_id(),
+            orig_task_id,
+            MAX_OOM_RETRIES,
+            oom.get_resume_operator_index());
+          if (_completion_handler) {
```
joosthooz, line 189: What we can try here, before failing, is having the downgrade executor spill as much batches as possible and then continue execution.

wmalpica: I agree that we do need something like that, but how much to spill would just be a guess. I want to add more of the foundational stuff we need first, so that this reactive downgrading is more accurate and "proper"

src/include/exec/kiosk.hpp
```diff
@@ -69,6 +69,15 @@ class kiosk {
     cv_.notify_all();
   }
 
+  // Re-enable the kiosk after a stop() so it can accept new tickets.
+  // Must only be called once all outstanding tickets have been released
+  // (i.e. after wait_all() has returned).
+  void reset_stopped()
```
aminaramoon: resume() ?

src/include/pipeline/completion_handler.hpp
```diff
@@ -73,6 +74,7 @@ class completion_handler {
     bool expected = false;
     if (_completed.compare_exchange_strong(expected, true)) {
       try {
+        _has_error.store(true);
```
aminaramoon, line 77: why do we need this ?

src/include/exec/kiosk.hpp
aminaramoon: reopen() ?

src/include/pipeline/completion_handler.hpp
wmalpica: It was to that it would be easy to query if it has an error

joosthooz (approved):

src/include/pipeline/gpu_pipeline_task.hpp
```diff
@@ -70,12 +71,16 @@ class gpu_pipeline_task_local_state : public sirius_pipeline_task_local_state {
   std::unique_ptr<op::operator_data> _input_data;  ///< Input data batches for the pipeline
   size_t _start_operator_index = 0;  ///< Operator index to resume from (0 = start of pipeline)
 
+  /// Number of times this task has been retried due to OOM (0 = first attempt).
+  uint32_t retry_count = 0;
+  /// Task ID of the original (non-retried) task; only meaningful when retry_count > 0.
+  uint64_t original_task_id = 0;
```
9prady9, line 77: better to store it using `std::optional<>` rather than having special name for `0` id.

src/op/scan/duckdb_scan_executor.cpp
```diff
@@ -97,6 +97,28 @@ void duckdb_scan_executor::set_task_creator(sirius::creator::task_creator* task_
...
+
+  // Join the manager thread so we know it has exited.
+  if (_manager_thread.joinable()) { _manager_thread.join(); }
+
+  // Wait for all in-flight thread-pool tasks to finish.
+  _kiosk.wait_all();
+
+  // Re-enable the kiosk and queue so the executor is ready for the next query.
+  _kiosk.resume();
+  _task_queue.reset();
```
9prady9, line 118: I think after kiosk and task-queue are interrupted, we should wait for manager thread, kiosk-wait-all and then drain the queue because drain queue uses `try_dequeue` it can potentially still have items even if drain returns right away.

Also mpc internal queue is not reset when we call `task_queue.reset`, it just marks the queue as active.

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -152,6 +168,42 @@ void gpu_pipeline_executor::manager_loop()
...
         }
 
+        // Determine retry count and original task ID for this rescheduled attempt.
+        auto* cur_local = dynamic_cast<gpu_pipeline_task_local_state*>(gpu_task->local_state());
+        uint32_t next_retry_count = 1;
+        uint64_t orig_task_id     = gpu_task->get_task_id();
+        if (cur_local && cur_local->retry_count > 0) {
+          next_retry_count = cur_local->retry_count + 1;
+          orig_task_id     = cur_local->original_task_id;
+        }
```
9prady9, line 178: if we use optional for the task id, then we can move all of this logic under the single condition right ?

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -230,6 +285,37 @@ void gpu_pipeline_executor::set_task_creator(sirius::creator::task_creator* task
 
 void gpu_pipeline_executor::drain_leftover_tasks() { _task_queue.drain(); }
 
+void gpu_pipeline_executor::drain_and_wait()
```
9prady9, line 288: same doubt as the other `drain_and_wait` mentioned in my earlier comment.

9prady9 (commented):
I have some comments, but looks like it was merged before I submitted them :D

PR #463, opened 2026-03-13
Update Benchmark Script

bwyogatama:
Update benchmark script to support multi-session on duckdb


mbrobbel (approved):

joosthooz (commented):
Can we make it so that both sirius and duckdb can use the multi-session parameter, and that the iterations are still in the same instance so that the cache still works? That makes it easy to run GDB on an individual query.

Issue #461, opened 2026-03-12
[Build] cudf::strings::like() deprecated warning + pixi resolves libcudf 26.02 instead of required 26.04


mandrews:
## Summary

Two related issues observed during build on commit `e0a42b0316b4fd36fca3560448ceed1fedbf2b2a` (HEAD -> dev, 2026-03-12):

1. The pixi environment resolves `libcudf 26.02.01` rather than the `>= 26.04` version stated in the README
2. `cudf::strings::like()` used in Sirius source is marked `[[deprecated]]` in the resolved cudf version, producing repeated compiler warnings

---

## Environment

| Component | Value |
|---|---|
| OS | Ubuntu 22.04.5 LTS, kernel `6.8.0-1048-gcp` |
| GPU | NVIDIA Tesla T4, CUDA 13.1 |
| Repo commit | `e0a42b0316b4fd36fca3560448ceed1fedbf2b2a` (dev, 2026-03-12) |
| libcudf resolved by pixi | **26.02.01** |
| libcudf required per README | **>= 26.04** |

---

## Issue 1 — pixi resolves libcudf 26.02 instead of 26.04

During CMake configuration, the following is logged:

```
-- Found cudf: .../cudf-config.cmake (found version "26.02.01")
```

The README states `libcudf >= 26.04` as a requirement, but the pixi lockfile resolves an older version. It's unclear whether `pixi.toml` is missing an explicit lower-bound pin for `libcudf`, or whether the lockfile needs to be regenerated.

## Issue 2 — `cudf::strings::like()` deprecated API

During compilation (step ~612/1296), the following deprecation warning fires repeatedly for both `LIKE` and `NOT LIKE` instantiations:

```
src/expression_executor/specializations/gpu_execute_function.cpp:133:40:
warning: 'std::unique_ptr<cudf::column> cudf::strings::like(
  const cudf::strings_column_view&,
  const cudf::string_scalar&,
  const cudf::string_scalar&,
  rmm::cuda_stream_view,
  rmm::device_async_resource_ref
)' is deprecated [-Wdeprecated-declarations]

    auto like = cudf::strings::like(
                ~~~~~~~~~~~~~~~~~~~^
```

Declared deprecated in:
```
.pixi/envs/default/include/cudf/strings/contains.hpp:177:
[[deprecated]] std::unique_ptr<column> like(
```

The affected file and line is:
- `src/expression_executor/specializations/gpu_execute_function.cpp:133`

---

## Questions for the Team

[...]

PR #465, opened 2026-03-13
Use string_view in cudf::strings::like to fix deprecation warnings

mbrobbel:
The cudf::strings::like API now accepts std::string_view instead of cudf::string_scalar for pattern and escape character parameters (rapidsai/cudf#20428).

As reported in #461.


joosthooz (approved):
Nice, thanks

PR #468, opened 2026-03-15
Fix TABLE_SCAN filter crash from projection_ids gaps

ducndh:
## Crash reproduction

This query crashes with `gpu_execution` on any parquet table with 3+ columns:

```sql
CREATE OR REPLACE TABLE t AS
  SELECT (i % 10)::INTEGER AS a, (i * 1.5)::DOUBLE AS b, (i % 200)::INTEGER AS c
  FROM generate_series(1, 100000) t(i);
COPY t TO '/tmp/test.parquet' (FORMAT PARQUET);
DROP TABLE t;
CREATE VIEW v AS SELECT * FROM read_parquet('/tmp/test.parquet');

CALL gpu_execution('SELECT SUM(b) FROM v WHERE a >= 0 AND c >= 50');
-- vector::_M_range_check: __n (which is 2) >= this->size() (which is 2)
-- *** SIGSEGV ***
```

## What goes wrong

1. DuckDB plans a parquet scan with `column_ids = [a, b, c]` and two WHERE filters: `a >= 0` and `c >= 50`
2. Since `a` ranges 0-9, `a >= 0` is always true — DuckDB's optimizer eliminates it
3. Column `a` gets removed from `projection_ids` since it's no longer needed → `projection_ids = [1, 2]` (gap at index 0)
4. The parquet scan only produces batch columns for projected indices, so the batch has **2 columns** (for `b` and `c`)
5. The surviving filter `c >= 50` has `column_index = 2` (its position in `column_ids`). The old code used this directly as the batch column index → **index 2 into a 2-column batch → OOB**

## Fix

Introduce `build_batch_column_map()` that maps each `column_ids` index to its actual batch position by replicating the same sorted-projection logic the parquet scan uses. Both the filter conversion and post-filter projection now go through this map instead of using raw indices.

Also: stop `task_creator` on exception in `gpu_pipeline_executor` to prevent the follow-up SIGSEGV from accessing destroyed operators.

### DIFF
--- a/src/op/sirius_physical_table_scan.cpp
+++ b/src/op/sirius_physical_table_scan.cpp
@@ -17,6 +17,7 @@
 #include "op/sirius_physical_table_scan.hpp"
 
 #include "expression_executor/gpu_expression_executor.hpp"
+#include "log/logging.hpp"
 #include "sirius_config.hpp"
 
 #include <cudf/concatenate.hpp>
@@ -26,6 +27,9 @@
 
 #include <cucascade/data/data_batch.hpp>
 #include <cucascade/data/gpu_data_representation.hpp>
+
+#include <algorithm>
+#include <format>
 
 namespace sirius {
 namespace op {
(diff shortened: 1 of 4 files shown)


src/op/sirius_physical_table_scan.cpp
```diff
@@ -151,25 +209,48 @@ std::unique_ptr<operator_data> sirius_physical_table_scan::execute(const operato
...
+                    projection_ids.size()));
+    }
+
     std::vector<std::shared_ptr<cucascade::data_batch>> projected_batches;
     projected_batches.reserve(output_batches.size());
 
     for (auto& batch : output_batches) {
       if (!batch) { continue; }
 
-      // Release the table from the batch (zero-copy: we're the sole consumer)
```
joosthooz: Consider keeping this comment

joosthooz (commented):
Thank you for contributing this! Just have 2 general questions: Does this also work for DuckDB table scans (not sure how it behaves compared to parquet, I think it might always just produce all columns), and does it still work if a column is selected multiple times?

src/op/sirius_physical_table_scan.cpp
ducndh: 1. Great point as yeah this also happens in DuckDB path. I ahve updated the fix to now always correct for whatever path exist. 
2. it should work if a column is selected multiple times
3. the comment said zero-copy, which is not entirely correct so I change to move. if you want it to be more illustrative about this current problem, I can extend it, but it might be redundant.

joosthooz (approved):
Thank you!

PR #469, opened 2026-03-15
Fix LEFT JOIN crash in gpu_processing materialize kernels

ducndh:
`gpu_processing` query with a LEFT JOIN followed by GROUP BY crashes:

```sql
CALL gpu_processing('
  SELECT l.name, SUM(r.value)
  FROM left_table l LEFT JOIN right_table r ON l.key = r.key
  GROUP BY l.name
');
-- SIGSEGV in materialize kernel
```

## What goes wrong

LEFT JOIN produces null entries for non-matching rows. The materialize kernels in `src/cuda/operator/materialize.cu` didn't handle null validity masks correctly — they read garbage data for null positions, causing out-of-bounds memory access in the CUDA kernels.

## Fix

Added null-awareness to the materialize kernels in `materialize.cu` and added a guard check in `gpu_materialize.cpp` before invoking the kernels.

### DIFF
--- a/src/cuda/operator/materialize.cu
+++ b/src/cuda/operator/materialize.cu
@@ -20,8 +20,16 @@
 
 #include <chrono>
 #include <cmath>
+#include <cudf/join/join.hpp>
 
 namespace duckdb {
+
+// Sentinel value for unmatched rows in LEFT/RIGHT/OUTER JOINs.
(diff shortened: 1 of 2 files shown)


wmalpica (approved):

wmalpica:
Thanks @ducndh , the fix looks good, but the PR needs to be rebased

PR #470, opened 2026-03-15
Fix nproc variable syntax in build commands

yayen-lin:


joosthooz (changes requested):
Hi Yayen-lin, thank you for opening a PR! Because `nproc` is not a set variable, I think we would want this to become `$(nproc)`

yayen-lin:
Oh oops thanks for catching that, I pushed the fixes!

mbrobbel (approved):

joosthooz (approved):

PR #473, opened 2026-03-17
Operator task memory consumption history and  improve peak memory consumption estimation

wmalpica:
Implements #472 

Requires https://github.com/NVIDIA/cuCascade/pull/92
If the above change in cuCascade is not desired, I can make a couple small changes to explicitly account for the host_parquet_representation.

This PR adds the following functions/concepts:
- local_state::get_task_consumption_basis(), which basically for most tasks would be the input number of bytes, but can mean different things for things like scans. 
- local_state::_bytes_to_materialize_input, which contains how many bytes are needed to put the input on the GPU. (this is 0 if the data is already on the GPU)
- global_state::get_memory_history(): this will allow you to `record()` the input task consuption basis and the peak memory consumption. 
- get_estimated_reservation_size() is updated to use the pipeline_memory_history::estimate_peak_memory() which uses a weighted average of historical peak/estimated ratios, and it also accounts for needing to pull data onto the GPU from host or disk.

Design considerations:
- The history is currently per pipeline for simplicity sake. As opposed to having a per operator history. The issue with this is that if a task fails in the middle of a pipeline, then the estimation logic is not as clean. This is ameliorated by the fact that if a task fails, we capture the initial task estimation basis so that it can be used in the task retry and therefore the memory estimation/reservation logic will apply the same as if the task retry was starting from the beginning of the pipeline. 
- If a task fails, we capture the peak consumption during failure, so that retries should then at least reserve as much memory as was being consumed during the OOM.  If the same task fails multiple times, it will record the history of the failure with the highest peak consumption, such that the next attempt will reserve more and ideally succeed. 
- If the data resides off the GPU, the memory estimation will add to the memory reservation how much it needs to hold the data on the GPU (_bytes_to_materialize_input). This value also gets subtracted from the peak consumption recorded for a task since its not really part of the operators themselves and if an operator receives a mixture of data that came from GPU and Host, we want the operator history to represent the actual operator consumption. 
- If materializing data onto the GPU requires any overhead (such as a decompression overhead like for host_parqut_representation), that overhead will not be explicitly accounted for. But if the peak consumption in a pipeline is actually due to this overhead and not from the peak consumption from an operator, then it will be captured by the operator history, which is not accurate, but until a better way of accounting for this is implemented, it does help with memory estimation and tracking. If we later introduce a compressed format that we can use during downgrading, then we dont have a good way to account for that. I have some ideas which can be layered on top of this implementation. One option is to have converters be able to acccuratelly represent their memory needs. Option 2 is that we can create historical tracking of conversions.

A couple things missing for followup PRs:
- This does not properly capture historicals for scan tasks. We need to figure out what to do about parquet scans in general. Right now they can use GPU, but then leave results in CPU. This is not what we want to do in the long run.
- We need to add a way to request downgrades from the downgrade executor


src/include/pipeline/pipeline_memory_history.hpp
```diff
@@ -0,0 +1,134 @@
...
+ * @brief A single observation from one completed task.
+ *
+ * For GPU pipeline tasks, estimated_bytes is the input data size.
+ * For scan tasks, estimated_bytes is the pre-history reservation estimate
+ * (e.g. approximate_batch_size or reserved_compressed_bytes).
+ */
+struct task_memory_record {
+  std::size_t estimated_bytes;              ///< Pre-history estimation basis
+  std::size_t peak_memory_bytes;            ///< Actual peak allocated bytes during execution
+  std::optional<std::size_t> output_bytes;  ///< Total output data size (nullopt if task OOM'd)
```
wmalpica, line 38: I am not actually using output_bytes right now, but I figure it could be useful in the future.

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -414,11 +442,41 @@ void gpu_pipeline_task::execute(rmm::cuda_stream_view stream)
...
+    }
+    throw;
+  }
+
+  // Record memory metrics for future reservation estimates
+  if (output_data) {
+    auto peak_bytes          = _allocator ? _allocator->get_peak_allocated_bytes(stream) : 0;
+    std::size_t output_bytes = 0;
+    for (const auto& batch : output_data->get_data_batches()) {
+      if (batch && batch->get_data()) { output_bytes += batch->get_data()->get_size_in_bytes(); }
```
joosthooz, line 480: With `get_size_in_bytes()` you want to get the full memory footprint here right? I've made a remark about that API in your cucascade PR https://github.com/NVIDIA/cuCascade/pull/92, because right now it doesn't include for example the metadata of a packed representation

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -437,8 +495,16 @@ std::size_t gpu_pipeline_task::get_input_size() const
...
-  // WSM TODO: this is a placeholder for the actual reservation size
-  return get_input_size() * 1;
+  auto input_size =
+    _local_state->cast<gpu_pipeline_task_local_state>().get_task_consumption_basis();
+
+  auto bytes_to_materialize_input =
+    _local_state->cast<gpu_pipeline_task_local_state>().get_bytes_to_materialize_input();
+  auto& global  = _global_state->cast<gpu_pipeline_task_global_state>();
+  auto estimate = global.get_memory_history().estimate_peak_memory(input_size);
+  if (estimate) { return *estimate + bytes_to_materialize_input; }
```
joosthooz, line 521: If the input data was not spilled, it is still accounted for in the reserved memory (it is not subtracted from the global nr of reserved bytes when the reservation object is destroyed, only when the memory is free'd). So I think we only have to reserve additional memory for the input if batches need to be fetched. Do you agree and is there a way to do that?

src/include/pipeline/gpu_pipeline_task.hpp
```diff
@@ -76,12 +76,46 @@ class gpu_pipeline_task_local_state : public sirius_pipeline_task_local_state {
...
+    return *_estimation_basis;
+  }
+
+  [[nodiscard]] std::size_t get_bytes_to_materialize_input() const
+  {
+    std::size_t input_size = 0;
+    if (_input_data) {
+      for (const auto& batch : _input_data->get_data_batches()) {
+        if (batch && batch->get_data() &&
+            batch->get_data()->get_current_tier() != cucascade::memory::Tier::GPU) {
```
joosthooz, line 117: This is where we only add to the reservation for batches that are not already in gpu memory? We should add some documentation about it to this function

src/op/scan/parquet_scan_task.cpp
```diff
@@ -379,6 +379,26 @@ parquet_scan_task::~parquet_scan_task()
   }
 }
 
+void parquet_scan_task::execute(rmm::cuda_stream_view stream)
```
joosthooz, line 382: Can we add this code to the base implementation of execute() instead? And then remove the overrides for both scan classes

src/op/scan/parquet_scan_task.cpp
```diff
@@ -379,6 +379,26 @@ parquet_scan_task::~parquet_scan_task()
...
+{
+  auto& l_state        = this->_local_state->cast<parquet_scan_task_local_state>();
+  auto estimated_bytes = l_state.get_reserved_compressed_bytes();
```
joosthooz, line 385: Why use `l_state.get_reserved_compressed_bytes();` here while the duckdb version uses `estimated_bytes = get_estimated_reservation_size();`?

src/include/pipeline/pipeline_memory_history.hpp
```diff
@@ -0,0 +1,132 @@
...
+      // has the same weight decay as 500MB vs 1000MB.
+      double log_ratio = std::abs(
+        std::log(static_cast<double>(rec.estimated_bytes) / static_cast<double>(estimated_bytes)));
+      double weight = 1.0 / (1.0 + log_ratio);
+
+      weighted_ratio_sum += ratio * weight;
+      weight_sum += weight;
+    }
+
+    if (weight_sum == 0.0) { return std::nullopt; }
```
9prady9: We already check for empty records in line 83. So, I assuming this for accumulation being actually close to zero. Comparing double with 0.0 is not a good idea, perhaps compare something like `abs() > some-epsilon` is what we are after here ?

If we are after it being closer to standard epsilon, we have `std::numeric_limits<T>::epsilon()` that we can use here.

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -387,7 +387,35 @@ void gpu_pipeline_task::execute(rmm::cuda_stream_view stream)
...
         first_op.get_operator_id());
     }
-    auto handle = lock_or_prepare_batch(batch, requested_memory_space, stream);
+    std::optional<cucascade::data_batch_processing_handle> handle;
+    try {
+      handle = lock_or_prepare_batch(batch, requested_memory_space, stream);
+    } catch (const rmm::out_of_memory& oom) {
+      auto peak_bytes  = allocator->get_peak_allocated_bytes(stream);
+      auto input_basis = local_state.get_task_consumption_basis();
+      // If the number of bytes consumed is greater than the reservation size, record the peak
```
joosthooz: But will the allocator not throw this exception when trying to allocate beyond the reservation? So can we ever see a peak here that is higher than the reservation?

src/op/scan/parquet_scan_task.cpp
joosthooz: Actually Claude says this is correct and the other is wrong. I have to say I don't fully understand why.

src/include/pipeline/pipeline_memory_history.hpp
```diff
@@ -0,0 +1,132 @@
...
+  [[nodiscard]] std::size_t size() const
+  {
+    std::lock_guard<std::mutex> lock(_mutex);
+    return _records.size();
+  }
+
+ private:
+  static constexpr std::size_t kMaxRecords = 64;
+
+  mutable std::mutex _mutex;
```
9prady9, line 157: how often is this mutex locked for reading vs storing records ? how much simultaneous access we are looking for here ?

The reason I ask is - a thought crossed my mind - is this location worth using shared mutex instead of regular mutex to enable simultaneous reads.

src/include/pipeline/pipeline_memory_history.hpp
```diff
@@ -0,0 +1,132 @@
...
+      weight_sum += weight;
+    }
+
+    if (weight_sum == 0.0) { return std::nullopt; }
+
+    double avg_ratio = weighted_ratio_sum / weight_sum;
+
+    auto result = static_cast<std::size_t>(static_cast<double>(estimated_bytes) * avg_ratio);
+
+    // Never estimate less than the estimation basis itself
```
joosthooz: Why not, what if the basis is off and we can converge to something better?
Do we want to add a trace print for now, so we can see how the new estimation code behaves?

test/cpp/pipeline/test_pipeline_memory_history.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+TEST_CASE("pipeline_memory_history single record produces correct estimate",
+          "[pipeline_memory_history][estimation]")
+{
+  pipeline_memory_history history;
+
+  // Record: estimated 100 bytes, peaked at 300 bytes → ratio = 3.0
+  history.record({100, 300, 200});
+
+  // For a new task with 100 bytes estimated, expect 100 * 3.0 * 1.2 = 360
```
joosthooz: The 1.2 values here and below might be outdated

src/include/pipeline/sirius_pipeline_task_states.hpp
```diff
@@ -103,6 +114,17 @@ class sirius_pipeline_task_local_state : public parallel::itask_local_state {
...
+  /**
+   * @brief Get the basis for estimating task memory consumption.
+   *
+   * This method allows for different task types to provide their own logic for providing something
+   * as a starting point for memory reservation estimation.
+   *
+   * @return The value to use as the basis for memory consumption estimation (e.g., input data size,
+   * number of rows, etc.)
+   */
+  [[nodiscard]] virtual std::size_t get_task_consumption_basis() const = 0;
```
9prady9, line 136: wondering if we should use an opaque object, even a simple struct here, to avoid changing APIs later if we want to capture more as basis-info ?

src/include/pipeline/gpu_pipeline_task.hpp
9prady9, line 117: this is more of a clarification for me, so we can have batches from multiple tiers grouped together ?

joosthooz (commented):
Thanks, I posted some questions, I hope they are useful

src/include/op/scan/parquet_scan_task.hpp
```diff
@@ -424,8 +435,12 @@ class parquet_scan_task : public pipeline::sirius_pipeline_itask {
...
   [[nodiscard]] size_t get_estimated_reservation_size() const override
   {
-    auto& l_state = this->_local_state->cast<parquet_scan_task_local_state>();
-    return l_state.get_reserved_compressed_bytes();
+    auto current_estimate =
+      this->_local_state->cast<parquet_scan_task_local_state>().get_task_consumption_basis();
+    auto& g_state = this->_global_state->cast<parquet_scan_task_global_state>();
+    auto refined  = g_state.get_memory_history().estimate_peak_memory(current_estimate);
+    if (refined) { return *refined; }
+    return current_estimate;
```
9prady9: move this implementation to cpp instead ?

src/include/op/scan/duckdb_scan_task.hpp
```diff
@@ -490,8 +495,12 @@ class duckdb_scan_task : public sirius::pipeline::sirius_pipeline_itask {
...
   std::size_t get_estimated_reservation_size() const override
   {
-    return this->_local_state->cast<duckdb_scan_task_local_state>()
-      .get_estimated_reservation_size();
+    auto current_estimate =
+      this->_local_state->cast<duckdb_scan_task_local_state>().get_task_consumption_basis();
+    auto& g_state = this->_global_state->cast<duckdb_scan_task_global_state>();
+    auto refined  = g_state.get_memory_history().estimate_peak_memory(current_estimate);
+    if (refined) { return *refined; }
+    return current_estimate;
```
9prady9: move to cpp ?

src/include/data/host_parquet_representation.hpp
```diff
@@ -200,11 +200,11 @@ class host_parquet_representation : public cucascade::idata_representation {
   [[nodiscard]] std::size_t get_size_in_bytes() const override { return _size_in_bytes; }
 
   /**
-   * @brief Gets the uncompressed size of the data represented by this representation.
+   * @brief Gets the uncompressed (logical) size of the data represented by this representation.
```
9prady9, line 203: is there any additional significance to calling them logical ? Wondering if we call it uncompressed directly .

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -529,7 +530,20 @@ void duckdb_scan_task::process_chunk(duckdb_scan_task_local_state& l_state)
...
+  if (output_data) {
+    std::size_t output_bytes = 0;
+    for (const auto& batch : output_data->get_data_batches()) {
+      if (batch && batch->get_data()) { output_bytes += batch->get_data()->get_size_in_bytes(); }
+    }
+    auto& g_state = _global_state->cast<duckdb_scan_task_global_state>();
+    g_state.get_memory_history().record({estimated_bytes, output_bytes, output_bytes});
+  }
+
   if (output_data) { publish_output(*output_data, stream); }
```
9prady9: Can't we do the following instead ? seems all of the logic should execute only if output_data is valid anyway.

```suggestion
  if (auto output_data = compute_task(stream); output_data) {
    auto estimated_bytes = get_estimated_reservation_size();
    std::size_t output_bytes = 0;
    for (const auto& batch : output_data->get_data_batches()) {
      if (batch && batch->get_data()) { output_bytes += batch->get_data()->get_size_in_bytes(); }
    }
    auto& g_state = _global_state->cast<duckdb_scan_task_global_state>();
    g_state.get_memory_history().record({estimated_bytes, output_bytes, output_bytes});
    
    publish_output(*output_data, stream);
  }
```

src/op/scan/parquet_scan_task.cpp
9prady9, line 385: similar update as the duckdb change, these are needed only if output_data is valid right ? or am I missing something here ?

9prady9 (commented):
left some comments, please check.

src/pipeline/gpu_pipeline_task.cpp
wmalpica: Yes, ideally I want the full memory footprint here, but since this is used for memory estimations for reservations, it does not have to be 100% accurate to do what we need

src/include/data/host_parquet_representation.hpp
wmalpica: No additional significance really. We could call it compressed vs uncompressed. Logical vs physical.

src/include/pipeline/gpu_pipeline_task.hpp
wmalpica: Yes, this is to see how many bytes need to be moved to the GPU prior to processing, and we need to add that to how much we need to reserve

wmalpica: Yes. It is possible that some batches were downgraded and some were not.

src/include/pipeline/pipeline_memory_history.hpp
wmalpica: If there were some records that had estimated_bytes == 0, then you could end up with that for loop actually accumulating nothing. THat is what that is checking for. In that case the weight_sum == 0.0 should work, despite being a double.

src/include/pipeline/pipeline_memory_history.hpp
wmalpica: I am more concerned about the estimation logic being off somehow and breaking things, which is why i like this max. But I could take it off.
As for trace logging, we do have some. Whenever we record when a task is completed, i am logging the peak consumption, input bytes, output bytes and reservation size

src/include/pipeline/sirius_pipeline_task_states.hpp
wmalpica: This is an internal API, if we need to change it Claude can help us doing easily. I dont want to introduce opaqueness for a just in case future thing

src/pipeline/gpu_pipeline_task.cpp
wmalpica: If the input data is already on the GPU (was not spilled), then its already accounted for (its not part of the free memory available). It is also not part of any reservation, since the reservation that created it has been freed. Its just allocated memory. So you are correct, `we only have to reserve additional memory for the input if batches need to be fetched`. That is what `bytes_to_materialize_input` represents. If the input data is already on the GPU, then `bytes_to_materialize_input` is 0. If not then we add that to our reservation needs.

src/pipeline/gpu_pipeline_task.cpp
wmalpica: No, we dont throw when we try to allocate past the reservation, only when the allocator cannot allocate due to its upper capacilty constraint.

src/op/scan/parquet_scan_task.cpp
wmalpica: We need to fix how a lot of this works for scan tasks in general. The scan executor works in a very different way, in fact these execute functions are not even being used in the code today. I made these updates just to pave the way towards the fixes that need to be made

src/op/scan/parquet_scan_task.cpp
wmalpica: I am not 100% sure either. I am besing myself partiall on how some of this works already

src/include/pipeline/pipeline_memory_history.hpp
wmalpica: There is basically one read for every write. For every task we do an estimate, and for every task we write a record. I dont think its worth it to use a shared mutex

src/include/pipeline/pipeline_memory_history.hpp
wmalpica: Either way, i changed the logic so that we should never end up with that by not allowing recording estimated_bytes == 0

src/include/pipeline/sirius_pipeline_task_states.hpp
9prady9: Claude :D coming up everywhere 👍🏾 tech debt is thing of past I guess

9prady9 (commented):
updates look good but we still have the `Attempted to dereference shared_ptr that is NULL` error from previous runs. Seems like we are using some `*` or `->` operator on shared_ptr while it wasn't set yet.

src/include/data/host_parquet_representation.hpp
9prady9: as long as we are consistent across different projects 👍🏾

src/include/pipeline/pipeline_memory_history.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+      // Look for an existing failure record with the same estimated_bytes
+      for (auto& rec : _records) {
+        if (rec.estimated_bytes == estimated_bytes && !rec.output_bytes.has_value()) {
+          rec.peak_memory_bytes = std::max(rec.peak_memory_bytes, peak_memory_bytes);
+          return;
+        }
+      }
+    }
+    this->record({estimated_bytes, peak_memory_bytes, std::nullopt});
```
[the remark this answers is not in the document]
9prady9: my understanding is you want to insert a new fresh record only when no failed records found. misunderstanding from co-pilot ?

src/include/pipeline/sirius_pipeline_task_states.hpp
```diff
@@ -100,9 +111,23 @@ class sirius_pipeline_task_local_state : public parallel::itask_local_state {
    */
   void set_reservation(std::unique_ptr<cucascade::memory::reservation> res)
   {
-    _reservation = std::move(res);
+    _reservation       = std::move(res);
+    _reservation_bytes = _reservation->size();
   }
```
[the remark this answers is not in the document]
9prady9: +1

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -277,18 +276,25 @@ std::unique_ptr<op::operator_data> gpu_pipeline_task::compute_task(rmm::cuda_str
       operator_input_output_data = run_one_operator(
         op, *operator_input_output_data, stream, pipeline, i, operators.size(), _allocator);
     } catch (const rmm::out_of_memory& oom) {
-      auto peak_bytes        = _allocator ? _allocator->get_peak_allocated_bytes(stream) : 0;
+      auto peak_bytes = _allocator ? _allocator->get_peak_allocated_bytes(stream) : 0;
+      // Subtract the peak allocated bytes to the input data to get the peak allocated bytes for the
+      // operators.
+      peak_bytes -= local_state._bytes_to_materialize_input;
```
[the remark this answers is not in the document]
9prady9: +1

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -387,7 +405,32 @@ void gpu_pipeline_task::execute(rmm::cuda_stream_view stream)
...
-    auto handle = lock_or_prepare_batch(batch, requested_memory_space, stream);
+    std::optional<cucascade::data_batch_processing_handle> handle;
+    try {
+      handle = lock_or_prepare_batch(batch, requested_memory_space, stream);
+    } catch (const rmm::out_of_memory& oom) {
+      auto peak_bytes  = allocator->get_peak_allocated_bytes(stream);
+      auto input_basis = local_state.get_task_consumption_basis();
+      auto& global     = _global_state->cast<gpu_pipeline_task_global_state>();
+      global.get_memory_history().record_on_failure(input_basis, peak_bytes);
+
```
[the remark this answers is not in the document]
wmalpica: In general we subtract _bytes_to_materialize_input from peak_bytes because we want to log the consumption AFTER the lock_or_prepare_batch. In general, lock_or_prepare_batch should never really OOM. If it is OOMing its because of memory fragmentation or because its a materialization that has some additional overhead, for example a host_parquet_representation. This code path is here, more to capture the host_parquet_representation situation. In this case we do want to capture the full peak consumption because the intent is to help the next task reserve more memory than it did when it OOMed. It is true that it would create a bit of inconsistency, but it would be an inconsistency which would tend towards guiding the memory reservation system to reserve more when there is an OOM, therefore helping prevent a future one.

src/include/pipeline/pipeline_memory_history.hpp
```diff
@@ -0,0 +1,155 @@
...
+        std::log(static_cast<double>(rec.estimated_bytes) / static_cast<double>(estimated_bytes)));
+      double weight = 1.0 / (1.0 + log_ratio);
+
+      weighted_ratio_sum += ratio * weight;
+      weight_sum += weight;
+    }
+
+    double avg_ratio = weighted_ratio_sum / weight_sum;
+
+    return static_cast<std::size_t>(static_cast<double>(estimated_bytes) * avg_ratio);
```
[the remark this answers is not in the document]
wmalpica: Good catch. That unit test was created before a minor design change. I deleted that unit test

src/include/pipeline/pipeline_memory_history.hpp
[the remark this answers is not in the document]
wmalpica: Copilot is just complaining about the atomicity of the operation (and its correct) because the way the code is now it locks, checks, releases the lock, then locks to record.  Claude had originally suggested the same thing copilot is suggesting, i just though it was a lot of code duplication for something unlikely and not dangerous. But on revision its not that much code to duplicate.

test/cpp/pipeline/test_gpu_pipeline_task_history.cpp
```diff
@@ -0,0 +1,597 @@
...
+
+  auto global_state =
+    std::make_shared<sirius::pipeline::sirius_pipeline_task_global_state>(ctx.pipeline);
+
+  // Task 1: execute successfully with 20 MB input and 20 MB execute allocation
+  auto task1 =
+    create_pipeline_task(f, global_state, std::move(input_batch), kReservationSize1, /*task_id=*/1);
+
+  task1->execute(stream);
+  task1->mark_as_rescheduled();
```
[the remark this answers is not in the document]
wmalpica: I mark as rescheduled so the destructor does not call mark_task_completed() (which would dereference the pipeline's null source operator). So its not something that happens in the engine normally, but it makes writing this unit test easier.

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -529,8 +530,20 @@ void duckdb_scan_task::process_chunk(duckdb_scan_task_local_state& l_state)
...
+    std::size_t output_bytes = 0;
+    for (const auto& batch : output_data->get_data_batches()) {
+      if (batch && batch->get_data()) { output_bytes += batch->get_data()->get_size_in_bytes(); }
+    }
+    auto& g_state = _global_state->cast<duckdb_scan_task_global_state>();
+    g_state.get_memory_history().record({estimated_bytes, output_bytes, output_bytes});
+
+    publish_output(*output_data, stream);
+  }
 }
```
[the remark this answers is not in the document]
wmalpica: Correct. This is not used in production. I put this code here to setup or be used  in a future PR

src/op/scan/parquet_scan_task.cpp
```diff
@@ -379,6 +379,25 @@ parquet_scan_task::~parquet_scan_task()
...
+    std::size_t output_bytes = 0;
+    for (const auto& batch : output_data->get_data_batches()) {
+      if (batch && batch->get_data()) { output_bytes += batch->get_data()->get_size_in_bytes(); }
+    }
+    auto& g_state = this->_global_state->cast<parquet_scan_task_global_state>();
+    g_state.get_memory_history().record({estimated_bytes, output_bytes, output_bytes});
+
+    publish_output(*output_data, stream);
+  }
+}
```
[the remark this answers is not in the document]
wmalpica: Correct. This is not used in production. I put this code here to setup or be used in a future PR

test/cpp/pipeline/test_gpu_pipeline_task_history.cpp
```diff
@@ -0,0 +1,597 @@
...
+  auto task_reservation2 = f.manager->request_reservation(
+    cucascade::memory::any_memory_space_in_tier{cucascade::memory::Tier::GPU}, estimation2);
+  REQUIRE(task_reservation2 != nullptr);
+  auto* local_state2_ptr =
+    dynamic_cast<sirius::pipeline::sirius_pipeline_task_local_state*>(task2->local_state());
+  REQUIRE(local_state2_ptr != nullptr);
+  local_state2_ptr->set_reservation(std::move(task_reservation2));
+
+  task2->execute(stream);
+  task2->mark_as_rescheduled();
```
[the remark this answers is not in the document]
wmalpica: I mark as rescheduled so the destructor does not call mark_task_completed() (which would dereference the pipeline's null source operator). So its not something that happens in the engine normally, but it makes writing this unit test easier.

9prady9 (approved):

PR #498, opened 2026-03-17
Use probing hash for mixed_join

joosthooz:
This PR adds some join optimizations targeting Q21 particularly, by converting the mixed_join into a probing hash combined with a post-processing step. This lifts the barrier on the large side of the join, allowing data to stream, overlapping a large part of the computation.

Some other optimizations are included like prioritizing scans on the build side of joins.
In the current state of this PR Q1 has a severe regression, I've got a fix for that but it sometimes stalls the engine. So once I fix that I will push it to this branch.

This code improves Q21 from 4.3s to 2.7s, and some other queries too (Q7, Q9, Q17).

### DIFF
--- a/src/op/sirius_physical_hash_join.cpp
+++ b/src/op/sirius_physical_hash_join.cpp
@@ -490,15 +490,6 @@ void sirius_physical_hash_join::build_pipelines(pipeline::sirius_pipeline& curre
 void sirius_physical_hash_join::update_join_exec_mode(int num_partitions, uint64_t build_side_bytes)
 {
   std::lock_guard<std::mutex> lg(op_state_mutex);
-  SIRIUS_LOG_INFO(
-    "update_join_exec_mode id {} join_type={} num_partitions={} build_side_bytes={} "
-    "current_mode={} max_build_bytes={}",
-    this->get_operator_id(),
-    duckdb::JoinTypeToString(join_type),
-    num_partitions,
-    build_side_bytes,
-    static_cast<int>(_join_mode),
-    _max_build_hash_table_bytes);
   if (num_partitions == 1 && join_type != duckdb::JoinType::RIGHT &&
       _join_mode != HASH_JOIN_MODE::BUILD_PROBE) {
     bool has_inequality = (num_equality_conditions < conditions.size());
@@ -513,13 +504,9 @@ void sirius_physical_hash_join::update_join_exec_mode(int num_partitions, uint64
       // Keep STANDARD mode.
     } else if (!has_inequality || join_type == duckdb::JoinType::INNER) {
       _join_mode = HASH_JOIN_MODE::BUILD_PROBE;
-      SIRIUS_LOG_INFO(
-        "sirius_physical_hash_join id {} switching to BUILD_PROBE mode with {} partitions and "
-        "build side size {} bytes{}",
-        this->get_operator_id(),
-        num_partitions,
-        build_side_bytes,
-        has_inequality ? " (with inequality post-filter)" : "");
+      SIRIUS_LOG_DEBUG("sirius_physical_hash_join id {} switching to BUILD_PROBE mode{}",
+                       this->get_operator_id(),
+                       has_inequality ? " (with inequality post-filter)" : "");
     }
   }
 }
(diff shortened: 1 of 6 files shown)


src/op/sirius_physical_hash_join.cpp
```diff
@@ -374,19 +490,24 @@ void sirius_physical_hash_join::build_pipelines(pipeline::sirius_pipeline& curre
 void sirius_physical_hash_join::update_join_exec_mode(int num_partitions, uint64_t build_side_bytes)
 {
   std::lock_guard<std::mutex> lg(op_state_mutex);
-  if (num_partitions == 1 && build_side_bytes < _max_build_hash_table_bytes &&
```
joosthooz, line 377: Note the `build_side_bytes < _max_build_hash_table_bytes` is removed here, according to Claude it is not needed because we are already checking `num_partitions == 1` which supposedly means we already checked the workload fits in memory. If that is true we can remove the entire config parameter. But actually I'm considering to keep the check here to give us an additional tuning parameter. If we keep it, we need to increase the size considerably to allow queries at sf1k to benefit from it (e.g. I've seen sizes of 6GB, we currently use 5GB limit, the default setting is 0.5GB) @wmalpica

src/planner/query.cpp
```diff
@@ -51,6 +55,20 @@ void query::build_indices()
       _operator_to_pipeline[&op_ref.get()] = pipeline;
     }
   }
+
+  // Prioritize build-side scans: hash join build partitions must complete before
+  // probe partitions can determine num_partitions, so starting build-side I/O
+  // first reduces the critical path.
+  std::stable_sort(_scan_operators.begin(), _scan_operators.end(), [this](auto* a, auto* b) {
```
wmalpica, line 62: If we have N joins, would this prioritize all N build side scans first? before any of the probe side scans? if so, then that will cause other problems.

src/pipeline/pipeline_executor.cpp
```diff
@@ -161,7 +161,15 @@ std::future<void> pipeline_executor::start_query()
     gpu_exec->set_completion_handler(_completion_handler.get());
   }
 
-  schedule_next_scan_tasks();
+  // Schedule all scans. Build-side scans appear first (sorted in query::build_indices()),
+  // so their I/O tasks are enqueued before probe-side scans.
+  {
+    std::lock_guard<std::mutex> lock(_priority_scans_mutex);
+    while (!_priority_scans.empty()) {
```
wmalpica, line 168: I think we need to be careful about scheduling too many scans all up front

src/op/sirius_physical_hash_join.cpp
wmalpica: The idea is that _max_build_hash_table_bytes is a tuning parameter where we can use it to limit how big is the data we are commiting to keep in GPU memory for the duration of this join operator. 
Also, num_partitions == 1 is a pre-requisite for this but the data also needs to fit in one batch, which is not guaranteed by num_partitions  == 1. I have a enforcer in the configuration settings that ensures that the concat size in bytes configurations has to be larger than _max_build_hash_table_bytes so that it guarantees that what arrives will be 1 batch

src/op/sirius_physical_hash_join.cpp
```diff
@@ -453,20 +580,21 @@ std::unique_ptr<operator_data> sirius_physical_hash_join::get_next_task_input_da
...
       throw std::runtime_error(
         "In sirius_physical_hash_join:get_next_task_input_data_for_build_probe: expected exactly 1 "
-        "partition and 1 batch in default (build) port in operator " +
+        "partition in build and probe ports in operator " +
         std::to_string(this->get_operator_id()));
     }
-    // When the hash table is not build yet, we will send both the build and probe side. To build
-    // the hash table and perform the first join.
+    // Pop the first probe batch and ALL build batches. The execute() method will
+    // concatenate multiple build batches into one table before building the hash table.
```
wmalpica, line 590: There is a concatenate operator which should be concatenating batches before the hash join operator. See my comment here https://github.com/sirius-db/sirius/pull/498/changes#r2949851427

joosthooz:
Superseded by https://github.com/sirius-db/sirius/pull/507, because many of the intricate changes in this PR are not really needed to see gains.

PR #366, opened 2026-02-27
Add `vcpkg` build support

mbrobbel:
Closes #547.

Adds `vcpkg` build support to build a fully statically linked extension for distribution:
```
ldd build/vcpkg-release/extension/sirius/sirius.duckdb_extension
          linux-vdso.so.1 (0x00007fb1b8676000)
          libdl.so.2 => /nix/store/l0l2ll1lmylczj1ihqn351af2kyp5x19-glibc-2.42-51/lib/libdl.so.2 (0x00007fb1a37f9000)
          librt.so.1 => /nix/store/l0l2ll1lmylczj1ihqn351af2kyp5x19-glibc-2.42-51/lib/librt.so.1 (0x00007fb1a37f4000)
          libpthread.so.0 => /nix/store/l0l2ll1lmylczj1ihqn351af2kyp5x19-glibc-2.42-51/lib/libpthread.so.0 (0x00007fb1a37ef000)
          libstdc++.so.6 => not found
          libm.so.6 => /nix/store/l0l2ll1lmylczj1ihqn351af2kyp5x19-glibc-2.42-51/lib/libm.so.6 (0x00007fb1a36f7000)
          libgcc_s.so.1 => /home/matthijs/sirius-vcpkg/.pixi/envs/vcpkg/lib/libgcc_s.so.1 (0x00007fb1a36c8000)
          libc.so.6 => /nix/store/l0l2ll1lmylczj1ihqn351af2kyp5x19-glibc-2.42-51/lib/libc.so.6 (0x00007fb1a3000000)
          /nix/store/l0l2ll1lmylczj1ihqn351af2kyp5x19-glibc-2.42-51/lib64/ld-linux-x86-64.so.2 (0x00007fb1b8678000)
```

PR #500, opened 2026-03-17
CI: Add benchmark runner Docker image and build workflow

mike-wendt:
## Summary

- Adds `.github/docker/benchmark-runner/Dockerfile` — a multi-arch benchmark runner image based on `nvidia/cuda:13.0.2-devel-ubuntu22.04` with pixi v0.65.0 pre-installed
  - Optional `BAKE_SCCACHE=true` build-arg pre-warms the sccache from a full dev build at image creation time, so subsequent benchmark runs only compile the delta
- Adds `.github/workflows/benchmark-runner-image.yml` — builds and publishes the image to `ghcr.io/sirius-db/sirius/benchmark-runner` as a multi-arch manifest (amd64 + arm64)
  - Triggers: PRs touching the Dockerfile or workflow (for validation), Dockerfile changes on `dev`, nightly cron (with sccache bake), manual `workflow_dispatch`
  - Native builds per arch (no QEMU), merge job creates the manifest
  - PRs push to `:pr-latest` (single overwritten tag) so changes can be tested before merge without touching `:latest`
  - Pushes to `:latest` and `:<sha>` on merge to `dev`
  - BuildKit layer cache stored in the registry (`type=registry`) to avoid consuming the 10 GB GHA cache limit
- Engineers can reproduce benchmark runs locally using `act`:
  ```
  act --container-options "--gpus all" \
      -P self-hosted=ghcr.io/sirius-db/sirius/benchmark-runner:latest
  ```

## Test plan
- [x] Add actions to allowlist to enable testing
- [x] Move PR to upstream branch so secrets are available to CI
- [x] Verify the PR build completes and `:pr-latest` is published
- [x] Pull `:pr-latest` on zeno-04 and verify the image works as expected (`nvidia-smi` passes)
- [x] Verify the multi-arch manifest covers both amd64 and arm64
- [ ] After merge, confirm nightly cron builds and publishes `:latest`

### DIFF
--- a/.github/docker/benchmark-runner/Dockerfile
+++ b/.github/docker/benchmark-runner/Dockerfile
@@ -1,0 +1,60 @@
+# Benchmark runner image for Sirius TPC-H benchmarks.
+#
+# Provides the pixi environment manager, which installs the full CUDA toolchain
+# and all Sirius build dependencies. The CUDA runtime is injected at container
+# startup by nvidia-container-toolkit on the host — no CUDA base layer needed.
+#
+# Built natively for linux/amd64 (T4 runners, developer workstations) and
+# linux/arm64 (DGX Spark) and published as a single multi-arch manifest at:
+#   ghcr.io/sirius-db/sirius/benchmark-runner:latest
+#
+# BAKE_SCCACHE=true triggers a full dev build during image construction to
+# pre-warm the sccache. Pass this in the nightly workflow only — a developer
+# pulling the image gets yesterday's dev cache and only compiles the delta.
+# Default is false so toolchain-only rebuilds are fast.
+#
+# Local usage with act:
+#   act --container-options "--gpus all" \
+#       -P self-hosted=ghcr.io/sirius-db/sirius/benchmark-runner:latest
+#
+# Or with a local sccache mount (skips cache bake, uses your own cache):
+#   act --container-options "--gpus all -v $HOME/.cache/sccache:/root/.cache/sccache"
+
+FROM nvidia/cuda:13.0.2-devel-ubuntu22.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install -y \
+    curl git ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install pixi — manages the full CUDA toolchain and all Sirius dependencies.
+# Version is pinned to match check.yml (pixi-version: v0.65.0).
+ARG PIXI_VERSION=0.65.0
+RUN curl -fsSL https://pixi.prefix.dev/install.sh \
+    | PIXI_VERSION=${PIXI_VERSION} bash -s -- --no-modify-path
+ENV PATH="/root/.pixi/bin:$PATH"
+
+# sccache cache directory. Populated by the nightly bake (BAKE_SCCACHE=true)
+# or by mounting a local cache directory at runtime.
+ENV SCCACHE_DIR=/root/.cache/sccache
+
+# --- Nightly sccache bake ---
+# Clone dev, run a full build to populate sccache, then discard the source and
+# build artifacts. Only the sccache directory is kept in the image layer.
+# This reduces incremental build time for PRs and on-demand benchmarks.
+ARG BAKE_SCCACHE=false
+ARG SIRIUS_REF=dev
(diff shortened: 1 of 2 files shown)


mike-wendt:
This is pointing to my repo for the time being. In order to move forward I'd like to merge this as-is with a follow up PR to move to this repo when packages are enabled. I have a request for this pending, I'm just unsure how long it will take to activate. Let me know what you think @mbrobbel in your review. Thanks in advance

mbrobbel (commented):
Thanks for setting this up Mike! Some questions/feedback;

- Maybe we can use https://github.com/prefix-dev/pixi-docker/?
- Do you want to merge this (with your PAT?)
- Instead of building in the container we could also build in the runner (it'll use/update our sccache GH cache) and then copy the result into the container with the required pixi env for runtime deps (see example in pixi-docker)?
- What do you think of the following approach: we produce `sirius.duckdb_extension` artifacts (cuda 12 and 13, arm64 and x86) for builds on `dev`. And then the benchmark flow is: grab the supported duckdb-cli and latest `dev` artifact, load the extension (`unsigned`), run the benchmarks. Initially we'll need the pixi env with runtime deps, but I'm working on a fully statically linked version in #366. Then we just need a compatible cuda driver installed.

mike-wendt:
> Thanks for setting this up Mike! Some questions/feedback;
> 
> * Maybe we can use https://github.com/prefix-dev/pixi-docker/?

This works for me, happy to leverage this instead

> * Do you want to merge this (with your PAT?)

Ideally I'd like switch this to the packages in the repo instead, there is an open request for this but no action yet

> * Instead of building in the container we could also build in the runner (it'll use/update our sccache GH cache) and then copy the result into the container with the required pixi env for runtime deps (see example in pixi-docker)?

This was exactly what I was thinking. In a new benchmark workflow I do have it leveraging the artifacts to seed things like the build and sccache. I just split the work up into this first PR as the level of changes were hard to track. I can rework this though that we do the implementation in one PR if that is more helpful. It will also give more time for the open request on rights to publish packages to be resolved. More details on the workflow implementation and use cases below

> * What do you think of the following approach: we produce `sirius.duckdb_extension` artifacts (cuda 12 and 13, arm64 and x86) for builds on `dev`. And then the benchmark flow is: grab the supported duckdb-cli and latest `dev` artifact, load the extension (`unsigned`), run the benchmarks. Initially we'll need the pixi env with runtime deps, but I'm working on a fully statically linked version in #366. Then we just need a compatible cuda driver installed.

I believe this works, my goal is to use GHA workflows to run in a few scenarios all leveraging the docker container as their base for a consistent benchmark environment. Here are the use cases I see:
1. Post-merge `dev` benchmark testing - from what you describe above I think your workflow fits this use case
2. Pre-merge `dev` benchmark testing OR on-demand PR benchmark testing - my idea was a benchmark GHA workflow that I'm working on checks to see if the artifacts exist from a previous build and pulls those into the container to skip the build process, then runs the benchmarks and uploads the results as artifacts
  - For the pre-merge side this could couple with an auto-merge bot that would check the results for regressions and merge if there are non-detected
3. Infra benchmark testing - leverage the container and the GHA workflow dispatch via `act` to run this locally, there is no local build so we use the workflow from 1 to get the artifacts and run
4. Dev local benchmark testing and troubleshooting - leveraging the container and the GHA workflow use dispatch via `act` to pull the local `sirius` build and do a test
  - The idea here that devs may see a regression in a commit and want to iterate on troubleshooting it locally with the benchmark runs as a means to do this

FWIW I'm basing the benchmark work around the tooling already in this repo as I know that will be the best way to keep it updated and maintained. Especially considering that we may move past TPC-H benchmarking in the near term, I was looking for a way to reuse the workflow for all of these use cases and their derivations, instead of maintaining a test suite for each location. Open to suggestions on this, I haven't posted the benchmark workflow work yet as I'm still trying to understand all of the metadata, query results, query plans, logs, and more that we'd want for reference and debugging. 

I should be clear that what is out of scope for this effort is aggregating and reporting the results. This eventual workflow will just need to capture the necessary data and then save it locally or in CI save it as an artifact. From there we'll have a separate reporting tool

Happy to discuss this further together if it helps

mike-wendt:
This works and publishes a [container](https://github.com/sirius-db/sirius/pkgs/container/sirius%2Fbenchmark-runner) but we will continue to discuss the best long-term approach here before moving forward

PR #503, opened 2026-03-18
Coalesce small scan batches to reduce GPU task overhead (#494)

### LINKED ISSUE #494: Support concatenating multiple scan outputs when smaller than default batch size
Support concatenating multiple scan output batches when they are smaller than the default batch size.

This reduces the overhead of processing many small batches by combining them into appropriately-sized batches.

Reference: #230

---
*Generated by Claude*

ducndh:
## Summary

Closes #494.

When scanning many small parquet files, each file produces a small GPU batch, leading to many tiny GPU pipeline tasks with high per-task scheduling and kernel launch overhead.

This change adds two complementary mechanisms:

1. Batch coalescing in get_next_task_input_data(): for single-port, non-sink operators (i.e. TABLE_SCAN), pop up to 32 batches (this is currently hard-coded so feel free to suggest fix) or DEFAULT_SCAN_TASK_BATCH_SIZE bytes from the port instead of one.

2. GPU table concatenation in TABLE_SCAN::execute(): when multiple batches are received, concatenate them via cudf::concatenate into a single table before filtering/projecting, enabling fewer, larger kernel launches.

I tested on TPC-H SF=10 by splitting a large parquet file into 100 small files then run both versions

## Testing results

TPC-H SF=10, warm runs (GPU-cached), Quadro RTX 6000:

| Scenario | Baseline | With fix | Change |
|---|---|---|---|
| Normal parquet (few large files) | 1.560s | 1.609s | +3% (noise) |
| Split parquet (100 small files) | 2.377s | 1.587s | **-33%** |

### DIFF
--- a/src/op/sirius_physical_operator.cpp
+++ b/src/op/sirius_physical_operator.cpp
@@ -19,7 +19,6 @@
 #include "gpu_executor.hpp"
 #include "pipeline/sirius_meta_pipeline.hpp"
 #include "pipeline/sirius_pipeline.hpp"
-#include "sirius_config.hpp"
 
 #include <cucascade/data/data_batch.hpp>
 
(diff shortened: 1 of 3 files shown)


src/op/sirius_physical_table_scan.cpp
```diff
@@ -111,7 +113,36 @@ std::unique_ptr<operator_data> sirius_physical_table_scan::execute(const operato
...
+      auto concatenated = cudf::concatenate(table_views, stream, space->get_default_allocator());
+      auto concat_rep =
+        std::make_unique<cucascade::gpu_table_representation>(std::move(concatenated), *space);
+      auto concat_batch   = std::make_shared<cucascade::data_batch>(0, std::move(concat_rep));
+      concatenated_holder = std::make_unique<operator_data>(
+        std::vector<std::shared_ptr<cucascade::data_batch>>{std::move(concat_batch)});
+      input_batches_ptr = &concatenated_holder->get_data_batches();
+    }
+  }
+  const auto& input_batches = *input_batches_ptr;
```
joosthooz: If we know that the input batches have been concatenated, we can remove the loops below. But then we may want to catch any errors that happened during the concatenation and throw so the task fails.

src/op/sirius_physical_operator.cpp
```diff
@@ -267,15 +268,41 @@ std::optional<task_creation_hint> sirius_physical_operator::get_next_task_hint()
...
+    constexpr size_t max_batches_per_task = 32;
+    while (true) {
+      auto batch = port_ptr->repo->pop_data_batch(::cucascade::batch_state::task_created);
+      if (!batch) { break; }
+      uint64_t batch_bytes = 0;
+      if (batch->get_data()) { batch_bytes = batch->get_data()->get_size_in_bytes(); }
+      accumulated_bytes += batch_bytes;
+      input_batch.push_back(std::move(batch));
+      ++batch_count;
+      if (accumulated_bytes >= config::DEFAULT_SCAN_TASK_BATCH_SIZE ||
```
joosthooz: If we change this function in this base class, the coalescing is also performed for other kernels besides scans (for example after a very selective filter), in which case we may want to use the `DEFAULT_CONCAT_BATCH_BYTES` instead. But for this PR, what might be better is to override this function in the table_scan class and only change the behavior there (and then do use this `DEFAULT_SCAN_TASK_BATCH_SIZE` parameter). I wonder how @wmalpica or @bwyogatama feel about it?

joosthooz (commented):
Thank you for this PR and the others! I've added some comment to get a small discussion going, your results look very good so I think we'd like to include this.

src/op/sirius_physical_table_scan.cpp
ducndh: Thanks for pointing that out, should have thought about how downstream get affected by the change. For the error catching, if cudf::concatenate throws (e.g. GPU OOM), the exception propagates up through the thread pool lambda in duckdb_scan_executor::manager_loop() where the existing catch(...) calls completion_handler->report_error(), so the task fails cleanly. I haven't tested the error catching carefully tho.

src/op/sirius_physical_operator.cpp
ducndh: I remember trying to catch the exact scan task with some conditions. However, overriding the table_scan is just a cleaner design. I have a branch testing out concatenation on partition, but other than very fast queries, I don't think it brings much value.

joosthooz (approved):
Thank you for addressing the feedback, I'm +1 on merging this but you might want to solicit a review from Bobbi or William!

wmalpica (approved):

PR #507, opened 2026-03-20
Allow larger build side for probing hashes, tuned scan scheduling at query startup

joosthooz:
This PR improves TPC-H query sum from 21.79s to 19.41s because of 2 changes:
- Decoupling the `max_build_hash_table_bytes` and `concat_batch_bytes` config parameters, so that `max_build_hash_table_bytes` is now allowed to be larger. When that happens, the partition operator will configure the following concat operator to concatenate all batches. This will allow the much more efficient probing join mode to be used more often, but at the same time other joins (standard & mixed) will still use the batch size that is configured with the `concat_batch_bytes`. Currently, the user would need to inflate the `concat_batch_bytes` setting to allow probing to be used for larger joins, but it will have negative impact in other areas. This change improves query sum to approx 20s.
- Small changes to scan scheduling: at query startup, we schedule up to 2 scans. Then in the task creator, we no longer always keep creating tasks for a scan until it is depleted. We only do that if there is only a single scan in the plan (to provide some more I/O parallelis). For the rest, we let the plan topology drive task creation (via the next_task_hint mechanism). This improves performance with an addition 0.6s.

### DIFF
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -103,6 +103,8 @@ void task_creator::prepare_for_query(const sirius::planner::query& query)
         operator_id, std::make_shared<pipeline::gpu_pipeline_task_global_state>(pipeline));
     }
   }
+  _num_scans_in_plan =
+    _scan_operator_global_state_map.size() + _parquet_scan_operator_global_state_map.size();
 }
 
 void task_creator::drain_pending_tasks()
@@ -286,8 +288,6 @@ void task_creator::manager_loop()
           size_t operator_id             = node->get_operator_id();
           auto parquet_task_global_state = _parquet_scan_operator_global_state_map.at(operator_id);
           auto* parquet_scan             = &node->Cast<op::sirius_physical_parquet_scan>();
-          size_t const num_scans_in_plan =
-            _scan_operator_global_state_map.size() + _parquet_scan_operator_global_state_map.size();
           while (true) {
             pipeline->mark_task_created();
             auto const partition_idx = parquet_task_global_state->get_next_rg_partition_idx();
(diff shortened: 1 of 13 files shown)


src/pipeline/pipeline_executor.cpp
```diff
@@ -161,7 +161,12 @@ std::future<void> pipeline_executor::start_query()
     gpu_exec->set_completion_handler(_completion_handler.get());
   }
 
-  schedule_next_scan_tasks();
+  constexpr int k_initial_scans = 2;
+  for (int i = 0; i < k_initial_scans && !_priority_scans.empty(); ++i) {
```
mbrobbel, line 166: Don't we need `std::lock_guard<std::mutex> lock(_priority_scans_mutex);`?

src/creator/task_creator.cpp
```diff
@@ -286,6 +286,8 @@ void task_creator::manager_loop()
           size_t operator_id             = node->get_operator_id();
           auto parquet_task_global_state = _parquet_scan_operator_global_state_map.at(operator_id);
           auto* parquet_scan             = &node->Cast<op::sirius_physical_parquet_scan>();
+          size_t const num_scans_in_plan =
+            _scan_operator_global_state_map.size() + _parquet_scan_operator_global_state_map.size();
```
mbrobbel: Should we move this out of the loop?

src/pipeline/pipeline_executor.cpp
joosthooz: Yes we do. Thanks for catching that

joosthooz: Should have made Claude do this (or review it) :laughing:

joosthooz:
@wmalpica following your comments on https://github.com/sirius-db/sirius/pull/498 (which is superseded by this), I've made a variant of this branch that doesn't perform the additional concatenation in the has_join. It instead does a special case in the concat operator itself, by looking ahead at the downstream join. But there's some additional code needed for (also for reserving enough memory), so you'll have to see here if you actually prefer that. Here's the diff to this branch: https://github.com/joosthooz/sirius/compare/allow-larger-probing-hashes...joosthooz:sirius:allow-larger-probing-hashes-concatlookahead?expand=1

src/creator/task_creator.cpp
joosthooz: Done

wmalpica (approved):

PR #509, opened 2026-03-20
Add `ENABLE_LEGACY_SIRIUS` CMake option to conditionally compile legacy code path

### LINKED ISSUE #496: Set build target (legacy or new Sirius)
Add a build configuration option to set the build target to either legacy Sirius or new Sirius.

This allows building only the needed code path, reducing build times and binary size.

Reference: #230

---
*Generated by Claude*

mbrobbel:
## Summary
- Add `ENABLE_LEGACY_SIRIUS` CMake option (default `OFF`) that controls whether the
  legacy Sirius code path (`gpu_processing` table function) is compiled
- Moves 46 legacy-only source files (`src/operator/`, `src/plan/`, `gpu_executor`,
  `gpu_physical_plan_generator`, etc.) into a separate `LEGACY_EXTENSION_SOURCES` list
- Guards legacy includes, structs, functions, and registration in `sirius_extension.cpp`
  and `sirius_extension.hpp` with `#ifdef SIRIUS_ENABLE_LEGACY`

## Test plan
- [ ] Build without the flag (default) — verify legacy sources are excluded and
      `gpu_processing` table function is not registered
- [ ] Build with `-DENABLE_LEGACY_SIRIUS=ON` — verify both `gpu_processing` and
      `gpu_execution` table functions are available

🤖 Created by Claude · Generated with [Claude Code](https://claude.com/claude-code)

Closes #496.


CMakeLists.txt
```diff
@@ -196,6 +180,34 @@ set(EXTENSION_SOURCES
     src/util/stream_check_wrapper.cpp
     src/util/segfault_backtrace_handler.cpp)
 
+# Legacy Sirius code path (gpu_processing entry point, plan generator,
+# context, and plan builders). The operators in src/operator/ and core
+# types (GPUColumn, GPUPipeline, etc.) remain in the main build because
+# the expression executor and CUDA kernels depend on them.
+set(LEGACY_EXTENSION_SOURCES
```
9prady9: Since we are doing this, I would recommend actually moving the legacy code into a `legacy` folder under `src` folder. It has a bunch of advantages going forward:

- We don't add any more code into legacy hierarchy.
- We can actively avoid using utilities or functions from this path.
- Even in CMake we can easily understand whats legacy source file or not by looking at the path prefix right away.

They are separated completely into a different folder hierarchy which enforces us to use `#include <legacy/.....hpp` type of includes. This by itself brings couple of additional advantages.

CMakeLists.txt
```diff
@@ -280,6 +296,12 @@ install(
   LIBRARY DESTINATION "${INSTALL_LIB_DIR}"
   ARCHIVE DESTINATION "${INSTALL_LIB_DIR}")
 
+if(ENABLE_LEGACY_SIRIUS)
+  message(STATUS "Legacy Sirius code path (gpu_processing) is ENABLED")
```
9prady9: perhaps `AUTHOR_WARNING` is better ? Normal users will know to ignore it but it can potentially catch developers attention

9prady9 (commented):
I understand it is in draft but thought of sharing some ideas before you consolidated on a particular approach.

I would recommend separating the legacy code into separate folder hierarchy (details more in the comments). It will help a lot in terms of source management and eventually making the removal a breeze if needed.

src/legacy/CMakeLists.txt
9prady9, line 1: nice, this makes it much more easier to remove it later :)

CMakeLists.txt
```diff
@@ -196,6 +150,14 @@ set(EXTENSION_SOURCES
     src/util/stream_check_wrapper.cpp
     src/util/segfault_backtrace_handler.cpp)
 
+# Legacy Sirius — all config lives in src/legacy/CMakeLists.txt.
+# Deleting src/legacy/ cleanly removes legacy build support.
```
9prady9, line 154: 👍🏾

.github/workflows/test.yml
```diff
@@ -86,14 +86,6 @@ jobs:
...
           ./build/release/duckdb -f scripts/tpch_to_parquet.sql
 
-      - name: Run sqllogic tests
-        run: |
-          ./build/release/test/unittest "$PWD/test/*" | tee test.log; EXIT=${PIPESTATUS[0]}
-          if grep -qE "FAILURES SUMMARY|Error in GPUExecuteQuery" test.log || [ "$EXIT" -ne 0 ]; then
-            echo "Test failures detected!"
-            exit 1
-          fi
-
```
9prady9, line 96: is this run via another workflow now ?

9prady9 (approved):
looks great, left one comment about removal in test GHA workflow file which I couldn't understand.

.github/workflows/test.yml
mbrobbel: Those tests were using the legacy function. I'm planning to work on sql tests for super sirius later.

9prady9: 👍🏾 , perhaps create an issue and link here so that is tracked.

PR #512, opened 2026-03-24
Implement FULL OUTER JOIN in gpu_processing via cudf::hash_join::full_join

ducndh:
gpu_processing query with a FULL OUTER JOIN falls back to CPU. FULL OUTER JOIN was only routed through the custom CUDA hash table kernels, which don't support it — JoinType::OUTER fell through to the CPU fallback path. However, cudf already has cudf::hash_join::full_join, which handles the full join semantics, including null-padding for non-matching rows on both sides.

Implementation:
Added cudf_hash_full_join() in cudf_join.cu following the same pattern as cudf_hash_left_join and cudf_hash_inner_join.
In gpu_physical_hash_join.cpp, route JoinType::OUTER through the cudf path: skip the custom hash table build in Sink,
call cudf_hash_full_join in Execute, and return FINISHED from GetData since cudf returns all row indices in one shot. Added outer_join_handled_in_execute flag in the header to guard the GetData early-exit

### DIFF
--- a/src/cuda/cudf/cudf_join.cu
+++ b/src/cuda/cudf/cudf_join.cu
@@ -351,4 +351,88 @@ void cudf_hash_left_join(vector<shared_ptr<GPUColumn>>& probe_keys,
   SIRIUS_LOG_DEBUG("CUDF Inner join result count: {}", count[0]);
 }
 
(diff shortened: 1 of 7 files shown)


wmalpica (approved):

wmalpica:
Thanks @ducndh , the fix looks good, but the PR needs to be rebased

PR #516, opened 2026-03-26
Add UNION ALL support for gpu_processing (legacy Sirius)

ducndh:
gpu_processing queries with UNION ALL fall back to CPU

Implementation:
  - Added GPUPhysicalUnion operator (src/legacy/operator/gpu_physical_union.cpp) following the same pipeline construction pattern as DuckDB's PhysicalUnion — creates a union pipeline that shares the same sink as the current pipeline, builds each child into its respective pipeline, and respects order-dependent sinks.
  - Added gpu_plan_set_operation.cpp to map LogicalSetOperation → GPUPhysicalUnion in the plan generator. Only UNION ALL is supported; EXCEPT/INTERSECT throw NotImplementedException for now but you can try to build them
  - Wired up CreatePlan(LogicalSetOperation&) in GPUPhysicalPlanGenerator to route LOGICAL_UNION through the new operator.
  - Refactored GPUResultCollection from raw DataChunk* array with manual new[]/delete[] to vector<unique_ptr<DataChunk>>, fixing a potential issue when multiple pipelines append results concurrently.
  - Fixed pipeline dependency scheduling in gpu_executor.cpp to check dependencies across all pipelines in a meta-pipeline, not just the base pipeline — without this, UNION ALL pipelines could be scheduled before their build-side dependencies completed.

### DIFF
--- a/src/include/legacy/gpu_physical_plan_generator.hpp
+++ b/src/include/legacy/gpu_physical_plan_generator.hpp
@@ -94,7 +94,7 @@ class GPUPhysicalPlanGenerator {
   // unique_ptr<GPUPhysicalOperator> CreatePlan(LogicalInsert &op);
   // unique_ptr<GPUPhysicalOperator> CreatePlan(LogicalCopyToFile &op);
   // unique_ptr<GPUPhysicalOperator> CreatePlan(LogicalExplain &op);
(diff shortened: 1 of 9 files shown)


mbrobbel:
Converting to draft. There are some conflicts.

PR #518, opened 2026-03-30
feat: Transparent GPU execution via optimizer extension

mbrobbel:
## Summary

- Adds transparent GPU execution so plain SQL can be intercepted and executed through Sirius without wrapping queries in `CALL gpu_execution('...')`
- Uses DuckDB extension hooks only:
  - `OptimizerExtension` captures an optimized logical-plan copy for eligible queries
  - `OnFinalizePrepare` generates the Sirius physical plan and replaces DuckDB's CPU physical plan with `PhysicalSiriusExecution`
- Keeps the explicit `CALL gpu_execution('...')` path
- Unsupported queries silently fall back to CPU
- Controlled per connection via `SET gpu_execution = true/false` (default: `true`)
- Uses `sirius.yaml` config discovery; if no config file is found, Sirius falls back to default config
- Legacy `sirius.cfg` is still recognized for compatibility

## Usage

```sql
LOAD 'sirius.duckdb_extension';

-- With ~/.sirius/sirius.yaml present, it is loaded automatically.
-- If no config file is found, Sirius initializes with default config.
SELECT * FROM lineitem WHERE l_quantity > 10 ORDER BY l_extendedprice LIMIT 100;

-- Disable transparent interception for the current connection:
SET gpu_execution = false;
```

### Transparent execution correctness

- restores `disabled_optimizers` correctly after transparent hooks run
- makes `gpu_execution` session-scoped instead of process-global
- fixes `PhysicalSiriusExecution` to keep fetched chunks alive instead of referencing freed storage
- skips transparent optimizer/query-lifecycle mutations for internal DuckDB connections
- adds transparent execution counters so tests can assert rebind / fallback / execution behavior directly

### Cross-connection safety for the new Super Sirius runtime

- serializes access to the shared `SiriusContext` query lifecycle across connections
- guards internal CPU fallback queries opened by `gpu_execution(...)` so they do not mutate shared Sirius state

Current behavior note:
- the current Super Sirius runtime is shared across connections on this branch, so Sirius-backed queries are serialized across connections for correctness
- true cross-connection concurrent Sirius execution is follow-up work

### Parquet / metadata-only regressions

- fixes the zero-column parquet path that could segfault on metadata-style queries such as `count(*)`
- preserves a carrier column through the zero-column table-scan path so filtered parquet batches remain valid

### Tests and docs

- transparent integration tests now compare actual row contents, not only row/column counts
- transparent and integration tests now assert whether Sirius actually rebound/executed or fell back
- docs updated to `gpu_execution` and `sirius.yaml`

## Validation

Ran locally on a GPU machine:

- `pixi run -- make -j8`
- `./build/release/extension/sirius/test/cpp/sirius_unittest "[transparent][integration]"`
- `./build/release/extension/sirius/test/cpp/sirius_unittest "Internal query guard preserves transparent execution state"`
- `./build/release/extension/sirius/test/cpp/sirius_unittest "gpu_execution - ungrouped count(*) parquet"`
- `./build/release/extension/sirius/test/cpp/sirius_unittest "gpu_execution hive partition - basic scan with partition columns"`

Not run in this PR update:

- full TPCH suite
- full multi-format suite
- Iceberg end-to-end validation in this latest pass


src/transparent/sirius_optimizer_extension.cpp
```diff
@@ -0,0 +1,72 @@
...
+  if (!duckdb::Config::ENABLE_TRANSPARENT_EXECUTION) { return; }
+
+  auto& context = input.context;
+  auto ctx      = context.registered_state->Get<duckdb::SiriusContext>("sirius_state");
+  if (!ctx || !ctx->is_initialized()) { return; }
+
+  // Disable optimizers that produce DuckDB-internal functions Sirius can't handle.
+  // The post-hook re-enables them so non-GPU queries aren't affected.
+  auto& disabled = duckdb::DBConfig::GetConfig(context).options.disabled_optimizers;
+  disabled.insert(duckdb::OptimizerType::IN_CLAUSE);
```
felipeblazing, line 54: why are IN_CLAUSE and COMPRESSED_MATERIALIZATION treated seperately from the ones in the config. Or are those disabled optimizers separate from the ones we are disabling for duckdb?

test/cpp/integration/test_gpu_execution_tpch.cpp
```diff
@@ -92,23 +95,40 @@ class GPUExecutionFixtureBase {
...
+    return rows;
+  }
+
   void compare_gpu_vs_cpu(const std::string& query,
                           std::optional<float> float_tolerance = std::nullopt)
   {
-    // Disable fallback so GPU errors are not silently hidden
-    con->Query("SET enable_duckdb_fallback = false;");
+    // Enable transparent GPU execution
+    con->Query("SET sirius_transparent_execution = true;");
```
felipeblazing: do we not still need to disable the duckdb fallback so it doesnt fall back to duckdb on failure?

test/cpp/integration/test_gpu_execution_tpch.cpp
```diff
@@ -122,54 +142,37 @@ class GPUExecutionFixtureBase {
...
+    auto& cpu_mat = cpu_result->Cast<duckdb::MaterializedQueryResult>();
+    auto gpu_rows = collect_rows(gpu_mat);
+    auto cpu_rows = collect_rows(cpu_mat);
+
+    for (duckdb::idx_t r = 0; r < gpu_rows.size(); r++) {
+      for (duckdb::idx_t c = 0; c < gpu_rows[r].size(); c++) {
+        if (float_tolerance.has_value()) {
+          // Check if this looks like a float value for tolerance comparison
+          try {
+            double gpu_d = std::stod(gpu_rows[r][c]);
```
felipeblazing: what if its a decimal value? we want to compare decimals exactly. could we not do something like use the duckdb to cudf converter than then compare tables?

felipeblazing: it shoudl be chekcing the  types

felipeblazing (changes requested):

test/cpp/integration/test_gpu_execution_tpch.cpp
bwyogatama: i think we can enable the fallback again now that GTC has passed?

src/transparent/sirius_optimizer_extension.cpp
bwyogatama: So this is duckdb specific optimization that doesn't apply to sirius. E.g. Compressed Materialization is to store intermediate result in a compressed duckdb format

bwyogatama:
Okay this looks good to me based on what i see i guess one thing is that let's name it set `gpu_execution` instead of `sirius_transparent_execution`

bwyogatama:
One other thing that I am thinking maybe we should start taking note is I believe the config that we modified here is global and affect all connection. For example, if i set `sirius_transparent_execution` to true, it will set it to true in all the connections. I guess it's fine because we assume there is only one connection right now but still worth noting

test/cpp/integration/test_transparent_execution.cpp
```diff
@@ -0,0 +1,181 @@
...
+    if (gpu_result->HasError()) {
+      UNSCOPED_INFO("Transparent GPU error: " << gpu_result->GetError());
+    }
+    REQUIRE_FALSE(gpu_result->HasError());
+
+    // Disable transparent execution and run on CPU.
+    con->Query("SET gpu_execution = false;");
+    auto cpu_result = con->Query(query);
+    con->Query("SET gpu_execution = true;");
+    REQUIRE(cpu_result);
```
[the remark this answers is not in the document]
mbrobbel: we'll do this in a follow-up

felipeblazing (approved):

PR #519, opened 2026-03-30
Refactor next port info

dhruv9vats:
Replaces `std::pair<sirius_physical_operator*, std::string_view>` with a named `next_port_info` struct having explicit `next_operator` and `next_operator_port_name` fields for readability and extensibility.

### DIFF
--- a/src/op/sirius_physical_concat.cpp
+++ b/src/op/sirius_physical_concat.cpp
@@ -210,15 +210,16 @@ void sirius_physical_concat::sink(const operator_data& output_data, rmm::cuda_st
   auto partitioned_output_data = dynamic_cast<const partitioned_operator_data*>(&output_data);
   auto partition_idx           = partitioned_output_data->get_partition_idx();
   for (auto& batch : partitioned_output_data->get_data_batches()) {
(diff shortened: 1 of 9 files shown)


src/op/sirius_physical_concat.cpp
```diff
@@ -210,15 +210,16 @@ void sirius_physical_concat::sink(const operator_data& output_data, rmm::cuda_st
   auto partitioned_output_data = dynamic_cast<const partitioned_operator_data*>(&output_data);
   auto partition_idx           = partitioned_output_data->get_partition_idx();
   for (auto& batch : partitioned_output_data->get_data_batches()) {
-    for (auto& [next_op, port_id] : next_port_after_sink) {
+    for (auto& next_port_info : next_port_after_sink) {
```
9prady9, line 213: nit: I think you can still use structured binding with structs.

mbrobbel (approved):

9prady9 (approved):

src/op/sirius_physical_concat.cpp
dhruv9vats: yes, but, we do this change to allow `struct next_port_info` to grow without having to change all access sites later.

PR #520, opened 2026-03-30
Cleanup executor class design

### LINKED ISSUE #502: Reduce code and logic duplication in gpu_pipeline_executor and downgrade_executor
This issue is a follow up for #438

9prady9:
Closes #502 

### Executor base class (`itask_executor`)

Pulls shared infrastructure out of `gpu_pipeline_executor`, `duckdb_scan_executor`, and
`downgrade_executor` into a single abstract base class:

- **Shared members moved to base**: `thread_pool`, `kiosk`, `interruptible_mpmc` task queue,
  manager thread, `_running` flag
- **Lifecycle methods moved to base**: `start`, `stop`, `schedule`, `drain_leftover_tasks`,
  `drain_and_wait`
- **Subclasses now only implement**:
  - `manager_loop()` — required, drives the dispatch loop
  - `get_per_thread_init`, `on_start`, `on_stop`, `on_stopped` — optional virtual hooks for
    executor-specific setup/teardown (e.g. setting CUDA device, starting a monitor thread)

### `bounded_thread_pool`

Merges the bounded counting semaphore (`kiosk`) and the unbounded worker pool (`thread_pool`) into
a single `exec::bounded_thread_pool` with only one scheduling mode (can be expanded as needed later):

- **Two-phase**: `reserve()` → `pool.dispatch(move(slot), fn)` — for executors that must confirm capacity
  before signalling a task request (used by `gpu_pipeline_executor`, `duckdb_scan_executor`,
  `task_creator`)
- **RAII `slot`**: releases back to the pool automatically on task completion — no need to move
  tickets into worker lambdas
- **Lifecycle**: `interrupt()` wakes all blocked callers without stopping in-flight work;
  `resume()` re-enables; `wait_all()` drains

### Cleanup

- Deleted `kiosk.hpp`, `thread_pool.hpp`, and `test_kiosk.cpp` — no longer used anywhere
- Added `test_bounded_thread_pool.cpp`


wmalpica (approved):

src/include/exec/bounded_thread_pool.hpp
```diff
@@ -0,0 +1,289 @@
...
+      std::lock_guard lock(mu_);
+      work_queue_.push([this, fn = std::move(fn)]() mutable noexcept {
+        try {
+          fn();
+        } catch (const std::exception& e) {
+          SIRIUS_LOG_ERROR("Exception in bounded_thread_pool task: {}", e.what());
+        } catch (...) {
+          SIRIUS_LOG_ERROR("Unknown exception in bounded_thread_pool task");
+        }
+        release_slot();
```
aminaramoon: we should notify after the lambda is destructed, otherwise we notify when the state of fn is not finished, and we might for example not be able to reserve memory and do unncessary downgrading.

src/include/exec/bounded_thread_pool.hpp
```diff
@@ -0,0 +1,289 @@
...
+  bounded_thread_pool(const bounded_thread_pool&)            = delete;
+  bounded_thread_pool& operator=(const bounded_thread_pool&) = delete;
+
+  /**
+   * @brief Schedule a function — blocks when at capacity.
+   *
+   * Returns false if interrupted before a slot becomes available.
+   */
+  bool schedule(absl::AnyInvocable<void()> fn)
+  {
```
aminaramoon: I don't know if this is a good api, maybe have a try_schedule or schedule, one wait and one return false if resources are not available

aminaramoon (changes requested):

src/include/exec/bounded_thread_pool.hpp
9prady9: We could refactor, but I would say - lets just remove it and add it when some usage is involved. I believe this was added to mirror the `thread_pool` class that was removed. What I found out is this usage style is actually not used anywhere in our code base right now.

src/include/exec/bounded_thread_pool.hpp
9prady9: good catch, let me work on that. Have a fix on mind, let me try it out.

src/include/exec/bounded_thread_pool.hpp
9prady9: Addressed in [c44a277a1ee106efb0bc9501086bfe9a75bf38d3](https://github.com/sirius-db/sirius/pull/520/commits/6abd2aff5ad1aa6bd27fdb822ccc131e9ac958ac)

src/include/exec/bounded_thread_pool.hpp
9prady9: I think I tried to address this much better here https://github.com/sirius-db/sirius/pull/520/commits/00f565132c20f33b1d7e9adebccfff546b520059 please have a look.

9prady9:
Just had a discussion with @aminaramoon future plans to use `kiosk` class for task book keeping during join implementations and other locations potentially. I will do a follow up PR to add it back with aforementioned info to let this PR in as it may collide with others as it touches all existing executors.

aminaramoon (approved):

aminaramoon (approved):

PR #529, opened 2026-03-31
feat: Add gpu_explain table function for query plan inspection

### LINKED ISSUE #381: sirius query explain
be able to explain sirius queries so that we generate the physical plan without running the query and can use to validate support for a query

mbrobbel:
## Summary

Closes #381

- Adds `gpu_explain('SELECT ...')` table function that generates the Sirius physical plan **without executing** the query, for plan inspection and support validation
- Returns key-value rows (`explain_key`, `explain_value`) matching DuckDB's EXPLAIN convention: the DuckDB logical plan + Sirius physical plan tree (or error message for unsupported queries)
- Adds `params_to_string()` overrides on 7 key operators for informative plan output (table_scan, filter, projection, hash_join, grouped_aggregate, order, limit)
- Extracts shared `optimizer_guard` (RAII) and `extract_optimized_plan()` utility, replacing 3 identical copies of PrepareConnection/CleanupConnection/ExtractPlan across gpu_execution, gpu_processing, and gpu_explain

### Usage
```sql
CALL gpu_explain('SELECT l_returnflag, sum(l_quantity) FROM lineitem GROUP BY l_returnflag');
```
```
┌──────────────────────┬───────────────────────────────────────────────────────┐
│     explain_key      │                     explain_value                     │
├──────────────────────┼───────────────────────────────────────────────────────┤
│ duckdb_logical_plan  │ ┌───────────────────────────┐ ...                    │
│ sirius_physical_plan │ HASH_GROUP_BY groups=[#0] aggs=[SUM, COUNT]  (est. 3)│
│                      │ └── TABLE_SCAN seq_scan [...]  (est. 6001215)        │
└──────────────────────┴───────────────────────────────────────────────────────┘
```

Unsupported queries report the error gracefully:
```sql
CALL gpu_explain('SELECT *, ROW_NUMBER() OVER() FROM lineitem');
-- Returns: error | "Window not supported"
```

### New files
| File | Purpose |
|------|---------|
| `src/gpu_explain.cpp` / `src/include/gpu_explain.hpp` | `gpu_explain` table function bind + execute |
| `src/util/sirius_plan_renderer.cpp` / `src/include/util/sirius_plan_renderer.hpp` | Recursive tree renderer for operator plans |
| `src/util/plan_extraction.cpp` / `src/include/util/plan_extraction.hpp` | Shared `optimizer_guard` + `extract_optimized_plan()` utility |
| `test/cpp/integration/test_gpu_explain.cpp` | 9 integration tests |

## Test plan
- [x] All 742 existing tests pass (78.8M assertions)
- [x] 9 new `[gpu_explain]` tests: schema, filter, join, aggregate, unsupported query (WINDOW), cardinality, tree connectors, NULL param, tree indentation
- [x] Pre-commit checks pass (clang-format, codespell, cmake-format)
- [x] CI

🤖 Created by Claude


bwyogatama:
This one the main issue is that the plan that it shows is actually not our final plan. Because in `sirius_engine::initialize_internal` we are modifying the pipeline that the `sirius_physical_plan_generator` produces.

So I think this one will show the plan produced by `sirius_physical_plan_generator` but will not show the pipeline produced by `sirius_engine::initialize_internal`

bwyogatama:
In general the main challenge is that when we insert new operator, we insert it when it's already in the form of pipeline instead of query plan. One option would be to show the pipeline instead of the query plan when doing explain and another option is to upstream all the modification we made in `sirius_engine::initialize_internal` to `sirius_physical_plan_generator`. Option 2 is better but requires breaking changes. Option 1 can be a temp solution.

bwyogatama:
I am doing some refactoring on #540 let me see if i can also do option 1 there

mbrobbel:
> I am doing some refactoring on #540 let me see if i can also do option 1 there

That looks more like working towards option 2? What is the reason we don't handle this in the plan generator?

bwyogatama:
The refactoring still doesn't upstream the operator to `sirius_physical_plan_generator` yet. 
We can handle it in the plan generator, it just requires breaking changes and it's not that urgent right now.

bwyogatama:
Let me make it a 2 phase refactoring: First phase for the non breaking and second phase for the breaking ones. I will start with first phase on #540 then will have a separate PR on the second phase

src/include/util/plan_extraction.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include "duckdb/main/config.hpp"
+#include "duckdb/planner/logical_operator.hpp"
+
+#include <set>
+
+namespace sirius::util {
+
+/// RAII guard that temporarily disables DuckDB optimizers incompatible with Sirius
+/// and restores the original settings on destruction.
+class optimizer_guard {
```
felipeblazing, line 29: Will this conflict with the work in https://github.com/sirius-db/sirius/pull/518

bwyogatama:
This is broken down into several phase so i will close this for now

PR #531, opened 2026-03-31
Expression executor with AST support for JIT compilation and operator fusion

### LINKED ISSUE #300: Avoid deep copy in expression executor `Execute()` method on a `BOUND_REFERENCE` expression
Because the `Execute()` interface in the expression executor returns a `std::unique_ptr<cudf::column>`, it must deep copy its inputs when executing a `BOUND_REFERENCE` expression. There should be a way around this.

kevkrist:
This PR introduces a new expression executor to utilize cuDF AST functionality, specifically the JIT support (via `compute_column_jit`) and the fused operator support (via `compute_column` -- a tree of AST operators is executed in a single kernel with a monolithic switch). It also supports the old operator-at-a-time model. These options are controlled via the `gpu_expression_executor::execution_strategy`. The executor logic essentially builds a tree whose nodes are AST trees and whose edges are 'AST breakers', namely operators that don't have AST equivalents (e.g., CASE, LIKE, SUBSTRING, etc.). The nodes are evaluated via JIT or operator fusion, depending on the `execution_strategy`, unless the execution strategy is `MATERIALIZE`, which as stated reverts to operator-at-a-time. The operator fusion approach trades slower kernels for less GMEM IO, and the JIT approach of course trades cold latency for fast hot execution. A knob is provided to control whether the use of AST functionality pays off: if the number of operators that can be translated to cuDF ASTs for a given expression node is smaller than `min_ast_size`, the executor reverts to the `MATERIALIZE` mode. 

The new executor never a) deep copies input columns (UNLESS an expression is simply a BOUND_REFERENCE, in which case a copy is necessary) or b) promotes scalars to columns (UNLESS an expression is simply a BOUND_CONSTANT, in which case this is again necessary).

Fixes #300

Currently, decimal type propagation through the AST tree is blocked until [this bug](https://github.com/rapidsai/cudf/issues/21980) on nested decimal evaluation is fixed.

The new expression executor is in the `experimental::sirius` namespace and is only plugged into the engine at the FILTER and PROJECT operators with the AST_INTERPRET strategy, based on the microbenchmarks below.

Here are some microbenchmark results (generated by Claude):
## `gpu_expression_executor` benchmark

Microbenchmark comparing the legacy `duckdb::sirius::GpuExpressionExecutor`
against the new `sirius::experimental::gpu_expression_executor` across its
three strategies (`MATERIALIZE`, `AST_INTERPRET`, `AST_JIT`), on two
expression shapes that bracket the realistic workload range.

### Methodology

- 50 hot iterations per strategy, each bracketed by
  `cudaStreamSynchronize`.
- One discarded warmup call before the hot loop for every strategy except
  `AST_JIT` — the first call for non-JIT strategies is dominated by
  allocator first-touch and kernel cache warmup, which is noise.
- `AST_JIT` reports its cold run because the first invocation captures
  cuDF's one-shot kernel compile, which is load-bearing to know.
- The cold numbers below assume a warm NVRTC on-disk cache. A truly
  first-ever compile on a clean system can run ~1s per expression; that
  cost is amortized across all processes after the first.

### Hardware

- GPU: NVIDIA GeForce RTX 4070 Ti (12 GB)
- Driver: 580.95.05
- Input size (both benchmarks): 4,194,304 rows (4M)

---

### Benchmark 1 — pure-arithmetic (full AST)

Best-case for `AST_JIT`: every node is AST-translatable, so the JIT fuses
the entire tree into a single kernel.

**Expression:**

```
((a + b) * (c - d)) + ((a - b) * (c + d)) + (a*c + b*d) - (a*d + b*c)
  PLUS
s = 'str_3'
```

- 11 INT32 binary ops (+, -, *), every node column-column (no constants).
- 1 string equality on a VARCHAR column.

**Input:** 4 × INT32 (`a`, `b`, `c`, `d`, range `[1,100]`) + 1 × VARCHAR
(`s`, range `"str_1".."str_10"`). 2 output columns.

**Results:**

| Strategy | Hot mean (ms) | Cold (ms) |
|---|---:|---:|
| legacy `GpuExpressionExecutor` | 2.601 | — |
| experimental `MATERIALIZE` | 1.636 | — |
| experimental `AST_INTERPRET` | 0.681 | — |
| experimental `AST_JIT` | 0.462 | 21.471 |

- `AST_JIT` is **5.6×** faster than legacy on the hot path and **1.5×**
  faster than `AST_INTERPRET`.
- Break-even for `AST_JIT` over `AST_INTERPRET`: 21.5 ms / 0.22 ms per
  call ≈ **~100 invocations**.

---

### Benchmark 2 — compound-boolean (mixed-mode)

Realistic predicate with arithmetic, numeric and string comparisons,
nested AND/OR, and a LIKE. LIKE has no cuDF AST equivalent, so it forces a
materialized sub-result — AST_JIT only gets to fuse the AST-capable portion
around it.

**Expression:**

```
(((a * b) - c) < 5000)
  AND (d == 'str_3')
  AND (((e * f) == 10) OR (g LIKE '%3%'))
```

Node breakdown: 3 INT32 arithmetic ops, 3 numeric/string comparisons,
1 LIKE (AST-incapable), 2 conjunctions (OR-2, AND-3).

**Input:** 5 × INT32 (`a`, `b`, `c`, `e`, `f`, range `[1,100]`) + 2 ×
VARCHAR (`d`, `g`, range `"str_1".."str_10"`). 1 output BOOL mask.

**Results:**

| Strategy | Hot mean (ms) | Cold (ms) |
|---|---:|---:|
| legacy `GpuExpressionExecutor` | 1.543 | — |
| experimental `MATERIALIZE` | 0.923 | — |
| experimental `AST_INTERPRET` | 0.742 | — |
| experimental `AST_JIT` | 0.739 | 25.638 |

- `AST_JIT` is **2.09×** faster than legacy on the hot path, but only
  **1.004×** faster than `AST_INTERPRET` (~4 µs per call).
- Break-even for `AST_JIT` over `AST_INTERPRET`: 25.6 ms / 0.004 ms per
  call ≈ **~6,400 invocations**.

---

### Takeaways

| | pure-arithmetic | compound-boolean |
|---|---|---|
| legacy hot | 2.601 ms | 1.543 ms |
| experimental MATERIALIZE hot | 1.636 ms (1.59×) | 0.923 ms (1.67×) |
| experimental AST_INTERPRET hot | 0.681 ms (3.82×) | 0.742 ms (2.08×) |
| experimental AST_JIT hot | 0.462 ms (5.63×) | 0.739 ms (2.09×) |
| AST_JIT cold | 21.5 ms | 25.6 ms |
| AST_JIT vs AST_INTERPRET | 1.47× faster | 1.004× faster |
| Break-even (calls) | ~100 | ~6,400 |

Speedup multipliers are relative to legacy on each benchmark row.

- **MATERIALIZE alone is a clear win over legacy** on both shapes
  (1.6–1.7×), so even workloads that skip the AST path benefit from the
  new executor.
- **AST_INTERPRET is the general-purpose sweet spot.** It beats legacy by
  2–4× without any compile tax, and on the compound-boolean case it is
  essentially tied with AST_JIT on the hot path.
- **AST_JIT only pays off when the expression is mostly AST-translatable.**
  On the pure-arithmetic tree it's 1.5× faster than AST_INTERPRET and
  amortizes its ~21 ms compile after ~100 calls. On the compound-boolean
  tree (with LIKE forcing materialization around it) JIT saves only
  ~4 µs per call and would need ~6,400 invocations to break even —
  rarely worth it.

Practical heuristic: default to `AST_INTERPRET`; escalate to `AST_JIT`
only when (a) the expression is mostly AST-translatable and (b) the hot
path will run the same expression many times.


src/expression_executor/specializations/gpu_execute_between.cpp
```diff
@@ -14,10 +14,130 @@
...
+#include <cudf/transform.hpp>
+
+namespace sirius::experimental {
+using execute_result = gpu_expression_executor::execute_result;
+
+execute_result gpu_expression_executor::execute(duckdb::BoundBetweenExpression const& expr,
+                                                execution_mode mode)
+{
+  if (_strategy != expression_executor_strategy::MATERIALIZE &&
+      (mode == execution_mode::AST || count_ast_ops(expr) >= _min_ast_size)) {
```
wmalpica, line 34: From the description on the PR i figured this should be:
` (mode == execution_mode::AST && count_ast_ops(expr) >= _min_ast_size)`
Meaning that it has to be AST mode AND the number of expressions has to be above the min

src/expression_executor/gpu_expression_executor.cpp
```diff
@@ -51,6 +54,420 @@ bool IsFixedWidth(cudf::data_type const& type)
...
+      "[execute_result] Attempted to get column view from non-column execute_result");
+  }
+  if (std::holds_alternative<cudf::column_view>(payload)) {
+    return std::get<cudf::column_view>(payload);
+  }
+  auto const& column_payload = std::get<std::unique_ptr<cudf::column>>(payload);
+  return column_payload->view();
+}
+
+std::unique_ptr<cudf::column> gpu_expression_executor::execute_result::get_column()
```
wmalpica: consider renaming this function to `release_column()` since it loses ownership

src/expression_executor/specializations/gpu_execute_between.cpp
kevkrist: Good question, and it's possible that it could be clearer with better naming of the enums. Always open to suggestions.

When `mode == execution_mode::AST`, that's the parent asking the current operator to add itself to the AST, and that request should be propagated downward as much as possible. So, when this is true, this operator doesn't need to worry about deciding for itself whether it should be AST -- the parent has already decided.

If `mode != execution_mode::AST` (i.e., `mode == execution_mode::MATERIALIZE`), the parent is asking the current operator to materialize its result, but the current operator now needs to decide for itself whether to materialize itself via AST or by operator-at-a-time. It makes this decision based on `count_ast_ops()`. The intuition is that AST gives more benefit the more operators it can execute in a single AST pass.

src/expression_executor/gpu_expression_executor.cpp
kevkrist: Agreed.

kevkrist (commented):
Renamed `get_column()` and clarified branching logic.

src/expression_executor/specializations/gpu_execute_comparison.cpp
```diff
@@ -14,20 +14,137 @@
...
+            "executor.");
+        default:
+          throw duckdb::InternalException(
+            "[expression_executor:comparison] Unrecognized comparison type : {}",
+            static_cast<int>(expr.GetExpressionType()));
+      }
+    };
+
+    auto left             = execute(*expr.left, execution_mode::AST);
+    auto right            = execute(*expr.right, execution_mode::AST);
```
joosthooz, line 68: This was changed in the "fix q17 bug" commit, what was the bug and what does the change mean? If there is a comparison in the expression, it will now always evaluate the left and right parts in AST mode, even if configured to use JIT?

src/op/sirius_physical_filter.cpp
```diff
@@ -54,14 +54,19 @@ std::unique_ptr<operator_data> sirius_physical_filter::execute(const operator_da
   const auto& input_batches = input.get_data_batches();
 
   // The executor uses the data_batch API to filter rows according to `expression`.
-  duckdb::sirius::GpuExpressionExecutor gpu_expression_executor(*expression.get());
+  sirius::experimental::gpu_expression_executor gpu_expression_executor(
+    expression.get(),
+    sirius::experimental::expression_executor_strategy::AST_INTERPRET,
```
joosthooz: The strategy is now hardcoded, is this something we would like to create an engine parameter for?

src/expression_executor/specializations/gpu_execute_between.cpp
```diff
@@ -14,10 +14,130 @@
...
+                         lower.get_temp_scalar_indices(),
+                         upper.get_temp_scalar_indices()},
+                        {input.get_temp_column_indices(),
+                         lower.get_temp_column_indices(),
+                         upper.get_temp_column_indices()});
+    return execute_result(std::move(result_column));
+  }
+
+  //===----------3: MATERIALIZE Mode, evaluate node with unary/binary ops----------===//
+  if (mode == execution_mode::AST) {
```
joosthooz: Shouldn't this be `if (mode == execution_mode::MATERIALIZE) {` ? Now it doesn't match the comment.

src/include/expression_executor/gpu_expression_executor.hpp
```diff
@@ -38,16 +39,373 @@
...
+   * MATERIALIZE mode to AST mode. If an expression subtree rooted at a given node produces an AST
+   * with N operators and N < min_ast_size, the expression will be evaluated operator-by-operator
+   * (in MATERIALIZE mode). Otherwise, the executor will try to evaluate the expression subtree by
+   * adding nodes to the AST tree.
+   */
+  gpu_expression_executor(
+    duckdb::Expression const* expression,
+    expression_executor_strategy strategy,
+    rmm::device_async_resource_ref resource_ref = cudf::get_current_device_resource_ref(),
+    rmm::cuda_stream_view stream                = cudf::get_default_stream(),
```
joosthooz, line 343: I think we want to minimize or even prevent the use of the default stream, so with that in mind we might not want to provide this default so that we don't accidentally use it

src/op/sirius_physical_filter.cpp
```diff
@@ -54,14 +54,19 @@ std::unique_ptr<operator_data> sirius_physical_filter::execute(const operator_da
...
+    cudf::get_current_device_resource_ref(),
+    stream);
 
   std::vector<std::shared_ptr<cucascade::data_batch>> output_batches;
   output_batches.reserve(input_batches.size());
 
   for (auto const& batch : input_batches) {
     if (!batch) { continue; }
-    auto filtered_batch = gpu_expression_executor.select(batch, stream);
+    // auto filtered_batch = gpu_expression_executor.select(batch, stream);
```
joosthooz: This commented line can be removed

src/op/sirius_physical_projection.cpp
```diff
@@ -40,14 +40,19 @@ std::unique_ptr<operator_data> sirius_physical_projection::execute(const operato
...
+    cudf::get_current_device_resource_ref(),
+    stream);
 
   std::vector<std::shared_ptr<cucascade::data_batch>> output_batches;
   output_batches.reserve(input_batches.size());
 
   for (auto const& batch : input_batches) {
     if (!batch) { continue; }
-    auto projected_batch = gpu_expression_executor.execute(batch, stream);
+    // auto projected_batch = gpu_expression_executor.execute(batch, stream);
```
joosthooz: Can also be removed

src/expression_executor/specializations/gpu_execute_comparison.cpp
```diff
@@ -14,20 +14,137 @@
...
+        case COMPARE_GREATERTHAN: return cudf::ast::ast_operator::GREATER;
+        case COMPARE_GREATERTHANOREQUALTO: return cudf::ast::ast_operator::GREATER_EQUAL;
+        case COMPARE_LESSTHAN: return cudf::ast::ast_operator::LESS;
+        case COMPARE_LESSTHANOREQUALTO: return cudf::ast::ast_operator::LESS_EQUAL;
+        case COMPARE_NOTEQUAL: return cudf::ast::ast_operator::NOT_EQUAL;
+        case COMPARE_DISTINCT_FROM:
+        case COMPARE_NOT_DISTINCT_FROM:
+          throw duckdb::NotImplementedException(
+            "[expression_executor:comparison] DISTINCT comparisons not supported in expression "
+            "executor.");
```
joosthooz: Does this mean that the old executor supports some expressions that the new doesn't Do we have/need a fallback mechanism for those? Or it is another reason to allow users to choose which executor (and strategy) to use

src/expression_executor/gpu_expression_executor.cpp
```diff
@@ -51,6 +54,429 @@ bool IsFixedWidth(cudf::data_type const& type)
...
+          auto const num_eqs = op_expr.children.size() - 1;
+          auto count =
+            num_ors + num_eqs + (op_expr.type == duckdb::ExpressionType::COMPARE_NOT_IN ? 1 : 0);
+          for (const auto& child : op_expr.children) {
+            count += count_ast_ops(*child);
+          }
+          return count;
+        }
+        case duckdb::ExpressionType::OPERATOR_COALESCE:
+          /// KEVIN: TODO
```
joosthooz: What is this TODO about, does it have a github issue? Consider creating and pointing to it here

joosthooz (commented):
Thank you, this looks great. Does it make sense to make a config setting for choosing the executor (and the strategy)? Is this new executor ready to be the default or should we consider it experimental (seeing some of the code is in an 'experimental' namespace and there is a bugfix we're waiting for)
General question: what is the difference between strategy and mode?

src/expression_executor/specializations/gpu_execute_between.cpp
kevkrist: yes good catch. There are some cases (e.g., CAST) where we need to check that the operation is AST-supported, in which case we need this check, since the parent might request AST mode but the child might not be able to execute in AST mode. So this is dead code here. I will check for similar cases.

src/expression_executor/specializations/gpu_execute_comparison.cpp
kevkrist: Oh thanks I just missed that DuckDB NULL_EQUAL is semantically equivalent to cuDF NOT_DISTINCT_FROM. Also COMPARE_DISTINCT_FROM is equivalent to NULL_NOT_EQUALS binary operator, and NOT + NULL_EQUAL ast operator, so both will be supported in this executor. I have also added new tests.

src/expression_executor/specializations/gpu_execute_comparison.cpp
kevkrist: The bug was that, even if the parent is not executing in AST mode, and thus not passing `execution_mode::AST` to its child as the `mode` parameter, the child might still execute in AST mode if it judges that doing so is worth it (if it can bundle a bunch of operations together into a single AST tree). So the `mode` input parameter might not be AST, even though in this path we are adding to the AST tree, and we have to explicitly request the children to add themselves to the AST.

There is a distinction between `execution_mode` and `expression_executor_strategy` that I didn't do a good job of making clear. `execution_mode` is whether the operator is adding its result to the AST tree or returning a materialized result. `expression_executor_strategy` tells the executor, either materialize every intermediate (operator-at-a-time), or use AST wherever possible, and when you need to evaluate the AST, use either the interpret (`compute_column()`) or JIT (`compute_column_jit()`) strategy.

src/expression_executor/gpu_expression_executor.cpp
kevkrist: Clarified and opened issue #635

src/op/sirius_physical_filter.cpp
kevkrist: I have added a configuration option to choose the strategy.
I have also added Issue #636 -- eventually the operator should decide based on state (statistics) which strategy to choose. The heuristics used should be informed by all the benchmarks you are adding :).

test/cpp/expression_executor/test_gpu_expression_executor.cpp
```diff
@@ -1677,6 +1677,61 @@ TEMPLATE_TEST_CASE("experimental select IS NULL and IS NOT NULL",
...
+TEMPLATE_TEST_CASE("experimental select COMPARE_NOT_DISTINCT_FROM",
+                   "[expression_executor][experimental]",
+                   mat_strategy,
+                   ast_interpret_strategy)
+{
+  constexpr auto strategy = TestType::value;
+  auto* space             = get_default_gpu_space();
+  REQUIRE(space != nullptr);
+
+  // Actual values: {10, NULL, 30, NULL, 50} — nulls must be distinct from 30
```
joosthooz: Why must the nulls be distinct from 30, isn't that an important case to test?

joosthooz (approved):

test/cpp/expression_executor/test_gpu_expression_executor.cpp
kevkrist: I think this is just poorly documented. This test covers that case: that NULL and 30 are distinct. I've updated the comment for more clarity.

PR #535, opened 2026-04-01
Add nightly cudf build and fix top_k_order sort regression

mbrobbel:
## Summary
- **Fix top_k_order sort ordering**: `cudf::top_k_order` does not guarantee sorted output per its API contract. Stable cudf 26.02 happened to return sorted indices but nightly 26.06 does not, causing wrong results when offset is applied in `top_n_merge`. Added `sort_by_key` after gather to ensure deterministic ordering.
- **Add nightly cudf CI build**: Infrastructure to build against `rapidsai-nightly` without polluting the main `pixi.lock`. A generation script derives `envs/nightly/pixi.toml` from the main `pixi.toml` (single source of truth for deps). Added `nightly-shell` and `nightly-run` pixi tasks, and an arm64/gcc/release nightly entry in the CI check matrix (`continue-on-error` so nightly breakage doesn't block merges).

## Test plan
- [x] All 716 unit tests pass with nightly cudf 26.06
- [x] `pixi run nightly-run make` builds successfully
- [ ] CI nightly matrix entry runs on arm64

🤖 Created by [Claude Code](https://claude.com/claude-code)


mike-wendt (approved):
LGTM, let's use the public `arm64` runners for now, if needed we can switch to our own runners once I get the cloud runners setup

mbrobbel:
We may need to update the required checks (because the names are different now). @mike-wendt can you help with that?

mike-wendt:
> We may need to update the required checks (because the names are different now). @mike-wendt can you help with that?

Yes I can get this updated, one sec

mike-wendt:
These are now updated
Screenshot 2026-04-08 at 11 39 04

mike-wendt:
Notified channels of this breaking change to merge other open PRs

PR #536, opened 2026-04-01
Use clang-debug preset for test workflow

### LINKED ISSUE #488: Switch CI build to run in debug mode
Switch the CI build to run the build in debug mode.

This will help catch issues earlier in CI that are only visible with debug assertions enabled.

Reference: #230

---
*Generated by Claude*

mbrobbel:
Closes #488. 

## Summary
- Switch test workflow from release to clang-debug build preset
- Remove TPC-H performance snapshot step (not needed for debug CI)
- GCC debug is excluded due to a known GCC 14 + CCCL constexpr bug at `-O0`, matching the check workflow's matrix

## Test plan
- [ ] CI builds successfully with `clang-debug` preset
- [ ] Unit tests pass on T4 GPU runner

🤖 Created by Claude


mbrobbel:
@wmalpica the tests now run much slower; I guess we should first try to reduce test times for these builds before we do this. What do you think?

mike-wendt:
See #488, no active request for this. Can reopen if needed

PR #546, opened 2026-04-02
Use cudf::distinct_hash_join for INNER joins with unique build keys in Super Sirius

### LINKED ISSUE #541: Use cudf::distinct_hash_join for unique build keys in Super Sirius
**Problem**
cuDF provides cudf::distinct_hash_join, a specialized hash join that eliminates duplicate-chain traversal when build-side keys are guaranteed unique. This is faster than the generic cudf::hash_join for FK-PK join patterns (e.g., lineitem JOIN orders ON l_orderkey = o_orderkey).

Super Sirius (gpu_execution) currently uses only cudf::hash_join / cudf::inner_join for all hash joins. The unique_build_keys and unique_probe_keys flags exist in the sirius_physical_hash_join header but are dead code — never set or read.

Legacy Sirius (gpu_processing) already uses cudf::distinct_hash_join for inner joins with unique build keys (commit 2bd3906), sourcing uniqueness from schema constraints via GPUColumn::is_unique. The optimization was never ported to Super Sirius during the ground-up rewrite.

**Proposed solution**
Propagate column-level uniqueness metadata through the Super Sirius physical plan at plan construction time, and use it to select cudf::distinct_hash_join when applicable.

When unique_build_keys is true:
- BUILD_PROBE mode (INNER): Construct cudf::distinct_hash_join instead of cudf::hash_join. The _hash_table member will need to become a variant or wrapper since the two types are distinct.
- STANDARD mode (INNER): Use cudf::distinct_hash_join for the one-shot join path instead of cudf::inner_join.

matal-nvidia:
Adds a column-level uniqueness propagation framework and uses it to select cudf::distinct_hash_join for INNER and LEFT joins when build-side keys are guaranteed unique (e.g., primary key / unique constraint).

Column property framework:
- New `column_property` struct on `sirius_physical_operator` tracks per-column guarantees (starting with is_unique)
- Table scan marks columns unique from DuckDB catalog constraints (UNIQUE / PRIMARY KEY)
- Properties propagate through filter, projection, limit, order, and aggregate operators at plan construction time
- Grouped aggregate marks grouping keys as unique (GROUP BY outputs are unique by definition)

Distinct hash join:
- New build_hash_table wrapper encapsulates both cudf::hash_join and cudf::distinct_hash_join behind a unified API
- BUILD_PROBE and STANDARD paths use distinct_hash_join when unique_build_keys is true
- Supports INNER (same return type) and LEFT (wrapper synthesizes trivial probe indices)
- FULL OUTER falls back to generic hash_join (cuDF has no distinct full join API)

Fixes #541

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,6 +104,7 @@ set(EXTENSION_SOURCES
     src/op/sirius_physical_filter.cpp
     src/op/sirius_physical_grouped_aggregate.cpp
     src/op/sirius_physical_grouped_aggregate_merge.cpp
+    src/op/build_hash_table.cpp
     src/op/sirius_physical_hash_join.cpp
     src/op/sirius_physical_limit.cpp
     src/op/sirius_physical_merge_sort.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,6 +303,7 @@ set(TEST_SOURCES
     test/cpp/operator/test_physical_top_n.cpp
     test/cpp/operator/test_physical_ungrouped_aggregate.cpp
     test/cpp/parallel/test_task_executor.cpp
+    test/cpp/planner/test_column_property_propagation.cpp
     test/cpp/pipeline/test_gpu_pipeline_executor.cpp
     test/cpp/pipeline/test_oom_reschedule.cpp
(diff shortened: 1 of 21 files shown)


mbrobbel:
Closed via #558.

PR #558, opened 2026-04-02
Alternative: use cudf::distinct_hash_join via logical plan uniqueness proof

mbrobbel:
## Summary

Alternative implementation for #541, taking a different approach from #546. I started working on this independently after seeing the issue — great to have two approaches to compare and discuss which design direction we want to go.

### Approach: logical plan uniqueness proof engine

Instead of propagating per-column metadata through the Sirius physical plan, this PR walks the **DuckDB logical plan subtree** at join creation time with a recursive `prove_unique_columns()` function. It returns the set of output column indices proven to form a unique key, then checks if the join key columns are a superset.

**Provable uniqueness sources:**
- `LogicalGet` + `PRIMARY KEY` constraint (mapped through `projection_ids`/`column_ids`)
- `LogicalAggregate` with single grouping set (GROUP BY output keys)
- Propagation through `LogicalFilter`, `LogicalOrder` (with `projection_map` handling), `LogicalLimit`, `LogicalTopN`
- `LogicalProjection` (direct `BoundReferenceExpression` pass-through only)

**Key design differences from #546:**

| | This PR | #546 |
|---|---|---|
| Uniqueness model | **Column set** (supports composite keys) | Per-column boolean |
| Where proof runs | DuckDB logical plan (before `create_plan`) | Sirius physical plan (during construction) |
| Constraint scope | PRIMARY KEY only | UNIQUE + PRIMARY KEY |
| `NOT DISTINCT FROM` | Excluded (needs `null_equality::EQUAL`) | Included |
| LEFT join handling | Direct column copy (no probe gather) | Host→device memcpy for probe indices |
| Files changed | 5 (localized) | 21 (framework across operators) |
| New abstractions | None | `column_property`, `build_hash_table` wrapper |

### Why the differences matter

- **Column sets vs per-column booleans**: A composite `PRIMARY KEY (a, b)` means the *pair* is unique, not each column individually. Per-column `is_unique` can't represent this. `GROUP BY a, b` has the same issue — marking both columns individually unique would incorrectly enable `distinct_hash_join` for a join on just `a`.
- **PRIMARY KEY only**: Plain `UNIQUE` allows multiple NULLs. With `null_equality::UNEQUAL`, cuDF's `distinct_hash_join` has undefined behavior on duplicate build keys — the cuDF docs are explicit about this. PRIMARY KEY guarantees NOT NULL + unique.
- **`NOT DISTINCT FROM` exclusion**: This comparison treats NULLs as equal, but the hash table is built with `null_equality::UNEQUAL`. Enabling the optimization for these joins would be a semantic mismatch.

### Execution paths

- **BUILD_PROBE mode**: creates `cudf::distinct_hash_join` instead of `cudf::hash_join` when `unique_build_keys && (INNER || LEFT)`
- **STANDARD mode**: constructs local `cudf::distinct_hash_join` for one-shot joins
- **LEFT join**: `gather_distinct_left_join_output()` copies probe columns directly and gathers only build columns with `NULLIFY` — avoids the identity gather that a synthesized probe index sequence would require

### Tests

8 planner-level test cases:
- PK detection (single + composite)
- Partial composite PK (negative)
- Plain UNIQUE exclusion (negative)
- Non-unique table (negative)
- GROUP BY uniqueness (single + multi-key subset negative)
- LEFT join with PK build side


matal-nvidia:
Thanks for taking this up, this approach seems much better than #546 given it can handle composite unique keys, is semantically more correct (disabling for UNIQUE and NOT DISTINCT FROM), and is cleaner.

src/planner/sirius_plan_comparison_join.cpp
```diff
@@ -14,28 +14,164 @@
...
 #include "op/sirius_physical_hash_join.hpp"
 #include "op/sirius_physical_nested_loop_join.hpp"
 #include "planner/sirius_physical_plan_generator.hpp"
 #include "sirius_context.hpp"
 
 namespace sirius::planner {
 
+/// Returns a set of output column indices proven to form a unique key for the
+/// given logical operator subtree, or an empty set if uniqueness cannot be proven.
+static std::unordered_set<duckdb::idx_t> prove_unique_columns(duckdb::LogicalOperator& op)
```
kevkrist, line 38: This is very cool! Are you leaving it as future work to consider uniqueness propagation through joins? The rule for JOINs seems to be:
INNER JOIN: Either side's unique columns are preserved in the output if and only if the other side's join keys are unique.
LEFT JOIN: Left side's unique columns are preserved in the output if and only if the right side's join keys are unique.
RIGHT JOIN: Switch left and right^
(ANTI) SEMI JOIN: Left side's unique columns are always preserved
FULL OUTER JOIN: never (?)
I understand that this is complicated and should arguably belong to future work, but I thought I'd ask the question here.

kevkrist (commented):
Very cool! Just one question.

src/planner/sirius_plan_comparison_join.cpp
mbrobbel: Great suggestion! Added to this PR: [cbd71b1](https://github.com/sirius-db/sirius/pull/558/commits/cbd71b17899516b5869a2d9fbbbe84d7d9169f9a):

| Join type | Left preserved | Right preserved |
  |---|---|---|
  | INNER | iff right keys unique | iff left keys unique |
  | LEFT | iff right keys unique | — |
  | RIGHT | — | iff left keys unique |
  | SEMI / ANTI | always | — |
  | RIGHT_SEMI / RIGHT_ANTI | — | always |
  | MARK / SINGLE | always | — |
  | FULL OUTER | — | — |

kevkrist (approved):

PR #561, opened 2026-04-02
refactor: replace DuckDB exceptions with Sirius exceptions in core engine

### LINKED ISSUE #551: Decouple from DuckDB: replace DuckDB exceptions with Sirius exceptions (Phase 2)
## Parent issue
Part of #545

## Description
Create `src/include/sirius/exception.hpp` with Sirius exception types inheriting from `std::runtime_error`:
- `sirius::internal_exception` (replaces `duckdb::InternalException`)
- `sirius::not_implemented_exception` (replaces `duckdb::NotImplementedException`)
- `sirius::invalid_input_exception` (replaces `duckdb::InvalidInputException`)

Replace usage in core Sirius files (~27 files). Keep DuckDB exceptions in the glue/adapter layer (planner, scan, result collection, extension entry point).

## Why
Exceptions are leaf dependencies — they don't pull in other DuckDB types. Replacing them removes `#include "duckdb/common/exception.hpp"` from many core files, which is one of the more transitive DuckDB header pulls.

## Scope
- 1 new header
- ~27 .cpp files + a few headers
- Low risk, each file can be converted independently

## Acceptance criteria
- [ ] Core operator/pipeline files use `sirius::` exceptions
- [ ] DuckDB exceptions only remain in glue layer files
- [ ] Full build + tests pass

---
Created by Claude

mbrobbel:
## Summary
- Adds `src/include/sirius/exception.hpp` with `sirius::internal_exception`, `sirius::not_implemented_exception`, and `sirius::invalid_input_exception` (inheriting from `std::runtime_error`)
- Replaces DuckDB exception usage in 15 core engine files (operators, pipeline, engine)
- Leaves DuckDB exceptions in glue layer (planner, scan, result collection, extension entry point)

Closes #551

## Test plan
- [ ] Full build succeeds
- [ ] `pre-commit run -a` passes
- [ ] Unit tests pass
- [ ] SQL logic tests pass


aminaramoon:
I think this can break duckdb fallback mechansim as well, please make sure this PR adds value and not breaks the fallback before merging

mbrobbel:
> I think this can break duckdb fallback mechansim as well, please make sure this PR adds value and not breaks the fallback before merging

This is just for exceptions in Sirius, and the fallbacks catch `std::exception&`:
https://github.com/sirius-db/sirius/blob/4a7299190f14f2619f72940731ef4b00f1cc9b83/src/sirius_extension.cpp#L296
https://github.com/sirius-db/sirius/blob/4a7299190f14f2619f72940731ef4b00f1cc9b83/src/sirius_extension.cpp#L421
https://github.com/sirius-db/sirius/blob/4a7299190f14f2619f72940731ef4b00f1cc9b83/src/sirius_engine.cpp#L212

9prady9 (commented):
Good direction — decoupling the core engine from DuckDB exception types is the right call as Sirius matures into a standalone engine.

Two architectural points that aren't tied to a specific line:

**No common base class**: All three exceptions inherit directly from `std::runtime_error` with no shared Sirius root type. Without a common `sirius_exception` base, any catch-all handler at glue-layer boundaries must list all three types individually. Consider:
```cpp
class sirius_exception : public std::runtime_error { using std::runtime_error::runtime_error; };
class internal_exception : public sirius_exception { ... };
class not_implemented_exception : public sirius_exception { ... };
class invalid_input_exception : public sirius_exception { ... };
```

**Exception translation at the DuckDB boundary**: When `not_implemented_exception` or `invalid_input_exception` propagates through `sirius_engine::execute()` to DuckDB's executor, DuckDB catches it as a generic `std::exception`. The original `duckdb::NotImplementedException` / `duckdb::InvalidInputException` produce categorized, user-friendly error messages; a plain `std::runtime_error` subclass may not. The PR summary says the glue layer keeps DuckDB exceptions, but `sirius_engine.cpp` is in the changed set. Is there a catch+translate in `sirius_interface.cpp` wrapping these back into appropriate DuckDB exception types? If not, this may be a user-facing regression.

> Claude review summary

cucascade
```diff
@@ -1 +1 @@
-Subproject commit 942c0bf0539b23ed2424a5178d757526d439e5b6
+Subproject commit cafd9b694b0f756b5694865bc11b2dbafc87f8ae
```
9prady9: This submodule bump appears unrelated to the exception refactoring. It should either live in a separate commit/PR or be explicitly called out in the PR description as intentional.

src/include/sirius/exception.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <cstdarg>
+#include <cstdio>
+#include <stdexcept>
+#include <string>
+
+namespace sirius {
+
+namespace detail {
+
+inline std::string format_message(const char* fmt, ...)
```
9prady9: `format_message` uses C-style `va_list`/`vsnprintf` in C++20 code, and the variadic template constructors forward into it. This works but loses compile-time type safety and pulls in `<cstdarg>` / `<cstdio>` unnecessarily. Since the project is already on C++20, `std::format` is available and a cleaner fit:
```cpp
#include <format>

template <typename... Args>
explicit internal_exception(std::format_string<Args...> fmt, Args&&... args)
  : std::runtime_error(std::format(fmt, std::forward<Args>(args)...))
{}
```
This eliminates the `detail::format_message` helper and both C headers entirely, and the format string is checked at compile time.

src/op/sirius_physical_result_collector.cpp
```diff
@@ -31,8 +31,10 @@
 #include <cucascade/memory/common.hpp>
 #include <cucascade/memory/memory_reservation_manager.hpp>
 
+// sirius exceptions
+#include <sirius/exception.hpp>
```
9prady9: All other files in this PR use quote-style includes (`"sirius/exception.hpp"`), but this file uses angle brackets (`<sirius/exception.hpp>`). Should be consistent with the rest of the codebase.

src/op/sirius_physical_result_collector.cpp
```diff
@@ -121,15 +123,15 @@ void sirius_physical_materialized_collector::sink(const operator_data& input_dat
 
   if (input_batches.empty()) {
     return;  // todo(kevin) we should handle this case properly
-    throw duckdb::InvalidInputException("[GPUPhysicalMaterializedCollector] input_batches is null");
+    throw invalid_input_exception("[GPUPhysicalMaterializedCollector] input_batches is null");
```
9prady9: This throw is dead code — control always exits via `return` on the line above (todo(kevin)). The exception type has been updated here, but the throw is still unreachable. Since this PR is already touching the line, worth either removing it or fixing the underlying TODO so the error path is actually reachable.

9prady9 (approved):
lgtm

PR #563, opened 2026-04-02
Add prepare_for_processing virtual method to operator_data

aminaramoon:
Extract batch locking logic from gpu_pipeline_task::execute into a virtual operator_data::prepare_for_processing method. The method iterates over all data batches, locking each into the requested memory space via the shared lock_or_prepare_batch utility, and returns the processing handles to the caller.

gpu_pipeline_task::execute now calls prepare_for_processing and handles OOM by recording memory metrics and raising oom_reschedule_exception, keeping exception-handling policy out of operator_data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ cufile.log
 .pixi/*
 !.pixi/config.toml
 compile_commands.json
+.cache
 
 # Zed editor settings
 .zed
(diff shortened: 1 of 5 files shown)


bwyogatama:
Merged into #573

bwyogatama:
Closed since we use #573

PR #565, opened 2026-04-02
Replace libconfig++ with YAML config and simplify config framework

### LINKED ISSUE #548: Remove use of libconfig
We should remove usage of libconfig to support #547, because (as far as I understand licenses) its LGPLv2 license is not compatible with this plan.

mbrobbel:
## Summary

- Replace libconfig++ (LGPL) with yaml-cpp (MIT)
- Rewrite config framework from template metaprogramming (~870 lines) to simple `yaml::reader` pattern (~280 lines)
- Config files change from `.cfg` to YAML `.yaml` format
- Add `SIRIUS_DISABLE` env var to explicitly disable Sirius initialization
- Remove extension lock (flock-based process lock)
- Auto-initialize Sirius with defaults when no config file exists
- Add human-readable byte suffixes (`8Gi`, `512M`, `1T`) for byte-valued config fields
- Warn users who still have a legacy `sirius.cfg` file to migrate to the new YAML format
- Add runtime warning for legacy Sirius users (`gpu_buffer_init`/`gpu_processing`) that `SIRIUS_DISABLE=1` is required — Super Sirius now claims most GPU and pinned host memory on extension load, breaking the legacy code path

Closes #548

## Motivation

**License**: libconfig++ is LGPL-licensed which is problematic for Sirius. yaml-cpp is MIT.

**Complexity**: The old `configuration_setter` / `custom_config_registrar` framework used heavy template metaprogramming to map config files to C++ structs. Adding a config field required editing two separate locations — the struct definition and a registrar specialization in a different file. The new pattern keeps deserialization next to the struct:

```cpp
struct gpu_mem_config {
  double usage_limit_fraction{0.95};  // default right here

  static void from_yaml(const YAML::Node& node, gpu_mem_config& cfg) {
    yaml::reader r(node, "memory.gpu");
    r.optional("usage_limit_fraction", cfg.usage_limit_fraction, yaml::fraction<double>{});
    r.reject_unknown();
  }
};
```

**Extension lock**: The flock-based lock prevented multiple processes from loading Sirius simultaneously, which was overly restrictive and caused issues in testing. Note: removing the lock means concurrent processes can now independently initialize against the same GPU. Each sizes its memory pools from its own snapshot of free VRAM, so concurrent runs may overcommit. This matches how most GPU libraries behave (e.g. PyTorch, cuDF).

**No config = failure**: Previously `gpu_execution()` would error if no config file existed. Now Sirius auto-initializes with sane defaults (95% GPU memory, 4 pipeline threads, etc.).

## Config file format

**Before** (`sirius.cfg`):
```
sirius = {
    topology = { num_gpus = 1; };
    memory = {
        gpu = { usage_limit_fraction = 0.9; };
        host = { capacity_bytes = 32000000000; };
    };
};
```

**After** (`sirius.yaml`):
```yaml
sirius:
  topology:
    num_gpus: 1
  memory:
    gpu:
      usage_limit_fraction: 0.9
    host:
      capacity_bytes: 32Gi
```

## Config file resolution (new)

Sirius searches in this order:
1. `SIRIUS_CONFIG_FILE` environment variable (errors if set but file missing)
2. `./sirius.yaml` in the current working directory
3. `~/.sirius/sirius.yaml` in the user's home directory
4. No file found → built-in defaults

When no YAML config is found but a legacy `sirius.cfg` exists in `./` or `~/.sirius/`, Sirius logs a warning pointing users to migrate to the new format.

## `SIRIUS_DISABLE` environment variable (new)

Set `SIRIUS_DISABLE=1` to prevent Sirius from initializing. This replaces the old pattern of unsetting/clearing `SIRIUS_CONFIG_FILE` to run CPU-only. Scripts that previously used `unset SIRIUS_CONFIG_FILE` for CPU baselines now use `export SIRIUS_DISABLE=1`.

**Legacy Sirius users** (`gpu_buffer_init`/`gpu_processing`): Super Sirius now auto-initializes on extension load and claims most GPU and pinned host memory upfront. This breaks the legacy code path, which needs to manage its own memory via `gpu_buffer_init`. Legacy users **must** set `SIRIUS_DISABLE=1` before loading the extension. A runtime warning is now emitted when Sirius initializes with defaults to alert users who may be affected.

## Human-readable byte suffixes (new)

Byte-valued config fields (marked with `yaml::bytes()` in code) accept suffixes following the Kubernetes/systemd convention:

| Decimal (SI) | Binary (IEC) |
|---|---|
| `K`/`KB` = 1000 | `Ki`/`KiB` = 1024 |
| `M`/`MB` = 1000² | `Mi`/`MiB` = 1024² |
| `G`/`GB` = 1000³ | `Gi`/`GiB` = 1024³ |
| `T`/`TB` = 1000⁴ | `Ti`/`TiB` = 1024⁴ |

Non-byte integer fields (e.g. `num_threads`, `num_gpus`) reject string suffixes.

## Breaking changes

1. **Config files must be converted** from `.cfg` to `.yaml` format
2. **Config file name**: `sirius.cfg` → `sirius.yaml`
3. **Cache setting**: `cache = false` → `cache: "none"`, `cache = "parquet"` → `cache: "parquet"`
4. **Extension lock removed** — multiple processes can now load Sirius simultaneously
5. **Disabling Sirius**: use `SIRIUS_DISABLE=1` instead of unsetting `SIRIUS_CONFIG_FILE`
6. **Legacy code path**: `gpu_buffer_init`/`gpu_processing` users must set `SIRIUS_DISABLE=1` before loading the extension — Super Sirius now claims GPU memory on load

## What was removed

| Removed | Lines | Reason |
|---------|-------|--------|
| `config_option.hpp` | 871 | Entire template framework replaced by `yaml_reader.hpp` (280 lines) |
| `extension_lock.cpp/.hpp` | 153 | No longer needed |
| All `.cfg` test data files | 9 files | Replaced with `.yaml` equivalents |

## Test plan

- [x] 15 `[config_opt]` unit tests (basic types, optional, required, validation, arrays, nested structs, enums, byte suffixes, unknown key detection, error messages)
- [x] 2 `[sirius][context]` integration tests (loading configurator.yaml and spaces.yaml)
- [ ] Full test suite: 740 test cases, 78M+ assertions
- [x] No remaining references to `libconfig`, `.cfg`, or `extension_lock` in source/test code
- [ ] Manual: verify `gpu_execution` works without any config file (defaults)
- [ ] Manual: verify `./sirius.yaml` is picked up from working directory
- [ ] Manual: verify byte suffixes work in a real config (e.g. `capacity_bytes: 64Gi`)
- [ ] Manual: verify `SIRIUS_DISABLE=1` prevents initialization
- [ ] Manual: verify legacy `.cfg` file triggers migration warning
- [ ] Manual: verify legacy `gpu_buffer_init` without `SIRIUS_DISABLE=1` shows runtime warning

🤖 Created by Claude


aminaramoon:
I like the unit string support. we lose type safety and array vs list distinction and that was the reason I chose cfg but I am okay with yaml too.

I am not going to review the PR, as long as all the configs we have are working, it is upto people to decide between yaml and cfg. I am fine either way.

here are some questions:

1. do we want to get rid of the extension lock or do we want to at least keep in and add try_lock and at least inform user about the extension being used buy another process, cause we will run into weird issues when running multiple extension of single gpu.

2. did you drop variant support in the configs ? that is used in many places but I didn't see it supported in yaml::reader

3. adding default config when config is missing means we are disabling legacy sirius, is that what we want to do ?

4. does it support writing the config back to yaml ?

5. just a design consideration. 

```
static void from_yaml(const YAML::Node& node, gpu_mem_config& cfg) {
    yaml::reader r(node, "memory.gpu");
    r.optional("usage_limit_fraction", cfg.usage_limit_fraction, yaml::fraction<double>{});
    r.reject_unknown();
 }
```
  
I think you want to keep it out of config, because you are effectively injecting a yaml dependency to every config file and that can increase the compilation time.

aminaramoon (commented):
left some comments, concerns, but LGTM given it is communicated with people that they need to change their configs.

mbrobbel:
>     1. do we want to get rid of the extension lock or do we want to at least keep in and add try_lock and at least inform user about the extension being used buy another process, cause we will run into weird issues when running multiple extension of single gpu.

Good question. I prefer simplifying here, but I also don't know what the weird issues (other than OOM) are that could happen? I guess we can provide useful error/help messages in the case of OOMs. In the case of multiple available GPUs in the system, the default config (when users don't select a specific GPU) could even attempt to find a GPU with available memory?

> 
>     2. did you drop variant support in the configs ? that is used in many places but I didn't see it supported in yaml::reader

It's preserved, for example here: https://github.com/sirius-db/sirius/pull/565/changes#diff-13bfd674eac0c34ed6d286f95e90397feaca9270cd79621f403274756b14c432R130-R147

>     3. adding default config when config is missing means we are disabling legacy sirius, is that what we want to do ?

Can you elaborate?

>     4. does it support writing the config back to yaml ?

No. Is there a use-case for that?

>     5. just a design consideration.
> 
> 
> ```
> static void from_yaml(const YAML::Node& node, gpu_mem_config& cfg) {
>     yaml::reader r(node, "memory.gpu");
>     r.optional("usage_limit_fraction", cfg.usage_limit_fraction, yaml::fraction<double>{});
>     r.reject_unknown();
>  }
> ```
> 
> I think you want to keep it out of config, because you are effectively injecting a yaml dependency to every config file and that can increase the compilation time.

yaml_reader.hpp is only included from sirius_config.cpp, not from any headers. And the free functions keep the yaml-cpp dependency isolated to just that unit.

9prady9:
@mbrobbel Have you considered `TOML` format ? The reason I ask about it as an alternative for yaml is that, yaml is very inconvenient with those strict indentations which can go wrong any time and very hard to spot.

mbrobbel:
> @mbrobbel Have you considered `TOML` format ? The reason I ask about it as an alternative for yaml is that, yaml is very inconvenient with those strict indentations which can go wrong any time and very hard to spot.

Yes I have considered it, but with the nesting we currently have it gets way more verbose than `yaml`.

9prady9:
> > @mbrobbel Have you considered `TOML` format ? The reason I ask about it as an alternative for yaml is that, yaml is very inconvenient with those strict indentations which can go wrong any time and very hard to spot.
> 
> Yes I have considered it, but with the nesting we currently have it gets way more verbose than `yaml`.

but isn't that a bit more okay to manage than the issue with the spaces ? I have been bitten quite hard with yaml spaces/tabs, hence my apprehension :)

docs/super-sirius/configuration.md
```diff
@@ -6,71 +6,93 @@ This document covers Super Sirius configuration: the `sirius_config` class, oper
...
+| Suffix | Base | Example | Bytes |
+|--------|------|---------|-------|
+| `K`, `KB` | 1000 | `500K` | 500,000 |
+| `Ki`, `KiB` | 1024 | `500Ki` | 512,000 |
+| `M`, `MB` | 1000² | `512M` | 512,000,000 |
+| `Mi`, `MiB` | 1024² | `512Mi` | 536,870,912 |
+| `G`, `GB` | 1000³ | `8G` | 8,000,000,000 |
+| `Gi`, `GiB` | 1024³ | `8Gi` | 8,589,934,592 |
+| `T`, `TB` | 1000⁴ | `1T` | 1,000,000,000,000 |
+| `Ti`, `TiB` | 1024⁴ | `1Ti` | 1,099,511,627,776 |
```
9prady9, line 47: very good to have such a table here 👍🏾

docs/super-sirius/configuration.md
```diff
@@ -6,71 +6,93 @@ This document covers Super Sirius configuration: the `sirius_config` class, oper
...
+
+Fractional values are supported (e.g. `1.5Gi`). Follows the Kubernetes/systemd convention.
```
9prady9: perhaps hyperlink the Kube convention URL for the unfamiliar readers ?

docs/gpu_execution.md
```diff
@@ -20,14 +20,20 @@ Note that if building consumes too much memory, try reducing the `CMAKE_BUILD_PA
...
+`gpu_execution` requires a config file in YAML format. Sirius searches for it in this order:
 
-An example config file is provided at [`test/cpp/integration/integration.cfg`](../test/cpp/integration/integration.cfg). See the [Configuration documentation](super-sirius/configuration.md) for a full reference of all options.
+1. `SIRIUS_CONFIG_FILE` environment variable (explicit path)
+2. `./sirius.yaml` in the current working directory
+3. `~/.sirius/sirius.yaml` in the user's home directory
+
+If no config file is found, Sirius initializes with built-in defaults (95% GPU memory, 4 pipeline threads, etc.). Set `SIRIUS_DISABLE=1` to prevent Sirius from initializing. Byte values support human-readable suffixes like `8Gi`, `512Mi`, `1T` (see [Configuration docs](super-sirius/configuration.md#byte-suffixes)).
+
+An example config file is provided at [`test/cpp/integration/integration.yaml`](../test/cpp/integration/integration.yaml). See the [Configuration documentation](super-sirius/configuration.md) for a full reference of all options.
```
9prady9: I think it is better to hyperlink to the `Config File Resolution` section from configuration.md here instead of duplicating the info here which can potentially lead to diverging info in future.

scripts/patch_config.py
```diff
@@ -1,170 +1,69 @@
...
 Backs up the original file before modifying it. Supports adding new keys
 into existing groups, creating intermediate groups as needed.
 
 Usage:
-    python patch_config.py sirius.cfg \\
+    python patch_config.py sirius.yaml \\
         --opt sirius.executor.pipeline.num_threads=4 \\
-        --opt sirius.executor.duckdb_scan.cache=true \\
+        --opt sirius.executor.duckdb_scan.cache=parquet \\
         --opt sirius.operator_params.scan_task_batch_size=536870912
```
9prady9, line 11: curious Q for me, Is my assumption CLI value takes precedence over the value in config file in such invocations ?

scripts/patch_config.py
```diff
@@ -1,170 +1,69 @@
...
+        if part not in current:
+            current[part] = {}
+        current = current[part]
+    current[parts[-1]] = value
 
 
 def main():
     parser = argparse.ArgumentParser(
-        description="Patch a libconfig file with --opt KEY=VALUE overrides.",
+        description="Patch a YAML config file with --opt KEY=VALUE overrides.",
```
9prady9, line 62: I think this answers my earlier query.

src/include/yaml_reader.hpp
9prady9, line 1: nit: lot of implementations in hpp right, move to cpp ?

src/sirius_config.cpp
```diff
@@ -147,167 +222,154 @@ struct disk_mem_config {
...
-};
+// ================ sirius_config ================= //
 
 sirius_config::sirius_config()
 {
   cucascade::memory::topology_discovery discovery;
   if (discovery.discover()) { _hw_topology = discovery.get_topology(); }
 }
 
+void sirius_config::apply_defaults()
```
9prady9, line 271: I might have missed it's invocation, where all do we call this ?

mbrobbel: https://github.com/sirius-db/sirius/pull/565/changes#diff-456aa5febd03185f6062a1a72e98b0d44e3b70788611764f61d115779841a8deR440

src/include/yaml_reader.hpp
mbrobbel: This header is only included from `sirius_config.cpp`. There is no compile-time benefit, it would just split the impl across files.

docs/gpu_execution.md
mbrobbel: [1ad863a](https://github.com/sirius-db/sirius/pull/565/commits/1ad863a21b14a38f544dbab374d761ef86bd4de6)

docs/super-sirius/configuration.md
mbrobbel: [1ad863a](https://github.com/sirius-db/sirius/pull/565/commits/1ad863a21b14a38f544dbab374d761ef86bd4de6)

9prady9 (approved):
lgtm

aminaramoon (approved):

PR #571, opened 2026-04-03
Add metadata scan task

### LINKED ISSUE #522: [IO Optimization]: separate metadata scan from physical scan
metadata scan: reading parquet/duckdb metadata in the operator, separate from parquet[duckdb] global state
pipeline needs to change, so that we have metadata scan + physical scan

### LINKED ISSUE #523: [IO Optimization]: Join Phsycial Scan + Table Scan tasks into an operator
Merge Physical Scan + Table Scan operators into one operator, read into GPU + filter

aminaramoon:
- **Add `sirius_parquet_metadata_scan_operator`**: a source operator that parses parquet file footers, translates filter expressions for  
  pushdown, and partitions row groups into batches sized by `approximate_batch_size`. Supports both AST filter pushdown (for row group      
  pruning + cuDF reader filters) and fallback to DuckDB expression evaluation.                                                              
  - **Add `sirius_gpu_parquet_scan_operator`**: a sink+source operator that accumulates partitioned metadata from pipeline 1, then serves   
  self-contained `parquet_scan_data` tasks for pipeline 2. Each task reads its partition via `cudf::io::read_parquet`, applies fallback     
  filter evaluation if AST translation failed, and prunes pure filter columns post-filtering.                                               
  - **Add shared scan data types**: `parquet_metadata_input`, `partitioned_parquet_metadata`, `parquet_scan_data`, and `row_group_range` in 
  `parquet_scan_operator_data.hpp`. `row_group_range` now stores explicit `vector<cudf::size_type>` indices instead of start/count, and     
  `reader_options` / `filter_expression` use `shared_ptr` for safe sharing across tasks.                                                    
  - **Add `scan_utils`**: extracts `convert_table_filters_to_expression` and `build_batch_column_map` from the legacy scan path into        
  reusable utilities.                                                                                                                       
  - **Add `prepare_for_processing` virtual method to `operator_data`**: enables the pipeline to prepare data batches for processing without 
  knowing the concrete type.                                                                                                                
  - **Add 13 unit tests** covering the full two-pipeline flow: basic scan, projection, diverse types (VARCHAR, DECIMAL, DATE),              
  integer/bigint filters, projection+filter, multi-file, partitioning behavior, empty filter results, and operator lifecycle.               
                                                                                                                                            
  ## Test plan                                                                                                                              
                                                                                                                                            
  - [x] All 13 new `[two_pipeline_scan]`, `[metadata_scan]`, and `[gpu_scan_operator]` tests pass                                           
  - [x] All 12 existing `[parquet_scan_task]` tests pass (no regression from `row_group_range` refactor)                                    
  - [x] Run full `[shared_context]` test suite                                                                                              
  - [x] Run `[gpu_execution]` integration tests                                                                                             
                                                                                                                                            
  Fixes #522, fixes #523


src/expression_executor/gpu_expression_translator.cpp
```diff
@@ -18,10 +18,196 @@
...
+#include <cudf/wrappers/timestamps.hpp>
+
 namespace sirius {
 
 using expr_ref = std::reference_wrapper<cudf::ast::expression const>;
 
+namespace {
+
+/// @brief Convert a cudf::ast::ast_operator to a human-readable string.
+std::string ast_operator_to_string(cudf::ast::ast_operator op)
```
aminaramoon, line 29: why hide this function ? do you think we need this converter elsewhere too at some point ?

src/expression_executor/gpu_expression_translator.cpp
```diff
@@ -18,10 +18,196 @@
...
+    case cudf::ast::ast_operator::GREATER_EQUAL: return ">=";
+    case cudf::ast::ast_operator::BITWISE_AND: return "&";
+    case cudf::ast::ast_operator::BITWISE_OR: return "|";
+    case cudf::ast::ast_operator::BITWISE_XOR: return "^";
+    case cudf::ast::ast_operator::LOGICAL_AND: return "AND";
+    case cudf::ast::ast_operator::NULL_LOGICAL_AND: return "null_and";
+    case cudf::ast::ast_operator::LOGICAL_OR: return "OR";
+    case cudf::ast::ast_operator::NULL_LOGICAL_OR: return "null_or";
+    case cudf::ast::ast_operator::IDENTITY: return "identity";
+    case cudf::ast::ast_operator::IS_NULL: return "IS_NULL";
```
aminaramoon, line 56: is this correct ? it looks different than others

src/expression_executor/gpu_expression_translator.cpp
```diff
@@ -18,10 +18,196 @@
...
+    case cudf::ast::ast_operator::EXP: return "exp";
+    case cudf::ast::ast_operator::LOG: return "log";
+    case cudf::ast::ast_operator::SQRT: return "sqrt";
+    case cudf::ast::ast_operator::CBRT: return "cbrt";
+    case cudf::ast::ast_operator::CEIL: return "ceil";
+    case cudf::ast::ast_operator::FLOOR: return "floor";
+    case cudf::ast::ast_operator::ABS: return "abs";
+    case cudf::ast::ast_operator::RINT: return "rint";
+    case cudf::ast::ast_operator::BIT_INVERT: return "~";
+    case cudf::ast::ast_operator::NOT: return "NOT";
```
aminaramoon, line 78: same here

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
```diff
@@ -146,85 +113,111 @@ size_t sirius_gpu_parquet_scan_operator::get_total_partitions() const
...
+  auto datasource = std::move(scan_data->datasource);
+  if (!datasource) { datasource = cudf::io::datasource::create(scan_data->file_path); }
+
+  // Build reader options for this partition's row groups.
+  auto opts = *scan_data->reader_options;
+  opts.set_source(cudf::io::source_info{datasource.get()});
+  opts.set_row_groups({scan_data->rg_range.row_group_indices});
+
+  // Read the parquet data onto the GPU.
+  auto [table, metadata] = cudf::io::read_parquet(opts, stream);
```
aminaramoon, line 175: this makes a copy, right ?

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
```diff
@@ -146,85 +113,111 @@ size_t sirius_gpu_parquet_scan_operator::get_total_partitions() const
...
-  SIRIUS_LOG_ERROR(
-    "[sirius_gpu_parquet_scan_operator] execute() is not yet fully integrated. "
-    "A dedicated parquet_gpu_scan_task type must provide the host-memory reservation context.");
-  throw std::runtime_error(
-    "[sirius_gpu_parquet_scan_operator] execute() is not yet fully integrated. "
-    "A dedicated parquet_gpu_scan_task type must provide the host-memory reservation context.");
+  // If datasource has not been set by prefetcher, create it now.  The datasource
+  // field on parquet_scan_data is a mutable transport slot that may be filled by the
+  // prefetcher before execution; if it's still null, we create one from the file path.
+  auto datasource = std::move(scan_data->datasource);
```
aminaramoon: why cannot we use the same datasource used for metadata scan ?

src/expression_executor/gpu_expression_translator.cpp
kevkrist: I found this very useful when debugging and figured others would as well as we increase our reliance on cudf::ast stuff. So you're right -- I've pulled those out.

src/expression_executor/gpu_expression_translator.cpp
kevkrist: I capitalized all verbal to_strings

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
kevkrist: It has to be a move since the copy constructor on a unique_ptr is deleted, right?
Looking more into it, the tuple unpacking is a move via 'structured binding'.

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
kevkrist: You're right -- the datasource from the metadata scan is now passed forward.

kevkrist (commented):
I also moved the attempt to translate the duckdb filter into the metadata scan operator's execute method so that 1) a proper stream is available and 2) computation is offloaded from the constructor.

PR #577, opened 2026-04-05
Auto-init git submodules on pixi shell activation

bwyogatama:
## Summary
- New users cloning without `--recurse-submodules` get broken symlinks in `.claude/commands/` and `.claude/skills/` because the `claude-tools` submodule isn't initialized
- Adds a check to `scripts/pixi_activate.sh` that auto-runs `git submodule update --init --recursive` when any submodule is uninitialized
- No-op if all submodules are already checked out

## Test plan
- [ ] Clone repo without `--recurse-submodules`, run `pixi shell`, verify submodules are initialized
- [ ] Run `pixi shell` on an existing checkout — confirm no extra output or delay

### DIFF
--- a/scripts/pixi_activate.sh
+++ b/scripts/pixi_activate.sh
@@ -20,3 +20,12 @@ rm -f "$project_root/duckdb/CMakeUserPresets.json"
 if [[ ! -e "$cmake_presets_dst" ]]; then
   ln -s "$cmake_presets_src" "$cmake_presets_dst"
 fi
+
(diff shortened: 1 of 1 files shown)


mbrobbel (commented):
I'm -1 on this change, mainly because I don't like it when scripts run git commands. It's also something that doesn't have anything to do with the activation of the Pixi environment.
Since this is about claude-tools, I think we should add a line to `CLAUDE.md` instead, to hint at the existence of this submodule for additional skills, so that it can fetch it if needed.

PR #579, opened 2026-04-06
Downgrade request

wmalpica:
Implements #474 

Downgrade Executor Redesign                                                                                                                                                                             
                                                                                                                                                                                                          
  Project Overview                                                                                                                                                                                        
                                                  
  The downgrade executor is Sirius's mechanism for relieving GPU memory pressure by migrating data_batch objects from GPU to HOST memory. Previously, it inherited from itask_executor and operated on a  
  queue-of-tasks model where each task downgraded a single batch. This PR redesigns the executor from scratch around a queue-of-requests model, where the unit of work is "free a target amount of memory
  (or satisfy a caller-defined predicate) by downgrading batches concurrently."                                                                                                                           
                                                  
  The redesign was executed in three phases (all in this PR) using get-shit-done for Claude:                                                                                                                                                              
  1. Foundation — Standalone executor with own thread pool, decoupled from itask_executor
  2. Request Execution & API — Predicate-driven incremental dispatch with async/sync public APIs                                                                                                          
  3. Lifecycle & Pipeline Integration — Retry-with-downgrade loop in gpu_pipeline_executor, SiriusContext init reordering, lifecycle tests
                                                                                                                                                                                                          
 Design Decisions:                                                                                                                                                                                        
- Drop itask_executor inheritance — The base class queue-of-tasks model doesn't fit queue-of-requests; fighting the abstraction added complexity.                                                         
                                                  
- Predicate as fundamental API — More flexible than byte-count-only; byte-based API is a thin wrapper. Enables callers like the pipeline executor to define success conditions (e.g., "I got my  reservation") rather than guessing how many bytes to free.
                                                                                                                                                                                                          
- std::future for async results — Simple, standard, no callback complexity; caller can poll or block.                                                                                                     
   
- Sequential request processing — Avoids contention between concurrent requests competing for the same batches.                                                                                           
                                                  
- weak_ptr candidates — Candidates collected for downgrade are held as weak_ptr and locked just before dispatch. Batches freed elsewhere between collection and dispatch are skipped, avoiding unnecessary lifetime extension and redundant downgrades.
                                                                                                                                                                                                          
- All requests require target_bytes — The executor needs to know approximately how many candidates to collect. Even predicate-driven requests must declare a target so candidate collection is bounded.

- Predicate checked after each batch — Enables earliest possible early-exit, minimizing unnecessary downgrades.                                                                                           
   
- Monitor uses fire-and-forget — The internal monitor loop pushes requests to its own queue rather than calling the blocking external API.                                                                
                                                  
- Pipeline executor uses predicate-based downgrade — Instead of a retry loop calling request_free_memory_and_wait(shortfall) repeatedly, the pipeline executor issues a single request_downgrade with a predicate that attempts make_reservation_or_null. The downgrade stops as soon as enough memory is free to satisfy the reservation — no over-freeing, no retry overhead.
                                                                                                                                                                                                          
  **Public API**                                      
         
```                                                                                                                                                                                                 
  // Async: enqueue request, returns future that resolves to actual bytes freed
  std::future<size_t> request_free_memory(size_t bytes);                                                                                                                                                  
   
  // Blocking: waits for completion, returns actual bytes freed                                                                                                                                           
  size_t request_free_memory_and_wait(size_t bytes);
                                                                                                                                                                                                          
  // Predicate-driven: collects ~target_bytes of candidates, dispatches downgrades
  // until predicate() returns true or candidates exhausted                                                                                                                                               
  std::future<size_t> request_downgrade(size_t target_bytes, std::function<bool()> predicate);                                                                                                            
   
  // Lifecycle (called by SiriusContext)                                                                                                                                                                  
  void start();                                   
  void stop();                                                                                                                                                                                            
  void drain();       
```                            

  request_free_memory — Wraps request_downgrade with a predicate that checks bytes_freed >= bytes. Fire-and-forget or .get() to block.                                                                    
   
  request_downgrade — The core API. The caller provides a predicate that defines "done" (e.g., "I successfully made a memory reservation"). The executor collects up to target_bytes worth of candidate   
  batches as weak_ptr, then dispatches downgrades to the thread pool. After each batch completes, the predicate is evaluated; if true, no new batches are dispatched (in-flight batches finish naturally).
   Returns total bytes freed.                                                                                                                                                                             
                                                  
  **Pipeline Integration**

  When gpu_pipeline_executor gets a partial reservation because there is not enough memory, it now issues a single predicate-based downgrade request:                                                    
   
```
  _downgrade_executor->request_downgrade(                                                                                                                                                                 
      target_bytes,                               
      [mem_space, bytes_needs, &new_reservation]() {
          auto res = mem_space->make_reservation_or_null(bytes_needs);                                                                                                                                    
          if (res && res->size() >= bytes_needs) {                                                                                                                                                        
              new_reservation = std::move(res);                                                                                                                                                           
              return true;                                                                                                                                                                                
          }                                       
          return false;                                                                                                                                                                                   
      }).get();      
```                             

  The target_bytes is max(shortfall * 1.25, bytes_needed / 4) — overcollecting candidates accounts for weak_ptr expiration and ensures enough are available to satisfy the reservation.       
   
  **Test plan**                                                                                                                                                                                               
                                                  
  - Unit tests: test_downgrade_executor.cpp — 12 test cases covering request APIs, predicate early-exit, partial fulfillment, concurrent dispatch                                                         
  - Lifecycle tests: test_downgrade_lifecycle.cpp — 6 test cases covering start/stop, drain, monitor loop, concurrency, CUDA streams


test/cpp/config/data/configurator.cfg
```diff
@@ -30,6 +30,7 @@ sirius = {
         downgrade = {
             num_threads = 1;
             thread_name_prefix = "sirius_downgrade_executor";
+            monitor_period_ms = 10;
```
9prady9, line 33: seems like we are adding the same value in most of these. Does our current cfg system set this as default if one is not provided ?

src/sirius_config.cpp
```diff
@@ -85,6 +85,19 @@ struct sirius::config::custom_config_registrar<sirius::exec::thread_pool_config>
...
+template <>
+struct sirius::config::custom_config_registrar<sirius::exec::downgrade_executor_config> {
+  static void config(sirius::config::configuration_setter& setter,
+                     sirius::exec::downgrade_executor_config& opt)
+  {
+    setter.add_config(
+      "num_threads", opt.thread_pool.num_threads, sirius::config::greater_than<size_t>{0});
+    setter.add_config("thread_name_prefix", opt.thread_pool.thread_name_prefix);
+    setter.add_config("cpu_affinity", opt.thread_pool.cpu_affinity_list);
+    setter.add_config("monitor_period_ms", opt.monitor_period_ms);
```
9prady9, line 97: I guess should have asked about the default value to monitor period here.

src/include/pipeline/pipeline_executor.hpp
```diff
@@ -67,10 +71,13 @@ class pipeline_executor {
    * @param mem_mgr Reference to the memory reservation manager
    * @param sys_topology Optional system topology info for CPU affinity
    */
```
9prady9, line 74: missing doc string for downgrade executors arg.

src/include/pipeline/gpu_pipeline_executor.hpp
```diff
@@ -59,7 +63,8 @@ class gpu_pipeline_executor : public sirius::parallel::itask_executor {
   explicit gpu_pipeline_executor(
     exec::thread_pool_config config,
     cucascade::memory::memory_space* mem_space,
-    exec::publisher<std::unique_ptr<task_request>> task_request_publisher);
+    exec::publisher<std::unique_ptr<task_request>> task_request_publisher,
+    sirius::parallel::downgrade_executor* downgrade_executor = nullptr);
```
9prady9, line 70: missing doc string here as well.

src/downgrade/downgrade_executor.cpp
```diff
@@ -16,93 +16,171 @@
...
-}
+  bool expected = false;
+  if (!_running.compare_exchange_strong(expected, true)) { return; }
 
-absl::AnyInvocable<void() noexcept> downgrade_executor::get_per_thread_init()
-{
   if (_memory_space) {
     auto device_id = _memory_space->get_device_id();
-    return [device_id]() noexcept { cudaSetDevice(device_id); };
+    cudaSetDevice(device_id);
```
aminaramoon: can we use rmm streams ? can we also use rmm::set_device_id_raii here ? it sets the evice back to what it was befor the call

src/downgrade/downgrade_executor.cpp
```diff
@@ -247,26 +317,37 @@ size_t downgrade_executor::run_downgrade_pass(std::vector<downgrade_repository_i
...
+
+std::future<size_t> downgrade_executor::request_downgrade(size_t target_bytes,
+                                                          std::function<bool()> predicate)
+{
+  auto req          = std::make_unique<downgrade_request>();
+  req->target_bytes = target_bytes;
+  req->predicate    = std::move(predicate);
+  auto future       = req->result.get_future();
+  _request_queue.push(std::move(req));
+  return future;
```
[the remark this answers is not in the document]
9prady9: this is a valid catch right ? we are not handling the push's boolean return value in any form. At the least we should log right

src/downgrade/downgrade_executor.cpp
```diff
@@ -247,26 +339,37 @@ size_t downgrade_executor::run_downgrade_pass(std::vector<downgrade_repository_i
...
+
+std::future<size_t> downgrade_executor::request_free_memory(size_t bytes)
+{
+  auto req          = std::make_unique<downgrade_request>();
+  req->target_bytes = bytes;
+  req->predicate    = [&freed = req->bytes_freed, bytes]() {
+    return freed.load(std::memory_order_relaxed) >= bytes;
+  };
+  auto future = req->result.get_future();
+  _request_queue.push(std::move(req));
```
9prady9: here too, push's bool return should let us at least log.

src/downgrade/downgrade_executor.cpp
```diff
@@ -35,74 +34,143 @@ downgrade_executor::downgrade_executor(
...
+                       stream   = rmm::cuda_stream_view{_stream},
+                       batch_size]() {
+                        downgrade_task task{batch, res_mgr};
+                        try {
+                          task.execute(stream);
+                          req_ptr->bytes_freed.fetch_add(batch_size, std::memory_order_relaxed);
+                          req_ptr->batches_downgraded.fetch_add(1, std::memory_order_relaxed);
+                          if (req_ptr->predicate && req_ptr->predicate()) {
+                            req_ptr->satisfied.store(true, std::memory_order_release);
+                          }
```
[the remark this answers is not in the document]
9prady9: has this ^ been already addressed ?

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -105,11 +110,87 @@ void gpu_pipeline_executor::manager_loop()
...
+        gpu_task->get_pipeline_id(),
+        gpu_task->get_task_id(),
+        target_bytes);
+
+      reservation.reset();  // release partial reservation before downgrade
+
+      std::unique_ptr<cucascade::memory::reservation> new_reservation;
+      auto* mem_space = _memory_space;
+      size_t freed    = 0;
+      std::mutex reservation_mutex;
```
9prady9, line 134: I couldn't quite get why this mutex is needed ? The lambda will ever be executed only from one thread right ?

test/cpp/config/data/configurator.cfg
wmalpica: yes it does

src/pipeline/gpu_pipeline_executor.cpp
wmalpica: The predicate is evaluated after every batch that is downgraded. These are happening in parallel in a thread pool. That is why i setup the mutex.

src/sirius_config.cpp
wmalpica: Yes the default is 10. Its set in config.hpp

src/downgrade/downgrade_executor.cpp
```diff
@@ -16,93 +16,171 @@
...
+
+  _pool->wait_all();
+  cancel_pending_requests();
+  _pool->stop();
+  _pool.reset();
 
-void downgrade_executor::on_stopped()
-{
   if (_stream) {
     cudaStreamDestroy(_stream);
```
aminaramoon: can we use rmm::streams instead

src/downgrade/downgrade_executor.cpp
```diff
@@ -114,15 +192,29 @@ void downgrade_executor::monitor_loop()
...
       if (amount > 0) {
-        // Collect all repositories from the manager
-        std::vector<downgrade_repository_info> repos;
-        _data_repo_mgr.for_each_repository(
-          [&repos](cucascade::shared_data_repository* repo) { repos.push_back({repo}); });
-        if (!repos.empty()) { run_downgrade_pass(std::move(repos), amount); }
+        auto req          = std::make_unique<downgrade_request>();
+        req->target_bytes = amount;
+        req->predicate    = [&freed = req->bytes_freed, amount]() {
+          return freed.load(std::memory_order_relaxed) >= amount;
```
aminaramoon, line 199: can we jjust load. instead of adding memory_order, I know it is claude, but it is better to not set it

src/downgrade/downgrade_executor.cpp
```diff
@@ -16,93 +16,171 @@
...
+                       &res_mgr = _reservation_manager,
+                       stream   = rmm::cuda_stream_view{_stream},
+                       batch_size]() {
+                        downgrade_task task{batch, res_mgr};
+                        try {
+                          if (task.execute(stream)) {
+                            req_ptr->bytes_freed.fetch_add(batch_size, std::memory_order_relaxed);
+                            req_ptr->batches_downgraded.fetch_add(1, std::memory_order_relaxed);
+                            if (req_ptr->predicate && req_ptr->predicate()) {
+                              req_ptr->satisfied.store(true, std::memory_order_release);
```
aminaramoon: can we get rid of all the memory_orders, `memory_order_release` is not the correct memory order here. let;s just `store()` and `load()`

src/downgrade/downgrade_executor.cpp
```diff
@@ -247,26 +339,47 @@ size_t downgrade_executor::run_downgrade_pass(std::vector<downgrade_repository_i
...
+                                                          std::function<bool()> predicate)
+{
+  auto req          = std::make_unique<downgrade_request>();
+  req->target_bytes = target_bytes;
+  req->predicate    = std::move(predicate);
+  auto future       = req->result.get_future();
+  if (!_request_queue.push(std::move(req))) {
+    SIRIUS_LOG_WARN("[downgrade] request_downgrade: queue inactive, dropping request for {} bytes",
+                    target_bytes);
+    req->result.set_value(0);
```
bwyogatama, line 378: Segfault here cause you set the value after moving req

src/include/downgrade/downgrade_executor.hpp
```diff
@@ -82,92 +101,82 @@ class downgrade_executor : public sirius::parallel::itask_executor {
...
 
  private:
-  cudaStream_t _stream{nullptr};
+  exec::downgrade_executor_config _config;
+  std::unique_ptr<exec::bounded_thread_pool> _pool;
+  exec::interruptible_mpmc<std::unique_ptr<downgrade_request>> _request_queue;
+  std::thread _processing_thread;
   std::thread _monitor_thread;
+  std::atomic<bool> _running{false};
+  cudaStream_t _stream{nullptr};
```
bwyogatama: Do we have a single cuda stream shared across thread pool?

src/downgrade/downgrade_executor.cpp
```diff
@@ -16,93 +16,171 @@
...
-void downgrade_executor::on_stopped()
-{
   if (_stream) {
     cudaStreamDestroy(_stream);
     _stream = nullptr;
   }
 }
 
-void downgrade_executor::manager_loop()
+void downgrade_executor::drain()
```
bwyogatama, line 94: When is drain() called in the codebase, i actually can't find it for downgrade_executor

bwyogatama (approved):

src/downgrade/downgrade_executor.cpp
wmalpica: Its called in SiriusContext::QueryEnd()

aminaramoon (approved):

PR #583, opened 2026-04-07
fix: add memory reservations before host allocations

### LINKED ISSUE #566: add missing host reservations
│ 2   │ sirius_physical_result_collector.cpp:156       │ sink()                  │ Makes a reservation but the builtin converter doesn't use it for the actual allocation │
  
  │ 3   │ host_parquet_representation.cpp:32             │ clone()                 │ Direct allocate_multiple_blocks() without reservation                                  
  │ 4   │ host_parquet_representation_converters.cpp:124 │ Cross-NUMA copy         │ Direct allocation without reservation

kevkrist:
Fixes #566                                                                                                                  
                                                                                                                              
  - **`host_parquet_representation::clone()`**: Added `make_reservation_or_null()` before `allocate_multiple_blocks()`, with a
   throw if the reservation fails. Previously allocated without any reservation tracking.                                     
  - **`convert_host_parquet_to_host_parquet` (cross-NUMA copy)**: Added `make_reservation_or_null()` before                   
  `allocate_multiple_blocks()`, passing the reservation to the allocator. Previously allocated without reservation tracking.  
  - **`sirius_physical_result_collector::sink()`**: Released the existing reservation before calling `convert_to`, so the     
  built-in converter's internal allocation can use the freed capacity. Previously the reservation held capacity that the      
  converter couldn't use (double-counting).                                                                                   
                                                                                                                              
  ## Test plan                                                                                                                
                                                                                                                              
  - [x] Added `"host_parquet_representation clone uses reservation"` — verifies reservation count returns to baseline after   
  clone                                                                                                                       
  - [x] Added `"host_parquet_representation clone fails when reservation unavailable"` — exhausts memory, confirms clone      
  throws                                                                                                                      
  - [x] Added `"host_parquet_representation cross-host copy uses reservation"` — verifies converter's reservation is released 
  after conversion                                                                                                            
  - [x] All 12 `[host_parquet_representation]` tests pass (265 assertions)

### DIFF
--- a/src/data/host_parquet_representation.cpp
+++ b/src/data/host_parquet_representation.cpp
@@ -17,6 +17,9 @@
 // sirius
 #include <data/host_parquet_representation.hpp>
 
+// cucascade
+#include <cucascade/memory/memory_reservation.hpp>
+
 namespace sirius {
 
 std::unique_ptr<cucascade::idata_representation> host_parquet_representation::clone(
@@ -28,8 +31,15 @@ std::unique_ptr<cucascade::idata_representation> host_parquet_representation::cl
     throw std::runtime_error("Cannot clone host_parquet_representation: no host memory resource");
   }
 
-  // Allocate new blocks for the copy
-  auto allocation_copy = host_mr->allocate_multiple_blocks(_size_in_bytes);
+  // Reserve memory before allocating to ensure proper memory tracking
+  auto reservation = get_memory_space().make_reservation_or_null(_size_in_bytes);
+  if (!reservation) {
+    throw std::runtime_error(
+      "Cannot clone host_parquet_representation: failed to reserve host memory");
+  }
+
+  // Allocate new blocks for the copy, tracked against the reservation
+  auto allocation_copy = host_mr->allocate_multiple_blocks(_size_in_bytes, reservation.get());
 
   // Copy data block by block
   const auto& src_blocks = _column_chunks->get_blocks();
--- a/src/data/host_parquet_representation_converters.cpp
+++ b/src/data/host_parquet_representation_converters.cpp
@@ -24,6 +24,7 @@
 // cucascade
 #include <cucascade/data/gpu_data_representation.hpp>
 #include <cucascade/memory/fixed_size_host_memory_resource.hpp>
+#include <cucascade/memory/memory_reservation.hpp>
 #include <cucascade/memory/memory_space.hpp>
 
 // cudf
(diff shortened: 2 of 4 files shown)


src/data/host_parquet_representation_converters.cpp
```diff
@@ -120,8 +121,16 @@ std::unique_ptr<cucascade::idata_representation> convert_host_parquet_to_host_pa
       "Target HOST memory_space does not have a fixed_size_host_memory_resource");
   }
 
+  // Reserve memory before allocating to ensure proper memory tracking
+  auto reservation = const_cast<cucascade::memory::memory_space*>(target_memory_space)
+                       ->make_reservation_or_null(data_size);
+  if (!reservation) {
+    throw std::runtime_error(
+      "Cannot corss-NUMA copy host_parquet_representation: failed to reserve host memory");
```
wmalpica, line 129: NIT: typo `corss` -> `cross`

src/data/host_parquet_representation.cpp
```diff
@@ -28,8 +31,15 @@ std::unique_ptr<cucascade::idata_representation> host_parquet_representation::cl
     throw std::runtime_error("Cannot clone host_parquet_representation: no host memory resource");
   }
 
-  // Allocate new blocks for the copy
-  auto allocation_copy = host_mr->allocate_multiple_blocks(_size_in_bytes);
+  // Reserve memory before allocating to ensure proper memory tracking
+  auto reservation = get_memory_space().make_reservation_or_null(_size_in_bytes);
+  if (!reservation) {
+    throw std::runtime_error(
```
wmalpica, line 37: Something I am wondering, once this PR is merged https://github.com/sirius-db/sirius/pull/579
We will want to have a host memory downgrade executor from which we can request memory. 
In this case, would we want to throw (and the caller handles it) or we want the downgrade executor go downgrade some data to free memory so that the reservation can be made?

My assumption is that it will depend case by case on the caller and the situation. I think this is fine for now, but we will have to figure it out

test/cpp/data/test_host_parquet_representation.cpp
```diff
@@ -674,6 +731,54 @@ TEST_CASE("host_parquet_representation clone then convert to GPU",
...
+  REQUIRE(host_space_1 != nullptr);
+
+  parquet_test_fixture fixture;
+  fixture.setup(200, "cross_host_reservation");
+
+  auto repr = fixture.build_representation(host_space_0);
+  REQUIRE(repr != nullptr);
+
+  // Record reservation state on target space before conversion
+  auto reserved_before = host_space_1->get_total_reserved_memory();
```
wmalpica, line 768: In these unit tests, can you check the peak consumption of the host resource. That should not be the same before and after the clones

wmalpica (changes requested):

kevkrist:
Closing per offline conversation in Sirius meeting.

PR #594, opened 2026-04-08
Update libcudf to 26.04 stable release

mbrobbel:
Move libcudf dependency from default feature to CUDA-specific features so the duckdb-python environment (which doesn't need GPU libs) can resolve independently.


9prady9 (approved):
lgtm @mbrobbel I was about to the first thing today :) Should we wait for any other changes or approvals ?

PR #615, opened 2026-04-10
Support decimals in `gpu_expression_translator` bc libcudf version bump

kevkrist:
We can now support flat decimal evaluation in the `gpu_expression_translator`. This will propagate to the parquet scan row group pruning and filter pushdown for parquet scans, in particular. Nested decimal evaluation is still waiting on this [bug fix](https://github.com/rapidsai/cudf/pull/21996), so remains disabled. Tests have been updated to reflect this change. Also, new tests for string types have been added.


test/cpp/expression_executor/test_gpu_expression_translator.cpp
```diff
@@ -162,29 +149,79 @@ std::unique_ptr<cudf::table> make_mixed_table(std::vector<int32_t> const& i32_va
...
-             values.data(),
-             sizeof(int16_t) * n,
-             cudaMemcpyHostToDevice);
+  auto const n = static_cast<cudf::size_type>(reps.size());
+  auto col = cudf::make_fixed_width_column(cudf::data_type{cudf::type_id::DECIMAL32, cudf_scale},
+                                           n,
+                                           cudf::mask_state::UNALLOCATED,
+                                           stream,
+                                           mr);
+  cudaMemcpy(
```
aminaramoon, line 163: where is the stream here ? why not async copies on the stream

aminaramoon (changes requested):

test/cpp/expression_executor/test_gpu_expression_translator.cpp
kevkrist: This is a test file, so not necessary here?

aminaramoon (approved):

PR #624, opened 2026-04-13
fix: prevent concurrent boundary computation in sort_sample

### LINKED ISSUE #593: Flaky test failure
We have a flaky issue

https://github.com/sirius-db/sirius/actions/runs/24127018920/job/70394748479

```
test/cpp/integration/test_gpu_execution_tpch.cpp:108: FAILED:
  REQUIRE_FALSE( gpu_result->HasError() )
with expansion:
  !true
with message:
  gpu_execution error: Invalid Error: SiriusExecuteQuery error: Invalid Error:
  CUDF failure at: /tmp/conda-bld-output/bld/rattler-build_libcudf/work/cpp/
  include/cudf/detail/null_mask.cuh:606: End index cannot be smaller than the
  starting index.

```

9prady9:
Closes #593

The `task_creator` exhausts all available batches in a while loop, dispatching one task per batch for sort_sample before any of them run. On the pipeline executor's thread pool, multiple tasks could enter `execute()` simultaneously with `_boundary_state=0` (not started), each racing to write _partition_boundaries via an async `cudf::gather`. The second writer frees the first task's GPU allocation mid-gather, producing corrupt boundary data. sort_partition then passes these corrupt boundaries to `cudf::lower_bound`, yielding non-monotone split positions whose slice index pairs have end < begin, triggering the below cuDF assertion
```
End index cannot be smaller than the starting index
```

Fix with a lock-free 3-state atomic (0=not started, 1=computing, 2=done): a compare_exchange elects exactly one winner; all other tasks pass-through without blocking. The pipeline framework's own `stream.synchronize()` (called after `execute()` returns) guarantees _partition_boundaries is fully materialized before Pipeline C and sort_partition start, so no explicit sync is needed in execute().

🤖 Created by Claude
Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/op/sirius_physical_sort_sample.cpp
+++ b/src/op/sirius_physical_sort_sample.cpp
@@ -51,7 +51,7 @@ sirius_physical_sort_sample::sirius_physical_sort_sample(
 std::optional<task_creation_hint> sirius_physical_sort_sample::get_next_task_hint()
 {
   // If boundaries already computed, use default behavior (process batches as they arrive)
-  if (_boundaries_computed.load()) { return sirius_physical_operator::get_next_task_hint(); }
+  if (_boundary_state.load(std::memory_order_acquire) == 2) { return sirius_physical_operator::get_next_task_hint(); }
 
   // Need to wait for N batches before computing boundaries
   auto port_ids = get_port_ids();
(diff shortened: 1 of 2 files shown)


src/op/sirius_physical_sort_sample.cpp
```diff
@@ -82,8 +81,18 @@ std::unique_ptr<operator_data> sirius_physical_sort_sample::execute(const operat
...
+    SIRIUS_LOG_DEBUG("Sort sample: passthrough ({} batches)", input_batches.size());
+    return std::make_unique<pipelineable_operator_data>(input.get_data_batches());
+  }
+
+  // Elect exactly one winner via CAS (0=not started → 1=computing).
+  // The task_creator's while loop dispatches one task per batch, so multiple tasks
+  // can be in-flight simultaneously. Losers passthrough without blocking — they don't
+  // need the boundaries themselves; sort_partition accesses them in a later pipeline.
+  int expected = 0;
+  if (!_boundary_state.compare_exchange_strong(expected, 1, std::memory_order_acq_rel)) {
```
mbrobbel, line 98: If `cudf::concatenate`, `sorted_order`, or `gather` throws `mm::out_of_memory` here, `gpu_pipeline_task::compute_task()` will reschedule the task and retry `execute()` from the same operator index. Because this CAS has already moved `_boundary_state` to 1, every retry (and every other in-flight `SORT_SAMPLE` task) takes the loser path and boundary computation is skipped permanently for the rest of the query.

We should reset `_boundary_state` back to 0 before the OOM exception propagates, so the rescheduled task (or another in-flight task) can win the election and reattempt the computation.

9prady9: good catch, looking into it.

wmalpica (approved):

mbrobbel (approved):

wmalpica (approved):

mbrobbel (approved):
Let's get this in.

PR #627, opened 2026-04-13
Added some initial microbenchmarks

joosthooz:
The benchmarks preferably generate their own data (or use test data that already exists in the repo), and in the end it would be nice if we can run them with default parameters every day and perform a parameter sweep once per week in CI.


bwyogatama:
@mike-wendt I believe we can close this PR as we have our own benchmark infrastructure now?

PR #632, opened 2026-04-14
doris: Update scripts for multiple physical GPUs

dhruv9vats:
Add `--separate-physical-gpus` to the `run-tpch.sh` script. And suggest the use of tpchgen-rs for data generation.


mbrobbel (approved):

cucascade
mbrobbel, line 1: Is this intentional?

mbrobbel (approved):

PR #633, opened 2026-04-14
doris: Fix std::bad_cast crash on concurrent GPU-cached exchange table scans

dhruv9vats:
When two scan threads race to read the same GPU-cached exchange table
(e.g., a 1-row scalar subquery result like min(ps_supplycost) in TPC-H
Q2), the losing thread returned bare operator_data instead of
pipelineable_operator_data. Downstream, publish_output() and
log_operator_data() used dynamic_cast<pipelineable_operator_data&> on
this, throwing std::bad_cast and killing the query.

Root cause: duckdb_scan_task.cpp:607 returned std::make_unique<operator_data>()
for the thread that lost the compare_exchange_strong race on the exhausted
flag. This type is not a pipelineable_operator_data, so:
  - duckdb_scan_task::publish_output() (line 726) crashes
  - gpu_pipeline_task::log_operator_data() (line 91) crashes

Fix:
  - Return pipelineable_operator_data with an empty batch vector instead of bare operator_data, matching the type contract of publish_output.
  - Harden log_operator_data to use pointer dynamic_cast with early return instead of throwing reference dynamic_cast (defense in depth).

### DIFF
--- a/src/op/scan/duckdb_scan_task.cpp
+++ b/src/op/scan/duckdb_scan_task.cpp
@@ -600,11 +600,13 @@ std::unique_ptr<op::operator_data> duckdb_scan_task::compute_task(rmm::cuda_stre
         // Use exhausted flag as a mutex — first thread wins.
         bool expected = false;
         if (!g_state._op.exhausted.compare_exchange_strong(expected, true)) {
-          // Another thread already claimed this — return empty (not nullptr!
-          // because the scan executor dereferences the return value).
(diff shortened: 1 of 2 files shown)


wmalpica (approved):

dhruv9vats:
fixed in [Checkpoint exchange-only capture path](https://github.com/sirius-db/sirius/commit/cad80ee2acafc5c7e991c8544e0e484219c97ffc)

Issue #188, opened 2026-01-23
[Memory Management] add get_next_downgradable function into data_repositories


wmalpica:
The get_next_downgradable function will provide the downgrader which batch should be downgraded next. 
We will likely want to have different interfaces first (see https://github.com/sirius-db/sirius/issues/187). That way, the order in which batches get provided to the downgrader, can depend on the specifics of different types of data_repositories

PR #637, opened 2026-04-14
Downgrade Executor can downgrade tasks. Task queue refactor.

wmalpica:
**Summary**

Implements #188 

Requires https://github.com/NVIDIA/cuCascade/pull/108

Thread-safe data infrastructure for the Sirius GPU SQL engine: a predicate-inspectable MPSC queue integrated as the task queue, and convertible_data abstractions enabling uniform, failure-safe memory-tier conversion across data batches and queued pipeline tasks. This also adds the ability for the downgrade executor to downgrade data form tasks in the task queue.                         
                                                                                                                                                             
  **New Components**                                                                                                                                             
                                                                                                                                                             
  inspectable_mpsc<T> (src/include/exec/inspectable_mpsc.hpp)                                                                                                
  Header-only thread-safe MPSC queue in sirius::exec namespace. Supports blocking/non-blocking push/pop, lifecycle control (interrupt/reactivate/drain), and predicate-based element inspection (pop_if, get_if, mutable_pop_if, mutable_get_if) with bidirectional search. Backed by std::deque<std::unique_ptr<T>>  with mutex + condition_variable.                
                                                                                                                                                             
  convertible_data interfaces (src/include/data/convertible_data.hpp)
  Abstract interfaces for uniform memory-tier conversion. convertible_data declares convert() and bytes_in_space(); convertible_data_provider declares  get_next_convertible(), get_all_convertible(), and get_bytes_in_space().                                                                                   
   
  convertible_data_batch (src/include/data/convertible_data_batch.hpp)                                                                                       
  Wraps cucascade::data_batch with failure-safe GPU/HOST conversion. Locks for in_transit, requests non-blocking reservations, converts via the converter  registry, and restores original batch state on any failure path. Includes convertible_data_batch_provider for repository-based batch discovery filtered by memory space and idle state.                    
                                                                                                                                                             
  convertible_gpu_pipeline_task (src/include/data/convertible_gpu_pipeline_task.hpp)                                                                         
  Wraps gpu_pipeline_task with RAII queue-return semantics — destructor always pushes the task back to the inspectable_mpsc queue. Includes a provider that uses mutable_pop_if for predicate-based task discovery by memory space.                                                                                    
                                                  
  **Removed Components**                                                                                                                                         
                                                  
  - downgrade_task (src/downgrade/downgrade_task.cpp, src/include/downgrade/downgrade_task.hpp) — replaced by convertible_data::convert()                    
  - gpu_pipeline_queue (src/include/pipeline/gpu_pipeline_queue.hpp, src/pipeline/gpu_pipeline_queue.cpp) - It was dead code
  - pipeline_queue (src/include/pipeline/pipeline_queue.hpp, src/pipeline/pipeline_queue.cpp)  - It was dead code                                                                
  - duckdb_scan_task_queue (src/include/op/scan/duckdb_scan_task_queue.hpp)  - It was dead code
  - itask_queue (src/include/parallel/task_queue.hpp)  - It was dead code                                                                                                        
                                                                                                                                                             
  **Changed Components**                                                                                                                                         
                                                                                                                                                             
  itask_executor — Internal queue type changed from interruptible_mpmc to inspectable_mpsc<itask>. Added get_task_queue_ptr() accessor.                      
   
  pipeline_executor — Task queue changed from interruptible_mpmc to inspectable_mpsc<itask>. Added get_pipeline_task_queue() accessors for downgrade executor wiring.        
                                                                                                                                                             
  downgrade_executor — Removed target_bytes from downgrade_request and request_downgrade API. Rewrote processing_loop with tiered provider-based candidate   
  fetching (data repositories ->  pipeline executor task queue). All conversion now flows through convertible_data::convert().
  Added predicate-driven completion checking and per-tier breakdown logging. Added set_task_queues() for deferred wiring.                                    
                                                  
  lock_or_prepare_batch (batch_lock_utils.hpp) — Refactored memory-space mismatch handling to perform direct convert_to<>() calls with explicit in_transit   
  locking, avoiding the convertible_data_batch::convert() path which uses non-blocking reservations inappropriate for the forward execution path.
                                                                                                                                                             
  SiriusContext::initialize — Wires task queue pointers from pipeline_executor and gpu_pipeline_executor into downgrade_executor after construction.         
   
  **Test Coverage**                                                                                                                                              
                                                  
  54 new tests (35 queue + 19 GPU integration) covering single/multi-threaded queue correctness, predicate search, GPU-to-HOST conversion, failure safety,   
  RAII ownership, and provider iteration. Full suite: 868+ tests passing.


src/downgrade/downgrade_executor.cpp
```diff
@@ -112,74 +126,198 @@ void downgrade_executor::processing_loop()
...
+    }
 
-    // 3. Incremental dispatch with predicate checking
-    for (auto& weak_batch : candidates) {
+    // === TIER 1: Data repositories ===
+    // Collect all repositories first so we can iterate (and later sort by priority).
+    // We use get_all_convertible() to snapshot eligible batches once per repo,
+    // avoiding re-scanning the same batch before its state changes from idle.
+    bool pool_interrupted = false;
+    auto repos            = _data_repo_mgr.get_repositories();
```
bwyogatama, line 170: So I think in our previous state of downgrade executor, we set priority for data repositories based on size and active partitions. It will prioritize downgrading starting from data repositories with the largest size and will prioritize downgrading non-active partition. I don't see this exist anymore in the current PR, maybe I'm wrong?

CMakeLists.txt
```diff
@@ -352,11 +349,14 @@ endif()
 set(TEST_SOURCES
     test/cpp/config/test_config.cpp
     test/cpp/config/test_context.cpp
+    test/cpp/data/test_convertible_data_batch.cpp
+    test/cpp/data/test_convertible_gpu_pipeline_task.cpp
```
bwyogatama, line 357: test/cpp/data/test_convertible_data.cpp is not added to the CMakeLists.txt

src/include/exec/inspectable_mpsc.hpp
```diff
@@ -0,0 +1,293 @@
...
+   * \param predicate Callable receiving T& (mutable) and returning bool.
+   * \param front_to_back If true, searches oldest-to-newest; if false,
+   *        newest-to-oldest.
+   * \return The matching element, or nullptr if no match found.
+   *
+   * Same as pop_if but the predicate receives a mutable reference,
+   * allowing state inspection that requires non-const access. Holds
+   * the mutex for the full scan duration.
+   */
+  std::unique_ptr<T> mutable_pop_if(std::function<bool(T&)> predicate, bool front_to_back)
```
bwyogatama, line 213: I don't fully understand why we need the mutable variant?

src/include/exec/inspectable_mpsc.hpp
```diff
@@ -0,0 +1,293 @@
...
+   * \param front_to_back If true, searches oldest-to-newest; if false,
+   *        newest-to-oldest.
+   * \return Raw pointer to the matching element, or nullptr if no match.
+   *
+   * Same as get_if but the predicate receives a mutable reference.
+   * Returned pointer is invalidated by any subsequent pop(), pop_if(),
+   * drain(), or other mutating queue operation. Safe under MPSC
+   * single-consumer.
+   */
+  T* mutable_get_if(std::function<bool(T&)> predicate, bool front_to_back)
```
bwyogatama: I don't think the `get_if` and `mutable_get_if` ever used anywhere?

src/include/data/convertible_gpu_pipeline_task.hpp
```diff
@@ -0,0 +1,335 @@
...
+   * @brief Get the total byte size of task data in the given memory space.
+   *
+   * Returns 0 because precise byte counting would require temporarily removing
+   * and re-inserting tasks, which is unsafe under concurrent producers. Callers
+   * needing exact totals should use get_all_convertible() + bytes_in_space().
+   *
+   * @param space The memory space to query.
+   * @return Always 0 (see rationale above).
+   */
+  std::size_t get_bytes_in_space(cucascade::memory::memory_space* /*space*/) const override
```
bwyogatama: So right now this might be misleading cause someone might think there is no data in the queue where it's actually something that we never suppose to call

src/downgrade/downgrade_executor.cpp
```diff
@@ -365,16 +383,13 @@ size_t downgrade_executor::request_free_memory_and_wait(size_t bytes)
...
-  req->predicate    = std::move(predicate);
-  auto future       = req->result.get_future();
+  auto req       = std::make_unique<downgrade_request>();
+  req->predicate = std::move(predicate);
+  auto future    = req->result.get_future();
   if (!_request_queue.push(std::move(req))) {
-    SIRIUS_LOG_WARN("[downgrade] request_downgrade: queue inactive, dropping request for {} bytes",
-                    target_bytes);
+    SIRIUS_LOG_WARN("[downgrade] request_downgrade: queue inactive, dropping request");
     req->result.set_value(0);
```
bwyogatama: Is this setting req after it is `moved` in the condition?

src/include/data/convertible_gpu_pipeline_task.hpp
```diff
@@ -0,0 +1,335 @@
...
+
+    const auto& batches = pipelineable->get_data_batches();
+    std::vector<std::size_t> totals(target_spaces.size(), 0);
+    bool any_converted = false;
+
+    for (const auto& batch : batches) {
+      if (!batch) { continue; }
+
+      // Skip batches already at a target space — no conversion needed
+      auto* batch_space      = batch->get_memory_space();
```
bwyogatama, line 126: I think `get_memory_space()` is not thread safe so there is a slight possibility that when you are calling this, there is another thread downgrading the batch at the same time, causing race condition.

This will probably be solved by the data batch lock refactor though

bwyogatama (approved):

src/downgrade/downgrade_executor.cpp
wmalpica: You are totally right. This is something i want to add in a followup PR. The plan for that is that I will get a prioritized list of data_repositories. The order of those repositories will be based on the order of execution of the operators in the graph, which I think is a much better way of prioritizing. 

Also, some of the things we were doing before, either are being implicitly being done or dont really mean anything.
For example, for data_repositories with multiple partitions, we always go from last to first partition, which is the opposite of how we process them in the engine.  So in a way, that accounts for the "Active" partitions component.  
Also he concept of an active partition before was "if a partition has a task that has been created". Now, we dont pull batches from data_repositories if a task has been created. We ignore those. We only "pull" those from data in the task queue.

src/include/exec/inspectable_mpsc.hpp
wmalpica: I wanted to encourage us to always use the non-mutable version. BUT depending on the predicate, there might be things you need to do there that could not be possible if we limit it to a const function. So I wanted that to be explicit in the name of the function.

src/include/exec/inspectable_mpsc.hpp
wmalpica: I thought i was going to use them. But i did not. I am removing them

src/include/data/convertible_gpu_pipeline_task.hpp
wmalpica: Good catch

src/downgrade/downgrade_executor.cpp
wmalpica: Good catch. Thanks. I swear i thought I had fixed this already....

src/include/data/convertible_gpu_pipeline_task.hpp
wmalpica: I agree, this is a similar issue to what we have in the lock and prepare batch for processing function

mbrobbel:
Shouldn't we merge the cucascade PR first?

PR #644, opened 2026-04-15
Use static cudart in vcpkg extension build

mbrobbel:
## Summary

- Statically link the CUDA runtime (`cudart_static`) instead of dynamically linking
  `libcudart.so` in the vcpkg extension build, so the loadable extension has no runtime
  dependency on a system CUDA installation.
- Set `CMAKE_CUDA_RUNTIME_LIBRARY=Static` globally for CUDA targets, pass it through to
  the RMM vcpkg port, and patch RMM's exported CMake targets to reference
  `CUDA::cudart_static` instead of `CUDA::cudart`.
- Add a `sirius_prefer_static_cudart()` CMake function that rewrites cucascade target
  link libraries from shared to static cudart after `add_subdirectory`.

## Test plan

- [x] Verify vcpkg build succeeds: `CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) make`
- [x] Confirm the loadable extension has no `libcudart.so` dependency:
      `ldd build/release/extension/sirius/sirius_loadable.duckdb_extension | grep cudart`
- [x] Run SQL logic tests: `make test`

🤖 Created by Claude

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,11 @@ if(VCPKG_BUILD)
   string(REGEX REPLACE "-I[^ ]*/include/cccl" "" CMAKE_CUDA_FLAGS
                        "${CMAKE_CUDA_FLAGS}")
 
+  # The loadable extension must not depend on libcudart.so in the vcpkg build.
+  # Set the CUDA language default to the static runtime before any CUDA targets
+  # are created.
+  set(CMAKE_CUDA_RUNTIME_LIBRARY Static)
+
   # Strip conda-injected -Wl,-rpath flags from linker flags and disable CMake's
   # RPATH mechanism so binaries don't carry a hardcoded path to the pixi/conda
   # env. All deps are statically linked, so no RPATH is needed. FORCE is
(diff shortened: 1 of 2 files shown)


vcpkg_ports/rmm/portfile.cmake
```diff
@@ -95,6 +96,16 @@ execute_process(
...
 
+# RMM still exports shared CUDA::cudart in its installed target file even when
+# built as a static library. That pulls libcudart.so into the final Sirius
+# loadable extension alongside libcudart_static.a. Rewrite the exported target
+# to prefer the static CUDA runtime for vcpkg consumers.
+file(READ "${CURRENT_PACKAGES_DIR}/share/rmm/rmm-targets.cmake" _rmm_targets)
+string(REGEX REPLACE "CUDA::cudart([;\">])" "CUDA::cudart_static\\1"
+                     _rmm_targets "${_rmm_targets}")
+file(WRITE "${CURRENT_PACKAGES_DIR}/share/rmm/rmm-targets.cmake"
+     "${_rmm_targets}")
```
9prady9, line 107: We should upstream a change to rmm's vcpkg port to export static version to avoid this patch eventually on our end.

CMakeLists.txt
```diff
@@ -91,6 +96,37 @@ set(WARNINGS_AS_ERRORS
...
+      get_target_property(_libs "${target_name}" "${_prop}")
+      if(NOT _libs OR _libs STREQUAL "_libs-NOTFOUND")
+        continue()
+      endif()
+
+      set(_patched_libs "${_libs}")
+      list(TRANSFORM _patched_libs REPLACE "^CUDA::cudart$"
+                                           "CUDA::cudart_static")
+      list(TRANSFORM _patched_libs REPLACE "^\\$<LINK_ONLY:CUDA::cudart>$"
+                                           "$<LINK_ONLY:CUDA::cudart_static>")
```
9prady9, line 111: We should create a ticket on our end to submit vcpkg features that lets us target the static versions of our dependencies instead of doing this regex replace on our end.

9prady9 (approved):
This solutions is good enough for now but we should upstream our requirements (generic) as vcpkg port features in respective upstreams. Lets create a ticket for that.

CMakeLists.txt
mbrobbel: I tried adding this to cucascade (https://github.com/NVIDIA/cuCascade/pull/105), but it doesn't really make sense, because you'd have to link everything statically. Normally with cucascade you'll have (cudf and) rmm which link to cudart dynamically.

vcpkg_ports/rmm/portfile.cmake
mbrobbel: There is no upstream rmm vcpkg port.

9prady9: wow, then we should add it :D I will add it to my bucket list

CMakeLists.txt
9prady9: Yeah, we need compatible ports across the broad.

PR #648, opened 2026-04-16
Quent Integration, Act 1: The Query Plan

dhruv9vats:
## Summary

Act one of integrating [Quent](https://github.com/rapidsai/quent), a Rust-based query-engine instrumentation framework, into Sirius. This act wires up plan-level telemetry — engine/worker bring-up, query lifecycle, and one-shot plan emission (operators, ports, edges). Per-task and per-batch events are scoped for subsequent acts.

Default behaviour is unchanged: telemetry is currently opt-in via a new config flag and uses a no-op exporter by default.

### Rust workspace
A new top-level `rust/` Cargo workspace hosts two crates:

| Crate | Purpose |
|---|---|
| `telemetry-model` 	| Declares the Sirius instrumentation model. |
| `telemetry-bridge` 	| `cxx`-based FFI bridge. Generates C++ headers (`telemetry-bridge/gen/*.rs.h`) and compiles a `staticlib` that links into the extension. |

CMake wires the bridge into the build (`rust/crates/telemetry/bridge/CMakeLists.txt`).

### New SQL surface

| SQL | Effect |
|---|---|
| `SET enable_quent = true;` | Switches the exporter from `noop` to `ndjson`. **Default: `false`.** |
| `SET quent_output_directory = '<path>';` | Where `ndjson` files are written. **Default: `telemetry_data`.** |
| `SET quent_engine_name = '<name>';` | Engine instance name reported in telemetry. **Default: `siriusDB`.** |
| `CALL sirius_set_query_label('<label>')` | Stashes a label on `SiriusContext`; the **next** Sirius query (transparent or `gpu_execution`) consumes it. |
| `CALL gpu_execution('…', query_label => '<label>')` | New named parameter; takes precedence over the stash. |

### Script changes

- Adds a new script `test/tpch_performance/run_tpch_parquet_and_generate_telemetry.sh` to run the tpch queries with telemetry enabled. Added as a separate script to not collude the existing benchmarking scripts.

## Behaviour with `enable_quent = false` (default)

- A `telemetry_context` is still constructed per `sirius_interface`, but with the `noop` exporter, so no files are written.
- All observer calls and plan emission still happen, but route to the noop exporter.
- No new dependencies on Quent's runtime exporters at the SQL layer — the user opts in.

## Behaviour with `enable_quent = true`

- Each `sirius_interface` writes one or more `*.ndjson` files into `quent_output_directory`.
- `query.instance_name` is the user-supplied label, or `"unnamed_query"`.


rust/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+    "crates/telemetry/model"
+]
+
+[workspace.package]
+version = "0.1.0"
+edition = "2024"
+
+[workspace.dependencies]
+serde = { version = "1", features = ["derive"] }
+uuid = { version = "1", features = ["v7", "serde"] }
```
mbrobbel: Those are re-exported by `quent_model`
```suggestion
```

rust/crates/telemetry/model/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version.workspace = true
+edition.workspace = true
+
+[dependencies]
+quent-attributes = { git = "https://github.com/rapidsai/quent.git", branch = "main" }
+quent-model = { git = "https://github.com/rapidsai/quent.git", branch = "main" }
+quent-query-engine-model = { git = "https://github.com/rapidsai/quent.git", branch = "main" }
+quent-stdlib = { git = "https://github.com/rapidsai/quent.git", branch = "main" }
+serde.workspace = true
+uuid = { workspace = true, features = ["serde", "v7"] }
```
mbrobbel:
```suggestion
```

rust/crates/telemetry/model/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "instrumentation-model"
+version.workspace = true
+edition.workspace = true
+
+[dependencies]
+quent-attributes = { git = "https://github.com/rapidsai/quent.git", branch = "main" }
+quent-model = { git = "https://github.com/rapidsai/quent.git", branch = "main" }
+quent-query-engine-model = { git = "https://github.com/rapidsai/quent.git", branch = "main" }
+quent-stdlib = { git = "https://github.com/rapidsai/quent.git", branch = "main" }
```
mbrobbel: I guess we should pin with `rev`?

rust/crates/telemetry/bridge/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "telemetry-bridge"
+version.workspace = true
+edition.workspace = true
+
+[lib]
+name = "telemetry_bridge"
+crate-type = ["staticlib"]
+
+[dependencies]
+cxx = "1"
```
mbrobbel, line 11: This (and `cxx-build`) deps should be re-exported by `quent_codegen`.

src/include/planner/query.hpp
```diff
@@ -35,8 +40,10 @@ class query {
    * @brief Construct a new query object.
    *
    * @param pipelines The ordered pipelines required to execute this query.
+   * @param engine The engine executing this query (provides telemetry context).
    */
-  explicit query(duckdb::vector<duckdb::shared_ptr<pipeline::sirius_pipeline>> pipelines);
+  query(duckdb::vector<duckdb::shared_ptr<pipeline::sirius_pipeline>> pipelines,
+        sirius_engine& engine);
```
mbrobbel: Why does this get `engine` here?

pixi.toml
```diff
@@ -26,6 +26,11 @@ pre-commit = "*"
 clang-tools = ">=21.1.8,<22"
 pkg-config = "*"
 mold = "*"
+libprotobuf = "*"
+rust = "1.94.*"
+rust-src = "1.94.*"
```
mbrobbel: Why do we need `rust-src`?

pixi.toml
```diff
@@ -26,6 +26,11 @@ pre-commit = "*"
...
+nodejs = ">=24"
+pnpm = "*"
```
mbrobbel: Why do we need those here?

pixi.toml
```diff
@@ -26,6 +26,11 @@ pre-commit = "*"
 clang-tools = ">=21.1.8,<22"
 pkg-config = "*"
 mold = "*"
+libprotobuf = "*"
```
mbrobbel, line 30: Is this for `protoc`?

mbrobbel (commented):
I know this is still in draft, but just some questions:

dhruv9vats:
I think it would make sense to use explicit context propagation. Such that multi-query sessions are naturally supported: https://github.com/rapidsai/quent/pull/117

pixi.toml
dhruv9vats: yes.

pixi.toml
```diff
@@ -36,6 +38,7 @@ libabseil = "*"
 duckdb = "=1.5.2"
 spdlog = "1.8.*"
 sqlite = "3.*"
+rust-src = "*"
```
dhruv9vats, line 41: this is needed for static analyzers to work (in IDEs)

bwyogatama (approved):

rust/crates/telemetry/model/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+        root: quent_query_engine_model::engine::Engine,
+        quent_query_engine_model::worker::Worker,
+        quent_query_engine_model::query_group::QueryGroup,
+        quent_query_engine_model::query::Query,
+        quent_query_engine_model::plan::Plan,
+        quent_query_engine_model::operator::Operator,
+        quent_query_engine_model::port::Port,
+        quent_stdlib::memory::Memory,
+        quent_stdlib::processor::Processor,
+        quent_stdlib::channel::Channel,
```
johanpel: I think these are unused

johanpel (approved):
Just one nit:

PR #651, opened 2026-04-16
Feature/gpu native scan task

ducndh:
# feat: GPU native scan task — GPU-side decode of `.duckdb` block storage

## :warning: DuckDB submodule change note (read first)

This branch **requires the GetRowGroupCollectionRef`,
`GetRowGroupsDirect`, `GetColumnDirect`, `GetCompressionType`,
`GetValidityData`)** that should be in the Duckdb submodule that is already in this PR, so you can just build from here and test things out.

This PR doesn't ask them for anything beyond the accessors they've already seen but is in private only

See the "mmap bypass: pure Sirius" section below for the design.

---

## Summary

Replaces DuckDB's CPU `duckdb_scan_task` table function with a Sirius-native
scan for block-backed `.duckdb` storage. The extension now decodes
DuckDB's on-disk compressed blocks (bitpacking, dictionary, FSST, RLE,
constant, uncompressed) directly on the GPU, parsing block metadata on the
device and skipping the buffer-manager Pin path for read-only databases.

**Results (warm, this branch vs main):**

| Workload | Main | This PR | Δ |
|---|---|---|---|
| TPC-H SF=10 GPU (RTX 6000) | ~10.5s | **5.37s** | −49% |
| TPC-H SF=100 GPU (GH200) | 8.4s | **5.33s** | −37%; **1.33× vs CPU** |
| ClickBench 10M GPU (RTX 6000) | 3.37s | **2.10s** | −38% |
| ClickBench 100-shard GPU (GH200) | OOM / 29.3s | **5.00s** | −83% |

GPU wins on 15/22 TPC-H queries at SF=100 on GH200.

---

### 1. `gpu_native_scan_task` (`src/op/scan/`, `src/creator/`)

Replaces the DuckDB table-function scan path for `.duckdb` block storage.
Walks `RowGroupCollection → RowGroup → ColumnData → ColumnSegment`, classifies
each segment by compression type, and dispatches to GPU decode kernels.
Keeps DuckDB's table-function path for parquet / non-block sources (the
fallback is in `src/fallback.cpp`, unchanged).

### 2. GPU decode kernels (`src/cuda/scan/`)

Per compression type:
- **Bitpacking** (int8 / int16 / int32 / int64) with fused GPU-side metadata
  parse; single kernel per batch of segments instead of one per segment
- **Dictionary** (including string dicts); vectorized gather
- **FSST**: three-phase chunked decode (offsets, prefix-scan, byte-emit)
  with per-phase kernel fusion
- **RLE**, **Constant**, **Uncompressed** (int + string)

Net decode time for TPC-H SF=100 warm Q1 is 7ms (1.1% of the 630ms wall
clock). We are no longer the bottleneck.

### 3. Parallel scan task scheduling

Scan launches N concurrent tasks (one per hardware thread up to the cudf
decode limit), each processing disjoint row-group segments. This saturates
the extension's scan thread pool and is the single biggest win at SF=10.

### 4. Batched two-pass string decode

Eliminates per-segment sync + kernel launch in the string decode path by
upper-bounding char allocation in pass 1 (device-side), then running a
single gather in pass 2.

### 5. mmap bypass: pure Sirius

Read-only `.duckdb` files get mmap'd and the extension indexes block data
directly, bypassing `BufferManager::Pin()` (which was 470ms / 74.6% of Q1
SF=100 wall). **Pure extension code — no DuckDB patch needed.**

Design:

```cpp
// single-slot atomic hot path (no mutex on the common case)
static std::atomic<mmap_file_state*> g_mmap_last{nullptr};

// keyed by BlockManager*; init under mutex, then published via atomic store
static std::mutex g_mmap_mutex;
static std::unordered_map<BlockManager*, unique_ptr<mmap_file_state>> g_mmap_cache;

try_get_mmap_for_table(storage):
  bm = &storage.GetAttached().GetStorageManager().GetBlockManager();
  last = g_mmap_last.load(acquire);
  if (last && last->owner == bm) return last;   // hot path, no allocation
  // slow path: lock → find or insert → mmap file → store->g_mmap_last
```

Block offset math: `BLOCK_START = 4096*3 + block_id * alloc_size + header`.
This is validated once per file via `std::call_once` against a `Pin()`'d
reference block — if DuckDB's layout changes, we detect and fall back
cleanly. `alloc_size` and `header_size` are read from
`ColumnSegment::block->GetBlockAllocSize()` / `GetBlockHeaderSize()` so they
track upstream changes automatically.

Gated on:
- `storage_manager.options.read_only == true`
- `!encryption_enabled`
- `.duckdb` single-file block manager (not parquet / other extensions)

Any mismatch falls back to `Pin()`.

### 6. Unified-memory auto-detect (GH200)

On GH200 with ATS, prefaulting the mmap with `MADV_POPULATE_READ` hurts
perf (cold: 16.4s → 2.4s avg per TPC-H SF100 query when we skip prefault).
`direct_block_scan.cpp` now auto-detects unified memory and skips prefault.
PCIe path (RTX 6000) already benefits from lazy faults — prefault was
removed in commit `938898f`.

---

## Rejected approaches

| Tried | Why reverted |
|---|---|
| H2D / compute interleaving via dual CUDA streams + ring buffer | Net regression at SF=10 on RTX 6000 (commit `e3e2719`). Scan bandwidth is saturated already; the extra stream synchronization cost more than the overlap gained. On GH200 the scan is already too fast that the decode now dominates but you can revisit if you have better design |
| `cudaMemcpyBatchAsync` for block transfers | ~20% slower than per-segment async copies. Observed 44µs inter-op gaps vs 1µs for naive per-segment path. Also I choose to batch it myself already. |
| `MADV_POPULATE_READ` prefault (everywhere) | Hurts GH200 cold by 6.9× (16.4s → 2.4s/query). Now gated to PCIe-only. |
| DuckDB `GetDirectBlockPointer()` submodule patch | Works, but we need buy in from DuckDB — pure-Sirius atomic matches perf within 2–3%. However, it would be nice if they provided it, as we won't need to maintain that part. |

---

## Where to pick this up

1. **SF=1000 GH200 run** — I don't have the data; theory says we continue
   to pull ahead of CPU. Block on data generation.
2. **Bigger-than-memory on GH200** — confident via ATS theory, no empirical
   run. Block on cache-flush methodology.

---

## Test plan

- [x] `make test` — all SQLLogicTests pass
- [x] `sirius_unittest` — unit tests pass
- [x] TPC-H SF=10 correctness: all 22 queries return DuckDB-CPU-identical results
- [x] TPC-H SF=100 correctness (GH200): all 22 queries return CPU-identical results
- [x] ClickBench 10M correctness: all 29 non-skipped queries match CPU
- [x] ClickBench 100-shard correctness (GH200): 16 tested queries match CPU
- [x] Works with `read_only=false` (falls back to `Pin()`)
- [x] Works with encrypted DB (falls back to `Pin()`)
- [x] Works with parquet sources (unchanged fallback path)


src/op/sirius_physical_hash_join.cpp
```diff
@@ -668,9 +668,38 @@ static std::unique_ptr<operator_data> gather_join_output(
...
 
+  // Validate string column offsets before gather — corrupt offsets cause cudf
+  // to compute negative buffer sizes that wrap to huge allocations.
+  auto validate_string_cols = [&stream](cudf::table_view const& tv, char const* label) {
+    stream.synchronize();
+    for (cudf::size_type c = 0; c < tv.num_columns(); c++) {
+      auto col = tv.column(c);
+      if (col.type().id() != cudf::type_id::STRING || col.size() == 0) continue;
+      auto offsets_col = col.child(0);
+      // Copy first and last offset to host to validate
```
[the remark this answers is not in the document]
ducndh: Great catch. Forgot to clear the diagnostic code.

src/op/sirius_physical_hash_join.cpp
```diff
@@ -668,9 +668,38 @@ static std::unique_ptr<operator_data> gather_join_output(
...
+      std::vector<int32_t> host_offsets(offsets_col.size());
+      cudaMemcpy(host_offsets.data(), offsets_col.head<int32_t>(),
+                 offsets_col.size() * sizeof(int32_t), cudaMemcpyDeviceToHost);
+      auto first = host_offsets.front();
+      auto last  = host_offsets.back();
+      auto total_chars = last - first;
+      bool monotonic = true;
+      for (size_t i = 1; i < host_offsets.size(); i++) {
+        if (host_offsets[i] < host_offsets[i-1]) { monotonic = false; break; }
+      }
```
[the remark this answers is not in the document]
ducndh: Great catch. Forgot to clear the diagnostic code.

src/include/cuda/scan/gpu_decode_batched_string.cuh
```diff
@@ -0,0 +1,44 @@
...
+/// @param device_staging  Optional: device staging buffer (from pipelined path).
+/// @param d_valid_count_out  If non-null, write valid count here (async, no sync).
+/// @return cudf string column
+std::unique_ptr<cudf::column> decode_string_column_batched(
+    sirius::op::scan::column_scan_result& col_scan,
+    rmm::cuda_stream_view stream,
+    rmm::device_async_resource_ref mr,
+    const std::unordered_map<int64_t, size_t>* device_blocks = nullptr,
+    uint8_t* device_staging = nullptr,
+    uint32_t* d_valid_count_out = nullptr);
```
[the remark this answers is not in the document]
ducndh: Did not think carefully about this but good point

src/include/cuda/scan/gpu_native_decode.cuh
```diff
@@ -0,0 +1,55 @@
...
+  std::unordered_map<int64_t, size_t> offsets;
+  size_t total_bytes = 0;
+};
+
+/// @brief Decode all columns from direct block scan results into a cudf::table on GPU.
+///
+/// Hybrid routing: each segment is decoded via GPU kernel (bitpacking, dictionary,
+/// uncompressed, constant) or falls back to CPU decode + H2D memcpy for unsupported
+/// compression types (RLE, FSST, ALP, etc.).
+///
```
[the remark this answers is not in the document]
ducndh: True

src/op/scan/direct_block_scan.cpp
```diff
@@ -0,0 +1,340 @@
...
+struct block_layout_info {
+  size_t alloc_size  = 0;
+  size_t hdr_size    = 0;
+  size_t block_start = 0;
+  bool   valid       = false;
+};
+
+static std::once_flag    g_layout_once;
+static block_layout_info g_layout;
+
```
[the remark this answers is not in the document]
ducndh: Might be the case but add guard just to be sure

joosthooz:
Hi duc, I just finished a measurement at SF200 on a B300 (our GB300 is being re-installed). This runs from a rather slow network disk.
```
Query   |   DuckDB Cold |   DuckDB Warm |   Sirius Cold |   Sirius Warm | Speedup (warm)
--------+---------------+---------------+---------------+---------------+---------------
Q1      |       259.55s |         0.54s |       272.25s |         0.60s |           .90x
Q2      |        28.90s |         0.25s |        30.50s |         0.25s |           .96x
Q3      |       117.04s |         0.66s |       134.42s |         0.71s |           .92x
Q4      |       139.11s |         0.59s |       145.90s |         0.61s |           .96x
Q5      |        46.48s |         0.63s |        45.91s |         0.68s |           .93x
Q6      |         0.18s |         0.17s |         0.20s |         0.17s |           .95x
Q7      |         0.72s |         0.70s |         0.73s |         0.72s |           .97x
Q8      |        78.68s |         0.55s |        83.24s |         0.59s |           .92x
Q9      |        13.32s |         2.19s |        14.16s |         2.33s |           .94x
Q10     |        43.53s |         1.04s |        46.76s |         1.06s |           .97x
Q11     |         6.79s |         0.09s |         7.85s |         0.09s |          1.01x
Q12     |         0.52s |         0.51s |         0.78s |         0.51s |          1.00x
Q13     |       105.11s |         1.97s |       109.29s |         1.96s |          1.00x
Q14     |         0.46s |         0.42s |         0.50s |         0.47s |           .89x
Q15     |         0.33s |         0.32s |         0.37s |         0.34s |           .91x
Q16     |         0.44s |         0.34s |         0.43s |         0.35s |           .97x
Q17     |         0.73s |         0.69s |         0.76s |         0.67s |          1.02x
Q18     |        16.70s |         1.68s |        17.33s |         1.72s |           .97x
Q19     |         0.81s |         0.82s |         0.86s |         0.81s |          1.01x
Q20     |         0.55s |         0.55s |         0.61s |         0.55s |           .99x
Q21     |         2.95s |         1.99s |         2.93s |         2.04s |           .97x
Q22     |         0.41s |         0.42s |         0.43s |         0.42s |           .99x
--------+---------------+---------------+---------------+---------------+---------------
TOTAL   |       863.32s |        17.11s |       916.22s |        17.66s |           .96x
```
I'll try running it from parquet to have another comparison point.

docs/super-sirius/benchmarks/2026-04-16_gh200/nsys/sirius_kernels.txt
```diff
@@ -0,0 +1,7 @@
+kernel                                                                  total_ms      calls       avg_us
```
joosthooz, line 1: All these benchmark files, they are interesting for the review but need to be removed before merging. If you think it is worth the time, consider adding a script to generate these so we can always (re)run them later!

docs/duckdb-pin-overhead-analysis.md
```diff
@@ -0,0 +1,151 @@
+# DuckDB BufferManager::Pin() Overhead Analysis
```
joosthooz, line 1: Same for these .md files; remove them when making the PR ready for merging

patches/0001-feat-add-GetDirectBlockPointer-for-zero-copy-block-a.patch
```diff
@@ -0,0 +1,184 @@
+From 8b2abc77c170045d8e2ebb3ec5e4cbc0d184d77d Mon Sep 17 00:00:00 2001
```
joosthooz, line 1: This file is a patch to duckdb, I think? We need to know what additional functionality we can rely on before we can merge code that uses it. I don't think we want to add a duckdb patching step into our build process. 
Additionally, we have to consider that Sirius can also be loaded as an extension. In that case, we have no control over how the duckdb that is loading us was built.
So this is something we have to talk about.

src/cuda/scan/gpu_decode_batched_string.cu
```diff
@@ -0,0 +1,1209 @@
+/*
```
joosthooz, line 1: General remark about the .cu files: I think all the .cu files that are currently in the `src/cuda` directory are only used by legacy Sirius. So we should move them. But besides that, I do think this is the proper location for these files so just keep them here 👍

src/cuda/scan/gpu_decode_batched_string.cu
```diff
@@ -0,0 +1,1209 @@
...
+  __shared__ uint32_t sh_dict_end;
+  __shared__ uint8_t sh_len[255];
+
+  if (threadIdx.x == 0) {
+    fsst_header_t hdr;
+    memcpy(&hdr, base, sizeof(hdr));
+    sh_bp_width = hdr.bitpacking_width;
+    sh_dict_end = hdr.dict_end;
+  }
+  __syncthreads();
```
joosthooz, line 248: This code produces some compiler warnings and Claude says it is a remnant of a refactor and can be removed. Can you look into it?

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,563 @@
+/*
```
joosthooz, line 1: I would like it if we have a tiny benchmark for these decoder kernels. But it's also ok to leave that for a follow-up.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,563 @@
...
+#include <cuda_runtime.h>
+
+#include <chrono>
+#include <cstring>
+#include <vector>
+
+namespace sirius::cuda::scan {
+
+//===----------------------------------------------------------------------===//
+// Fused parse-and-decode kernel — metadata parsed on GPU, zero H2D for metadata
```
joosthooz, line 21: It seems there is a fused and a batched implementation, is it correct that the fused version is not used at all? If that is the case we should consider removing it, we shouldn't keep unused code in the repo, or at least keep it behind a compile-time config or a run-time parameter if it has value in certain scenarios

src/cuda/scan/gpu_decode_fsst.cu
```diff
@@ -0,0 +1,334 @@
...
+  if (threadIdx.x == 0 && blockIdx.x == 0) {
+    d_sentinel_slot[0] = d_prefix_offsets[row_count - 1] + d_lengths[row_count - 1];
+  }
+}
+
+//===----------------------------------------------------------------------===//
+// Host-side API
+//===----------------------------------------------------------------------===//
+
+void gpu_decode_fsst(
```
joosthooz, line 190: I can't find any usages of this API, how is it hooked up?

src/cuda/scan/gpu_decode_dictionary.cu
```diff
@@ -0,0 +1,363 @@
...
+
+  // Copy string bytes
+  for (uint32_t b = 0; b < str_len; ++b) {
+    d_chars[out_offset + b] = str_ptr[b];
+  }
+}
+
+//===----------------------------------------------------------------------===//
+// Host-side API
+//===----------------------------------------------------------------------===//
```
joosthooz, line 79: Move a little bit lower, to just above the function

src/cuda/scan/gpu_decode_dictionary.cu
```diff
@@ -0,0 +1,363 @@
...
+    const uint32_t* __restrict__ d_lengths,
+    uint32_t* __restrict__ d_sentinel_slot,          // &d_offsets[row_count] as uint32
+    uint32_t row_count)
+{
+  if (threadIdx.x == 0 && blockIdx.x == 0) {
+    d_sentinel_slot[0] = d_prefix_offsets[row_count - 1] + d_lengths[row_count - 1];
+  }
+}
+
+void gpu_decode_dictionary(
```
joosthooz, line 94: I can't find any usages of this

src/cuda/scan/gpu_decode_batched_string.cu
```diff
@@ -0,0 +1,1209 @@
...
+  unsigned char zeroTerminated;
+  unsigned char len[255];
+  unsigned long long symbol[255];
+};
+
+/// Compact FSST decoder for GPU upload — just len + symbol, no version header.
+struct fsst_decoder_compact {
+  uint8_t len[255];
+  unsigned long long symbol[255];
+};
```
joosthooz, line 175: Would it make more sense to have these in gpu_decode_fsst.cu?

src/cuda/scan/gpu_decode_fsst.cu
```diff
@@ -0,0 +1,334 @@
+/*
```
joosthooz, line 1: It looks like this file is not being used and gpu_decode_batched_string.cu contains a newer version of this code. Can you either delete this file or (maybe nicer) break out the fsst-related code from that file and move it here? (and of course remove the old code from this file)

src/cuda/scan/gpu_native_decode.cu
```diff
@@ -0,0 +1,677 @@
...
+                       d_block ? const_cast<void*>(static_cast<const void*>(d_block)) : nullptr,
+                       /*skip_block_copy=*/d_block != nullptr,
+                       d_rle_cumsum,
+                       RLE_CUMSUM_CAP);
+        break;
+      }
+      default:
+        throw std::runtime_error("gpu_native_decode: unsupported compression type " +
+                                 std::to_string(static_cast<int>(seg.compression)) +
+                                 " for fixed-width column — falling back to CPU scan");
```
joosthooz, line 504: Are you sure there is a fallback path? If not update this message.

src/include/cuda/scan/gpu_native_decode.cuh
```diff
@@ -0,0 +1,42 @@
...
+
+namespace sirius::cuda::scan {
+
+/// @brief Decode all columns from direct block scan results into a cudf::table on GPU.
+///
+/// Pre-stages every unique block referenced by any column to device memory in one
+/// coalesced bulk H2D (runs of contiguous block_ids share a single cudaMemcpyAsync),
+/// then decodes all segments from device staging.  Supported compression types are
+/// decoded via GPU kernels (bitpacking, dictionary, FSST, RLE, uncompressed,
+/// constant); unsupported segments cause the scan to fall back to duckdb_scan_task.
```
joosthooz, line 26: I can't find a fallback path, can you check if it really exists or update this comment?

src/op/sirius_physical_hash_join.cpp
```diff
@@ -804,6 +805,17 @@ static std::unique_ptr<operator_data> resolve_mark_join_result(
       make_data_batch(std::move(output_cudf_table), *left_batch->get_memory_space())});
 }
 
+std::size_t sirius_physical_hash_join::get_extra_memory_estimate() const
```
joosthooz, line 808: Do we want to add this? My feeling would be to not add these get_extra_memory_estimate and related changes in this PR. If the scan has some special considerations we should override the existing functions

src/op/scan/duckdb_scan_task.cpp
```diff
@@ -32,11 +32,105 @@
 #include <cucascade/memory/memory_reservation.hpp>
 
 // duckdb
+#include <duckdb/catalog/catalog_entry/duck_table_entry.hpp>
```
joosthooz, line 35: It seems this PR also contains a lot of changes to make the current (CPU) duckdb scan paths work again, and improve their performance. Could you create a separate PR for that? Right now this PR is way too large, and due to the dependency on the duckdb patch we can't merge it right now. But these fixes would be really nice to merge soon.

joosthooz (changes requested):
Hi @ducndh I did a first review pass, but the PR is so large that I haven't read all the code yet. But before I go further there's some items I want to discuss, like splitting off the work on fixing the current duckdb scans into a separate PR.

docs/super-sirius/benchmarks/2026-04-16_gh200/nsys/sirius_kernels.txt
ducndh: Yah good points about these. I will centralize some benchmarking and scripting effort too

docs/duckdb-pin-overhead-analysis.md
ducndh: This is mostly for me, tasking a new Claude conversation, but those should not be in the PR.

patches/0001-feat-add-GetDirectBlockPointer-for-zero-copy-block-a.patch
ducndh: This PR was mostly to get some verifications run on other machines that my existing boxes and some proof of concepts, so I know exactly what we need from DuckDB people, so I have to hardcore patching DuckDB to get it running end-to-end. Will remove these in the future PR

src/cuda/scan/gpu_decode_batched_string.cu
ducndh: I will look through the cu files to create a PR to move the one for legacy Sirius, too. Great catch.

src/cuda/scan/gpu_decode_batched_string.cu
ducndh: Yeah, this one is my initial implementation of using one thread to do everything to make sure everything can be done on the GPU safely and correctly. Should have cleaned this up more carefully.

src/cuda/scan/gpu_decode_bitpacking.cu
ducndh: I am taking a look at the all decoder kernel and also their performance on different data characteristics to make sure that we don't miss anything, but I will leave a number on top of the file for future bookkeeping.

src/cuda/scan/gpu_decode_bitpacking.cu
ducndh: I was trying to be clever about reusing some code that led to a lot of trash but since tons of things go through a batch interface should be more strict in cleaning theses.

src/cuda/scan/gpu_decode_fsst.cu
ducndh: Put this inside batch string too should have cleaned this up. Great catch.

src/cuda/scan/gpu_decode_dictionary.cu
ducndh: Yah will take a more careful look at styling to keep it more consistent with the whole code base

src/cuda/scan/gpu_decode_dictionary.cu
ducndh: Yah dead code that got move into batch sorry for that

src/cuda/scan/gpu_decode_batched_string.cu
ducndh: Hmm I will have to think about this a bit more. Everything is jumble in batch right now and FSST is a string compression so it sometimes need special treatment. Thanks for pointing this out.

src/cuda/scan/gpu_native_decode.cu
ducndh: Good catch — the wording mixes plan-time viability fallback (check_viability in gpu_native_scan_task.cpp sets viable_ = false on unsupported compressions and the executor routes to duckdb_scan_task) with runtime fallback (which doesn't exist — the throw is a defensive assert that check_viability stays in sync with the decoder switch). 

I will update the comment on the header to make the runtime path clear: "unsupported segments are rejected at plan time (check_viability), and the scan falls back to duckdb_scan_task; reaching a throw in the decoder indicates a viability-check drift bug.

src/include/cuda/scan/gpu_native_decode.cuh
ducndh: Yah same as above

src/op/sirius_physical_hash_join.cpp
ducndh: Yah this might not be the best place to do this. I will review if I can override existing function to resolve this

src/op/scan/duckdb_scan_task.cpp
ducndh: Yah this should not be mangle up in here. good point

ducndh:
@joosthooz I really appreciate your review. Since DuckDB is changing some of their APIs for us so we don't have to do some direct nmap and some hacky changes, this PR is a huge mangled up of things with a sheer focus on making things run decently fast for this path and we have better io library so I don't have to roll my own, I will split this into smaller PR and phase to easier review and keep track of.  I will keep this open until I get the new split PRs up and we have the Parquet number in for progress bookkeeping. 

I will make sure that the split PRs will not have dead code littering everywhere.

PR #654, opened 2026-04-17
Add config option to disable parquet row group pruning

joosthooz:
Introduces enable_row_group_pruning (default true) so users can opt out of the filter-pushdown stats-based row group pruning added in #363

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -44,6 +44,8 @@ uint64_t Config::DEFAULT_SCAN_TASK_VARCHAR_SIZE = 256ULL;
 
 uint64_t Config::MAX_SORT_PARTITION_BYTES = 0;  ///< 0 = auto (33% of available GPU memory)
 
+bool Config::ENABLE_ROW_GROUP_PRUNING = true;
+
 std::string Config::LOG_LEVEL = "info";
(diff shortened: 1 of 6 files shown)


mbrobbel (commented):
Just curious: what is the use-case for disable?

joosthooz:
> what is the use-case for disable?

So that we're not caching pre-filtered data for hot runs; with pruning enabled the results for hot runs are too optimistic because it uses tables where the rowgroup filtering was already performed in the cold run.

kevkrist:
We can either wait for #620, which replaces the relevant code here with `sirius_parquet_metadata_scan_operator::execute()`, or we can merge it now and I can either add it to #620 or issue #655. Do you have a preference?

joosthooz:
I don't have a preference, we can wait for #620 or if there will be another consensus by the team about how to allow replicating GTC results that will be also fine with me!

bwyogatama:
@joosthooz #620 is merge so you can continue this PR

mbrobbel:
There are some conflicts. Converting to draft.

bwyogatama:
let's close this issue and re-open again if we think it's necessary since the scan infrastructure has changed a lot since

PR #656, opened 2026-04-17
Support COALESCE in gpu_expression_executor

### LINKED ISSUE #635: Support COALESCE expressions in the expression executor.

kevkrist:
## Summary
- Implement `OPERATOR_COALESCE` in `sirius::gpu_expression_executor` using `cudf::replace_nulls` iteratively across children. First child is materialized (scalar lifted to a column), and each subsequent child replaces residual nulls — scalar replacements use the scalar overload, column replacements use the column overload. Result is wrapped via `materialize_as_ast_column` so COALESCE composes inside AST-capable parents.
- Expand `COMPARE_IN` numeric dispatch to cover INT8, UINT{8,16,32,64}, BOOL8, DECIMAL{32,64,128}, and TIMESTAMP_{D,S,MS,US,NS}. BOOL8 dispatches via `uint8_t` because `std::vector<bool>::data()` is deleted.
- Tighten a couple of AST-mode wraps in `gpu_execute_case.cpp` and `gpu_execute_function.cpp` so materialized children are registered as AST leaves consistently.

## Notes
- Short-circuit: children are materialized only if there are nulls in current result. `count_ast_ops` returns 0 for COALESCE, matching the materialize-only path.

## Test plan
- [x] `build/release/extension/sirius/test/cpp/sirius_unittest \"[expression_executor][experimental]\"` — 73 test cases / 3219 assertions pass.
- [x] New `TEMPLATE_TEST_CASE(\"experimental execute COALESCE\")` across `mat`, `ast_interpret`, `ast_jit` strategies covering:
  - col + scalar fallback fills every null
  - col + col leaves residual nulls where both are null
  - col + col + scalar chain fills every null
  - col + col + col leaves residual nulls where all three are null
- [x] New `TEMPLATE_TEST_CASE(\"experimental select COALESCE nested in predicate\")` — `COALESCE(col0, 0) > 20` exercises AST parent → materialized child handoff.
- [x] SQLLogicTests covering COALESCE end-to-end through the planner.

Fixes #635


src/expression_executor/specializations/gpu_execute_operator.cpp
```diff
@@ -162,10 +250,8 @@ execute_result gpu_expression_executor::execute(duckdb::BoundOperatorExpression
             not_expr, output.get_temp_scalar_indices(), output.get_temp_column_indices()));
         }
         case duckdb::ExpressionType::OPERATOR_COALESCE:
-          /// KEVIN: TODO
-          throw duckdb::NotImplementedException(
-            "[gpu_expression_executor] execute called on an unsupported COALESCE operator "
-            "expression.");
+          throw std::runtime_error(
```
9prady9: I think we are using sirius exceptions now, defined under include, please check that header.

src/expression_executor/specializations/gpu_execute_case.cpp
```diff
@@ -23,6 +23,7 @@
 #include <duckdb/planner/expression/bound_function_expression.hpp>
 
 // cudf
+#include <cudf/ast/expressions.hpp>
 #include <cudf/column/column_factories.hpp>
```
9prady9:
```suggestion
```
can be removed now that we don't use scalar factory

src/expression_executor/specializations/gpu_execute_case.cpp
```diff
@@ -114,17 +117,10 @@ execute_result gpu_expression_executor::execute(duckdb::BoundCaseExpression cons
...
-      result_column = cudf::make_column_from_scalar(
-        current_result.get_scalar(), _input_table.num_rows(), _stream, _mr);
-    } else {
-      result_column =
-        std::make_unique<cudf::column>(current_result.get_column_view(), _stream, _mr);
-    }
-    return materialize_as_ast_column(std::move(result_column));
+    // The caller wants an AST node.
+    // Since at least one copy_if_else has been executed, current_result must have an owned column.
+    D_ASSERT(current_result.is_owned_column());
```
9prady9: is a scalar result not possible anymore ? wouldn't release cause a throw if that happens ? 

Perhaps may be actually throw instead of D_ASSERT ?

src/expression_executor/specializations/gpu_execute_function.cpp
```diff
@@ -144,13 +144,11 @@ execute_result gpu_expression_executor::execute(duckdb::BoundFunctionExpression
...
+    // Re-enter execute with MATERIALIZE mode to get the result as a column, then add to the AST
+    // tree.
     auto result = execute(expr, execution_mode::MATERIALIZE);
-    if (result.is_scalar()) {
-      return materialize_as_ast_column(
-        cudf::make_column_from_scalar(result.get_scalar(), _input_table.num_rows(), _stream, _mr));
-    }
-    return materialize_as_ast_column(
-      std::make_unique<cudf::column>(result.get_column_view(), _stream, _mr));
+    D_ASSERT(result.is_owned_column());
```
9prady9: same concern as here -> https://github.com/sirius-db/sirius/pull/656/changes#r3122696578

src/expression_executor/specializations/gpu_execute_operator.cpp
kevkrist: I've done my best to replace all the duckdb exceptions with sirius exceptions in the expression executor.

src/expression_executor/specializations/gpu_execute_case.cpp
kevkrist: It should be impossible. I've added a throw here and below.

src/expression_executor/specializations/gpu_execute_function.cpp
kevkrist: See above.

kevkrist (commented):
Thanks for this!

9prady9 (approved):
lgtm

PR #658, opened 2026-04-17
refactor: promote sirius::gpu_expression_executor to primary path

### LINKED ISSUE #646: Deprecate legacy expression executor
Move the legacy GpuExpressionExecutor to the legacy folder and replace all occurrences of the legacy executor with the new executor. Drop the experimental namespace qualifier from the new executor.

kevkrist:
## Summary                                                                                                                                
  - Promote the experimental expression executor (`sirius::experimental::gpu_expression_executor`)
    to be the primary one used by Super Sirius, and drop the `experimental` namespace — it is now                                           
    `sirius::gpu_expression_executor`.                                             
  - Move the legacy `duckdb::sirius::GpuExpressionExecutor` (plus its specializations and the
    CUDA `GpuDispatcher`) to `src/legacy/expression_executor/` and `src/legacy/cuda/expression_executor/`,                                  
    behind the existing `ENABLE_LEGACY_SIRIUS` CMake flag. Legacy sources are now wired in via a                                            
    new `src/legacy/CMakeLists.txt` and a guarded append in the root `CMakeLists.txt`.                                                      
  - Update Super Sirius call sites (`sirius_physical_filter`, `sirius_physical_projection`,                                                 
    `sirius_physical_nested_loop_join`, `sirius_physical_table_scan`,                                                                       
    `sirius_gpu_parquet_scan_operator`) to use the new API.               
  - strategy parsing is removed — the executor reads the default from                                                                       
    `duckdb::Config::EXPRESSION_EXECUTOR_STRATEGY` internally.                     
  - Turn `Config::EXPRESSION_EXECUTOR_STRATEGY` into a typed enum                                                                           
    (`sirius::expression_executor_strategy`) instead of a raw string. The new                                                               
    `expression_executor_strategy.hpp` header defines the enum plus `string_to_strategy` /                                                  
    `strategy_to_string` helpers. The `SET expression_executor_strategy = '...'` DuckDB variable                                            
    still accepts the same three string values (`materialize`, `ast_interpret`, `ast_jit`).                                                 
  - Rewrite `docs/super-sirius/expression-executor.md` to document the strategy system                                                      
    (MATERIALIZE / AST_INTERPRET / AST_JIT), the tree-of-AST-trees model, `min_ast_size`,                                                   
    AST breakers, and where per-expression-class specializations live. Minor touch-ups to                                                   
    README/operators/scan/optimizations docs.                                                                                               
                                                                                                                                            
  ## Test plan                                                                                                                              
  - [x] Build with default `ENABLE_LEGACY_SIRIUS=OFF` — legacy sources excluded.                                                            
  - [x] Build with `-DENABLE_LEGACY_SIRIUS=ON` — legacy sources compiled via                                                                
        `src/legacy/CMakeLists.txt` and linked into the extension.                                                                          
  - [x] `sirius_unittest` expression_executor suite — 70/70 pass.                                                                           
  - [x] `tpch-sirius.test` (legacy path, `SIRIUS_DISABLE=1`) — passes; the 5 fallback                                                       
        messages are the pre-existing DECIMAL128 scalar throw from `src/legacy/gpu_columns.cpp`                                             
        (introduced in #160), unrelated to this refactor.                                                                                   
  - [ ] TPC-H SF1 via `gpu_execution` on each of the three strategies                                                                       
        (`ast_interpret`, `ast_jit`, `materialize`).                                                                                        
                                                                                                                                            
  Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com> 

Fixes #646


docs/super-sirius/operators.md
```diff
@@ -93,15 +93,15 @@ These operators process data in a single pass without buffering.
 
 Applies a predicate expression to filter rows.
 
-- **GPU execution:** `GpuExpressionExecutor::select(batch, stream)` — evaluates the boolean expression and compacts rows using cuDF filtering
+- **GPU execution:** `gpu_expression_executor::select(batch, stream)` — evaluates the boolean expression and compacts rows using cuDF filtering
```
9prady9:
```suggestion
- **GPU execution:** `gpu_expression_executor::select(batch)` — evaluates the boolean expression and compacts rows using cuDF filtering
```

docs/super-sirius/operators.md
```diff
@@ -93,15 +93,15 @@ These operators process data in a single pass without buffering.
...
 - **Key members:** `expression` (filter predicate)
 
 ### `sirius_physical_projection` — `PROJECTION`
 **File:** `src/include/op/sirius_physical_projection.hpp`
 
 Evaluates a list of expressions to produce output columns.
 
-- **GPU execution:** `GpuExpressionExecutor::execute(batch, stream)` — evaluates each expression, producing a new table with projected columns
+- **GPU execution:** `gpu_expression_executor::execute(batch, stream)` — evaluates each expression, producing a new table with projected columns
```
9prady9, line 104:
```suggestion
- **GPU execution:** `gpu_expression_executor::execute(batch)` — evaluates each expression, producing a new table with projected columns
```

docs/super-sirius/expression-executor.md
```diff
@@ -6,14 +6,50 @@ This document covers the GPU expression execution subsystem used by FILTER and P
...
+
+| Strategy | How it executes | cuDF API |
+|----------|-----------------|----------|
+| `MATERIALIZE` | Every expression node becomes a single kernel. Intermediate results are materialized as `cudf::column`s. | `cudf::unary_operation`, `cudf::binary_operation`, etc. (one per node) |
+| `AST_INTERPRET` (default) | Builds a `cudf::ast::tree` and interprets it with a single monolithic kernel. | `cudf::compute_column` |
+| `AST_JIT` | Builds a `cudf::ast::tree` and JIT-compiles it into a fused kernel. | `cudf::compute_column_jit` |
+
+### Tree of AST Trees
+
+Not every DuckDB expression has a cuDF AST equivalent — these are called **AST breakers** (e.g. `CASE`, `LIKE`, `SUBSTRING`, unsupported `CAST` types). For AST strategies, the executor walks the DuckDB expression and greedily builds AST subtrees up to each breaker. When it hits a breaker, it materializes that subtree as a `cudf::column`, stashes it in `_temp_columns`, and references it from the enclosing AST subtree via a `cudf::ast::column_reference`.
```
9prady9:
```suggestion
Not every DuckDB expression has a cuDF AST equivalent — these are called **AST breakers** (e.g. `CASE`, `LIKE`, `SUBSTRING`, unsupported `CAST` types). For AST strategies, the executor walks the DuckDB expression and greedily builds AST subtrees up to each breaker. When it hits a breaker, it materializes that subtree as a `cudf::column`, stashes it internally, and references it from the enclosing AST subtree via a `cudf::ast::column_reference`.
```

CMakeLists.txt
```diff
@@ -276,6 +272,10 @@ set(CUDA_SOURCES
     src/cuda/utils.cu)
 # cmake-format: on
 
+if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/src/legacy/CMakeLists.txt")
```
9prady9, line 278: we need to add check for build legacy as well here.

test/cpp/expression_executor/test_gpu_expression_executor.cpp
9prady9, line 1: do we want to drop the legacy code tests right away ?

src/include/expression_executor/expression_executor_strategy.hpp
```diff
@@ -0,0 +1,73 @@
...
+ * @brief Returns the canonical string representation of an expression_executor_strategy.
+ */
+inline std::string_view strategy_to_string(expression_executor_strategy s)
+{
+  switch (s) {
+    case expression_executor_strategy::MATERIALIZE: return "materialize";
+    case expression_executor_strategy::AST_INTERPRET: return "ast_interpret";
+    case expression_executor_strategy::AST_JIT: return "ast_jit";
+  }
+  return "ast_interpret";
```
9prady9: was this intended as default behavior ?

src/sirius_extension.cpp
```diff
@@ -915,7 +917,7 @@ void SiriusExtension::InitialGPUConfigs(DBConfig& config)
     "Strategy for the experimental gpu_expression_executor: 'materialize', 'ast_interpret', or "
```
9prady9:
```suggestion
    "Strategy for the gpu_expression_executor: 'materialize', 'ast_interpret', or "
```

src/include/expression_executor/expression_executor_strategy.hpp
9prady9, line 1: this header is included in almost all translation units via config.hpp, perhaps we move the cpp implementations to cpp file or move these utilities else where ?

9prady9: or another alternative is to move the enum to it's own header ?

src/include/expression_executor/expression_executor_strategy.hpp
kevkrist: No this code should be unreachable, so I've marked it as such.

src/include/expression_executor/expression_executor_strategy.hpp
kevkrist: Good call -- I've moved the implementations to a .cpp source file.

CMakeLists.txt
kevkrist: When I add this gate:
```
if(ENABLE_LEGACY_SIRIUS AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/src/legacy/CMakeLists.txt")
  list(APPEND CUDA_SOURCES ${SIRIUS_LEGACY_CUDA_SOURCES})
endif()
```
without a corresponding gate on the .cpp files at 245, I get compilation errors because the .cu files aren't compiled. If I  add the gate at 245 as well, sirius fails to compile, because there is code not in legacy (e.g. cpu_cache.cpp) that still refers to legacy sources (e.g., 'gpu_columns.cpp'). So unfortunately, without extensive changes, I think it has to stay this way.

test/cpp/expression_executor/test_gpu_expression_executor.cpp
kevkrist: This is consistent with the rest of test/cpp/ — none of the legacy code path is unit tested there. Removing these just aligns the legacy tests with the existing pattern.

src/expression_executor/expression_executor_strategy.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+std::string_view strategy_to_string(expression_executor_strategy s)
+{
+  switch (s) {
+    case expression_executor_strategy::MATERIALIZE: return "materialize";
+    case expression_executor_strategy::AST_INTERPRET: return "ast_interpret";
+    case expression_executor_strategy::AST_JIT: return "ast_jit";
+  }
+  assert(false && "Unrecognized expression_executor_strategy");
+  __builtin_unreachable();
```
9prady9, line 50: 👍🏾

CMakeLists.txt
9prady9: If it isn't working, it is what it is although unfortunate we have to compile unused sources. We can revisit this later. thanks for looking into it.

9prady9 (approved):
lgtm

PR #661, opened 2026-04-20
[WIP] Draft design of caching framework in super sirius

9prady9:
This is the first draft of caching design me and claude came up with with current code & some generic info as inputs.

Appreciate any and all feedback. Wanted to add it here for easy collaboration. If the design is ratified and accepted, we can continue this as the caching design doc in super sirius.

### DIFF
--- a/docs/super-sirius/caching.md
+++ b/docs/super-sirius/caching.md
@@ -1,0 +1,1050 @@
+# Super Sirius Batch Cache Manager — Design Document
+
+**Status:** Draft v3, for review
+**Scope:** Single-node batch cache manager with cluster-ready extension points
+**Target language:** C++20
+**Supersedes:** the `cache_level` enum in `op/scan/config.hpp` and the ad-hoc wrapping logic in `parquet_scan_task.cpp`
+
+---
+
+## 1. System Context and Assumptions
+
+- **Workload.** OLAP, columnar analytics over largely static tables. Plan re-execution and retries are common; the same chunk is read many times across queries. Cross-query reuse is therefore valuable, not incidental.
+- **Hardware.** One host, one or more GPUs, NUMA-aware host memory. All GPUs on the node run inside a single Super Sirius process and share the process-global cache manager described here. GPUs on the node may or may not be NVLink-connected; the design does not assume NVLink.
+- **Existing substrate.** Everything described in §1–§7 of `data-batch-caching.md`: the `cucascade::data_batch` abstraction, swappable `idata_representation`s, `cached_shared_representation<T>` for shared ownership, cucascade's host pool and reservation manager, and the byte-range cache used by the parquet scan. This design sits on top of those primitives; it does not replace any of them.
+- **Transfer substrate.** A GPU-to-GPU transfer capability is assumed to exist behind a backend-agnostic abstraction. The abstraction — not a specific product — is the dependency. Acceptable backends include NIXL, a homegrown shim over `cudaMemcpyPeer` and a host-bounce fallback, or a future libkvikio-based transport with extensions. No specific backend is mandated.
+- **Concurrency regime.** Many pipelines may run concurrently across multiple GPUs. Multiple pipelines may look up the same chunk at the same time. The hot-path lookup must not serialize on a global mutex.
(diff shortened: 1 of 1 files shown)


docs/super-sirius/caching.md
```diff
@@ -0,0 +1,1050 @@
...
+
+---
+
+## 3. Executive Summary — What Is Reused from the Current Codebase
+
+The central claim of this design is that Super Sirius already has the right *primitives* for a non-copying, tier-aware batch cache. What is missing is a **policy layer** sitting above those primitives. The table below states the reuse position precisely.
+
+| Concept in this design         | Primitive it rides on                                         | Status          |
+|--------------------------------|---------------------------------------------------------------|-----------------|
+| Unit of caching                | `cucascade::data_batch` with swappable `idata_representation` | Reused as-is    |
```
wmalpica, line 51: A data_batch in practice is a set of columns. The unit of caching has to be at the column level. 

This is so that if you have:
SELECT colA, colB from table
and 
SELECT colB, colC from table.
you want to be able to use colB from the Cache.

And perhaps more specifically, it has to be at the column level and rowgroup level. 
So that SELECT colA, colB from table WHERE colA > 1000
and SELECT colB, colC from table WHERE colC > 1000 may have something in common

wmalpica (changes requested):

ducndh:
Thanks for writing this up. 

I was trying to build a cache system this weekend, but my scope is narrower than yours: read-only analytical, Parquet + DuckDB-native, single-session-first. Within that, three measurements drive the whole design for me:  
- Query cost breaks into ~equal thirds: transport + decode + execution (Hopper collapses transport; x86 keeps it at ~1/3).                                           - Filter reapplication is ~free relative to decode.
- The same (column × rowgroup) is scanned by many queries with different predicates —TPC-H Q1/Q3/Q6/Q14/Q15/Q20 all scan lineitem with different l_shipdate predicates. 

That pushes me toward a shape where cache sits below filter, above decode (predicates lift downstream, a decoded column serves every query shape); the unit is (column × rowgroup) — which is also what @wmalpica flags in R51; eviction is one measured number (decode cost) rather than a 4-axis framework; and cuCascade's existing tier system handles movement while the cache just holds shared_ptr<data_batch> and contributes evict-first candidates.

Though I'm still tinkering on it — will share the actual design doc and benchmark numbers once it's stable and free of problems, but essentially, my system records where things are staying with a unit of (rowgroup x col), and a scan manager that will manage how to get a coherent table view to pass to the downstream operator based on where these entries live.
                                                                                        
There are some concerns about this draft from my narrow point of view, based on measurements I have done:

- filter_fingerprint in the key binds predicate to cache identity. For the 6 lineitem queries above, that's 6 entries instead of 1 — paying ~1/3 decode N times for ~0 reapply savings. Lifting filters downstream is the quantitative win for me.                - "Chunk" (multi-column) granularity — cross-projection queries share nothing.
- 4-axis policy framework (admission × eviction × migration × prefetch) is a lot since the cost is mostly decode + transport (or just maybe decode in the future), and prefetching might be a hard thing to do with GPU memory pressure. However, I might just not think about the problem from a correct point of view.
- epoch broadcast invalidation — versioning itself is real, and I handle it with stable identifiers in the key, but a broadcast channel for mid-session mutations feels unnecessary for a read-only analytical workload. However, if this is something we need to think carefully about at the beginning, because applying a patch would break our system, feel free to push back.
                                                                                        
Though I do agree on these points:   
- One logical cache with per-entry tier state — same outcome, just implicit via shared_ptr + cuCascade rather than explicit tier slots.
- Source-agnostic key with per-source construction.
- Your cache_stats_snapshot observability structure is cleaner than what I have right now (which is a bunch of print lines)                                                                                        
  
TLDR: (column × rowgroup) without filter pushdown should be the main unit, and the cost structure in my mind is a bit different. I am happy to discuss this more or just open a WIP PR if you want to take a look at the code directly.

docs/super-sirius/caching.md
9prady9: I think `chunk_identity` and `cache_key` which result in creating the `cache_entry` already accounts for this https://github.com/sirius-db/sirius/pull/661/changes#diff-fec5f264d2a0bcd487e96d0afc8d0175f6c80331bbdf44e01458586bc50614f8R560

9prady9:
@ducndh thank you for the thorough feedback, let me go over it and will get back to you.

9prady9:
closing this for now, I am not actively work on this draft right now and other devs started on it. So no relevance right now.

PR #662, opened 2026-04-20
scan: name-based DuckDB→parquet column mapping in metadata scan

kevkrist:
## Summary

Behavior-preserving refactor of `sirius_parquet_metadata_scan_operator` that replaces its implicit assumption *"DuckDB primary index == parquet column-chunk index"* with an explicit, name-based lookup through the parquet file metadata. This is a prerequisite for nested-column projection support (follow-up PR) and also eliminates a latent correctness hazard for any file whose physical column order is not identical to DuckDB's logical column order.

## Motivation

`sirius_parquet_metadata_scan_operator.cpp` currently indexes into `row_group.columns[col_idx]` where `col_idx` is a *DuckDB primary index* (i.e., an index into `scan_op->names` / `scan_op->returned_types`). This only works when the parquet file's column-chunk order matches DuckDB's logical column order one-to-one — which is true for the flat-schema test corpus, but:

- Fails for any file where the writer's physical order differs from the reader's projection order.
- Assumes a 1:1 correspondence between DuckDB columns and parquet column chunks, which is only true for **flat** schemas (no STRUCT / LIST). This blocks PR 2.

Today's downstream `parquet_scan_task.cpp` (the non-pipeline path) already does the name-based lookup — see `parquet_scan_task.cpp:169` (\"the order of set_column_names ...\"). This PR brings the metadata operator in line with that approach.

## Scope (what changes)

- Replace direct indexing (`row_group.columns[col_idx]`) with a DuckDB-index → parquet-leaf-chunk-index resolution built from the parquet schema's `path_in_schema`.
- Introduce a single-file helper (tentatively `src/op/scan/parquet_schema_mapping.{hpp,cpp}`) exposing:
  - `leaf_indices_for_column(schema, column_name) -> std::vector<std::size_t>` — returns all parquet leaf column-chunk indices whose `path_in_schema[0]` matches `column_name` (case-insensitive to mirror DuckDB).
  - For this PR, each call returns exactly 1 entry — the flat-schema guard is preserved (PR 2 removes it).
- Mirror the `CUDF_VERSION_NUM >= 2604` branch from `parquet_scan_task.cpp:408-412` in `sirius_parquet_metadata_scan_operator.cpp:194` (currently unversioned — works on 26.04 only).
- Unit tests exercising: column order swaps, case differences between DuckDB and parquet names, projection of a subset of columns in non-file order.

## Out of scope

- Nested types (STRUCT / LIST / DICTIONARY) — #663.
- Iceberg — separate follow-up.
- Any changes to `parquet_scan_task.cpp` (already name-based).

## Verification

- [x] `HivePartitioningIndex.index` is a DuckDB-level column index, not a parquet schema index (`duckdb/src/common/multi_file/multi_file_reader.cpp:270-279`). Metadata operator's name-based mapping must therefore resolve **before** consulting parquet metadata — unchanged from today.
- [x] `CUDF_VERSION_NUM` guard pattern established at `src/op/scan/parquet_scan_task.cpp:408-412` and `test/cpp/data/test_host_parquet_representation.cpp:557-560`.
- [x] Confirm: name matching should be case-insensitive (DuckDB convention) — to be verified against current hive-partition lookup behavior in `HivePartitioning::Parse()`.

## Test plan

- [x] Existing pipeline-scan SQL tests (flat schemas) continue to pass — no behavioral change expected for files whose column order matches.
- [x] New unit test: parquet file written with columns in order `[c, a, b]`, scan with projection `[a, c]` — assert row-group metadata for the correct leaves is emitted.
- [x] New unit test: case-only name difference between DuckDB and parquet — assert mapping succeeds.
- [x] `test/cpp/scan/test_metadata_gpu_scan_operators.cpp` and `test/cpp/pipeline/test_metadata_gpu_pipeline_task_counting.cpp` still build and pass.

---


src/op/scan/parquet_schema_mapping.cpp
```diff
@@ -0,0 +1,36 @@
...
+std::vector<std::size_t> leaf_indices_for_column(cudf::io::parquet::FileMetaData const& metadata,
+                                                 std::string const& column_name)
+{
+  std::vector<std::size_t> result;
+  if (metadata.row_groups.empty()) { return result; }
+
+  auto const& columns = metadata.row_groups.front().columns;
+  for (std::size_t i = 0; i < columns.size(); ++i) {
+    auto const& path = columns[i].meta_data.path_in_schema;
+    if (!path.empty() && path.front() == column_name) { result.push_back(i); }
```
ducndh, line 31: Does this work for a Parquet file that is not written by DuckDB? I remember DuckDB handle normalization for us but we might need to figure something out for Parquet in the wild.

ducndh:
Thanks for this PR, since this ordering problem has been on my mind for a while.

I suppose the same problem still exists for parquet_scan_task; we might need to be careful there about column ordering too, but you have already handled that, that is awesome.

ducndh (approved):

src/op/scan/parquet_schema_mapping.cpp
kevkrist: It should work for any Parquet file, not just DuckDB-written ones, becuase I think DuckDB's Parquet reader populates names by reading the file's schema directly, without normalization (I could be wrong?), so the strings line up byte-for-byte with what's in the file. If there are pathological parquet files out there for which this won't work (duplicate names?), let's handle them when we run into them.

PR #663, opened 2026-04-20
scan: support nested column projection in metadata scan

kevkrist:
## Summary

Extends the parquet metadata-scan pipeline to support projection of **STRUCT** and **LIST** columns in addition to flat primitives. Builds on #662's name-based DuckDB→parquet column mapping by generalizing the metadata scan's single-leaf-per-column assumption to a multi-leaf-per-column one.

## Motivation

Before this PR, the metadata-scan pipeline enforced a flat-schema guard (`"Parquet scans with projections currently only support flat projected columns."`) because each projected DuckDB column was assumed to map to exactly one parquet column chunk. That breaks for:

- **STRUCT** — one DuckDB column maps to N parquet leaf chunks (one per field), all sharing `path_in_schema[0]`.
- **LIST** — parquet's 3-level list encoding produces leaves at a deeper path (`.list.element` etc.); again, one DuckDB column → one or more leaves.

Any projection of a nested column fell back to DuckDB CPU.

## Changes

1. **Drop the flat-schema guard** in the metadata scan `execute()` path.
2. **Multi-leaf expansion**: for each selected column, call `leaf_indices_for_column` (already generalized to `std::vector<size_t>` in #662) and push every leaf index — including pure-filter leaves — into the chunk-index sets.
3. **No post-read reassembly needed** — cuDF's parquet reader, given a top-level column name, materializes the nested `cudf::column` natively (option (a) from design). `execute()` downstream projection is unchanged.

## Tests

11 new C++ unit tests in `test/cpp/scan/test_metadata_gpu_scan_operators.cpp` (tagged `[two_pipeline_scan][nested]`, 572 assertions):

- Project STRUCT / LIST / STRUCT-inside-STRUCT / LIST-of-LIST / LIST-inside-STRUCT
- Mixed flat + nested projection
- Nested column across multiple files
- Table with nested column but flat-only projection (regression guard)
- Flat filter pushdown alongside nested projection
- Edge cases: LIST with empty rows, STRUCT with NULL rows

## Known gaps (not addressed here)

- **Legacy `parquet_scan_task` path still has the flat-only guard** (`src/op/scan/parquet_scan_task.cpp:466`).
- No SQLLogic / end-to-end `gpu_execution` test — the new tests drive the operators directly.
- No negative test for *filter on a nested column*. The expression executor is known to error rather than crash, but that behavior isn't locked in by a test.

## Out of scope

- **MAP** — not present in `sirius::type_id`.
- **DECIMAL128 in nested contexts.**
- **Nested columns as hive partition columns** — hive key parser only handles scalars.

---

### DIFF
--- a/src/op/scan/sirius_parquet_metadata_scan_operator.cpp
+++ b/src/op/scan/sirius_parquet_metadata_scan_operator.cpp
@@ -18,7 +18,7 @@
 #include <log/logging.hpp>
 #include <op/scan/hive_partition.hpp>  // build_partition_inject_fn
 #include <op/scan/parquet_scan_operator_data.hpp>
-#include <op/scan/parquet_scan_task.hpp>  // detail::make_selected_column_indices, detail::projected_columns_are_flat
+#include <op/scan/parquet_scan_task.hpp>       // detail::make_selected_column_indices
 #include <op/scan/parquet_schema_mapping.hpp>  // detail::leaf_indices_for_column
 #include <op/scan/scan_utils.hpp>
 #include <op/scan/sirius_gpu_parquet_scan_operator.hpp>
@@ -247,12 +247,6 @@ std::unique_ptr<operator_data> sirius_parquet_metadata_scan_operator::execute(
       cudf::host_span<uint8_t const>(footer_buffer->data(), footer_buffer->size()),
       *result->reader_options);
     auto metadata = reader.parquet_metadata();
-    if (_is_projected && !detail::projected_columns_are_flat(metadata, _projected_column_names)) {
-      /// TODO: Support nested column schemas with projection.
-      throw std::runtime_error(
-        "[sirius_parquet_metadata_scan_operator] Parquet scans with projections currently only "
-        "support flat projected columns.");
-    }
 
     //===----------Resolve selected DuckDB columns to parquet column chunk indices----------===//
     // row_group.columns is indexed in parquet schema-leaf order (preorder), which can differ from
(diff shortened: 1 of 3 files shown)


src/op/scan/sirius_parquet_metadata_scan_operator.cpp
```diff
@@ -261,19 +255,18 @@ std::unique_ptr<operator_data> sirius_parquet_metadata_scan_operator::execute(
...
-        // projected_columns_are_flat (checked above) guarantees exactly one leaf per name.
-        if (leaves.size() != 1) {
-          throw std::runtime_error(
-            "[sirius_parquet_metadata_scan_operator] Projected column '" +
-            _projected_column_names[k] +
-            "' did not resolve to exactly one parquet leaf in file: " + file_path);
+        if (leaves.empty()) {
+          throw std::runtime_error("[sirius_parquet_metadata_scan_operator] Projected column '" +
+                                   _projected_column_names[k] +
+                                   "' not found in parquet file: " + file_path);
```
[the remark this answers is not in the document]
kevkrist: The implementation is intentionally case-sensitive: by the time leaf_indices_for_column runs, the names come from DuckDB's parquet bind, which already extracted them from the file schema verbatim, so we're matching file-schema names to themselves. Making it case-insensitive would mask rather than help the one pathological case (two file-schema columns differing only in case). The case-sensitive behavior is locked in by a dedicated unit test. I'll correct the #662 description.

src/op/scan/sirius_parquet_metadata_scan_operator.cpp
```diff
@@ -261,19 +255,18 @@ std::unique_ptr<operator_data> sirius_parquet_metadata_scan_operator::execute(
...
         }
-        selected_chunk_indices.push_back(leaves.front());
-        if (_pure_filter_column_indices.contains(_selected_column_indices[k])) {
-          pure_filter_chunk_indices.insert(leaves.front());
+        bool const is_pure_filter =
+          _pure_filter_column_indices.contains(_selected_column_indices[k]);
```
ducndh, line 266: I think this one might conflict with how #706 resolve the logic when rebasing. `_pure_filter_column_indices` here is indexed by primary/selected-column index, but `_plan.pure_filter_batch_positions()` returns batch positions (D-space). A mechanical substitution might silently break — pure-filter leaves won't get pushed into `pure_filter_chunk_indices`, the uncompressed-byte accounting will count them then row-group partitions get too large. The test might still produces correct rows so the unit tests won't catch it hrere but I can entriely be offbase.

test/cpp/scan/test_metadata_gpu_scan_operators.cpp
```diff
@@ -1088,3 +1089,678 @@ TEST_CASE("two-pipeline scan - DECIMAL filter falls back to DuckDB expression ex
   }
   REQUIRE(total_rows == 100);
 }
```
ducndh: I don't know if you prefer a negative test but I think a test that check explicit filter-on-nested-column will throw instead of crashing might be nice.

ducndh (approved):
Yah great a PR and I don't think my work affects anything in this one

PR #667, opened 2026-04-21
CI: Fail merge_group runs when there are errors with TPC-H sf1

mike-wendt:
Depends on #670

Right now the performance comparison of TPC-H sf1 runs at the end of the GPU test, but it does not fail if any of the queries fail. This changes our approach to fail only on `merge_group` runs if there are any detected query errors. This should help prevent future PRs that break the TPC-H test

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,3 +99,10 @@ jobs:
             cat "$RUN_DIR/comparison.txt"
             echo '```'
           } >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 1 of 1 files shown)


mbrobbel (approved):

mike-wendt:
Failure due to upstream packaging changes from RAPIDs, will wait and try again later today

mbrobbel:
> Failure due to upstream packaging changes from RAPIDs, will wait and try again later today

We'll need a follow-up PR for cucascade to support both stable and nightly, and then another here for the same.

PR #675, opened 2026-04-22
Introducing Sirius IO, Prefetching and Caching Framework

aminaramoon:
# Introduce `sirius::io` — reactor-driven I/O with a pinned prefetching cache

Adds the `sirius::io` subsystem: a `cudf::io::datasource`-compatible I/O stack built on io_uring reactors, with a pinned-memory prefetching cache and a pluggable backend seam.

## Summary

- `sirius_datasource` — implements `cudf::io::datasource`, `supports_device_read() = true`, delegates to a shared `sirius_ioctx`.
- `sirius_ioctx` — abstract shared context; owns the optional `prefetching_cache` and the reactor pool. Every `device_read` consults the cache first and falls through to backend I/O on miss.
- `templated_ioctx<Reactor>` — generic ioctx implementation. Handles request splitting, aligned 1 MiB chunking, round-robin dispatch across reactors, and sync/async adapters.
- `uring_reactor` / `uring_ioctx` / `uring_io_object` — concrete io_uring backend. One thread per reactor, `O_DIRECT` device reads through pinned bounce slots, buffered host reads on the same ring via a tagged `user_data` pointer.
- `prefetching_cache` — pinned-memory, chunk-based, lock-free per-entry state machine with background worker + evictor threads and tiered LRU buckets.
- `buffer_pool` — growable multi-slab pool of 1 MiB pinned chunks.
- `admission_control` — RAII slot handed out against a fixed in-flight budget (default 2 GiB worth of chunks).

## Extensibility model

Two C++20 concepts define the plug-in contract:

- `io_object_c<O, Handle>` — derives from `sirius_io_object`, exposes `host_handle()` / `device_handle()` of type `Handle`.
- `io_reactor_c<R>` — associated types (`native_handle_type`, `io_object_type`, `device_read_req_type`, `host_read_req_type`) plus 5 operations: `enqueue_bulk`, `host_read`, `host_read_async`, `shutdown`, static `align_to_physical`.

A new backend is: your `io_object` + your reactor + `templated_ioctx<your_reactor>`. `uring_ioctx = templated_ioctx<uring_reactor>` is the first instantiation; `cufile_ioctx` would be the next.

## Design points

- **Cache seam lives in the base class.** `sirius_ioctx::device_read{,_async}` is non-virtual — it consults `_cache` and falls through to pure-virtual `device_read_io{,_async}`. Backends never see the cache; the cache never sees the backend. `supports_device_read()` stays `true` even when the cache serves the read, because the final copy is still `cudaMemcpyAsync` from pinned host memory to device.
- **Packed atomic state machine.** `entry_state` encodes a 4-bit state enum and a 28-bit pin count in a single `atomic<uint32_t>`. Every transition is one CAS, which closes the TOCTOU gap between "is this entry readable?" and "bump the pin count." Readers park in `wait_while_loading()` using `atomic::wait` and are woken by `notify_all()` on completion.
- **Request aggregation via `request_context`.** One logical read fans out into N chunk sub-requests; each sub-request decrements `pending`; the last one fires the user's completion handler. Error reporting is single-writer (`failed.exchange`) so partial failures don't race.
- **Batch dispatch, amortized wakes.** `templated_ioctx::enqueue_device_read` groups chunks per-reactor with a rotating round-robin start and dispatches one `enqueue_bulk` per non-empty group. Collapses N wake-notifies to at most M (reactor count) and uses the cheaper `moodycamel::enqueue_bulk` path. Cross-reactor parallelism preserved for ranges with ≥ M chunks.
- **Multi-GPU safe.** `device_read_req` carries the caller's `device_id`; reactor threads `cudaSetDevice()` before issuing the H2D copy. Bounce slabs are `cudaHostAllocPortable` so they're reachable from any CUDA context.
- **Evictor as backpressure service.** When `buffer_pool.allocate_bulk` can't satisfy the worker, the worker posts an `eviction_request` (promise + chunk count) and blocks on the future. The evictor walks LRU buckets coldest-first, returns chunks to the pool, then resolves the promise. Worker retries allocation. Pool exhaustion is never a silent failure.
- **Tiered LRU with age drift.** 5 buckets; `refresh_cache()` is the caller's only input to the aging signal. Score is `(NUM_BUCKETS-1) + n_total_request − cache_age`, clamped. Never-consumed entries get a floor of 1 to avoid being first out; consumed-or-stale entries can drop below 0. Raw score `< -5` gets evicted on the spot during candidate drain, skipping the bucket walk.
- **Admission control has a deadlock escape.** A request larger than the total budget is granted the full budget when no other slots are outstanding, so oversized reads make progress instead of waiting forever.
- **Cooperative shutdown.** Worker/evictor are `std::jthread` with `stop_token`; `admission_control::acquire` and the cache's semaphore wait both honor it. Destruction order: worker joined before evictor before pool.
- **io_uring feature detection.** Ring init attempts `SINGLE_ISSUER | DEFER_TASKRUN` and transparently falls back to plain flags.

## File layout

```
io/
  types.hpp                       // sirius_io_object, sirius_ioctx, io_datasource, request_context, req structs
  sirius_datasource.{hpp,cpp}     // cudf::io::datasource impl
  templated_ioctx.hpp             // templated_ioctx<Reactor> + concepts
  admission_control.{hpp,cpp}     // RAII slot / budget backpressure
  prefetching_cache.{hpp,cpp}     // buffer_pool, entry_state, pinned_view, prefetching_cache
  uring/
    uring_reactor.{hpp,cpp}       // uring_io_object + single-threaded reactor
    uring_ioctx.{hpp,cpp}         // ring_pool + templated_ioctx<uring_reactor> specialization
```

## Constants (in `types.hpp`)

| Name | Value | Role |
|---|---|---|
| `CHUNK_SIZE` | 1 MiB | Bounce-buffer / cache chunk size |
| `NUM_CHUNKS` | 32 | Bounce slots per reactor |
| `IO_BLOCK_SIZE` | 4096 | `O_DIRECT` alignment |
| `CHUNKS_PER_SLAB` | 500 | Pinned chunks per `buffer_pool` slab |

## Notes for reviewers

- **Concept usage is load-bearing.** `templated_ioctx` calls only the operations named by `io_reactor_c`; adding a new method to the concept is a breaking change for every backend. Please flag additions.
- **Locking hierarchy in the cache:** L0 `prefetching_cache::_map_mtx` → L1 `file_entry::mtx` → (independent) lock-free `cache_entry` atomics. Don't invert.
- **`io_completion_handler` is `std::function`**, so admission_control slots (move-only) are captured via `shared_ptr<slot>` in the worker's IO continuation. See `worker_loop` epilogue — the `slot_holder` pattern deliberately extends the slot's lifetime to the callback's end.
- `uring_io_object` opens two fds per file (buffered + `O_DIRECT`). Callers holding one must keep the object alive.
- No test suite in this PR. Integration tests land in a follow-up; the subsystem has been exercised end-to-end through cuDF's Parquet reader on NVMe workloads.

## Follow-ups (not in this PR)

- Tests (unit for `entry_state` / `buffer_pool` / `admission_control`; integration via cuDF readers).
- `cufile_reactor` as the second backend instantiation to validate the concept boundary.
- Metrics export for `prefetching_cache::summary()` counters.


src/include/io/templated_ioctx.hpp
```diff
@@ -0,0 +1,402 @@
...
+  {
+    auto& tobj = as_typed(obj);
+    size       = std::min(size, tobj.size() > offset ? tobj.size() - offset : size_t{0});
+    if (size == 0) {
+      handler(0, nullptr);
+      return;
+    }
+    auto ctx         = std::make_shared<request_context>();
+    ctx->handler     = std::move(handler);
+    ctx->total_bytes = size;
```
bwyogatama, line 168: I understand that currently we are doing 1MB chunks so we will never run into this issue.

But should we add a guard in this API that size has to be < 4 GB (`unsigned` limit for `io_uring`) just for safety in case someone call `host_read_async` API directly?

src/io/uring/uring_reactor.cpp
```diff
@@ -0,0 +1,333 @@
...
+      io_uring_sqe* sqe = io_uring_get_sqe(ring.get());
+      if (!sqe) break;
+      auto& req = pending.front();
+      spdlog::trace("reactor submit slot={} fd={} file_off={} io_size={:.2f}MB",
+                    si,
+                    req.handle,
+                    req.file_off,
+                    to_mb(req.io_size));
+      io_uring_prep_read(
+        sqe, req.handle, _bounce[si].buf.get(), (unsigned)req.io_size, (__u64)req.file_off);
```
bwyogatama, line 214: Same here, should we throw an error if `req.io_size` > 4 GB? 

We will not run into this issue since we chunk it into 1 MB but maybe someone wants to call `host_read_async` directly

src/io/uring/uring_reactor.cpp
```diff
@@ -0,0 +1,333 @@
...
+        if (res < 0) {
+          req.ctx->chunk_failed(std::make_exception_ptr(
+            std::runtime_error("reactor read: " + std::string(strerror(-res)))));
+          sinfo = slot_info{};
+        } else {
+          size_t rd     = (size_t)res;
+          size_t actual = rd > req.data_off ? std::min(req.data_size, rd - req.data_off) : 0;
+          if (actual > 0 && !req.ctx->failed.load(std::memory_order_relaxed)) {
+            _bounce[si].cuda_done.store(false, std::memory_order_relaxed);
+            if (req.device_id >= 0) cudaSetDevice(req.device_id);
```
bwyogatama, line 276: Shall we add error checking in all the cuda calls? I believe we did that with cucascade as well

src/include/io/uring/uring_ioctx.hpp
```diff
@@ -0,0 +1,91 @@
...
+namespace sirius::io {
+
+// ---------------------------------------------------------------------------
+// ring_pool
+// ---------------------------------------------------------------------------
+
+/**
+ * @brief Pool of @c io_uring instances for host (buffered) reads.
+ */
+class ring_pool {
```
bwyogatama, line 35: Is this class actually used anywhere?

src/io/prefetching_cache.cpp
```diff
@@ -0,0 +1,1038 @@
...
+    lru_insert(raw, bucket);
+    raw->in_lru = true;
+  }
+}
+
+// ===========================================================================
+// evictor_loop
+// ===========================================================================
+
+void prefetching_cache::evictor_loop(std::stop_token stop)
```
bwyogatama, line 730: Do we need to add try catch block on the `evictor_loop` and `worker_loop` to avoid deadlock if it terminates?

src/io/uring/uring_reactor.cpp
```diff
@@ -0,0 +1,333 @@
...
+      io_uring_sqe* sqe = io_uring_get_sqe(ring.get());
+      if (!sqe) break;
+      auto* req = new host_read_req_type(std::move(pending_host.front()));
+      pending_host.pop_front();
+      spdlog::trace("reactor host submit fd={} offset={} size={:.2f}MB",
+                    req->handle,
+                    req->offset,
+                    to_mb(req->size));
+      io_uring_prep_read(sqe, req->handle, req->dst, (unsigned)req->size, (__u64)req->offset);
+      io_uring_sqe_set_data64(sqe, reinterpret_cast<uint64_t>(req) | HOST_TAG);
```
bwyogatama, line 233: This pointer tagging approach works because bit 64 is 0 in the current pointers. Is this always the case?

I also heard that in future generation of ARM, they will use PAC which will cryptographically signs pointers for security, occupying the upper bits

bwyogatama (approved):

src/include/io/prefetching_cache.hpp
```diff
@@ -0,0 +1,628 @@
...
+// ---------------------------------------------------------------------------
+//
+// Packs a 4-bit state enum and a 28-bit reader pin count into a single
+// atomic uint32_t.  Every transition is a single CAS (or store), which
+// eliminates the TOCTOU race between checking state and modifying pin_count.
+//
+// State machine:
+//
+//         try_start_queueing()     try_start_loading()
+//   empty ─────────────────► queued ──────────────────► loading
```
wmalpica, line 122: NIT: this diagram is illegible

wmalpica (approved):

aminaramoon:
I am working on pr comments and waitin for yu's comments

src/io/prefetching_cache.cpp
```diff
@@ -0,0 +1,1038 @@
...
+        io_dsts.emplace_back(e->chunks[i], sz);
+      }
+    }
+
+    // io_completion_handler is std::function (copy-constructible), so the
+    // move-only slot must be wrapped in a shared_ptr to survive the copy.
+    auto slot_holder = std::make_shared<admission_control::slot>(std::move(budget_slot));
+
+    {
+      auto& obj_ref = *item.io_obj;
```
ran-yuan-rui, line 1006: Minor - Can we define an explicit shutdown order for cache vs backend reactors? The cache owns worker threads and callbacks that call back into the ioctx, but _cache lives in the base class while backend resources live in derived classes. We need a guarantee that cache workers are drained before reactors/file handles are destroyed.

src/io/sirius_datasource.cpp
```diff
@@ -0,0 +1,273 @@
...
+{
+}
+
+size_t sirius_datasource::size() const { return _io_object->size(); }
+
+bool sirius_datasource::supports_device_read() const { return true; }
+
+bool sirius_datasource::is_device_read_preferred(size_t) const { return true; }
+
+size_t sirius_datasource::host_read(size_t offset, size_t size, uint8_t* dst)
```
ran-yuan-rui, line 184: design need confrim  - supports_device_read() and is_device_read_preferred() currently always return true. I think supports_device_read() may be acceptable if every backend provides a functional fallback path, but is_device_read_preferred() should probably be backend-specific.
For example, a true GDS backend may want to prefer device reads, but an S3 backend would likely use HTTP range GET plus host staging plus H2D copy, where preferring device reads may be misleading or slower. Could we delegate is_device_read_preferred(size_t) to the underlying sirius_ioctx, with different defaults per backend?

src/include/io/templated_ioctx.hpp
```diff
@@ -0,0 +1,402 @@
...
+
+// ---------------------------------------------------------------------------
+// Concept: io_object_c
+// ---------------------------------------------------------------------------
+//
+// A concrete io_object for use with templated_ioctx<Reactor> must:
+//   - derive from sirius_io_object
+//   - expose device_handle() / host_handle() returning Handle
+
+template <class O, class Handle>
```
ran-yuan-rui, line 43: design to confirm - I think we should clarify the layering here. templated_ioctx<Reactor> currently requires the IO object to expose host_handle() / device_handle(), and uring_io_object implements that with POSIX fds plus an O_DIRECT fd. That is a good fit for local-file / fd-like backends, but it should probably be treated as an implementation helper rather than the stable backend abstraction.

For S3/object-store backends there is no native fd or O_DIRECT handle. Those backends can still implement sirius_ioctx directly, but they should not be forced into the templated_ioctx<Reactor> model. Can we document or structure this so sirius_ioctx is the stable abstraction, and templated_ioctx is just one backend strategy?

ran-yuan-rui:
design to confirm - The framework introduces sirius_ioctx / sirius_datasource, but the parquet scan path still constructs datasources directly with cudf::io::datasource::create(file_path). If the goal is to make this IO framework reusable across local files, S3, future GDS, etc., I think we need a datasource factory/registry seam here. and I will address that part in my S3 PR, so want  to post here first.

Ideally the scan layer should not know whether a URI is file://, s3://, or another backend. It should ask a registry/factory to create the appropriate cudf::io::datasource from the URI. That would let S3 and future backends plug in without adding backend-specific logic to the scan operators.

ran-yuan-rui (approved):

src/include/io/templated_ioctx.hpp
aminaramoon: it doesn't need to be a handle to a file, as you mentioned it doesn't exist in S3, but it can be a key to whatever registry you need to grab the url, bucket+object->url -> key = handle
you can lookup the GET or HEAD urls using your backend defined handle

src/io/sirius_datasource.cpp
aminaramoon: it returns true, because if you don't support device read you are guaranteed to get poor performance from your IO backend. if device_read is not supported, the read for parquest pages becomes
`host_read -> copy to bpu -> host_read -> copy to gpu -> ... `
vs
`gpu_read_async, gpu_read_async, gpu_read_async, ... -> then decomp/decode

src/include/io/templated_ioctx.hpp
aminaramoon: like for example, if your S3 work, the handle should give you a base url, and then u use the offset to add the ranges you need form the object.

Issue #189, opened 2026-01-23
change data_repository state machine and locking system


wmalpica:
@aminaramoon has a proposal for this which involves using shared and unique locks, and relying less on the existing state machine like set of states.

mbrobbel:
Closed via  #689

PR #689, opened 2026-04-23
Data batch refactor

wmalpica:
This closed this issue  #189  and more! 

This PR modifies Sirius to use the new tri-class implementation of `data_batch` which is now: `data_batch`, `read_only_data_batch` and `mutable_data_batch`. Which can be seen here https://github.com/NVIDIA/cuCascade/pull/117

One of the greatest benefits of this refactor can be seen in `lock_or_prepare_batch` which had some posible race conditions and the logic was complex. Not its very simple. 

API Changes:
Most API changes are actually from https://github.com/NVIDIA/cuCascade/pull/117

convertible_data_batch and convertible_gpu_pipeline_task  `convert` function now has a bool flag to make it blocking or non-blocking.

pipelineable_operator_data now can be constructed with a `std::vector<std::shared_ptr<::cucascade::data_batch>> data_batches` or `std::vector<::cucascade::read_only_data_batch> read_only_data_batches` and you can get either from it as well. Internally it will produce one or the other as necessary.


src/downgrade/downgrade_executor.cpp
```diff
@@ -204,7 +205,7 @@ void downgrade_executor::processing_loop()
            &host_target_stats,
            &disk_target_stats]() mutable {
             try {
-              auto result = cand->convert(targets, exc_stream, res_mgr);
+              auto result = cand->convert(targets, exc_stream, res_mgr, false);
```
bwyogatama, line 208: Seeing this got me thinking,
Same `shared_ptr<data_batch>` can be selected for downgrade in multiple repositories. So there can be a downgrade race on the same `data_batch` from two different repositories, should we check before converting that the source space is still the same as the original source space that we intended to downgrade. I can see this messing up the downgrade predicate

src/include/data/convertible_gpu_pipeline_task.hpp
```diff
@@ -250,65 +249,61 @@ class convertible_gpu_pipeline_task_provider : public convertible_data_provider
...
 
-    for (const auto& batch : pipelineable->get_data_batches()) {
-      if (batch && batch->get_memory_space() == space &&
-          batch->get_state() == cucascade::batch_state::task_created) {
-        return true;
-      }
+    for (const auto& batch : operator_data->get_data_batches()) {
+      if (!batch) { continue; }
+      if (batch->get_state() != cucascade::batch_state::idle) { continue; }
+      auto ro = batch->to_read_only();
```
bwyogatama: So just to note here that between line 304 and 305, there can potentially be another thread acquiring a mutable lock, hence blocking `to_read_only()`.

When I take a look at who is the caller of `has_matching_batches`, I don't think it will cause deadlock or anything but it might cause performance degradation cause `mutable_pop_if` (the caller) is holding the lock for the task queue.

Which means that if `to_read_only` is blocking, the whole task queue is blocked.

bwyogatama: This is a very low chance thought so maybe can be ignored, but worth commenting

src/include/data/convertible_gpu_pipeline_task.hpp
```diff
@@ -79,65 +79,64 @@ class convertible_gpu_pipeline_task : public convertible_data {
...
+   * the task is destroyed -- this is expected during query teardown.
    */
   ~convertible_gpu_pipeline_task() override
   {
-    if (_task) {
-      if (!_queue.push(std::move(_task))) {
-        SIRIUS_LOG_WARN("convertible_gpu_pipeline_task: queue interrupted, task destroyed");
-      }
-    }
+    if (_task) { _queue.push(std::move(_task)); }
```
bwyogatama, line 87: This is not really related to this PR, but if we fail to convert `convertible_gpu_pipeline_task` then we just push it back into the queue. After that it will go back to the beginning of the loop in the downgrade executor and the same task will be popped into a `candidate` again? And we just continuously doing that? Just want to make sure that this `livelock` behavior is intentional, not that I have a better solution for this

bwyogatama (approved):

src/downgrade/downgrade_executor.cpp
wmalpica: If there is a race condition, then this convert will be a no-op and it will return nothing, which means, it will know if it did not free anything and the predicate will still work. That being said, i realize that the converter function internally then has to check. The reason i would rather do it this way, is that for the conversion of a task which may have more than one batch, then checking if its the same space as the source space is not a yes or no, it can be partial.

src/include/data/convertible_gpu_pipeline_task.hpp
wmalpica: Good catch. I am making it a try_to_read_only

src/include/data/convertible_gpu_pipeline_task.hpp
wmalpica: This is a good point. I am going to add a counter in the downgrade_executor to only attempt to downgrade a max number of tasks equal to the size of the task queue at the beginning to prevent an infinite loop

PR #693, opened 2026-04-24
Refactor legacy source cleanup

mbrobbel:
## Summary
Follow-up to #692.

- move legacy-only CPU cache, buffer manager, fallback, print, and offset helper headers/sources under the legacy tree
- make ENABLE_LEGACY_SIRIUS=OFF publish empty legacy source/include/test lists
- remove stale legacy includes from the current execution path and gate legacy-only test helpers

## Verification
- normal release preset regenerates without Sirius src/legacy object targets or include/legacy compile paths
- legacy-release preset regenerates with legacy objects, include/legacy, SIRIUS_ENABLE_LEGACY, and test_cpu_cache
- targeted normal and legacy utility/test/legacy objects compile

Full make is still blocked by the existing cucascade API mismatch in src/downgrade/downgrade_executor.cpp:170: get_repositories() no longer exists.


9prady9 (approved):

joosthooz (approved):
Great, this speeds up the build a lot!

PR #705, opened 2026-04-24
Restore stable and nightly RAPIDS compatibility

mbrobbel:
It seems #681 was affected by https://www.githubstatus.com/incidents/zsg1lk7w13cf so I'm going to try figure out what happened.


mbrobbel:
Looks like this merged cleanly after all.

PR #706, opened 2026-04-24
Refactor parquet column indexing chaos

kevkrist:
# Consolidate scan bookkeeping into `scan_plan`

## Summary

Two correctness fixes in the parquet scan path, plus a refactor that collapses
nine scattered member fields and three helper-function duplications into a single
`scan_plan` struct. The refactor also fixes a latent bug that had the AST filter
pushdown commented out.
- **Refactor:** New `scan_plan` owns the canonical representation of what the
  scan reads, how it assembles output, and how filters map from column_ids
  space to batch space.

---

## The mental model

Three index spaces appear throughout the scan pipeline. They were tangled
together in the old code; naming them explicitly was the key to the refactor.

```
 Parquet file         DuckDB schema           column_ids list           Data batch
 (on disk)            (full table cols)       (this scan's cols)        (reader output)

  year=2024/               P=0  id               C=0  id                 D=0  id
    month=01/              P=1  name             C=1  name               D=1  name
     data.parquet          P=2  amount  ─────►   C=2  amount  ─────►     D=2  amount
      id | name | amount   P=3  year             C=3  year  (hive)       (year removed)
      ...                  P=4  month            C=4  month (hive)       (month removed)

    P = primary index (DuckDB schema position)
    C = column_ids position (scan-local)
    D = batch position (reader output, post-hive-removal)
```

Hive partitions live in P- and C-space but **never in D-space** — they're
synthesized from the file path post-read.

The `BoundReferenceExpression.index` that the AST translator passes to its
resolver is in **D-space**. The old `_column_name_by_ref` was indexed by C.
These coincided by accident whenever `projection_ids` was dense.

---

### Before: nine fields, three helpers, drift-prone

The metadata scan operator carried parallel views of the same question:

```cpp
std::vector<std::size_t>        _selected_column_indices;
bool                            _is_projected;
std::vector<std::string>        _projected_column_names;
bool                            _has_filter;
std::vector<std::string>        _column_name_by_ref;          // C-indexed, latently wrong
std::vector<std::size_t>        _post_filter_projection_ids;
std::unordered_set<std::size_t> _pure_filter_column_indices;
std::unordered_set<std::size_t> _hive_partition_index_set;
std::vector<hive_partition_column> _hive_partition_columns;
```

Plus `make_selected_column_indices`, `build_batch_column_map`, and
`build_partition_inject_fn`, each independently re-deriving overlapping subsets.

### After: one plan, built once

```cpp
struct scan_plan {
  struct data_column      { size_t primary_idx; std::string name; };
  struct partition_column { size_t primary_idx; std::string name; logical_type type; };
  struct output_entry     { enum source_t { DATA, PARTITION } source; size_t idx; };

  std::vector<data_column>      data_columns;       // D order
  std::vector<partition_column> partition_columns;  // injection metadata
  std::vector<output_entry>     output_layout;      // one entry per output col, in DuckDB order

  std::vector<std::optional<size_t>> batch_position_by_column_id;  // C -> D
  std::unordered_set<size_t>         partition_primary_indices;   // for filter-skip
};
```

`build_scan_plan` walks `projection_ids` once and populates all of it
consistently.

### One closure replaces three steps

The old GPU scan operator did *apply filter -> prune post_filter_projection_ids -> inject partitions*.
The new `build_inject_fn` closure walks `output_layout` exactly once:

```
output_layout:  [DATA(0)]  [DATA(1)]  [PARTITION(0)]
                   |          |            |
                   v          v            v
post-read batch: [id]      [name]     (make_column_from_scalar("2024"))
                   |          |            |
                   +----------+------------+
                              |
                              v
                         final output:  id | name | year
```

- `DATA(k)` entries `std::move` from the batch at position `k`.
- `PARTITION(k)` entries synthesize a scalar-backed column from the file path.
- Unused `data_cols[k]` (pure-filter columns read but not output) fall out of
  scope and free.

Pure-filter pruning and hive injection used to be two independent steps with a
subtle ordering dependency. Now they're the same pass.

### Identity short-circuit

For `SELECT *` (no partitions, no pure-filter cols) the closure would be a
pure-overhead move loop. Worse, for `SELECT count(*)` both `data_columns` and
`output_layout` are empty and a naive closure produces a 0-column table — which
regresses the count aggregation to 0.

`build_inject_fn` detects the identity case and returns `nullptr` so the GPU
scan operator's existing `if (_hive_partition_inject_fn)` guard skips assembly
entirely and the reader's output flows through untouched.

### AST resolver bug — fixed by construction

The old resolver:

```cpp
auto name_resolver = [this](duckdb::idx_t ref_index) {
  return _column_name_by_ref.at(ref_index);  // indexed by C
};
```

`ref_index` from `BoundReferenceExpression.index` is D-space, set in
`convert_table_filters_to_expression` via `batch_column_map[column_index]`.
When projection reorders or drops columns, C != D and the resolver returned the
wrong name (or OOB on the partition-filter case). Works only by coincidence
when `projection_ids` is the identity.

New resolver:

```cpp
auto name_resolver = [this](duckdb::idx_t ref_index) {
  return _plan.batch_column_name(ref_index);  // D-indexed, by construction
};
```

`batch_column_name(D)` is just `data_columns.at(D).name`. Single source of
truth for the D->name mapping, so nothing can drift.

### Operator shrinks

Metadata scan operator constructor, 85 lines of interleaved hive-removal /
filter-building / projection-building logic, is now:

```cpp
_plan = build_scan_plan(column_ids, projection_ids, names, returned_types,
                        this->types.size(), partition_indices);

if (auto inject_fn = _plan.build_inject_fn()) {
  _gpu_scan->set_hive_partition_inject_fn(std::move(inject_fn));
}

if (table_filter_set && !table_filter_set->filters.empty()) {
  auto batch_column_map  = _plan.make_batch_column_map();
  auto duckdb_expression = convert_table_filters_to_expression(
    *table_filter_set, column_ids, returned_types, batch_column_map,
    _plan.partition_primary_indices);
  if (duckdb_expression) _duckdb_filter_expression = std::move(duckdb_expression);
}
```

Three clauses, each doing one thing, no load-bearing ordering between them.


src/op/scan/scan_plan.cpp
```diff
@@ -0,0 +1,248 @@
...
+//===----------------------------------------------------------------------===//
+// scan_plan accessors
+//===----------------------------------------------------------------------===//
+
+bool scan_plan::is_projected() const
+{
+  // The scan is "projected" — the reader needs to be told which columns to
+  // read — whenever the set of data columns is not a plain read-everything.
+  // Callers use this to decide whether to call reader_options.set_column_names.
+  return !data_columns.empty();
```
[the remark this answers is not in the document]
kevkrist: Compromising with a single flag set in the factory:

  needs_reader_projection = !projection_ids.empty() || !partition_indices.empty()

This is O(1) instead of three loops.

src/op/scan/scan_utils.cpp
```diff
@@ -69,34 +69,22 @@ duckdb::unique_ptr<duckdb::Expression> convert_table_filters_to_expression(
...
-    }
+    auto primary_idx = column_ids.at(column_index).GetPrimaryIndex();
     if (skip_primary_indices.count(primary_idx)) {
       SIRIUS_LOG_DEBUG(
         "TABLE_SCAN filter: skipping filter on primary_idx={} (hive partition or equivalent)",
         primary_idx);
       continue;
     }
-    auto col_type = returned_types[primary_idx];
+    auto const col_type = returned_types.at(primary_idx);
```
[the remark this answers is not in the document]
kevkrist: .at() declutters the code and is preferred.

src/op/scan/sirius_parquet_metadata_scan_operator.cpp
```diff
@@ -53,95 +54,45 @@ sirius_parquet_metadata_scan_operator::sirius_parquet_metadata_scan_operator(
...
-        }
-      }
-    }
+  // Projection and filter pushdown both rely on column names to drive the reader;
+  // refuse early if we'd be forced into name-based paths without them.
+  if ((!projection_ids.empty() || (table_filter_set && !table_filter_set->filters.empty())) &&
+      names.empty()) {
+    throw sirius::internal_exception(
+      "[sirius_parquet_metadata_scan_operator] Projection or filter pushdown requires column "
+      "names to be provided.");
```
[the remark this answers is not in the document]
kevkrist: No -- we need names to correctly resolve column indexing in `detail::leaf_indices_for_column()`

kevkrist (commented):
Adopting some recommended changes.

src/include/op/scan/scan_plan.hpp
```diff
@@ -0,0 +1,187 @@
...
+/// @param returned_types      DuckDB schema column types.
+/// @param output_types_size   The scan operator's types.size() — used to split
+///                            projection_ids into output vs pure-filter.
+/// @param partition_indices   Hive partition indices advertised by the bind data.
+scan_plan build_scan_plan(duckdb::vector<duckdb::ColumnIndex> const& column_ids,
+                          duckdb::vector<duckdb::idx_t> const& projection_ids,
+                          duckdb::vector<std::string> const& names,
+                          duckdb::vector<sirius::logical_type> const& returned_types,
+                          std::size_t output_types_size,
+                          duckdb::vector<duckdb::HivePartitioningIndex> const& partition_indices);
```
ducndh, line 187: Hive partitioning might not be the only type of partition indices we need to be care about. Iceberg and some other formats have partition indices that this will ignore. Can you add a TODO so I can circle back after reviewing how other library do this and implement this class more generally?

src/include/op/scan/scan_plan.hpp
```diff
@@ -0,0 +1,187 @@
...
+ * Three index spaces appear throughout the scan pipeline:
+ *   P = primary index       (position in DuckDB's full schema @c names / @c returned_types)
+ *   C = column_ids position (position in the operator's @c column_ids list)
+ *   D = batch position      (position in the reader's output table, after hive removal)
+ *
+ * Hive-partition columns live in P-space but are not in the parquet file, so
+ * they never appear in D-space. They are injected post-read into the final
+ * output in the order column_ids expects.
+ */
+struct scan_plan {
```
ducndh, line 59: ON #574 there is a special type of parquet files that is need for schema-evolution tracking. `scan_plan` treats every column as either "in `partition_indices`" or "data column." For files with heterogeneous schemas (`{a, b}` vs `{a, b, c}`), the reader gets told to read `c` from a file that doesn't have it and dies at concat. PR #574 fixed this on the legacy `parquet_scan_task` path (`_per_file_column_names` + per-file projection). New path will need the equivalent. Suggestion: add `per_file_data_columns: vector<vector<size_t>>`, populated only when files differ; `build_inject_fn` closes over it.

src/op/scan/scan_plan.cpp
```diff
@@ -0,0 +1,270 @@
...
+{
+  std::vector<std::string> names;
+  names.reserve(data_columns.size());
+  for (auto const& c : data_columns) {
+    names.push_back(c.name);
+  }
+  return names;
+}
+
+std::vector<duckdb::idx_t> scan_plan::make_batch_column_map() const
```
ducndh, line 60: If yuor style permtis, a comment saying that `batch_position_by_column_id` is already in the  cleaner `vector<optional<size_t>>` shape and a future cleanup of the filter-translator API should drop the sentinel might be hlepful here

ducndh:
Great change the logic should stay together. The existing problem is mostly because of my design is still decouple too much iceberg 1-to-1 to parquet. I will look at the parquet sacn task struct on DuckDB and Velox to come up with a cleaner general design to make parquet scan task more extensible and less decoupling

kevkrist:
See [comment here](https://github.com/sirius-db/sirius/pull/663#pullrequestreview-4180588795).

kevkrist:
> Great change the logic should stay together. The existing problem is mostly because of my design is still decouple too much iceberg 1-to-1 to parquet. I will look at the parquet sacn task struct on DuckDB and Velox to come up with a cleaner general design to make parquet scan task more extensible and less decoupling

I think this should be merged and you can refactor as you need.

ducndh:
I am just saying that this might have a conflict with the current ice berg scan task, but you can merge this, and I will fix the iceberg path on top of this

kevkrist:
Thanks Duc. Can you approve?
________________________________

ducndh (approved):

src/include/op/scan/scan_plan.hpp
kevkrist: This seems to be mostly an iceberg use case. I have deliberately deferred iceberg integration into the new scan path because it magnifies this change and has been actively changing. I've opened issue #717 to track this integration.

kevkrist (commented):
Thanks Duc!

PR #710, opened 2026-04-26
scan: GPU-native decode kernel (Part 1 — numeric codecs)

ducndh:
First of three split PRs for the decode kernels in feature/gpu-native-scan-task. 

What's here:                                                                                                                                                                         
- `pinned_bounce.cuh`: 2-slot per-thread pinned ring for pageable→device H2D                                                                                                        - `device_scratch.cuh`: bump-allocated TLS device arena (zero allocator calls steady-state)                                                                                  
- `gpu_decode_bitpacking.cu`: batched decoder, one CTA per 2048-row group one launch per column (was N per-segment launches)                                                                                                                                 
- `gpu_native_decode.cu`: dispatcher (BITPACKING / UNCOMPRESSED / fixed-width); throws `viability invariant violated` on anything which is a defensive assert against the upstream check_viability out of - `direct_block_scan.hpp` (types only — `.cpp` lands in later PR
- 5 unit tests / 779 assertions for the bitpacking decoder                                                                                                                           

What's not here yet:                                                                                                                                                                     
- String codecs (dict / FSST / DICT_FSST / uncompressed-string) → next PR
- RLE → next PR                                                                                                                       
- ALP / ALPRD floats → next of next kernel PR
- direct_block_scan implementation, gpu_native_scan_task → the PR after all the kernel PR
                                                                                                                                                                                                                                                                                          
Tests: `sirius_unittest "[scan]"` → 11/11 (6 baseline + 5 new), 1092 asserts.


src/cuda/scan/gpu_native_decode.cu
```diff
@@ -0,0 +1,665 @@
...
+    case duckdb::LogicalTypeId::UINTEGER: return cudf::data_type(cudf::type_id::UINT32);
+    case duckdb::LogicalTypeId::UBIGINT: return cudf::data_type(cudf::type_id::UINT64);
+    case duckdb::LogicalTypeId::FLOAT: return cudf::data_type(cudf::type_id::FLOAT32);
+    case duckdb::LogicalTypeId::DOUBLE: return cudf::data_type(cudf::type_id::FLOAT64);
+    case duckdb::LogicalTypeId::BOOLEAN: return cudf::data_type(cudf::type_id::BOOL8);
+    case duckdb::LogicalTypeId::DATE: return cudf::data_type(cudf::type_id::TIMESTAMP_DAYS);
+    case duckdb::LogicalTypeId::TIMESTAMP:
+      return cudf::data_type(cudf::type_id::TIMESTAMP_MICROSECONDS);
+    case duckdb::LogicalTypeId::VARCHAR: return cudf::data_type(cudf::type_id::STRING);
+    case duckdb::LogicalTypeId::HUGEINT: return cudf::data_type(cudf::type_id::INT64);
```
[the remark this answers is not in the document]
ducndh: Throw here might be a safer way but we might need to revisit type system later

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,275 @@
...
+  const uint8_t* block_base = desc.d_block + desc.block_offset;
+  T* out                    = d_output + desc.global_row_offset;
+
+  // Shared metadata — written by thread 0, read by all after barrier
+  __shared__ uint8_t sm_mode;
+  __shared__ uint32_t sm_data_offset;
+  __shared__ uint32_t sm_width;
+  __shared__ uint32_t sm_row_count;
+  __shared__ T sm_frame;
+  __shared__ T sm_aux;
```
felipeblazing, line 82: add a comment defining what these all do. sm_aux is not clear to me without further investigation for example

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,275 @@
...
+                                 void* d_output,
+                                 rmm::cuda_stream_view stream)
+{
+  if (num_segments == 0) return;
+
+  // Upload descriptors to device via thread-local scratch arena. After the
+  // first few batches the arena is sized to hold this and we skip
+  // malloc/free entirely.
+  size_t desc_bytes = num_segments * sizeof(batched_bp_seg_desc);
+  auto desc_alloc   = arena_allocate(desc_bytes, stream.value());
```
felipeblazing: We have a fixed size allocator that hyu can use to get blocks of 1MB you should use that an dsuballocate from that if the scratch arena is smaller, this is untracked memory otherwise.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,275 @@
...
+  auto* d_descs     = static_cast<batched_bp_seg_desc*>(desc_alloc.ptr);
+  bounce_h2d_async(d_descs, descs, desc_bytes, stream.value());
```
felipeblazing, line 310: why do we need a wrapper function when asynchronous apis already exist?

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,275 @@
...
+
+  constexpr uint32_t BLOCK_DIM = 256;
+  constexpr uint32_t max_width = sizeof(T) * 8;
```
felipeblazing: where is this number coming from

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,275 @@
...
+  // +1 guard word for 3-word unpack_value reads on 64-bit types.
+  constexpr uint32_t max_packed_words = (BP_META_GROUP_SIZE * max_width + 31) / 32 + 1;
+  constexpr size_t shmem_bytes        = max_packed_words * sizeof(uint32_t);
+
+  kernel_decode_bitpacking_batched<T><<<num_segments, BLOCK_DIM, shmem_bytes, stream.value()>>>(
```
felipeblazing, line 336: how are we guaranteeing that shmem_bytes is within the gpus capacity?  How did we pick block_size here?

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,275 @@
...
+    d_descs, static_cast<T*>(d_output), num_segments);
+
+  if (desc_alloc.needs_free) { cudaFreeAsync(d_descs, stream.value()); }
+}
+
+void gpu_decode_bitpacking_batched(const batched_bp_seg_desc* descs,
```
felipeblazing, line 340: cant there be bigger types like in the case of 128bit decimals?

src/cuda/scan/gpu_native_decode.cu
```diff
@@ -0,0 +1,618 @@
...
+using sirius::op::scan::column_scan_result;
+
+//===----------------------------------------------------------------------===//
+// Helpers
+//===----------------------------------------------------------------------===//
+
+namespace {
+
+/// Map DuckDB LogicalType to cudf data_type.
+cudf::data_type to_cudf_type(const duckdb::LogicalType& type)
```
felipeblazing, line 76: These are already defined elsewhere in the codebase

src/cuda/scan/gpu_native_decode.cu
```diff
@@ -0,0 +1,618 @@
...
+    default: break;
+  }
+  // Defensive assert: PR E's check_viability owns keeping unsupported types
+  // out of this dispatcher.  Reaching here means viability fell out of sync.
+  throw std::runtime_error(
+    "gpu_native_decode: viability invariant violated — unsupported DuckDB type " + type.ToString());
+}
+
+/// Get byte size of a DuckDB physical type.
+uint32_t get_type_size(duckdb::PhysicalType pt)
```
felipeblazing: I am pretty sure we also have utility functions for this

src/cuda/scan/gpu_native_decode.cu
```diff
@@ -0,0 +1,618 @@
...
+    case duckdb::PhysicalType::FLOAT: return 4;
+    case duckdb::PhysicalType::INT64:
+    case duckdb::PhysicalType::UINT64:
+    case duckdb::PhysicalType::DOUBLE: return 8;
+    case duckdb::PhysicalType::INT128: return 16;
+    default: return 0;
+  }
+}
+
+bool is_signed_type(duckdb::PhysicalType pt)
```
felipeblazing: this too if not it should be put where these utility functions are not defined in a .cu file

src/cuda/scan/gpu_native_decode.cu
```diff
@@ -0,0 +1,618 @@
...
+                          stream.value());
+        } else {
+          // Block wasn't pre-staged (block_id < 0 etc.) — per-segment H2D.
+          bounce_h2d_async(d_dest, seg.data_ptr, seg_bytes, stream.value());
+        }
+        break;
+      }
+      case duckdb::CompressionType::COMPRESSION_CONSTANT: {
+        const uint8_t* vs = seg.data_ptr ? seg.data_ptr : seg.constant_data;
+        launch_fill_constant(d_dest, vs, type_size, seg.row_count, stream.value());
```
felipeblazing: why doesnt launch_fill_constant use cudf::scalar then you can type dispatch to the specific type of scalar  to get the right value before you launch the kernel or pass in the device pointer from the kernel and just dereference it. 

https://docs.rapids.ai/api/libcudf/stable/classcudf_1_1scalar#details

look at fixed_width_scalar and numeric_scalar

felipeblazing (changes requested):
Still reviewing but wanted to get what I had so far in

src/cuda/scan/gpu_decode_bitpacking.cu
ducndh: Yah add a per-mode comment block above the __shared__ decls listing what sm_aux means in each of CONSTANT/CONSTANT_DELTA/FOR/DELTA_FOR.

src/cuda/scan/gpu_decode_bitpacking.cu
ducndh: Yah it was small then but it was careless of me. replaced the arena alloc with rmm::device_buffer(desc_bytes, stream, mr)

src/cuda/scan/gpu_decode_bitpacking.cu
ducndh: Pageable→device cudaMemcpyAsync silently goes synchronous via the driver's pinned bounce (~450 µs/call on A100); the wrapper user-space-bounces through TLS pinned slots so the DMA hits the async fast path. Could also be possible that there is more elegant solution.

src/cuda/scan/gpu_decode_bitpacking.cu
ducndh: BLOCK_DIM=256 matches the per-segment kernel's occupancy on SM 75/80 but yah could make this a knob

src/cuda/scan/gpu_decode_bitpacking.cu
ducndh: Yah should throws here great catch

src/cuda/scan/gpu_native_decode.cu
ducndh: Yah change to re use now.

src/cuda/scan/gpu_native_decode.cu
ducndh: Yah at the point of implentation for ulti I was heavily relied on Claude should have check the code base more carefully

src/cuda/scan/gpu_native_decode.cu
ducndh: Yah change to DuckDB function

src/cuda/scan/gpu_native_decode.cu
ducndh: Yah change to stay with RAPIDs convention

src/cuda/scan/gpu_decode_bitpacking.cu
ducndh: I think this one shold be safe since it peaks at 16kB and I think most GPU can accomodate that but will add an assert to make sure it does not break silently

src/cuda/scan/gpu_decode_bitpacking.cu
felipeblazing: and are you using the mr which corresponds to the memory space you are currently operating on?

src/cuda/scan/gpu_decode_bitpacking.cu
felipeblazing: why are we using pageable memory? Can't we use the fixed size allocator that we default to for host?

src/cuda/scan/gpu_decode_bitpacking.cu
ducndh: Yes — `mr` is `rmm::device_async_resource_ref`; descriptor goes via `rmm::device_buffer(desc_bytes, stream, mr)` so it shares whatever device pool the caller threads through. Same `mr` is used for `data_buf` and `null_mask` in `gpu_native_decode.cu:353,286,461`, so the descriptor lives on the same reservation arena as the column output it's decoding into.

src/cuda/scan/gpu_decode_bitpacking.cu
ducndh: The pageable bit is the caller's std::vector<batched_bp_seg_desc> — built on the C++ heap at the descriptor build site. Cleanest fix is for that build site to materialize directly into fixed_size_host_memory_resource blocks (The descriptor build site lives in PR #714); then the wrapper drops away, and we call cudaMemcpyAsync straight, with both ends tracked.

However, PR #714 also has an untracked cudaMallocAsync for fallback block staging that should swap to rmm::device_buffer(..., stream, mr) at the same time. 

Thanks for pointing these out. I will scourge that PR for unaccounted memory too.

ducndh:
I was looking through resolving the comments on memory accounting, so I took the chance to do some kernel optimizing: switched FOR / DELTA_FOR global stores from blocked to striped layout. Each warp now writes one cache line per iteration instead of eight and routes every store through __stwt to bypass L1 on writes that the kernel never rereads.

PR #713, opened 2026-04-26
Fix row-group pruning in parquet metadata scan and memory reservation for parquet gpu scan

kevkrist:
### Bug fix 1: AST filter scalar lifetime (use-after-free)

Fix `cudaErrorIllegalAddress` crash caused by failure to enforce stream synchronization between pipeline streams (metadata scan -> gpu scan).

**Fix:** Add `stream.synchronize()` after AST filter translation to establish a
full memory barrier. Without this, downstream `GPU_PARQUET_SCAN` tasks access the
scalar device memory on a different CUDA stream without a dependency on the
producing stream. CUDA's async memory allocator uses stream-ordering semantics, so
the allocations may appear invalid from the consuming stream's perspective.

- Re-enabled AST filter translation (previously commented out with "bug hiding here")
- Added `stream.synchronize()` after `translate_expression_with_names()`
- Validated: TPC-H SF50 all 22 queries × 5 iterations pass without `cudaErrorIllegalAddress`

### Bug fix 2: Zero-byte reservations for GPU scan tasks

`GPU_PARQUET_SCAN` tasks were receiving 0-byte memory reservations because
`get_task_consumption_basis()` only recognized `pipelineable_operator_data` inputs
(GPU data batches). Scan tasks carry `parquet_scan_data` (file paths and metadata),
so their input size was always 0 — blinding the reservation system to their actual
GPU memory needs.

**Fix:** Add a virtual `get_estimated_size_in_bytes()` to the `operator_data` base
class so the reservation system can query any input data for its GPU memory
footprint without downcasting.

- `operator_data` base returns 0 (metadata-only subclasses)
- `pipelineable_operator_data` sums its data batches' uncompressed sizes
- `parquet_scan_data` returns `rg_range.reserved_uncompressed_bytes`
- `gpu_pipeline_task_local_state::get_task_consumption_basis()` simplified from
  a manual `dynamic_cast` + loop to a single `_input_data->get_estimated_size_in_bytes()` call

### Refactor: Decouple GPU scan completion detection

Replace the bespoke `_finalized` / `finalize_partitions()` signaling between
the metadata scan and GPU scan operators with the standard port-based
`is_pipeline_finished()` check that every other operator already uses.

**Before:** The metadata scan operator held a raw pointer to the GPU scan and
called `finalize_partitions()` from its `finalize_operator()` override, setting
a `_finalized` bool to tell `get_next_task_hint()` when to stop waiting.

**After:** `get_next_task_hint()` checks the existing "handoff" port's
`src_pipeline->is_pipeline_finished()` under a single lock. No cross-operator
state mutation needed.

- Removed `finalize_partitions()`, `_finalized` from `sirius_gpu_parquet_scan_operator`
- Removed `finalize_operator()` override from `sirius_parquet_metadata_scan_operator`
- Rewrote `get_next_task_hint()` to use port-based completion detection
- Updated pipeline converter comments, operator docstrings, and tests

### DIFF
--- a/src/op/scan/sirius_gpu_parquet_scan_operator.cpp
+++ b/src/op/scan/sirius_gpu_parquet_scan_operator.cpp
@@ -61,37 +61,29 @@ void sirius_gpu_parquet_scan_operator::accumulate_metadata(
   }
 }
 
-void sirius_gpu_parquet_scan_operator::finalize_partitions()
-{
-  std::lock_guard<std::mutex> lock(_metadata_mutex);
-  _finalized = true;
-}
-
 //===----------------------------------------------------------------------===//
 // Scheduling interface
 //===----------------------------------------------------------------------===//
 std::optional<task_creation_hint> sirius_gpu_parquet_scan_operator::get_next_task_hint()
 {
-  {
-    std::lock_guard<std::mutex> lock(_metadata_mutex);
+  std::lock_guard<std::mutex> lock(_metadata_mutex);
 
-    // 1. Work available? Dispatch immediately, even if metadata pipeline
-    //    is still producing.
-    if (_next_partition_idx < _partition_index.size()) {
-      return task_creation_hint{TaskCreationHint::READY, this};
-    }
-
-    // 2. No work and metadata pipeline is done — we are finished.
-    if (_finalized) { return std::nullopt; }
+  // 1. Work available? Dispatch immediately, even if metadata pipeline
+  //    is still producing.
+  if (_next_partition_idx < _partition_index.size()) {
+    return task_creation_hint{TaskCreationHint::READY, this};
   }
 
-  // 3. Metadata pipeline is still running — defer to it.
+  // 2. Metadata pipeline still running? Wait on it.
   auto it = ports.find("handoff");
-  if (it != ports.end() && it->second && it->second->src_pipeline) {
+  if (it != ports.end() && it->second && it->second->src_pipeline &&
+      !it->second->src_pipeline->is_pipeline_finished()) {
     if (auto upstream = it->second->src_pipeline->get_source()) {
       return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, upstream.get()};
     }
   }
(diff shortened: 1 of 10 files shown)


src/op/scan/sirius_parquet_metadata_scan_operator.cpp
```diff
@@ -200,35 +200,40 @@ std::unique_ptr<operator_data> sirius_parquet_metadata_scan_operator::execute(
...
+      // different stream may hit cudaErrorIllegalAddress due to CUDA stream-ordering
+      // semantics in the async memory allocator.
+      gpu_expression_translator translator(stream, cudf::get_current_device_resource_ref());
+      auto name_resolver = [this](duckdb::idx_t ref_index) -> std::string {
+        return _column_name_by_ref.at(ref_index);
+      };
+      auto optional_filter =
+        translator.translate_expression_with_names(*_duckdb_filter_expression, name_resolver);
+      stream.synchronize();
+      if (optional_filter) {
```
9prady9: we can move the stream sync inside the if guard right ?

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
9prady9, line 1: this PR's diff doesn't handle it, but when I also asked model to review this - it correctly pointed that `get_next_task_input_data` function increments under _metadata_mutex, then reads _partition_index[idx] after releasing it. Meanwhile accumulate_metadata() can append and reallocate the vector - can you please check this since you already fixed couple of bugs :)

9prady9 (commented):
nice catch, have left a comment for sync and one request, please have a look.

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
kevkrist: Yes good catch! Now I grab the entry in the lock.

kevkrist (commented):
Thanks for this!

aminaramoon (approved):

PR #714, opened 2026-04-26
scan: final connector piece for DuckDB native scan task on GPU

ducndh:
## Overview

Adds the DuckDB GPU-native scan task that lets `gpu_execution` read DuckDB-tables directly on the GPU instead of pulling rows through DuckDB's CPU table-function API.                                                                                                                                                                                                       

## PR #651 split explanation:                                                                                                                                                                                    

PR #651 was a 5000-LOC feature branch (`feature/gpu-native-scan-task`) that shipped GPU-native scan as one giant blob. It's being split into reviewable chunks:                                                                                                                                                                                                                   
                                                         
#709 — CPU-side scan fixes (lifecycle, projection)                                                                                                                                                                      
#710 — GPU decode foundation + numeric codecs        
#711 — string / RLE / ALP / ALPRD decoders + dispatcher                                                                                                                                                                 
#712 — direct block scan walker (data path + mmap)                                                                                                                                                                      
THIS — integrates #710 + #711 + #712 into a working scan 
later — sirius_io migration (depends on #675)                                                                                                                                                                           

This PR is the final connector: #710 + #711 produced the GPU kernels, #712 the segment walker, but neither was actually wired into `gpu_execution`. This PR adds `duckdb_native_scan_task` (the operator that owns the lifecycle), routes `SELECT ... FROM duckdb_table` through it when viable, and falls back to the CPU `duckdb_scan_task` cleanly when it isn't.                                                                                                                                                       

In total, we walk segment trees → pin blocks → bulk H2D → GPU decode → emit cudf::table. No `init_global`, no per-row materialization through DuckDB's TableFunction.                                                                                                                                                                                                   
                                                                                                                                                                                                                            Bypassing `init_global` means we lose the implicit handling DuckDB does for sampling, virtual columns, dynamic join filters, etc. Two checks refuse the fast path when we can't handle something:                                                                                                                                                                                 
1. Operator-level escape gate (`task_creator`): refuse on `dynamic_filters`, `extra_info.sample_options`, or any non-rowid virtual column.
2. Per-segment compression check (`walk_duckdb_native_metadata`): refuseunsupported data / validity codecs.                                                                                                                                                                                  
When either fires, we fall back to `duckdb_scan_task` to avoid silent error.

This PR also contains rowid columns (late_materialization rewrite of LIMIT/TOP_N) that are *implemented* over #651, not refused: the dispatcher emits a synthetic BIGINT column via `thrust::sequence` per row group.                                                                                                                                                                                         
                                                                                                                                                                                                                            ## Merge plan                                                                                                                                                                                                             

This branch stacks on #712 (which will need to be changed after DuckDB gives us the API and we update DuckDB version) and on #711 (which contains #710's decoders). Suggested merge order, in priority:                                                                                                                                                                                       
1. #710 lands → #711 rebases onto upstream/dev.                                                                                                                                                                         2. #711 lands → #712 rebases.                        
3. #712 lands → this PR rebases. Single linear history.                                                                                                                                                                 

(though #711 and #712 can be reviewed independently as this PR is just the connection between those twos)                                                                                                                   
                                                                                                                                                                                                                            
A future PR promotes `walk_duckdb_native_metadata` (the free function in this PR) into a real metadata-scan operator and hooks #675's prefetching cache. The Level-1 split header in commit #2 is prep for this — no runtime perf change today, just signature stability.

Optional follow-up: dynamic join filter consumption (~200 LOC), OPTIONAL TOP_N row-group reordering — both currently escape to CPU.                                                                                                                                                                  
                                                                                                                                                                                                                            
## Tests:
I add a few tests for ALP decode since existing TPCH don't have that type of compression. Other can accept the existing tests as equivalent end to end test of this task


ducndh:
when #714 rebases onto #712's new tip, conflict in direct_block_scan.cpp resolves to #712's stat-based mmap impl + #714's validity walk + ROARING.

joosthooz:
Hi @ducndh, if I understood the structure of the PRs correctly, this final PR contains all the changes of the preceding PRs (except the CPU scan fixes), so the diff will shrink as those others are being merged in.
What I did is check out this branch and build it on the GB300 just to try it out. On Q1 I'm getting OOM errors right from the start at sf100. Q13 shows `Error executing query: batched string decode: total_chars wrapped negative (column exceeds cuDF's INT32_MAX-byte string-column limit): -1636968161`.
For the rest of the queries, here's the results:
```
============================================================
  Results Summary  (SF100)
============================================================

Query   |   DuckDB Cold |   DuckDB Warm |   Sirius Cold |   Sirius Warm | Speedup (warm)
--------+---------------+---------------+---------------+---------------+---------------
Q2      |         0.16s |         0.12s |         0.30s |         0.16s |           .77x
Q3      |         0.58s |         0.33s |         0.61s |         0.39s |           .83x
Q4      |         0.37s |         0.23s |         0.43s |         0.30s |           .75x
Q5      |         0.60s |         0.60s |         0.56s |         0.51s |          1.16x
Q6      |         0.06s |         0.05s |         0.23s |         0.20s |           .23x
Q7      |         0.49s |         0.50s |         0.51s |         0.50s |           .99x
Q8      |         0.28s |         0.27s |         0.54s |         0.49s |           .54x
Q9      |         1.27s |         1.34s |         0.67s |         0.62s |          2.14x
Q10     |         0.63s |         0.61s |         1.18s |         0.53s |          1.13x
Q11     |         0.07s |         0.06s |         0.13s |         0.10s |           .62x
Q12     |         0.16s |         0.21s |         0.30s |         0.28s |           .74x
Q15     |         0.17s |         0.20s |         0.25s |         0.23s |           .84x
Q16     |         0.29s |         0.31s |         0.26s |         0.15s |          2.07x
Q17     |         0.20s |         0.25s |         0.49s |         0.49s |           .50x
Q18     |         1.43s |         1.41s |         0.74s |         0.73s |          1.91x
Q19     |         0.33s |         0.33s |         0.44s |         0.43s |           .76x
Q20     |         0.27s |         0.23s |         0.33s |         0.32s |           .69x
Q21     |         1.27s |         1.25s |         1.03s |         0.99s |          1.26x
Q22     |         0.21s |         0.20s |         0.13s |         0.13s |          1.51x
--------+---------------+---------------+---------------+---------------+---------------
TOTAL   |         8.84s |         8.48s |         9.16s |         7.59s |          1.11x
```

ducndh:
Thank you so much for your help. This is mostly for end-to-end testing and benchmarking. 

Though the bad news is that the memory errors can come from 3 different places: cuCascade has a memory accounting problem for cross-thread/stream (I have already sent the PR there); cudf 26.04, which we using right now instead of the version on the old PR, has unaccounted memory (which will be fixed in their later patch); I make mistake while splitting and porting the old code to dev. 

I will also look into trying to replicate the error and investigate the slower query to understand where the problem is, but I some of the fixes might be out of my hands, and I need advice on how to proceed.

PR #707, opened 2026-04-24
feat: introduce sirius::ast type hierarchy scaffold

9prady9:
Part of #666. Closes #694.

First atomic PR in the Phase 7b series. Pure additive — zero callers, zero behavior change, no DuckDB includes in the new public surface.

## What's added

- 12 new public headers under `src/include/expression/ast/`:
  - `node.hpp` — struct wrapping `std::variant<reference, constant, comparison, conjunction, between, case_expr, cast, unary_op, coalesce, in_list, function_call>` with `holds<T>()` / `get<T>()` accessors and a forwarding ctor.
  - 11 per-node headers, each with a forward `struct node;` and `std::unique_ptr<node>` recursive children.
- `test/cpp/expression/test_ast_scaffold.cpp` — Catch2 unit test (tag: `[ast_scaffold]`) covering instantiation of every node kind, recursive trees, and move-only `static_assert`s.
- One-line `CMakeLists.txt` addition to register the test.

## Why a struct wrapper, not a `using` alias

`std::variant<Ts...>` requires every `T` complete at instantiation. Per-node headers carry `std::unique_ptr<node>` recursive children that need to forward-declare `node`, and type aliases can't be forward-declared. Wrapping the variant in `struct node` breaks the cycle cleanly.

## Why the catch-all `operator_` was split

DuckDB's `BoundOperatorExpression` lumps seven SQL operators (NOT, IS NULL, IS NOT NULL, TRY, COALESCE, IN, NOT IN) into one node type whose `children` field means different things depending on the kind tag. Sirius's AST splits those by operand shape so the type system carries the contract:

- `unary_op` (kinds: `not_`, `is_null`, `is_not_null`, `try_`) — holds a single `unique_ptr<node> child`.
- `coalesce` — holds a `vector` of children; no kind tag needed.
- `in_list` — holds a probe, a vector of values, and a `negated` bool that replaces the `COMPARE_IN` / `COMPARE_NOT_IN` enum pair.

Same seven SQL semantics, reorganized so each node's operand structure is explicit.

## Acceptance

- [x] `grep -rn "duckdb/" src/include/expression/ast/` → 0 hits
- [x] `sirius_unittest "[ast_scaffold]"` passes
- [x] Full `sirius_unittest` suite passes (no regression)
- [x] Builds on default and `-DENABLE_LEGACY_SIRIUS=ON`
- [x] Pre-commit hooks

Issue #666, opened 2026-04-21
DuckDB Decoupling: phase 7b changes of issue 556


9prady9:
Sub-issue of #556. Follow-up to #665 (phase 7a).

## Description

Replace the `duckdb::Expression`-backed opaque wrapper from phase 7a with a Sirius-native expression IR — a proper AST type hierarchy under `sirius::`, mirroring the phase 6 precedent of `sirius::logical_type` / `sirius::type_id`.

- Plan generator becomes the sole translator from `duckdb::Expression` trees into the Sirius IR.
- GPU expression executor and all specializations dispatch on Sirius IR nodes directly — no `duckdb::BoundXxxExpression` references anywhere below the plan-builder boundary.
- Enables non-DuckDB frontends and removes the last semantic coupling to DuckDB in operator evaluation.

## Scope

Design-heavy, 50+ files. Rough breakdown:

- **New IR type hierarchy** under `src/include/expression/ast/` — base `sirius::ast::node` plus node types covering what the executor handles today: `reference`, `constant`, `comparison`, `conjunction`, `between`, `case`, `cast`, `operator_`, `function`.
- **Constant representation** — replace `duckdb::Value` payloads (used by `BoundConstantExpression`) with a Sirius-native typed variant keyed on `sirius::type_id`.
- **Function resolution** — `BoundFunctionExpression` today carries a DuckDB scalar-function descriptor and bind data. Design a Sirius-side function identifier (enum or registry handle) sufficient for the set of functions the executor supports.
- **Translator** (plan-generator side) — per-DuckDB-node conversion into the Sirius IR, plus validation / "unsupported → fallback" signalling.
- **Executor rewrite** — retarget the nine `src/expression_executor/specializations/gpu_execute_*.cpp` files and `gpu_expression_translator.cpp` to dispatch on Sirius IR.
- **Wrapper removal** — retire `sirius::expression` PIMPL from #665; operator headers store Sirius IR nodes directly.

## Prerequisite

#665 must land and stabilize first — the wrapper surface defines the exact boundary the native IR will replace.

PR #716, opened 2026-04-27
feat: sirius::function_id closed enum (Phase 3 of #666 — closes #696)

### LINKED ISSUE #696: Phase 7b PR 3: closed sirius::function_id enum replacing FUNC_STR macros
Sub-issue of #666. Depends on PR 1 (AST scaffold). Sibling of PR 2.

## Description

Replace the 27 `*_FUNC_STR` string macros in `src/expression_executor/specializations/gpu_execute_function.cpp` and the `supported_ast_functions` allowlist in `src/include/expression_executor/ast_supported_types.hpp` with a closed `sirius::function_id` enum.

## Scope

- `src/include/expression/function_id.hpp` — `enum class sirius::function_id : uint16_t { add, sub, mul, div, int_div, mod, substring, like, not_like, contains, prefix, suffix, strlen, length, regexp_replace, year, month, day, hour, minute, second, millisecond, microsecond, date_trunc, row, struct_pack, error };` (27 entries — exact set the executor supports today).
- `src/expression/function_id.cpp` — `std::optional<function_id> from_duckdb_function_name(std::string_view)` and `std::string_view to_duckdb_function_name(function_id)`.
- `supported_ast_functions` becomes `std::array<function_id, 6>{add, sub, mul, div, int_div, mod}` in `ast_supported_types.hpp`.
- `gpu_execute_function.cpp` converts `expr.function.name` to `function_id` at entry, then switches on the enum. All `*_FUNC_STR` macros deleted.

## Acceptance

- [ ] All `*_FUNC_STR` macros deleted from `gpu_execute_function.cpp`.
- [ ] `supported_ast_functions` is typed as `std::array<function_id, 6>`.
- [ ] `gpu_execute_function.cpp` and `gpu_execute_case.cpp` dispatch on the enum (single string→enum conversion at entry).
- [ ] `sirius_unittest` passes (behavior byte-equivalent).
- [ ] `tpch-sirius.test` passes on release.
- [ ] Builds on both `ENABLE_LEGACY_SIRIUS=OFF` and `ON`.
- [ ] No TPC-H SF1 regression.

## Requirements traced
- REQ-AST-03

9prady9:
Closes [#696](https://github.com/sirius-db/sirius/issues/696). Phase 3 of the 11-phase expression-framework rework tracked in [#666](https://github.com/sirius-db/sirius/issues/666).

**Stacked on [#715](https://github.com/sirius-db/sirius/pull/715) (Phase 2 — sirius_value), which is stacked on [#707](https://github.com/sirius-db/sirius/pull/707) (Phase 1 — sirius_ast_scaffold).** Until those merge, the Files Changed view includes all three PRs' changes.

## What this lands

Introduces `sirius::function_id`, a closed `enum class : uint16_t` with 27 entries covering arithmetic (6), string (9), temporal (9), and struct/control (3) functions. Cardinality is locked at compile time.

Two free functions provide bidirectional name mapping between DuckDB's string-based function names and `function_id`:
- `from_duckdb_function_name` — returns `std::nullopt` for unrecognised names; both `"substring"` and `"substr"` resolve to `function_id::substring`.
- `to_duckdb_function_name` — returns the canonical DuckDB spelling via a direct index into a reverse table (no search).

A compile-time consistency check (`consteval` + `static_assert`) validates that the forward and reverse tables agree on exactly one canonical name per enum value. Adding a new `function_id` entry without updating both tables is a build error.

The GPU executor's function dispatch is switched from 28 string-comparison macros to a single `from_duckdb_function_name` call at entry, followed by enum equality checks. The `supported_ast_functions` allowlist also switches from `string_view` to `function_id`.

31 Catch2 tests cover round-trips for all 27 entries, the `substr`/`substring` alias, and unknown-name `nullopt` cases.

## Scope locks honoured

- Legacy `gpu_processing` path (`src/legacy/`) — byte-unchanged.
- `gpu_execute_case.cpp` — byte-unchanged (`error` function migration deferred to Phase 6).

## Test plan

- [x] Clean build, no warnings
- [x] `[ast_function_id]` — 31/31 pass
- [x] `[ast_value]` — Phase 2 regression gate green
- [x] `[ast_scaffold]` — Phase 1 regression gate green
- [x] Full pre-existing suite green (1060/1060)


src/expression/function_id.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <string_view>
+#include <utility>
+
+namespace sirius {
+
+namespace {
+
+// D-02: forward table — 28 entries (substring has TWO DuckDB-side aliases).
+// Linear scan; called once per BoundFunctionExpression at executor entry.
+inline constexpr std::array<std::pair<std::string_view, function_id>, 28> kForwardTable = {{
```
kevkrist: I think the `inline` is redundant -- static linkage is guaranteed by anonymous namespace. Same below.

src/expression/function_id.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <cstddef>
+#include <optional>
+#include <string_view>
+#include <utility>
+
+namespace sirius {
+
+namespace {
+
+// D-02: forward table — 28 entries (substring has TWO DuckDB-side aliases).
```
kevkrist: This is maybe just a convention I'm not familiar with -- what is D-02? Or rather, what are these references pointing at?

src/expression_executor/specializations/gpu_execute_function.cpp
```diff
@@ -47,80 +48,53 @@
 #include <regex>
 #include <string>
 
-// There has to be a better way to extract the function semantics than string comparison!
```
kevkrist, line 50: Yay!

src/expression_executor/specializations/gpu_execute_function.cpp
```diff
@@ -411,9 +385,8 @@ execute_result gpu_expression_executor::execute(duckdb::BoundFunctionExpression
       num_rows, std::move(child_cols), 0, rmm::device_buffer{}, _stream, _mr);
   }
 
-  // If we reach here, it means the function is not supported in the expression executor
   throw not_implemented_exception(
-    "[gpu_expression_executor:function] execute called on unsupported function: %s", func_string);
+    "[gpu_expression_executor:function] execute called on unsupported function: {}", func_string);
```
kevkrist: The throw from the failure at line 60 and here now mean slightly different things. The first says we don't have this function name in our table of supported function names. The second (here) says either a) we do have it in our table but we don't have an implementation for it or b) the duckdb "error" function was called. Reaching a) indicates a likely developer error and reaching b) should probably log a different exception string. Consider explicitly handling "error" as a case and/or refining the exception string here.

src/expression/function_id.cpp
```diff
@@ -0,0 +1,95 @@
...
+// D-02: reverse table — 27 entries, indexed by enum value (NOT searched).
+inline constexpr std::array<std::string_view, 27> kReverseTable = {
+  "+",           "-",           "*",           "/",          "//",
+  "%",           "substring",   "~~",          "!~~",        "contains",
+  "prefix",      "suffix",      "strlen",      "length",     "regexp_replace",
+  "year",        "month",       "day",         "hour",       "minute",
+  "second",      "millisecond", "microsecond", "date_trunc", "row",
+  "struct_pack", "error",
+};
+
```
kevkrist, line 75: Consider a consteval function that walks the forward and reverse tables to ensure strict agreement (claude):
```
consteval bool foo() {
  auto constexpr enum_count = kReverseTable.size();
  std::array<int, kEnumCount> count{};
  for(auto const& [name, id] : kForwardTable) {
    auto const idx = static_cast<std::size_t>(id);
    if(idx >= enum_count) { return false; }
    if(name == kReverseTable[idx]) { ++canonical_count[idx]; }
  }
  for(auto const c : canonical_count) {
    if(c != 1) { return false; }
  }
  return true;
}
static_assert(foo());
```

kevkrist (approved):
This is great thanks!
Just some nits.

src/expression_executor/specializations/gpu_execute_function.cpp
9prady9: Addressed this, please check

PR #718, opened 2026-04-27
Sirius source manager

aminaramoon:


src/scan_manager/parquet_split_provider.cpp
```diff
@@ -0,0 +1,299 @@
...
+    return future;
+  }
+
+  auto remaining = std::make_shared<std::atomic<std::size_t>>(batches.size());
+  for (auto& batch : batches) {
+    pool.schedule([this, batch = std::move(batch), &connector, remaining, promise]() {
+      try {
+        run_batch(batch, connector);
+      } catch (const std::exception& e) {
+        SIRIUS_LOG_ERROR("[parquet_split_provider] metadata scan task failed: {}", e.what());
```
kevkrist, line 187: Do you intend to swallow and not propagate this error? A query that should fail will produce partial results instead.

Maybe have the catch capture the exception into a shared std::exception_ptr and have the last task call promise->set_exception() if any task failed.

src/include/scan_manager/parquet_split_provider.hpp
```diff
@@ -0,0 +1,133 @@
...
+   *                                the original operator). Used to derive the
+   *                                post-filter projection ids.
+   * @param table_filter_set        Filter set for row-group pruning / pushdown.
+   * @param partition_indices       Hive partition indices, if any.
+   * @param approximate_batch_size  Target uncompressed bytes per row-group
+   *                                partition.
+   * @param max_file_processed      Maximum number of files handled by one
+   *                                metadata task.
+   */
+  parquet_split_provider(
```
kevkrist, line 74: Align member init order with declaration order?

src/include/scan_manager/parquet_split_provider.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <duckdb/common/multi_file/multi_file_data.hpp>
+#include <duckdb/common/types.hpp>
+#include <duckdb/common/vector.hpp>
+
+#include <cstddef>
+#include <memory>
+#include <string>
+#include <unordered_set>
+#include <vector>
+
```
kevkrist, line 35:
```suggestion
#include <optional>
```

src/include/scan_manager/parquet_split_provider.hpp
```diff
@@ -0,0 +1,133 @@
...
+class TableFilterSet;
+}  // namespace duckdb
+
+namespace sirius::scan_manager {
+
+/**
+ * @brief Split provider that parses parquet metadata and emits one
+ *        @c parquet_scan_data per row-group partition.
+ *
+ * This provider replaces the previous sirius_parquet_metadata_scan_operator:
```
kevkrist: Remove reference to code that is deleted in this PR.

src/include/scan_manager/parquet_split_provider.hpp
```diff
@@ -0,0 +1,133 @@
...
+  ///        per partition into @p connector.
+  void run_batch(file_batch const& batch, split_connector& connector);
+
+  std::vector<std::string> _file_paths;
+  std::vector<std::size_t> _selected_column_indices;
+  bool _is_projected;
+  std::vector<std::string> _projected_column_names;
+  bool _has_filter;
+  std::shared_ptr<duckdb::Expression> _duckdb_filter_expression;
+  std::vector<std::string> _column_name_by_ref;
```
kevkrist, line 120: This is unused since it looks like Claude removed the AST code that was temporarily commented out. It's been uncommented and is live now. Just noting here that your merge will need to address this.

src/include/scan_manager/sirius_scan_manager.hpp
```diff
@@ -0,0 +1,102 @@
...
+    op::scan::sirius_gpu_parquet_scan_operator* op);
+
+  /// \brief Run providers sequentially: start each, wait on its future, advance.
+  void run_driver_loop();
+
+  exec::thread_pool_config _config;
+  std::unique_ptr<exec::thread_pool> _thread_pool;
+  std::unordered_map<op::scan::sirius_gpu_parquet_scan_operator*, std::unique_ptr<split_provider>>
+    _providers_by_op;
+  std::vector<op::scan::sirius_gpu_parquet_scan_operator*> _scan_op_order;
```
kevkrist: Could this be simplified with a vector<pair<op, provider>> since at sirius_scan_manager.cpp:64-65 you populate the map and the vector at the same time?

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
```diff
@@ -42,84 +43,48 @@ namespace sirius::op::scan {
...
-      !it->second->src_pipeline->is_pipeline_finished()) {
-    if (auto upstream = it->second->src_pipeline->get_source()) {
-      return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, upstream.get()};
-    }
-  }
-
-  // 3. No work, pipeline done — finished.
-  return std::nullopt;
+  if (_split_connector->is_closed()) { return std::nullopt; }
+  return task_creation_hint{TaskCreationHint::READY, this};
```
kevkrist, line 83: Is there another WAITING state between when split_connector queue is empty but not yet closed?

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -0,0 +1,144 @@
...
+void sirius_scan_manager::start()
+{
+  if (_thread_pool) { return; }
+  _thread_pool = std::make_unique<exec::thread_pool>(
+    _config.num_threads, _config.thread_name_prefix, _config.cpu_affinity_list);
+}
+
+void sirius_scan_manager::stop()
+{
+  if (_driver_thread.joinable()) { _driver_thread.join(); }
```
kevkrist, line 140: If, e.g., remote storage is unresponsive, stop will hang here because the driver thread will fail to join waiting on its future from parquet_split_provider::start(). Probably fine now but should have force-stop functionality at some point.

src/include/op/scan/parquet_scan_operator_data.hpp
kevkrist: Doc references partitoned_parquet_metadata which no longer exists.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -0,0 +1,144 @@
...
+    _providers_by_op.emplace(op, std::move(provider));
+    _scan_op_order.push_back(op);
+
+    SIRIUS_LOG_DEBUG("[sirius_scan_manager::prepare_for_query] registered op_id={}",
+                     op->get_operator_id());
+  }
+
+  if (_scan_op_order.empty()) { return; }
+
+  if (!_thread_pool) {
```
kevkrist, line 75: Consider moving this to the beginning, so that operators don't get installed but orphaned connectors in case someone forgot to call start().

src/scan_manager/split_connector.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <utility>
+
+namespace sirius::scan_manager {
+
+split_connector::split_connector()  = default;
+split_connector::~split_connector() = default;
+
+void split_connector::push_split(std::unique_ptr<op::operator_data> split)
+{
+  {
```
kevkrist, line 32: Is it worth checking that the split_connector is closed? I can see that push after close cannot happen by construction currently, but it might be sane to guard against future corruptions. Also that split is not null?

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
```diff
@@ -42,84 +43,48 @@ namespace sirius::op::scan {
...
 //===----------------------------------------------------------------------===//
 sirius_gpu_parquet_scan_operator::sirius_gpu_parquet_scan_operator(
-  duckdb::vector<sirius::logical_type> types, duckdb::idx_t estimated_cardinality)
+  duckdb::vector<sirius::logical_type> types,
+  duckdb::idx_t estimated_cardinality,
+  std::unique_ptr<parquet_scan_info> scan_info)
   : sirius_physical_operator(
-      SiriusPhysicalOperatorType::GPU_PARQUET_SCAN, std::move(types), estimated_cardinality)
+      SiriusPhysicalOperatorType::GPU_PARQUET_SCAN, std::move(types), estimated_cardinality),
+    _split_connector(std::make_unique<scan_manager::split_connector>()),
```
kevkrist, line 51: Should we close() the default split connector in the constructor? If the scan_manager doesn't install a non-default one, this remains open, and get_next_task_input_data() will block forever on the connector's condition variable. Again, this doesn't currently happen by construction, but it might be worth adding for, e.g., unit tests.

kevkrist (commented):
Very cool!
Do you think there should be unit tests?
 - split_connector
 - parquet_split_provider
 - sirius_scan_manager
Otherwise everything is small and partly a reflection of my effort to understand this PR.

PR #721, opened 2026-04-27
Feature: Adding ability to pin tables in gpu memory

aminaramoon:
``` -- Pin lineitem with 4 columns. tier is accepted; current implementation
  -- always pins to the first GPU memory space.

  CALL pin_table(
      '/home/aaramoon/Documents/tpch/sf10/parquet/lineitem.parquet',
      name   = 'lineitem',
      tier   = 'gpu',
      cols   = ['l_orderkey', 'l_quantity', 'l_extendedprice', 'l_shipdate']
  );

  -- Sample query: TPC-H Q6-style filter on l_shipdate using the 4 pinned columns.
  -- The path inside read_parquet must equal the path used in pin_table so the
  -- scan_manager swaps in cached_split_provider.
 SELECT
          SUM(l_extendedprice * l_quantity) AS revenue,
          COUNT(*)                          AS line_count
      FROM read_parquet('/home/aaramoon/Documents/tpch/sf10/parquet/lineitem.parquet')
      WHERE l_shipdate >= DATE '1994-01-01'
        AND l_shipdate <  DATE '1995-01-01'
        AND l_quantity <  24;
 

  -- When you're done with the pin:
  CALL unpin_table('lineitem');
```


src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -109,33 +197,28 @@ std::unique_ptr<split_provider> sirius_scan_manager::create_provider_for(
...
+    auto it = _providers_by_op.find(op);
+    if (it == _providers_by_op.end()) { continue; }
     auto* connector = op->get_split_connector();
     if (connector == nullptr) { continue; }
 
     try {
-      auto future = provider->start(*_thread_pool, *connector);
-      future.get();
+      auto future = it->second->start(*_thread_pool, *connector);
+      future.wait();
```
aminaramoon: that is a mistake done by claude, wait doesn't proopage exception, get does, I will fix it now

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -109,33 +197,28 @@ std::unique_ptr<split_provider> sirius_scan_manager::create_provider_for(
...
-      // resolves — otherwise the closed-but-undercount connector looks like normal end-of-stream
-      // to the gpu scan operator and the query returns wrong results silently.
-      _driver_error = std::current_exception();
-      // Provider already closed its own connector before setting the exception; close all
-      // remaining connectors so any other in-flight gpu scan ops drain immediately rather
-      // than blocking on dead providers we will not start. close() is idempotent.
-      close_all_connectors();
-      return;
+      // Make sure the consumer is unblocked even on failure.
+      connector->close();
```
bwyogatama, line 214: So now if there is an error, we close the connector and just go to the next provider? Why didn't we fallback or return an error?

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -90,6 +87,97 @@ std::unique_ptr<split_provider> sirius_scan_manager::create_provider_for(
...
+    auto selected_indices =
+      op::scan::detail::make_selected_column_indices(info->column_ids, info->projection_ids);
+
+    std::vector<std::vector<std::shared_ptr<cudf::column>>> columns_per_request;
+    columns_per_request.reserve(selected_indices.size());
+    for (auto idx : selected_indices) {
+      auto const& col_name = info->names[idx];
+      auto it              = entry.data_batches_by_column.find(col_name);
+      if (it == entry.data_batches_by_column.end()) {
+        throw std::runtime_error("[sirius_scan_manager::create_provider_for] pinned entry '" +
```
bwyogatama, line 106: So we will throw an error if there is a mismatch on the column? Shouldn't this be falling back to the regular path instead of throwing?

src/scan_manager/sirius_scan_manager.cpp
aminaramoon: actually, I have a bunch of fixes related to connector in the ther PR, maybe I should bring them here. the issue here is scan manager should not handle the error itself, it should pass the error the its consumer, ( gpu_parquest_scan_operator), that way the error in parsing metadata will be propogated to the egninee the way an error in a operator would, and that way, we don't need to worry about error here at all, we should just capture them and pass them to the connector

src/scan_manager/sirius_scan_manager.cpp
aminaramoon: yes, it should fallback.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -147,30 +230,87 @@ void sirius_scan_manager::start()
...
+  // Compute the total row count of the incoming tables before releasing them
+  // (release() empties the table; num_rows() would then return 0).
+  std::size_t new_num_rows = 0;
+  for (auto const& table : data_tables) {
+    if (table) { new_num_rows += static_cast<std::size_t>(table->num_rows()); }
+  }
+
+  auto existing_it = _pinned_entries.find(name);
+  if (existing_it != _pinned_entries.end()) {
+    if (existing_it->second.num_rows == new_num_rows) {
```
bwyogatama, line 256: When combining existing and new
1. Should we also check the file path to make sure it's from the same parquet?
2. Is the row alignment for each batch guaranteed to be the same between existing and new?

src/include/op/scan/parquet_scan_operator_data.hpp
```diff
@@ -139,4 +142,41 @@ class parquet_scan_data : public op::operator_data {
...
+//===----------------------------------------------------------------------===//
+/**
+ * @brief Input to a GPU parquet scan task served from a pinned (cached) entry.
+ *
+ * The cached_split_provider produces one of these per cached batch. It carries
+ * a zero-copy data_batch (whose gpu_table_representation is a view over pinned
+ * columns) plus the post-filter / projection metadata that the scan operator
+ * applies in execute(). The gpu_memory_space lives on the wrapped data_batch.
+ */
+class scan_cached_operator_data : public op::operator_data {
```
bwyogatama, line 164: IIRC there is `prepare_for_processing()` and `get_estimated_size_in_bytes()` in `operator_data`
For `get_estimated_size_in_bytes(),` isn't this supposed to be used in the reservation of subsequent operators? and right now it's just set to 0 which might mess that up.
For `prepare_for_processing()`, i guess it just won't work for multi gpu case, cause one column can be pinned in one GPU but the task is executed in another. Can be addressed in another PR, but worth noting

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
```diff
@@ -176,12 +164,95 @@ std::unique_ptr<operator_data> sirius_gpu_parquet_scan_operator::execute(
...
+      SIRIUS_LOG_DEBUG(
+        "[sirius_gpu_parquet_scan_operator] Applied duckdb filter expression on cached batch.");
+    }
+
+    if (has_projection) {
+      auto columns = has_filter ? table->release() : std::vector<std::unique_ptr<cudf::column>>{};
+      std::vector<std::unique_ptr<cudf::column>> selected;
+      selected.reserve(cached->post_filter_projection_ids.size());
+      if (has_filter) {
+        for (auto const idx : cached->post_filter_projection_ids) {
```
bwyogatama: Hmmm not 100% sure but i think the `post_filter_projection_ids` could be on a different index space than the cached table column indices

Cause the `post_filter_projection_ids` is derived from `projection_ids` which I think is indexing into the actual column indices in the relation (like if the relation has 5 columns then the index is 0-4). But the one that is cached might be a subset of those column indices.

Not 100% positive, but worth saying

CMakeLists.txt
```diff
@@ -239,6 +239,7 @@ set(EXTENSION_SOURCES
     src/planner/sirius_plan_projection.cpp
     src/planner/sirius_plan_recursive_cte.cpp
     src/planner/sirius_plan_top_n.cpp
+    src/scan_manager/cached_split_provider.cpp
```
bwyogatama, line 242: There is two other new files need to be added to the CMakeLists right? `pin_table.cpp `and `test_pin_table.cpp`?

bwyogatama (approved):

src/scan_manager/sirius_scan_manager.cpp
aminaramoon: 1. matches_scan_info does exactly that.
since I don;t take offset, it should be. it is only the first n_rows

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
aminaramoon: @kevkrist can you confirm if that is the case?

kevkrist: I do believe this is a conflation of D-space (columns[id]) and C-space (projection_ids, which are positions in column_ids). (See scan_plan.hpp/cpp). This works whenever the projected set is the leading N columns, which DuckDB seems to generate consistently, but it will silently fail otherwise and does represent a semantic confustion.

Recommendation: in create_provider_for(), build a scan_plan the same way parquet_split_provider_does and attach it to scan_cached_operator_data, and for the post-filter assembly, use the inject_fn it provides as in the non-cached path.

src/include/op/scan/parquet_scan_operator_data.hpp
```diff
@@ -139,4 +142,41 @@ class parquet_scan_data : public op::operator_data {
...
+      post_filter_projection_ids(std::move(post_filter_projection_ids))
+  {
+  }
+
+  /// Cached data batch viewed by the scan. Owning shared_ptr keeps the pinned
+  /// columns alive for the lifetime of the task.
+  std::shared_ptr<cucascade::data_batch> batch;
+  /// Filter expression, same shape as parquet_scan_data::filter_expression. May
+  /// be empty (i.e. variant default) when no filter applies.
+  std::variant<std::shared_ptr<translated_expression>, std::shared_ptr<duckdb::Expression>>
```
kevkrist: I think you can ditch this variant and just store a shared_ptr of duckdb expression.

src/include/scan_manager/cached_split_connector.hpp
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * @brief Connector backed by a fixed vector of pre-pinned cached batches.
+ *
+ * Unlike split_connector — which is a producer/consumer queue — this connector
+ * hands out entries from an immutable vector supplied at construction. Each
+ * call to next_split() returns the entry at the current cursor and advances
+ * the cursor atomically; once the cursor passes the end of the vector,
+ * next_split() returns std::nullopt.
+ */
+class cached_split_connector {
```
kevkrist: This isn't used anywhere?

src/include/scan_manager/sirius_scan_manager.hpp
```diff
@@ -83,34 +111,34 @@ class sirius_scan_manager {
...
-  /// driver's own close on success is safe.
+  /// \brief Pin the entry for a table.
   ///
-  /// Lifetime caveat: connectors are operator-owned (unique_ptr). Callers MUST
-  /// invoke this BEFORE the gpu scan operators are destroyed — typically from
-  /// the engine's catch handler before drain_after_error joins the task pools.
-  /// Calling it from SiriusContext::terminate() is unsafe because DuckDB tears
-  /// down the physical plan first, so the operator pointers in _providers
-  /// would dangle.
-  void close_all_connectors();
```
kevkrist, line 100: Might be worth documenting here if it's still possible but will be fixed: if a downstream operator throws, the driver might still be mid-provider, and parked task_creator workers will hang without a wakeup on the connector during drain_after_error.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -109,33 +199,28 @@ std::unique_ptr<split_provider> sirius_scan_manager::create_provider_for(
...
     try {
-      auto future = provider->start(*_thread_pool, *connector);
+      auto future = it->second->start(*_thread_pool, *connector);
       future.get();
     } catch (const std::exception& e) {
       SIRIUS_LOG_ERROR("[sirius_scan_manager] driver: provider failed: {}", e.what());
-      // Capture the failure so the engine can rethrow it after task_scheduler's future
-      // resolves — otherwise the closed-but-undercount connector looks like normal end-of-stream
-      // to the gpu scan operator and the query returns wrong results silently.
-      _driver_error = std::current_exception();
```
kevkrist, line 124: A provider failure here gets swallowed, the connector gets closed normally, and the query returns incomplete results from the scan, no?

PR #731, opened 2026-04-28
Sirius Scan Manager

kevkrist:
Supersedes #718 by rebasing with new changes.


ducndh:
Hi Kevin —This looks great as the `split_provider` / `split_connector` / `sirius_scan_manager` triple is a clean extraction of the existing logic. Two things I'd like to ask, because they'll affect how iceberg so I just want to discuss first though your choice if I can do the refactor myself on later PR or you can do it in this.
(one of them is kinda a necessary but none blocking and one of them is I think a cleaner pattern in my opinion)

## Extract parquet's row-group enumeration as a callable helper

The thing that actually blocks iceberg-on-scan_manager is whether parquet's split-emission logic is callable from another provider.

The row-group enumeration lives privately inside `parquet_split_provider::run_batch`. That's fine for parquet alone, but iceberg is *defined* by being a layer over parquet — manifest + delete-file resolution → then "scan these specific row groups in these parquet files". The natural shape for `iceberg_split_provider` is:

```cpp
void iceberg_split_provider::run_batch(...) {
  auto data_files = read_manifests(...);
  auto delete_masks = resolve_deletes(...);
  for (auto& df : data_files) {
    parquet::emit_splits(df, delete_masks[df], connector);  // <-- shared helper
  }
}
```

That requires `parquet::emit_splits` (or similar) to be a free function or public static helper, not a private method on `parquet_split_provider`.

It is a quick fix but just want to let you know about this and I'm happy to do it on my PR

## Generalize the factory signature

`create_provider_for(sirius_gpu_parquet_scan_operator*)` is parquet-typed. To add iceberg / DuckDB-native, the stopgap right now is overload
```cpp
create_provider_for(sirius_iceberg_scan_operator*);
create_provider_for(sirius_duckdb_native_scan_operator*);
```

That works but means changing on `sirius_scan_manager` every time someone adds a format. What do you think about this factory-like class to make it less tangled with each other

```cpp
struct scan_info_base {
  enum class kind { parquet, iceberg, duckdb_native };
  virtual kind get_kind() const = 0;
};

// every scan operator carries one
sirius_physical_operator::get_scan_info() -> scan_info_base*

// factory becomes
create_provider_for(sirius_physical_operator* op) {
  switch (op->get_scan_info()->get_kind()) { ... }
}
```

One method, dispatched by data instead of by overload so new format just need to add a kind value + a case so manager stops accumulating knowledge of every format.

This isn't blocking — happy to live with overloads — just flagging because it's much easier to fix the API while #731 has one consumer than after it has three.

## Clarifying Questions

- `prepare_for_query` runs providers **sequentially in registration order** on the driver thread. Is the eventual plan to parallelize across pipelines, or is sequencing intentional (e.g. so pipeline order = scan order for memory budgeting)? I just want to know the road map for this.
- The deletion of `sirius_parquet_metadata_scan_operator` — did the column-mapping (#662) and row-group pruning logic move to `parquet_split_provider::run_batch`, or to bind-time? I want to mirror your pattern Iceberg and DuckDB or we can discuss if you are not sure about this yet.

mike-wendt (commented):
CODEOWNERS - Approved for GHA Test workflow changes only

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+void sirius_scan_manager::run_driver_loop()
+{
+  for (auto& [op, provider] : _providers) {
+    auto* connector = op->get_split_connector();
+    if (connector == nullptr) { continue; }
+
+    try {
+      auto future = provider->start(*_thread_pool, *connector);
+      future.get();
```
bwyogatama, line 118: So maybe I'm misunderstanding something here but it seems to me that `run_driver_loop` is iterating through providers sequentially and blocks until each provider done reading the metadata?

Is this intentional? Shouldn't this be asynchronous?

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -0,0 +1,146 @@
...
+  _thread_pool = std::make_unique<exec::thread_pool>(
+    _config.num_threads, _config.thread_name_prefix, _config.cpu_affinity_list);
+}
+
+void sirius_scan_manager::stop()
+{
+  /// KEVIN: if the split_provider fails to deliver its future (e.g., unresponsive remote storage),
+  /// stop() will hang here on the join().
+  if (_driver_thread.joinable()) { _driver_thread.join(); }
+  if (!_thread_pool) { return; }
```
bwyogatama, line 156: Agree with Kevin's comment here on the hang, and also shouldn't we stop the thread pool first before joining the driver thread?

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -0,0 +1,146 @@
...
+    if (connector == nullptr) { continue; }
+
+    try {
+      auto future = provider->start(*_thread_pool, *connector);
+      future.get();
+    } catch (const std::exception& e) {
+      SIRIUS_LOG_ERROR("[sirius_scan_manager] driver: provider failed: {}", e.what());
+      // Provider already closed the connector before setting the exception;
+      // closing again here is idempotent and covers synchronous start() throws.
+      connector->close();
```
bwyogatama: So here we close the connector when there is an error and we just iterate through the next provider? Shouldn't we throw or fallback to DuckDB? Otherwise the scan operator will think that the metadata scan is done cause the connector is already closed and it will produce undercount results

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -0,0 +1,146 @@
...
+    if (source->type != ::sirius::op::SiriusPhysicalOperatorType::GPU_PARQUET_SCAN) { continue; }
+
+    auto* op                = &source->Cast<op::scan::sirius_gpu_parquet_scan_operator>();
+    auto already_registered = std::any_of(
+      _providers.begin(), _providers.end(), [op](auto const& entry) { return entry.first == op; });
+    if (already_registered) { continue; }
+
+    auto provider = create_provider_for(op);
+    if (!provider) {
+      // No scan_info parked on the operator (e.g. tests construct the operator
```
bwyogatama: Consider putting a warning LOG here, cause I think this should only happen on tests

bwyogatama (approved):

mike-wendt:
Force pushed a rebase to get the tests to clear to unblock the PR, feel free to force push again with your own rebase. Just trying to help out

mike-wendt (approved):
Approved for CODEOWNERS - `.github/` files

wmalpica (approved):

src/scan_manager/sirius_scan_manager.cpp
kevkrist: That is correct. This synchronicity will be removed in #740.

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
```diff
@@ -42,84 +43,55 @@ namespace sirius::op::scan {
...
-      return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, upstream.get()};
-    }
-  }
-
-  // 3. No work, pipeline done — finished.
-  return std::nullopt;
+  if (_split_connector->is_closed()) { return std::nullopt; }
+  // Will return READY even when the queue is empty but not yet closed — get_next_task_input_data()
+  // will block on the connector's cv. This parks a worker but avoids needing a scheduler-visible
+  // wake-up signal from the scan_manager.
```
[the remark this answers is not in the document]
kevkrist: Decided to keep the parked-worker design with the trade-off as documented for now.

The deadlock-on-shutdown is fixed by adding `sirius_scan_manager::close_all_connectors()` and calling  it before `task_scheduler::drain_after_error()` in the engine's catch handlers, and before `task_creator_->sop_thread_pool()` in `SiriusContext::terminate()`, so any parked `task_creator` thread is woken before the thread pool joins.

src/scan_manager/sirius_scan_manager.cpp
kevkrist: For first comment, see response to Copilot. For second, if we stop the thread pool first, `remaining` in `parquet_split_provider::start()` never reaches zero, and the promise is never set, so the driver thread will fail to join.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -0,0 +1,146 @@
...
+    _config.num_threads, _config.thread_name_prefix, _config.cpu_affinity_list);
+}
+
+void sirius_scan_manager::stop()
+{
+  /// KEVIN: if the split_provider fails to deliver its future (e.g., unresponsive remote storage),
+  /// stop() will hang here on the join().
+  if (_driver_thread.joinable()) { _driver_thread.join(); }
+  if (!_thread_pool) { return; }
+  _thread_pool->stop();
```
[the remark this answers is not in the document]
kevkrist: Leaving the producer-side cancellation as a TODO (see comment).
Regarding the consumer-side cancellation, `close_all_connectors()` prevents deadlock in the task creator.

src/scan_manager/sirius_scan_manager.cpp
kevkrist: Fixed.

kevkrist:
@ducndh Let's add Iceberg/Duckdb support incrementally rather than up-front here. Your recommendations are solid.

PR #732, opened 2026-04-28
Feature/single node multi gpu2

felipeblazing:


src/creator/task_creator.cpp
```diff
@@ -378,8 +439,97 @@ void task_creator::manager_loop()
...
+              // under the hood, and cuco tables must live on a single device —
+              // a stream bound to GPU A touching a counter built under GPU B
+              // trips cudaErrorInvalidValue at counter_storage.cuh. Routing on
+              // partition_idx keeps every task of a given partition on one GPU
+              // while still spreading partitions across GPUs.
+              if (auto* partitioned =
+                    dynamic_cast<op::partitioned_operator_data*>(pipelineable_input);
+                  partitioned && _sys_topology && !_sys_topology->gpus.empty()) {
+                auto n_gpus         = _sys_topology->gpus.size();
+                auto idx            = partitioned->get_partition_idx() % n_gpus;
```
wmalpica, line 436: We should have partitioned_operator_data have a new member like `target_cpu_idx`, so that its calculated elsewhere. Then we can also add more complex logic. For example, if there is only one partition, we might want the GPU to be different for different operators, that way we help balance the load between GPUs. 

This can be done in a followup if we want

src/creator/task_creator.cpp
```diff
@@ -378,8 +439,97 @@ void task_creator::manager_loop()
...
+                preferred_device_id = static_cast<int>(_sys_topology->gpus[idx].id);
+              }
+              if (!preferred_device_id.has_value() && pipelineable_input &&
```
wmalpica, line 439: All this logic to identify the preferred device id should be refactored into its own function and have unit tests

src/sirius_config.cpp
```diff
@@ -213,6 +213,11 @@ struct host_mem_config {
 
   void setup_configurator(cucascade::memory::reservation_manager_configurator& builder) const
   {
+    // MGPU-05: cucascade builds one numa_region_pinned_host_memory_resource
+    // per distinct NUMA node when the configurator sees this call. Relied
+    // upon by SiriusContext::initialize() which asserts host_spaces.size()
+    // == topology.num_numa_nodes on the default path. YAML configs may
+    // override by explicitly setting per-space numa_id.
     builder.use_host_per_numa();
```
wmalpica, line 238: I know its not new exactly on this PR, but right now as it is `memory.host.capacity_bytes` configuration is really per numa region not per process. This is confusing. We should either rename it capacity_bytes_per_numa  or change it so that capacity_bytes is divided between the numa regions

src/op/sirius_physical_hash_join.cpp
```diff
@@ -606,7 +618,15 @@ static join_side_keys_result prepare_join_keys(
   cudf::table_view table = get_cudf_table_view(input_batch);
 
   if (!cast_necessary) {
-    result.keys = table.select(key_col_indices);
+    // Phase 12-02: filter key_col_indices to the valid range [0, num_columns())
+    // so a stale index from the SORT-as-HASH_JOIN partitioner doesn't throw
+    // std::out_of_range from cudf::table_view::select.
+    std::vector<cudf::size_type> valid_indices;
```
wmalpica: This does not make sense. What changed in the code that would make the key_col_indices now fall outside the range?
I am worries that this code was added because something was failing that has a different root cause that needs to be addressed

docs/super-sirius/multi-gpu-architecture.md
```diff
@@ -0,0 +1,245 @@
...
+
+The engine assumes a single process pinning all visible GPUs (`CUDA_VISIBLE_DEVICES` controls which GPUs Sirius can use). There is no notion of distributed multi-node execution in this codebase.
+
+## Tier Hierarchy
+
+Every byte of data lives in exactly one **memory tier** at any moment. Tiers, fastest to slowest:
+
+| Tier | Backed by | Capacity per host | Typical purpose |
+|------|-----------|-------------------|-----------------|
+| `GPU` | `cuda_async_memory_resource` (one pool per device) | Bounded by GPU device memory × `reservation_fraction_per_gpu` (default 0.75) | Active query data |
```
wmalpica: I think this should be 
| `GPU` | `cuda_async_memory_resource` (one pool per device) | Bounded by GPU device memory × `usage_limit_fraction`| Active query data |
| `HOST` | NUMA-local pinned host memory | Set by config: memory.host.capacity_bytes (which is per numa host memory resource| Downgrade target when GPU is full |

docs/super-sirius/multi-gpu-architecture.md
```diff
@@ -0,0 +1,245 @@
...
+| `DISK` | On-disk file pool via cucascade's `idisk_io_backend` | Configured via `set_disk_mounting_point(gpu_id, capacity, path)` | Last-resort downgrade target |
+
+`HOST` is partitioned by NUMA node — each GPU has a paired host region on its NUMA-local memory controller for fast downgrade. This is configured via `cucascade::memory::reservation_manager_configurator::use_host_per_gpu()` at startup.
+
+Conceptually:
+
+```
+        ┌─────────┐  ┌─────────┐
+        │  GPU 0  │  │  GPU 1  │      ← Tier::GPU
+        └────┬────┘  └────┬────┘
```
wmalpica:
```
┌─────────┐  ┌─────────┐  ┌─────────┐  ┌─────────┐
│  GPU 0  │  │  GPU 1  │  │  GPU 2  │  │  GPU 3  │  ← Tier::GPU
└────┬────┘  └────┬────┘  └────┬────┘  └────┬────┘
     └─────┬──────┘            └─────┬──────┘
           │                         │
      ┌────┴────┐               ┌────┴────┐
      │ Host    │               │ Host    │
      │ (NUMA0) │               │ (NUMA1) │ ← Tier::HOST (per-GPU NUMA-local)
      └────┬────┘               └────┬────┘
           └─────────────┬───────────┘
                    ┌────┴────┐
                    │  Disk   │              ← Tier::DISK (shared pool)
                    └─────────┘
```

docs/super-sirius/pipeline-execution.md
```diff
@@ -111,6 +111,192 @@ The **destructor** calls `pipeline->mark_task_completed()` to update pipeline co
...
+
+`std::map` gives an ascending-by-`device_id` iteration order, which makes `_gpu_executors.begin()` deterministic instead of hash-bucket-dependent and makes `std::advance(it, idx)` walk a fixed sequence.
+
+**Per-query reset.** `src/pipeline/task_scheduler.cpp:156-160`, inside `task_scheduler::prepare_for_query`:
+
+```cpp
+// Reset SCHED-RR counter so the round-robin walk is reproducible across
+// iterations of the same query (cache=table_gpu warm path keys cache
+// entries by device_id; without this reset the second iteration's source
+// tasks would assign to a different GPU and miss the cache entries).
```
wmalpica, line 245: We should improve this in a followup. cache hits should be dependent on the data, perhaps using a chunk/file/table id and not be related to the order in which a counter is incremented

src/pipeline/sirius_pipeline_converter.cpp
```diff
@@ -1085,6 +1090,37 @@ void sirius_pipeline_converter::link_join_partition_siblings()
...
+{
+  // Pull num_gpus from the build context (populated from sirius_engine's
+  // hardware topology at convert time). Single-GPU runs keep the default
+  // min of 1 (no-op). For multi-GPU we force a floor equal to num_gpus on
+  // big-enough inputs; small_table_bytes keeps tiny aggregations on a
+  // single GPU to avoid cross-device overhead.
+  const int num_gpus = build_ctx_.num_gpus;
+  if (num_gpus <= 1) return;
+  // Heuristic threshold: below ~16 MiB per GPU the partition overhead
+  // dominates. Configurable later if we find a workload where this matters.
```
wmalpica, line 1102: we should make this an operator config option. And instead of having this function which iterates over all the operators, we should follow the same pattern we use for all operator config options, which is to set it on the operator construction. See for example how its set for sirius_physical_concat and sirius_physical_partition in `sirius_pipeline_converter::split_intermediate_joins`. This way we follow the same pattern.

src/pipeline/task_scheduler.cpp
```diff
@@ -153,6 +153,11 @@ void task_scheduler::prepare_for_query(duckdb::shared_ptr<planner::query> query)
   for (auto& [device_id, gpu_exec] : _gpu_executors) {
     gpu_exec->drain_leftover_tasks();
   }
+  // Reset the round-robin counter so the walk is reproducible across
+  // iterations of the same query (cache=table_gpu warm path keys cache
+  // entries by device_id; without this reset the second iteration's source
+  // tasks would assign to a different GPU and miss the cache entries).
+  _no_pref_rr_counter.store(0, std::memory_order_relaxed);
```
wmalpica, line 160: We should figure out how to make the cache hit determinism be purely on whether or not we are querying the same data. It should be unrelated to this counter, otherwise its not much of a cache

src/sirius_context.cpp
```diff
@@ -213,6 +246,137 @@ void SiriusContext::initialize(const sirius::sirius_config& config)
...
+  if (!gpu_ioctxs_.empty()) {
+    // Match the spelling of `kFileScheme` from datasource_factory.cpp
+    // verbatim (anonymous namespace constant).
+    static constexpr std::string_view kFileScheme = "file";
+    auto lowest_gpu                               = std::min_element(
+      gpu_ioctxs_.begin(), gpu_ioctxs_.end(), [](auto const& a, auto const& b) {
+        return a.first < b.first;
+      });
+    datasource_registry_.register_ioctx(std::string{kFileScheme}, lowest_gpu->second);
+    spdlog::info(
```
wmalpica: We have standard logging functions, we should not use spdlog directly. There are multiple instances of this

.ai-helper/commands.yaml
```diff
@@ -0,0 +1,144 @@
+version: 1
```
wmalpica, line 1: we want to delete this file, right?

wmalpica (approved):
Looks good, lets merge this soon and address issues immediatelly after

src/sirius_extension.cpp
bwyogatama: Not from this PR but we should remove this

docs/super-sirius/multi-gpu-architecture.md
```diff
@@ -0,0 +1,245 @@
...
+```sql
+CALL pin_table('lineitem');                    -- distribute chunks across GPU memory spaces (default)
+CALL pin_table('lineitem', tier => 'host');    -- pin into host memory only
+```
+
+The pin pipeline (`src/scan_manager/`):
+
+1. **Open the parquet files.** `PinTableFunction` resolves the file scheme through `datasource_registry_`, gets per-GPU `sirius_datasource` instances backed by `uring_ioctx`.
+2. **Enumerate chunks.** Reads parquet metadata to discover row-group chunks.
+3. **Distribute round-robin.** `chunk_memory_spaces[i] = gpu_spaces[i % gpu_spaces.size()]` — chunk `i` is assigned to the `i`-th GPU's memory_space. This is the **PIN-MGPU-01 invariant**.
```
bwyogatama: This is actually per-file instead of per-chunk

src/sirius_extension.cpp
```diff
@@ -821,19 +844,54 @@ void SiriusExtension::PinTableFunction(ClientContext& context,
...
       if (read_column_names.empty()) {
         read_column_names.reserve(chunk.metadata.schema_info.size());
         for (auto const& col_info : chunk.metadata.schema_info) {
           read_column_names.push_back(col_info.name);
         }
       }
+      file_rows_read += chunk_rows;
       if (data.args.tier == "host") {
-        cucascade::gpu_table_representation gpu_repr(std::move(chunk.tbl), gpu_mem_space);
+        cucascade::gpu_table_representation gpu_repr(std::move(chunk.tbl), gpu_mem_space, stream_view);
```
bwyogatama: The `gpu_mem_space` and `host_mem_space` is still set to GPU 0 here? Is this intentional?

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -100,9 +113,20 @@ std::unique_ptr<split_provider> sirius_scan_manager::create_provider_for(
   try {
     for (auto const& [pinned_name, entry] : _pinned_entries) {
       if (!matches_scan_info(entry)) { continue; }
-      if (entry.memory_space == nullptr) {
+      // Validate the per-chunk memory_space vector before building the
+      // cached_split_provider. Empty vector means the pinned entry has no
+      // chunks (unusual but legal for empty parquet files); null entries
+      // inside the vector violate the chunks-at-index-i invariant.
+      if (entry.chunk_memory_spaces.empty()) {
```
bwyogatama: Right now the `chunk_memory_spaces` is always empty for `pin_tier` = "host"

bwyogatama (approved):
Most of my reviews are related to the multi-GPU functionality around pinning tables, which I think is not working yet. This can be addressed in a future PR.

One more thing before we merge, should we test if SF1k SNSG on GB300 is still working on top this PR? Just to avoid degradation. I'm happy to test it if people think this is necessary

src/creator/task_creator.cpp
```diff
@@ -359,8 +406,93 @@ void task_creator::manager_loop()
             auto gpu_pipeline_task_global_state = _gpu_operator_global_state_map.at(operator_id);
             auto local_state =
               std::make_unique<pipeline::gpu_pipeline_task_local_state>(std::move(input_data));
+
+            // pipelineable_input remains valid here: the cast happened before
+            // the move into local_state, and unique_ptr move transfers
+            // ownership without relocating the object.
+            {
+              std::optional<int> preferred_device_id;
```
bwyogatama, line 422: Don't we need to set `preferred_device_id` for table that got pinned in specific GPU? 
Right now the pinned batches is tagged with their own `chunk_memory_space` so there is some locality there, but currently we only take into account the locality for `pipelineable_operator_data`. Shouldn't we need to do that for `scan_cached_operator_data` as well?

.ai-helper/commands.yaml
felipeblazing: no its a utility file used by a system that can be used to develop the engine. even if i am the only one using it

src/creator/task_creator.cpp
aminaramoon: I agree with william

src/data/host_parquet_representation_converters.cpp
aminaramoon, line 1: if it helps getting this PR faster into dev, this file will be removed from the source very soon, so I owoudn;t buffer too mich about it

src/include/creator/task_creator.hpp
```diff
@@ -197,6 +201,19 @@ class task_creator {
   std::unique_ptr<duckdb::ThreadContext> _thread_context;
   std::unique_ptr<duckdb::ExecutionContext> _execution_context;
   std::mutex _global_state_mutex;  // Protect concurrent access to the map
+
+  /// System topology for NUMA-aware GPU routing (non-owning, may be null)
+  const cucascade::memory::system_topology_info* _sys_topology{nullptr};
```
aminaramoon, line 206: shouldn;t all of this logic  leave inside reservation_manager

manager->reservation_for_downgrade(src, destiation)

interanlly uses the cost matrix to decide which space is most suitable ?

src/include/data/host_parquet_representation.hpp
aminaramoon, line 1: this will be deadcode soon, don't spend too much on it if you can.

src/include/op/scan/parquet_scan_task.hpp
```diff
@@ -245,6 +264,39 @@ class parquet_scan_task_global_state : public pipeline::sirius_pipeline_task_glo
...
+   * to the per-GPU sirius::io::sirius_ioctx bound to that GPU's CUDA context.
+   * Used by compute_task() (hot path, routed by preferred_device_id) and by
+   * initialize_from_files() (planning-time, first-available GPU).
+   *
+   * @return A const reference to the map.
+   */
+  [[nodiscard]] std::unordered_map<int, std::shared_ptr<sirius::io::sirius_ioctx>> const&
+  get_gpu_ioctxs() const
+  {
+    return _gpu_ioctxs;
```
aminaramoon, line 281: what is thi for ? one ioctx should be good for all of your GPUs.

aminaramoon (commented):
I have a bunch of commetns about the scan work

1. scan_parquest_task, duckdb scan tasks will be deadcode soon. I think it makes sense to drop them completely if it helps you clean-up this PR
2. ioctx is gpu agnostic, you only need a single ioctx for all your gpus readers.

src/creator/task_creator.cpp
felipeblazing: i agree lets do tis in a follow up

src/sirius_config.cpp
felipeblazing: lets change that in a follow up

src/sirius_extension.cpp
felipeblazing: no this is a bug

mike-wendt (approved):
CODEOWNERS approval only for changes to `.github/workflows/test.yml`

- CI `test` workflow will now use mutli-GPU nodes to support the multi-GPU tests in this PR
- Once the merge conflicts are resolved, CI will run

docs/super-sirius/multi-gpu-architecture.md
```diff
@@ -0,0 +1,246 @@
...
+5. **Record the pinned entry.** `pinned_entry { data_batches_by_column, chunk_memory_spaces }` is stored on the `sirius_scan_manager`. The `chunk_memory_spaces` vector is parallel to `data_batches_by_column[col_idx]` — `chunk_memory_spaces[i]` is the owning space for chunk `i` regardless of which column you're looking at.
+
+Repeat invocations of `pin_table('lineitem', ...)` are idempotent — duplicates dropped, existing `chunk_memory_spaces` preserved (Phase 22 Pitfall 3 invariant: any merge must verify `chunk_memory_spaces` integrity).
+
+When the new HOST-tier pinning path is used (`2e197c6` upstream feature, integrated in Phase 24), `pin_table` constructs a `cucascade::host_data_representation` and stores it in `pinned_entry`'s host-mode slot; subsequent scans go through a host-mode `cached_split_provider` that slices the host chunks per query and converts back to GPU only when a scan task starts. The GPU-tier and HOST-tier pinning paths coexist as parallel code paths — both maintain the `chunk_memory_spaces` invariant.
```
9prady9, line 125: `integrated in Phase 24` is this yet to be done ? or already done ?

docs/super-sirius/multi-gpu-architecture.md
```diff
@@ -0,0 +1,246 @@
...
+See [`pipeline-execution.md`](pipeline-execution.md) "Per-task-device contract under SCHED-RR" for the deeper contract.
+
+## Cross-GPU Data Movement
+
+When an operator must consume data on GPU A that lives on GPU B (e.g., a hash join's probe side has chunks scattered across all GPUs), the engine performs a **GPU↔GPU transfer** via `cucascade::convert_gpu_to_gpu` (in `cucascade/src/data/representation_converter.cpp`).
+
+The transfer chooses one of two paths empirically:
+
+1. **Direct peer DMA** (`cudaMemcpyPeerAsync`) — fastest, used when `probe_peer_dma_works(src, dst)` returns true. Real peer access requires both GPUs to have driver-level P2P enabled AND the hardware to actually honor it.
+2. **Host-staging** (`cudaMemcpyAsync(DtoH)` → host buffer → `cudaMemcpyAsync(HtoD)`) — fallback for hardware where peer DMA is empirically broken (e.g., the consumer-grade RTX 6000 Ada we use for development, which advertises P2P but silently fails DMA in both directions).
```
9prady9, line 163: is that host bounce buffer a pinned one or regular one ?

docs/super-sirius/multi-gpu-architecture.md
```diff
@@ -0,0 +1,246 @@
...
+- **Probe device-context restore** (Phase 23 Plan 23-07): `run_p2p_probe_locked` ends with a paired `cudaSetDevice` to restore the caller's device context, not a hardcoded `cudaSetDevice(0)` which would clobber the caller's RAII guard.
+
+## Multi-GPU-Safe Parquet I/O
+
+The I/O path is **kvikio-free** (GATE-22.1-A invariant). The reasoning: cudf's bundled `file_source` factory uses kvikio, which binds the file handle to whichever CUDA context was active at construction time. In multi-GPU execution that's a hidden source of corruption — a file_handle bound to GPU 0 will silently funnel reads through GPU 0 even when the consumer is on GPU 1.
+
+The Sirius path:
+
+1. **All file reads go through `sirius_ioctx::make_datasource(io_object)`.** Never `cudf::io::datasource::create(path)` and never `cudf::io::source_info{path}`. A grep gate enforces this in `src/`.
```
9prady9, line 179: `grep` gate ? I think we should enforce this more concretely right ?

docs/super-sirius/multi-gpu-architecture.md
```diff
@@ -0,0 +1,246 @@
...
+2. **`sirius_ioctx` is per-GPU.** Construction binds `cudaSetDevice(gpu_id)` via RAII; the underlying `uring_reactor` runs in a worker thread that holds that device's context.
+3. **Schemes are resolved through `datasource_registry_`.** A scan task that wants to read `file:///path/to/x.parquet` passes the URL to the registry, which returns a factory bound to the per-GPU `sirius_ioctx`. No silent fallback.
+4. **Pin-table chunk reads route through the owning GPU's ioctx.** Chunk `i` reads through `gpu_ioctxs_.at(chunk_memory_spaces[i]->get_device_id())`. The read lands directly in the target GPU's pinned memory region.
+
+The 7 historical sites that bypassed this contract (kvikio-fallback in `sirius_gpu_parquet_scan_operator`, `sirius_extension`, `iceberg_metadata_reader`, `datasource_factory`, `parquet_split_provider`) were all migrated to `sirius_ioctx::make_datasource` over Phase 22.1. The unit-test fallback in `parquet_split_provider:295` (D-08 site #7) was deleted outright.
```
9prady9, line 184: Does this mean we are yet to do this ? If so, we should create issues to not loose track of these in docs.

docs/super-sirius/multi-gpu-architecture.md
9prady9, line 1: I think this file still needs a proper overhaul on it's content like we discussed in the PR review meet.

- A lot of references to phases which it mentions in future tense confusing if the task is done or we are to create issues to follow up.
- A lot of non-multi GPU info is also added that can be trimmed.
- Concurrency invariants table has entries like GATE, grep gates. I think we should create an issue and handle that enforcement later if not in this PR.
- A clean mermaid diagram could help a lot.

In current form this feels like plan document than a architecture doc which should be addressed.

.ai-helper/commands.yaml
9prady9, line 1: this will be removed before merge right ?

docs/super-sirius/multi-gpu-architecture.md
9prady9: Also, similar to `docs/super-sirius/README.md` this document could definitely use a TOC in the beginning since it has lot of content it covers and it would make it easy to refer to any specific topic later.

docs/super-sirius/pipeline-execution.md
```diff
@@ -111,6 +111,192 @@ The **destructor** calls `pipeline->mark_task_completed()` to update pipeline co
...
+
+**The hazard this exposes.** Several operators read `valid_batches[0]->get_memory_space()` (or an equivalent expression on a single input batch) as the authoritative target memory space, then perform their concat/merge/sort directly on that space. Pre-Phase-14, this was *accidentally* safe — every batch in the input vector was already on the implicit "default GPU" because every upstream task was dispatched to that same default. Under SCHED-RR, that accident is gone. If an operator reads `batches[0]->get_memory_space()` without a guarantee that *all* batches in the input vector are colocated on that space, it can silently produce wrong results, mis-allocate, or skip data on the other GPU.
+
+The fix is not to patch every read site to detect cross-GPU input. The fix is the upstream contract below: every operator's input batches are colocated by the task scheduler **before** the operator's `execute()` runs, so reading `batches[0]->get_memory_space()` is a SAFE alias for the task's reservation device.
```
9prady9, line 132: is this for a follow up as well ?

docs/super-sirius/pipeline-execution.md
9prady9, line 1: Requires same treatment as `docs/super-sirius/multi-gpu-architecture.md`

.gitignore
9prady9, line 1: removed by mistake ?

9prady9 (commented):
stopping the review since it is already merged but most of points added till now are still relevant.

PR #734, opened 2026-04-29
CI: Remove GHA Runner Spot Usage

mike-wendt:
While we have dropped down to only a few spot-interruptions in a 24 hr period these are still very disruptive to devs. Switching to full on-demand usage for the time being to remove the failures associated with spot-interruptions until we can handle these more gracefully


mbrobbel (approved):

mike-wendt:
Pulling in changes from #735, once this passes I'll merge

PR #735, opened 2026-04-29
CI: Disable GHA sccache on self-hosted runners

mike-wendt:
This was triggered by a version mismatch where our updated `sccache` action was installing `0.15.0` of `sccache` but our `pixi.toml` and lock file installed `0.14.0`. When the stats call was made, it would error with the version mismatch.

`0.15.0` feedstock was just merged in conda-forge which should fix the issue once the packages our built. To ensure we don't run into this again, I'm adding a step to configure `sccache` based on the env vars present. Self-hosted runners configure `sccache` for S3 so they should not have GHA `sccache` active at the same time.

Once the `0.15.0` `sccache` packages are available I will update the `pixi.toml` an lock file as well as a part of this PR

### Fixes
- [x] Disable GH `sccache` on self-hosted runners
- [x] Update `pixi.toml` with new `0.15.0` version for `sccache`
- [x] Update `pixi.lock` with new `0.15.0` version for `sccache`

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -104,6 +104,16 @@ jobs:
             ${{ matrix.cudf-channel == 'nightly' && 'nightly-runner'
               || matrix.environment }}
 
+      - name: Configure sccache backend
+        run: |
+          if [[ -n "${SCCACHE_BUCKET}" ]]; then
+            echo "SCCACHE_BUCKET DEFINED -> Self Hosted Runner, Disable GHA sccache"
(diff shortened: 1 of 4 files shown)


mike-wendt:
`0.15.0` has not propagated to the CDN, will have to wait ~30 mins to try the lock file update again

mike-wendt:
Lock file change in, now we should see ✅ for all checks

mike-wendt:
Will admin merge this once all checks pass to get the team unblocked

PR #736, opened 2026-04-30
scan: DuckDB format GPU-native decode kernel (Part 1: Template dispatcher + uncompressed + constant)

ducndh:
First of a 10-PR series wiring up a DuckDB-native GPU scan path. Establishes the per-column decode dispatcher contract; subsequent PRs plug in kernels via one switch case in `dispatch_data_run`.                                                                                                                                  

**Public API.** `gpu_decode_table(cols, stream, mr)` takes per-column descriptors that point at on-device segment bytes and returns a `cudf::table`. Codec metadata stays inside segment bytes — kernels parse their own headers, so the descriptor types stay frozen across codec PRs.                                                                                                             

**Codecs shipped.** UNCOMPRESSED (single-segment cudaMemcpyAsync DMA fast path, multi-segment batched-memcpy kernel) and CONSTANT (cudf::type_dispatcher + device-pointer broadcast kernel; the cudf alternative would do ≥3 D2H syncs per call). Validity uses `cudf::create_null_mask` + a single batched `cudf::batch_null_count` covering every mask in the table.                                                                                                                           

**Measure Performance:** UNCOMPRESSED 1×32MiB on Quadro RTX 6000 (sm_75) — 504 GiB/s rd+wr, CONSTANT 32MiB at 535 GiB/s write                                                                                                                                            

17 unit tests under `[scan][decode]` + 5 microbenches `[!benchmark][scan][decode]`.


src/cuda/scan/gpu_native_decode.cu
```diff
@@ -0,0 +1,536 @@
...
+                          rmm::cuda_stream_view stream)
+{
+  for (auto const& seg : run.segments) {
+    if (seg.row_count == 0) continue;
+    if (seg.bytes_size < type_size) {
+      throw std::runtime_error("gpu_decode_table: CONSTANT segment bytes_size (" +
+                               std::to_string(seg.bytes_size) + ") < type_size (" +
+                               std::to_string(type_size) + ")");
+    }
+    launch_broadcast_constant(
```
kevkrist: I feel that, just as with the batched memcpy, we should be able to batch constant segments together, no?

src/cuda/scan/gpu_native_decode.cu
```diff
@@ -0,0 +1,536 @@
...
+      throw std::runtime_error("gpu_decode_table: validity row_offset (" +
+                               std::to_string(seg.row_offset) + ") not byte-aligned");
+    }
+    size_t bytes = (size_t{seg.row_count} + 7) / 8;
+    if (size_t{seg.bytes_size} < bytes) {
+      throw std::runtime_error("gpu_decode_table: validity segment bytes_size (" +
+                               std::to_string(seg.bytes_size) + ") < required " +
+                               std::to_string(bytes));
+    }
+    RMM_CUDA_TRY(cudaMemcpyAsync(
```
kevkrist: Again, we should batch these together, if possible.

kevkrist (commented):
A few requests.

src/cuda/scan/gpu_native_decode.cu
ducndh: Each segment is 120k rows and the chance that we do multiple constant segment at once (the chance of a constant segment is already quite rate) is pretty low so I will leave it here since batching may introduce complexity about typing across them (I will delay this until we hit a bad case of this)

src/cuda/scan/gpu_native_decode.cu
ducndh: This one actually costs a decent amount of performance. I can do a follow up PR with a hand roll batched kernel or do you want to land this now?

kevkrist (approved):

src/cuda/scan/gpu_native_decode.cu
kevkrist: I left a comment.

src/cuda/scan/gpu_native_decode.cu
kevkrist: I added cub batched memcpy here.

PR #737, opened 2026-04-30
scan: DuckDB format GPU-native decode kernel (Part 2: Bit Packing CONSTANT / CONSTANT_DELTA / FOR / DELTA_FOR)

ducndh:
Second of the 10-PR DuckDB-native scan series. (after PR #736 merge, this should get much shorter) Plugs BITPACKING into the dispatcher established by `decode-foundation` via one switch case in `dispatch_data_run`; per-segment metadata (mode, width, FOR base, delta offset) lives in segment bytes and is parsed on device, so the descriptor surface stays codec-agnostic
 
 **Modes covered:** CONSTANT, CONSTANT_DELTA, FOR, DELTA_FOR (the 4 concrete modes DuckDB's AUTO resolves to). INVALID metadata or any out-of-bounds parse demotes the group to a no-op rather than falling through to a default decode 
                                                                                                                                     
**Stores:** CONSTANT / CONSTANT_DELTA vectorise to int4 (16-byte) `__stwt` when the type fits; FOR / DELTA_FOR keep striped scalar `__stwt`. On sm_75 the `__stwt` annotation is a wash but show no regression but I haven't test on sm_80+ where L1-bypass might be a win.  
                                                                                                                                                        
**Performance measure:** BITPACKING FOR int64 width=8, 4M rows, sm_75 — **278 GiB/s write**. Full numbers in `test/data/decode_baselines/sm75.json`.                                                                                                                            

11 unit tests under `[scan][decode][bitpacking]` (mode coverage + multi-segment + multi-group + 4 defensive guards). 3 microbenches


src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+  }
+  __syncthreads();
+
+  uint32_t const rc = sm_row_count;
+  auto const mode   = static_cast<BitpackingMode>(sm_mode);
+
+  //===--------------------------------------------------------------------===//
+  // CONSTANT — broadcast `sm_aux` to every row.
+  //
+  // We don't go through `cudf::type_dispatcher` like the UNCOMPRESSED-CONSTANT
```
kevkrist: Delete the rest of this comment. It's clear why cudf::type_dispatcher is not used.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+  // path does, because the constant value is parsed *inside this kernel* from
+  // the segment's metadata; reaching for a separate dispatcher launch would
+  // require staging the parsed value back to a device scalar and re-launching.
+  //===--------------------------------------------------------------------===//
+  if (mode == BitpackingMode::CONSTANT) {
```
kevkrist, line 271: Consider using a switch on the mode and dispatching to a corresponding __device__ function for each mode please to enhance modularity and readability.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+    static_assert(sizeof(T) <= 8 && 16u % sizeof(T) == 0u,
+                  "BITPACKING kernel only instantiates for type sizes in {1,2,4,8}; "
+                  "extending the dispatcher to a non-conforming type needs a scalar "
+                  "fallback path here");
+    T const val            = sm_aux;
```
kevkrist: In the previous PR, there was an alignment check for using the vectorized int4 path. Why is there no alignment check here? I believe there needs to be one.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+    constexpr uint32_t TPV = 16u / sizeof(T);
```
kevkrist: If the 16 is always referring to sizeof(int4), consider replacing 16 with sizeof(int4) to makre the connection clear.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+  // CONSTANT — broadcast `sm_aux` to every row.
+  //
+  // We don't go through `cudf::type_dispatcher` like the UNCOMPRESSED-CONSTANT
+  // path does, because the constant value is parsed *inside this kernel* from
+  // the segment's metadata; reaching for a separate dispatcher launch would
+  // require staging the parsed value back to a device scalar and re-launching.
+  //===--------------------------------------------------------------------===//
+  if (mode == BitpackingMode::CONSTANT) {
+    static_assert(sizeof(T) <= 8 && 16u % sizeof(T) == 0u,
+                  "BITPACKING kernel only instantiates for type sizes in {1,2,4,8}; "
```
kevkrist: This comment doesn't explain why 16 % sizeof(T) == 0 is required. It seems, just based on the comment, that 8 % sizeof(T) would do. So I'm guessing this has something to do with int4 vectorization?

test/cpp/scan/decode_test_utils.hpp
```diff
@@ -0,0 +1,144 @@
+/*
```
kevkrist, line 1: Delete this file, per Joost's comment offline, and document performance in the PR comments.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+                               h_descs,
+                               num_groups * sizeof(bp_group_desc),
+                               cudaMemcpyHostToDevice,
+                               stream.value()));
+
+  // Live shmem footprint per CTA, in 32-bit words:
+  //   BP_META_GROUP_SIZE values × max_width bits / 32, rounded up.
+  // The +1 guard word satisfies `unpack_value`'s 3-word read contract for
+  // 64-bit types when a value spans words [w, w+1, w+2].
+  constexpr uint32_t max_width        = sizeof(T) * 8u;
```
kevkrist:
```suggestion
  auto constexpr max_width        = ::cuda::std::numeric_limits<T>::digits;
```

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+  constexpr uint32_t max_packed_words = (BP_META_GROUP_SIZE * max_width + 31u) / 32u + 1u;
```
kevkrist:
```suggestion
  auto constexpr max_packed_words = ::cuda::ceil_div(BP_META_GROUP_SIZE * max_width, 32) + 1;
```
Also consider making 32 a named static constant like WORD_BITS so it's clear what 32 means in contexts where that is its meaning.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+  constexpr size_t shmem_bytes        = max_packed_words * sizeof(uint32_t);
+
+  // Default per-CTA dynamic shmem cap on Turing/Ampere is 48 KB. For T up
+  // to int64 the live footprint peaks at ~16 KB. The static_assert catches
+  // any future BP_META_GROUP_SIZE / wider-T regression at compile time
+  // rather than as an opaque "previous unspecified launch failure" at the
+  // next stream sync.
+  static_assert(shmem_bytes <= 48u * 1024u,
```
kevkrist: We shouldn't be hard-coding architecture-specific constraints into the code. Add a logging statement indicating the shared memory footprint instead.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+
+  //===--------------------------------------------------------------------===//
+  // CONSTANT_DELTA — out[i] = frame + i*delta.
+  //===--------------------------------------------------------------------===//
+  if (mode == BitpackingMode::CONSTANT_DELTA) {
+    // sizeof(T) constraint enforced by the static_assert in the CONSTANT
+    // branch above — reach here only via the same template instantiations.
+    T const frame          = sm_frame;
+    T const delta          = sm_aux;
+    constexpr uint32_t TPV = 16u / sizeof(T);
```
kevkrist: sizeof(int4) / sizeof(T)

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+    }
+    return;
+  }
+
+  uint32_t const width = sm_width;
+  T const frame        = sm_frame;
+
+  extern __shared__ uint32_t shmem[];
+  uint8_t const* packed_bytes = seg_base + sm_data_offset;
+  uint32_t packed_words       = (rc * width + 31u) / 32u;
```
kevkrist: Please use ::cuda::ceil_div() here

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+
+  // Stage the packed-bytes slice into shmem via `cp.async.ca.shared.global`
```
kevkrist: This comment can be trimmed to just `Stage the packed-bytes slice into SMEM via async memcpy.
The whole alignment issue can also be averted if we use cooperative groups:
```
#include <cooperative_groups.h>
#include <cooperative_groups/memcpy_async.h>

namespace cg = cooperative_groups;
auto block = cg::this_thread_block();

cg::memcpy_async(block,
                 shmem,
                 packed_bytes,
                 packed_words * sizeof(uint32_t));

if (threadIdx.x == 0) shmem[packed_words] = 0u;

cg::wait(block); 
```
The runtime will pick the best transaction size based on the pointer's alignment. Consider this replacement if there is no performance regression.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+
+  if (mode == BitpackingMode::FOR) {
+    // Striped layout: in iteration v, all blockDim.x threads write
+    // contiguous output positions [v*blockDim.x .. v*blockDim.x+blockDim.x).
+    // One coalesced cache-line transaction per warp per iteration.
+#pragma unroll
+    for (uint32_t v = 0; v < VPT; ++v) {
+      uint32_t idx = v * blockDim.x + threadIdx.x;
+      if (idx >= rc) break;
+      __stcs(out + idx, static_cast<T>(frame + unpack_value<T>(shmem, idx, width)));
```
kevkrist: This is T + T, so why the static_cast? Same for thread_data below.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+/// segment of `run`. Returns a host-side vector ready to be uploaded.
+std::vector<bp_group_desc> build_group_descs(gpu_codec_run const& run)
+{
+  std::vector<bp_group_desc> descs;
+  // Each segment contributes ceil(row_count / BP_META_GROUP_SIZE) groups;
+  // the reserve is a tight upper bound when every segment is a full group.
+  descs.reserve(run.segments.size() * 2);
+  for (auto const& seg : run.segments) {
+    if (seg.row_count == 0) continue;
+    uint32_t num_groups = (seg.row_count + BP_META_GROUP_SIZE - 1) / BP_META_GROUP_SIZE;
```
kevkrist: Use ::cuda::ceil_div here

src/include/cuda/scan/gpu_decode_bitpacking.cuh
```diff
@@ -0,0 +1,85 @@
...
+#include <cstddef>
+#include <cstdint>
+
+namespace sirius::cuda::scan {
+
+/// Number of values per metadata group within a bitpacked segment. Fixed by
+/// DuckDB's on-disk format — every metadata-group entry encodes the mode and
+/// data offset for exactly this many rows (the last group of a segment may
+/// be short).
+inline constexpr uint32_t BP_META_GROUP_SIZE = 2048;
```
kevkrist: Why not static constexpr? Also, if this is a duckdb constant, can we alias that here, rather than redefine the value ourselves?

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+    sm_data_offset = 0;
+    sm_row_count   = 0;
+
+    uint64_t metadata_end = 0;
+    if (seg_bytes >= sizeof(uint64_t)) { memcpy(&metadata_end, seg_base, sizeof(uint64_t)); }
+
+    // The metadata trailer holds one uint32 entry per group, ending at
+    // metadata_end; entry K (0-based) lives at metadata_end - (K+1) * 4.
+    // The trailer must fit *and* lie inside the segment buffer.
+    bool metadata_ok =
```
kevkrist, line 182: I think a description, maybe even an ascii diagram, of the layout of a segment (header, data, metadata) would be very helpful. Also the layout of the data buffer for a group. I see that there is a comment above but more elaboration would be helpful.

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,588 @@
...
+            sm_aux         = v2;
+            sm_data_offset = data_off + 3u * sizeof(T);
+            break;
+          }
+          default:
+            // Unknown / INVALID mode — leave sm_mode at its default INVALID.
+            break;
+        }
+
+        // sizeof(T)*8 is the kernel's hard upper bound for unpack_value;
```
kevkrist: This comment makes it seem like it's an artifact of the kernel that this check is here, when it really just indicates that the segment is malformed.

kevkrist (commented):
All just nits / cosmetic. Looks good to me. Good performance as well (90% memory throughput) given the instruction footprint. This might change on new architecutres, but there might not be any way around it either. Good job!

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,587 @@
...
+
+  //===--------------------------------------------------------------------===//
+  // CONSTANT_DELTA — out[i] = frame + i*delta.
+  //===--------------------------------------------------------------------===//
+  if (mode == BitpackingMode::CONSTANT_DELTA) {
+    // sizeof(T) constraint enforced by the static_assert in the CONSTANT
+    // branch above — reach here only via the same template instantiations.
+    T const frame          = sm_frame;
+    T const delta          = sm_aux;
+    constexpr uint32_t TPV = 16u / sizeof(T);
```
kevkrist: Again -- we need an alignment check here.

test/data/decode_baselines/sm75.json
```diff
@@ -9,6 +9,25 @@
     "purpose": "Per-arch microbench baseline. Codec PRs gate against these numbers (>5% throughput regression blocks merge). Update by re-running the bench."
```
kevkrist: Per Joost's offline comments, let's delete this file and add this data to the PR description.

test/cpp/scan/test_gpu_decode_bitpacking.cpp
```diff
@@ -0,0 +1,440 @@
+/*
```
kevkrist, line 1: From claude:
Test coverage gap: FOR and DELTA_FOR are not exercised across multi-group segments. Add at least one multi-group case where group 1 is FOR (and ideally one where group 1 is DELTA_FOR with a non-trivial delta_offset).
DELTA_FOR is only tested for int32 and int64. uint8/uint16 prefix-sum wrapping isn't locked down by a regression test. If you don't expect those types in real workloads from DuckDB, document that in the dispatcher; if you do, add a small uint16 case (1024 rows, width=8, mild deltas).

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,587 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Build per-group descriptors covering every metadata group across every
+/// segment of `run`. Returns a host-side vector ready to be uploaded.
+std::vector<bp_group_desc> build_group_descs(gpu_codec_run const& run)
+{
+  std::vector<bp_group_desc> descs;
+  // Each segment contributes ceil(row_count / BP_META_GROUP_SIZE) groups;
+  // the reserve is a tight upper bound when every segment is a full group.
+  descs.reserve(run.segments.size() * 2);
```
kevkrist: The comment isn't helping me here. Why is x2 a tight upper bound in that case?

src/cuda/scan/gpu_decode_bitpacking.cu
kevkrist: You can ignore this for now if you want. This is purely cosmetic / code simplification.

src/cuda/scan/gpu_decode_bitpacking.cu
kevkrist: Added a comment that the required alignment is guaranteed.

src/cuda/scan/gpu_decode_bitpacking.cu
kevkrist: It does. I've added a comment and moved the static_assert to the top of the file, since it asserts a kernel-global invariant.

src/cuda/scan/gpu_decode_bitpacking.cu
kevkrist: I think the x2 is a heuristic assuming each segment will produce 2 metadata groups on average. We don't know a priori how many segments there will be.

src/cuda/scan/gpu_decode_bitpacking.cu
kevkrist: Will refactor later

src/cuda/scan/gpu_decode_bitpacking.cu
```diff
@@ -0,0 +1,587 @@
...
+  uintptr_t const src_addr = reinterpret_cast<uintptr_t>(packed_bytes);
+  bool const can_use_16b   = (src_addr & 15u) == 0 && (packed_words & 3u) == 0;
+  if (can_use_16b) {
+    uint32_t const quad_count = packed_words >> 2;  // packed_words / 4
+    for (uint32_t i = threadIdx.x; i < quad_count; i += blockDim.x) {
+      __pipeline_memcpy_async(&shmem[i * 4], packed_bytes + i * 16u, 16);
+    }
+  } else {
+    for (uint32_t i = threadIdx.x; i < packed_words; i += blockDim.x) {
+      __pipeline_memcpy_async(&shmem[i], packed_bytes + i * sizeof(uint32_t), sizeof(uint32_t));
```
kevkrist: This assumes that the source ptr packed_bytes satisfies 4B alignment requirements. I t produces cudaMisalignedAddress errors when this fails to be true. This was exposed with tests using narrower values of T.

kevkrist (commented):
One bug found and fixed.

kevkrist (approved):

PR #738, opened 2026-04-30
Multifile parquet splits

kevkrist:
This PR changes the parquet partitioning to support bundling multiple files into a single parquet read -> data batch.


mike-wendt (approved):
Approved for .github/ changes

PR #740, opened 2026-04-30
Enable IO framework in sirius scan manager

aminaramoon:
This PR addresses comments on previous ScanManager PR, and Io Framework PR
as well as enabling the prefetching and IoContext in Sirius

---

- It introduces scope_dispatcher for static_thread_pool. better interface for managing tasks in thread-pool, will soon replace bounded_thread_pools.
- It introduces minor change to Io reactors. enables cache path for host reads as well.

----
DO NOT MERGE YET:
1. I am still investigating intermittent failure on TPCH when running with PR
2. I am still investigating why we get more OOMs with this PR compared to dev.
3. the overal cold perfromance with prefetching seems to be 30-50% better than our current cold numbers, but cold numbers without prefetching doesn't translate to signifcant wins on the e2e runs. I am investigating why 2-4X better parqeust performance compared to default cudf path in microbenchmarks doesn't translate to overal win without prefetching.

### DIFF
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -16,6 +16,7 @@
 
 #include "creator/task_creator.hpp"
 
+#include "cucascade/memory/common.hpp"
 #include "log/logging.hpp"
 #include "op/scan/cpu_source_task.hpp"
 #include "op/scan/duckdb_scan_executor.hpp"
@@ -33,6 +34,7 @@
 #include <duckdb/execution/execution_context.hpp>
 #include <duckdb/parallel/thread_context.hpp>
 
+#include <limits>
 #include <optional>
 
 namespace sirius::creator {
@@ -344,14 +346,17 @@ void task_creator::manager_loop()
 
         } else {
           // Create all possible tasks until all ports are empty
-          while (!node->all_ports_empty()) {
+          auto is_gpu_parquet_scan = node->type == op::SiriusPhysicalOperatorType::GPU_PARQUET_SCAN;
+          std::size_t count = is_gpu_parquet_scan ? 1 : std::numeric_limits<std::size_t>::max();
+          // need to exhaust input batches until all ports are empty
+          while (!node->all_ports_empty() && count-- > 0) {
             auto task_lock  = pipeline->get_task_creation_lock();
             auto input_data = node->get_next_task_input_data();
             auto* pipelineable_input =
               dynamic_cast<op::pipelineable_operator_data*>(input_data.get());
             if (!input_data ||
                 (pipelineable_input && pipelineable_input->get_data_batches().empty())) {
-              // do data to create task for
+              // no data to create task for
               break;
             }
 
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -346,10 +346,12 @@ void task_creator::manager_loop()
 
         } else {
           // Create all possible tasks until all ports are empty
-          auto is_gpu_parquet_scan = node->type == op::SiriusPhysicalOperatorType::GPU_PARQUET_SCAN;
-          std::size_t count = is_gpu_parquet_scan ? 1 : std::numeric_limits<std::size_t>::max();
-          // need to exhaust input batches until all ports are empty
-          while (!node->all_ports_empty() && count-- > 0) {
+          // TODO(amin) : do this based on the operator hint
+          // auto is_gpu_parquet_scan = node->type ==
+          // op::SiriusPhysicalOperatorType::GPU_PARQUET_SCAN; std::size_t count =
+          // is_gpu_parquet_scan ? 1 : std::numeric_limits<std::size_t>::max(); need to exhaust
+          // input batches until all ports are empty
+          while (!node->all_ports_empty()) {
             auto task_lock  = pipeline->get_task_creation_lock();
             auto input_data = node->get_next_task_input_data();
             auto* pipelineable_input =
--- a/src/include/exec/scoped_dispatcher.hpp
+++ b/src/include/exec/scoped_dispatcher.hpp
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2025, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <absl/functional/any_invocable.h>
+#include <exec/thread_pool.hpp>
+
+#include <concepts>
+#include <condition_variable>
+#include <mutex>
+#include <type_traits>
+#include <utility>
+
+namespace sirius::exec {
+
+class scoped_dispatcher {
+ public:
+  scoped_dispatcher(static_thread_pool& pool, std::size_t max_concurrent_task = 0)
+    : pool_(pool),
+      max_inflight_(max_concurrent_task == 0 ? pool.num_threads()
+                                             : std::min(max_concurrent_task, pool.num_threads()))
+  {
+  }
+
+  ~scoped_dispatcher()
+  {
+    // User contract: drain (via wait_for_all, optionally preceded by
+    // request_stop) before destruction. Loud assert beats silent UAF.
+    std::lock_guard lk(mu_);
+    assert(inflight_ == 0 && pending_.empty() &&
+           "scoped_dispatcher destroyed with outstanding work; "
+           "call request_stop()/wait_for_all() first");
+  }
+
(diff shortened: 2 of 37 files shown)


ran-yuan-rui:
Hi @aminaramoon — quick review comments and questions

### 1.To avoid duplicating work, I'd suggest taking on the S3 GPU SQL surface layer integration to cover:

- `read_parquet('s3://...')` → `sirius_read_parquet(...)` rewrite
- `sirius_read_parquet` footer-only bind / schema inference
- pipeline-converter glue to pass the S3 URI into `scan_manager`
- `scan_manager` S3 wiring
- `SiriusContext` config injection
- `ioctx` lifecycle / teardown fixes

Sounds good to you?

### 2.What API shape should SQL bind use to get the S3-capable ioctx from `sirius_scan_manager`?

Options I can imagine:

- `sm.io_ctx_for(path)`
- `sm.s3_io_ctx()`
- map lookup by scheme
- something else

This affects how `sirius_read_parquet` bind creates the S3 datasource for footer-only schema inference.

### 3. What is the canonical source for S3 object-store config?

Specifically endpoint / region / credentials:

- YAML `object_store_config` only?
- `SET s3_*` handlers?
- both?

If both exist, which one should treat as canonical during bind?

### 4. Should mixed local + S3 queries be supported in the first integration?

Example:

```sql
SELECT ...
FROM read_parquet('/local.parquet') l
JOIN read_parquet('s3://bucket/remote.parquet') r
ON ...
```
Should runtime routing support mixed backends per scan/path, or should the first version allow only a single backend per query?

### 5. Should S3 and uring share one `buffer_pool` / prefetch cache resource?

S3 has high-latency network refetches, while uring local reads are much cheaper. Do we need per-backend cache budgets / eviction policy, or is one shared FSMR-backed pool acceptable for now?

### 6. Should `scan_manager` expose a `thread_pool()` accessor?

Options:

- expose accessor and let `s3_ioctx` use it
- keep S3 thread pool internal and inject through `s3_config`
- keep all thread pools hidden behind `scan_manager` lifecycle

### 7 Do we need a unified shutdown contract across ioctx backends?

For example:

- uring reactor joins worker threads
- S3 async paths may rely on `shared_from_this` / callback lifetime
- multiple ioctxs may need an explicit shutdown order
Should `scan_manager` enforce a common `ioctx` shutdown order / lifecycle contract?


```

File: src/include/io/io_context.hpp [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+  /// Convenience: @c create_io_object + @c make_datasource in one shot.
+  std::unique_ptr<cudf::io::datasource> open_datasource(std::string path)
+  {
+    return make_datasource(create_io_object(std::move(path)));
+  }
+
+  /// Whether this backend can serve reads for @p path.  Backends should
+  /// validate scheme/protocol support and any backend-specific
+  /// preconditions (e.g. file existence for local-disk backends).
+  [[nodiscard]] virtual bool supports(std::string_view path) const = 0;
```
ran-yuan-rui, line 70: +1 on `supports(path)` as a virtual rather than scheme-keyed lookup — dispatch collapses to a `find_if` over a vector of ioctx, knowledge stays co-located with each backend, and multi-instance scenarios
(two s3_ioctx with different endpoints, path-style vs virtual-hosted bucket layouts) work without any external dispatch layer.

Side note: this makes the `datasource_registry` I shipped in PR1 #746
structurally redundant — once `supports()` lives on the ioctx,
scheme-keyed lookup has nothing left to do. Happy to leave that class
dormant rather than thread it through.

src/creator/task_creator.cpp
```diff
@@ -313,58 +313,68 @@ void task_creator::manager_loop()
...
           SIRIUS_LOG_DEBUG("Task Creator: scheduling cpu_source_task, dest_repos={}",
                            destination_data_repositories.size());
-          task_lock.unlock();
           _task_scheduler->schedule(std::move(task));
-
         } else {
-          // Create all possible tasks until all ports are empty
-          while (!node->all_ports_empty()) {
-            auto task_lock  = pipeline->get_task_creation_lock();
+          auto is_gpu_parquet_scan = node->type == op::SiriusPhysicalOperatorType::GPU_PARQUET_SCAN;
```
aminaramoon: I am going to properly address this

src/include/exec/scoped_dispatcher.hpp
```diff
@@ -0,0 +1,160 @@
...
+  using task_t = absl::AnyInvocable<void()>;
+
+  template <typename F>
+  task_t wrap(F&& f)
+  {
+    return [this, f = std::forward<F>(f)]() mutable {
+      if (!stop_.stop_requested()) {
+        try {
+          invoke_with_optional_stop_token(f, stop_.get_token());
+        } catch (...) { /* swallow */
```
wmalpica, line 127: Does this mean that any task we run with this scoped_dispatcher will swallow any exception it throws? that does not seem like a good idea

src/include/scan_manager/split_provider.hpp
```diff
@@ -16,50 +16,151 @@
...
+  schedule_worker(scheduler, worker_state::create(connector));
+}
+
+template <typename Scheduler>
+void split_provider::schedule_worker(Scheduler& scheduler, std::shared_ptr<worker_state> state)
+{
+  scheduler.enqueue([this, state = std::move(state), &scheduler]() {
+    try {
+      auto splits = create_split();
+      if (splits.empty()) { return; }
```
bwyogatama: Correct me if i'm wrong but this assumes that if the split is empty then the `split_provider` for that scan should be done and the next `schedule_worker` will not be called

Wouldn't there be cases where the `split_provider` return empty split, but the scan is not done yet? For example, what if all the row groups in the split is pruned, wouldn't that also return an empty split?

src/scan_manager/split_connector.cpp
```diff
@@ -50,6 +54,9 @@ std::optional<std::unique_ptr<op::operator_data>> split_connector::get_next_spli
 {
   std::unique_lock<std::mutex> lock(_mutex);
   _cv.wait(lock, [this] { return !_splits.empty() || _closed; });
+  // Drained and closed. If the producer captured an error via close(exception_ptr),
+  // surface it here so the consumer sees the failure rather than a silent EOF.
+  if (_exception) { std::rethrow_exception(_exception); }
```
bwyogatama, line 58: Should we drained the queue before throwing the exception?

src/io/uring/uring_reactor.cpp
```diff
@@ -268,26 +348,78 @@ void uring_reactor::worker_loop()
...
+        if (actual > 0 && !req.ctx->failed.load(std::memory_order_relaxed)) {
+          _bounce[si].cuda_done.store(false, std::memory_order_relaxed);
+          if (req.device_id >= 0) cudaSetDevice(req.device_id);
+          cudaMemcpyAsync(req.dst,
+                          (uint8_t*)_bounce[si].buf.get() + req.data_off,
+                          actual,
+                          cudaMemcpyHostToDevice,
+                          req.stream);
+          cudaLaunchHostFunc(req.stream, cuda_copy_cb, &_cb_args[si]);
+          sinfo.state = slot_state::COPYING;
```
bwyogatama: Would it make sense to check for CUDA errors in these cuda calls? If one of them failed, the callback never gets called, and the reactor can hang?

src/io/prefetching_cache.cpp
```diff
@@ -206,22 +225,29 @@ pinned_view& pinned_view::operator=(pinned_view&& o) noexcept
...
+    _entry.reset();
+  } else {
+    // Async path (device reads): defer release_read via a host callback so
+    // it fires only after the caller's stream reaches this point.  Any
+    // cudaMemcpyAsync submitted earlier against this entry's pinned chunks
+    // therefore completes before pin_count drops to zero — at which point
+    // the entry transitions in_use → cached and becomes evictable.
+    auto args   = std::make_unique<release_callback_args>();
+    args->entry = std::move(_entry);
+    cudaLaunchHostFunc(_stream, &release_read_host_callback, args.release());
```
bwyogatama: Same here, would it make sense to check for CUDA errors?

src/io/io_context.cpp
```diff
@@ -0,0 +1,230 @@
...
+        auto copied = copy_pinned_slices_to_device(slices, dst, stream);
+
+        cuda::cuda_event event(cudaEventDisableTiming);
+        event.record(stream);
+        return std::async(std::launch::deferred, [copied, e = std::move(event), stream]() mutable {
+          e.synchronize();
+          return copied;
+        });
+      } catch (...) {
+        return std::async(std::launch::deferred, []() -> size_t { throw; });
```
bwyogatama: Capture exception then `rethrow` rather than just `throw`?

src/scan_manager/split_connector.cpp
bwyogatama: The comment says "drained and close" after all

src/io/uring/uring_reactor.cpp
```diff
@@ -103,6 +114,13 @@ cudf::io::text::byte_range_info uring_reactor::align_to_physical(
   return {static_cast<int64_t>(a_start), static_cast<int64_t>(a_end - a_start)};
 }
 
+bool uring_reactor::supports(std::string_view path)
```
bwyogatama, line 129: Maybe a good place to check if the file system also support `O_DIRECT`?

bwyogatama (approved):

PR #742, opened 2026-05-02
add S3 datasource support to parquet scan path

ran-yuan-rui:
Adds end-to-end S3 (and any future object-store) backend for parquet scans: `gpu_execution('SELECT ... FROM read_parquet(''s3://bucket/key'')')` now routes through `s3_ioctx` (libcurl + SigV4) instead of cudf's default datasource — Sirius's vcpkg KvikIO build has KvikIO_REMOTE_SUPPORT=OFF, so cudf default throws "Unsupported URL scheme" on any non-file URI. Hooks into the Sirius IO framework (#675) and is meant to be the reference implementation for the per-scheme `sirius_ioctx` contract.

== New IO subsystem ==

* `io::uri_parser` — RFC-3986-ish parser; rejects bare relative paths (callers that need leniency use the helper below).
* `io::datasource_factory::create(uri, registry, config)` — strict entry point; routes by URI scheme to the registered `sirius_ioctx`. Local files (`/abs` or `file:///abs`) keep using cudf's default datasource.
* `io::datasource_factory::create_for_parquet_scan` — lenient sibling used by every parquet-scan IO site; bare relative paths bypass the parser to keep DuckDB's iceberg / hive fixtures (which deliver relative paths via MultiFileBindData) working.
* `io::object_store_config` — engine-config struct for endpoint, region, access/secret keys, max connections, request timeout.

== S3 backend ==

* `io::s3::sigv4` — AWS Signature V4 (path-style, MinIO + real AWS).
* `io::s3::s3_io_object` — `sirius_io_object` carrying `(bucket, key, size)`.
* `io::s3::s3_ioctx` — `sirius_ioctx` implementation:
    - host_read / host_read_async via libcurl Range GETs with a connection pool (max_connections, request_timeout configurable).
    - device_read_io bounces through host buffers + cudaMemcpyAsync (S3 has no native device path).
    - HEAD-then-GET so `s3_io_object::size()` stays noexcept.

== Engine integration ==

* `sirius_engine::datasource_registry()` — process-shared registry, populated lazily from `sirius_config::object_store_config`. Any process that never sets `s3_endpoint` keeps zero-cost behaviour: registry stays empty, no libcurl pull-in at runtime.
* SET handlers (`s3_endpoint`, `s3_region`, `s3_access_key`, `s3_secret_key`, `s3_max_connections`, `s3_request_timeout_seconds`) in sirius_extension fold straight into `object_store_config`.
* `sirius_read_parquet` SQL function: rewrites `read_parquet('s3://...')` at planner time; `SiriusReadParquetBind` runs schema inference via a temp local materializer (`materialize_remote_parquet_for_bind`) and stores the original URI on `scan_op.parameters[0]` for scan time.

== Scan-path adaptation ==

Bulk of the diff. Bridges the IO framework into the existing parquet, iceberg, and hive scan paths in a way that doesn't leak engine internals into scan_manager / SiriusContext:

* `op::scan::parquet_scan_info::open_datasource` — `std::function` populated by `sirius_pipeline_converter` capturing `&engine_`. The closure carries the dispatch logic to where it's needed; scan_manager and SiriusContext stay oblivious to the IO framework.
* `parquet_split_provider` — ctor takes the closure (default-empty for tests / standalone callers); the metadata-scan loop dispatches through it instead of `cudf::io::datasource::create` directly.
* `parquet_scan_task::initialize_from_files` and `compute_task` — both call sites use `create_for_parquet_scan`. iceberg inherits this path via `iceberg_scan_task_global_state`; hive keeps using split_provider. Both backends now share one dispatch rule.

== Connector exception propagation (fixes a latent #731 bug) ==

`sirius_scan_manager::run_driver_loop` previously called only `connector->close()` when a provider threw — `close()` is the end-of-stream signal, so consumers silently saw 0 rows and downstream GPU pipelines hung waiting on input that never came. With S3 the throw becomes routine (auth errors, transient 5xx), so this latent bug is now load-bearing.

* `split_connector::set_error(std::exception_ptr)` — new API; clears pending splits and stores the error under the same lock as the queue.
* `split_connector::get_next_split` — rethrows the stored error on the first wake after `set_error`; subsequent calls behave like a closed-empty queue.
* `is_closed` / `has_more_splits` — both treat a pending error as "still has work", preventing the gpu scan op's early-exit shortcut from skipping error delivery.
* `run_driver_loop` catches `(...)`; calls `set_error(std::current_exception())` before `close()`. The exception unwinds through the existing `task_creator` catch and reaches `task_scheduler::terminate_query`, so the query fails cleanly instead of hanging.

== Build, infra, tooling ==

* CMakeLists / vcpkg: libcurl + AWS-SigV4 deps, sirius IO static libs.
* Makefile: `s3-up` / `s3-down` targets driving MinIO docker-compose.
* `test/cpp/integration/s3/` — MinIO fixture rig (compose, env.sh, fixture upload script, sirius.yaml).
* Dev tools (opt-in via Makefile): `parquet_benchmark`, `stress_prefetching`, `verify_device_read`, `multi_gpu_datasource`.
* `pixi.toml` / `pixi.lock` / vcpkg triplets — toolchain bumps.

== Tests ==

* `test/cpp/io/test_uri_parser.cpp` — URI parsing edge cases.
* `test/cpp/io/test_datasource_factory.cpp` — strict + lenient entry points.
* `test/cpp/io/test_device_read_flags.cpp` — device-read capability.
* `test/cpp/io/s3/{test_sigv4,test_s3_ioctx,test_s3_integration, test_s3_parquet_integration}.cpp` — SigV4 vectors, handle pool, range GETs, MinIO end-to-end.
* `test/cpp/integration/test_gpu_execution_s3.cpp` — full gpu_execution over `s3://` (basic scan, projection+filter, aggregate, missing object, bad credentials, no-httpfs).
* `test/cpp/scan/test_split_connector.cpp` — connector close + error semantics.
* `test/cpp/scan/test_parquet_scan_task.cpp` — factory-routed metadata fetch.
* `test/cpp/operator/test_parquet_scan_via_factory.cpp` — operator-level factory integration.
* `test/cpp/engine/test_registry_bootstrap.cpp` — lazy registry population.
* `test/cpp/config/test_object_store_config.cpp` — config plumbing.
* `test/sql/datasource/set_s3_transport.test` — DuckDB SET round-trip.

== Out of scope / follow-ups ==

* Iceberg delete-file IO (`iceberg_metadata_reader.cpp`) still uses cudf default directly — relative-path-only today
* GDS / RDMA backends 
* Credential-provider abstraction  — current handlers take static keys via SET; AWS SDK / IMDS / STS chain comes later.
* Honest device-read capability reporting — `s3_ioctx` currently hard-codes `true` and bounces; should opt out so cudf doesn't steer toward an unnecessary copy.


ran-yuan-rui:
Got review feedback from Amin, I agreed that I will need break current PR into 3+ small PRs, and since Amin is working on how to registry and enable reactors to sirius_engine, I plan to remove that part of logic out of this PR and at the same time to divide into 3 PRs for better review:

- PR1 — Foundation:uri_parser,datasource_factory+datasource_registry(as a plain class,notattached tosirius_engine),object_store_config. Zero S3 code, zero libcurl, just unit tests.

- PR2 — AWS auth layer:aws_credentials+credential_providerinterface +static_credential_provider, plus asigv4refactor that consumes them. Independent of PR1, can review in parallel. Worth doing now rather than deferring becauseaws_credentialsneedssession_token/expires_at—x-amz-security-tokeninjection has to land with the signer or it's a retrofit later.

- PR3 — S3 reactor:s3_io_object,s3_ioctx(libcurl handle pool, sync/async host & device reads), MinIO integration tests. Depends on PR1 + PR2. 

Explicitly dropping from current PR (Amin's change will handle):

- Thedatasource_registry_member + accessor onsirius_engine— agreed it doesn't belong there given engine is per-query

- All SET handlers insirius_extension.cpp(s3_endpoint,s3_region,s3_access_key,s3_secret_key,s3_transport)

- sirius_read_parquetSQL function + bind-time materializer

- parquet_scan_info::open_datasourceclosure +parquet_split_providerctor change + pipeline-converter injection

- split_connector::set_error+ scan_manager error-propagation change — saw your note on redoing connector error handling

**Open issue** @bwyogatama  @aminaramoon 
After these 3 PRs, the end-to-end testing will jut be S3 reactor end-to-end，and will not be Sirius SQL/gpu_execution end-to-end like current PR742

ran-yuan-rui:
See first new PR at https://github.com/sirius-db/sirius/pull/746/

If we align on the plan which I just outline in comments then I will close current 742

ran-yuan-rui:
Based on the feedback, I closed this PR with new plan as below:
PR1 — Foundation:uri_parser,datasource_factory+datasource_registry(as a plain class,notattached tosirius_engine),object_store_config. [Zero S3 code, zero libcurl, just unit tests.](https://github.com/sirius-db/sirius/pull/746)
PR2 — AWS auth layer:aws_credentials+credential_providerinterface +static_credential_provider, plus asigv4refactor that consumes them. Independent of PR1, can review in parallel. Worth doing now rather than deferring becauseaws_credentialsneedssession_token/expires_at—x-amz-security-tokeninjection has to land with the signer or it's a retrofit later.
PR3 — S3 reactor:s3_io_object,s3_ioctx(libcurl handle pool, sync/async host & device reads), MinIO integration tests. Depends on PR1 + PR2.

PR #743, opened 2026-05-04
scan: introduce duckdb-native metadata walker

ducndh:
A metadata-only walker for the duckdb-native scan path. Output is a data struct (`duckdb_native_metadata`) consumed by follow-up PRs that build the split provider and scan operator. No bytes read and no blocks pinned in this yet

### DuckDB storage model 

A table is sliced horizontally into **row groups** (~122 880 rows each). Within a row group, each column has its own chain of segments — contiguous row ranges sharing one compression codec, tiled by `segment_start`. Validity is its own child column with its own segment chain, distinguished by `column_path` (`"[c, 0]"` vs `"[c]"`).

For example

```
Table "lineitem"  (e.g. 15M rows)
│
├── Row Group 0   (rows 0…122_879           horizontal slice of ALL columns)
│   │
│   ├── Column 0  "l_orderkey"      column_id=0
│   │   ├── Segment 0   segment_start=0       segment_count=2048   BitPacking
│   │   ├── Segment 1   segment_start=2048    segment_count=2048   BitPacking
│   │   ├── Segment 2   segment_start=4096    segment_count=4096   RLE
│   │   └── Segment 3   segment_start=8192    segment_count=114688 Dictionary
│   │
│   ├── Column 0  "l_orderkey"      column_path "[0, 0]"  ◄── validity child
│   │   └── Segment 0   segment_start=0       segment_count=122880 Empty Validity
│   │
│   ├── Column 1  "l_partkey"       column_id=1
│   │   └── Segment 0   segment_start=0       segment_count=122880 Dictionary
│   │
│   └── Column 9  "l_shipdate"      column_id=9
│       └── Segment 0   segment_start=0       segment_count=122880 Constant   (block_id = -1)
│
├── Row Group 1   (rows 122_880…245_759)
│   └── ... fresh segment chain per column, INDEPENDENT of Row Group 0
│
└── Row Group N
```

Key relationships:

- **column_id** is table-wide; the same id refers to the same column across every row group.
- **row_group_index** is the horizontal slice index (0, 1, 2, …).
- **segment_start** is row-group-local — the row offset *within the row group* where this segment begins. Segments tile the row group's rows linearly with no overlap.
- A row group's segment chain for one column is **independent** of any other row group — different boundaries, different codecs.
- Constant segments may have `block_id = -1` (the value lives in stats, not on disk).

### Block layout 
Each segment's payload lives in DuckDB's BlockManager-allocated blocks (256 KB each by default). The walker records three fields per segment that downstream I/O resolves into byte ranges:

- `block_id` — the **primary** block holding the segment payload. May be `-1` for blockless layouts (e.g. Constant segments, whose value lives in stats and not on disk).
- `block_offset` — byte offset within the primary block where this segment's compressed payload starts. Multiple small segments can share a single block (each at its own offset).
- `additional_blocks` — overflow blocks for variable-width payloads.

Variable-width compression breaks the "one block per segment" assumption. **FSST** has a symbol table separate from the per-row symbol indices; **Dictionary** has a string heap separate from the index column. When the auxiliary structure won't fit in the primary block, DuckDB allocates more blocks and lists their ids in `additional_blocks`. The downstream scan operator must read **all** of `[block_id] + additional_blocks` to assemble the segment's bytes before dispatching decompression.

For example:

```
Segment 3  (Dictionary, "l_orderkey", row group 0)
  block_id          = 44               primary: per-row dictionary indices
  additional_blocks = [45, 46]         overflow: the string heap
  block_offset      = 0                starts at the top of block 44

  Scan operator (follow-up PR) does, roughly:
    bytes  = read(file, block_44 + offset_0, segment_compressed_size)
           ‖ read(file, block_45 + 0, full_block)
           ‖ read(file, block_46 + 0, full_block)
    then dispatches Dictionary decode against the assembled buffer.
```

The walker just records the ids. Block-id → file-byte-range conversion is the BlockManager's job and lives at the I/O boundary in the follow-up scan-operator PR.

## How metadata is filled

```
                 ┌──────────────────────────────────────────┐
                 │            duckdb::DataTable             │
                 └──────────────────────────────────────────┘
                       │                              │
   GetColumnSegmentInfo│                              │GetPartitionStats
        (per segment)  ▼                              ▼ (per row group)
   ┌──────────────────────────┐         ┌──────────────────────────┐
   │ ColumnSegmentInfo        │         │ PartitionStatistics      │
   │  • column_id             │         │  • row_start             │
   │  • row_group_index       │         │  • partition_row_group ──┼─► GetColumnStatistics
   │  • column_path "[c]"     │         └──────────────────────────┘    └─► StringStats::
   │      vs "[c, 0]"         │                                              MaxStringLength
   │      ▲                   │                                                │
   │      └─ comma rule:      │                                                │ (VARCHAR only,
   │         data vs validity │                                                │  dict-family
   │  • block_id + add'l      │                                                │  codecs need it)
   │  • block_offset          │                                                │
   │  • segment_start, count  │                                                │
   │  • compression: string ──┼─► reverse-map → CompressionType enum           │
   └──────────────────────────┘                                                │
              │                                                                │
              │  bucket by (rg_idx, projected_col, data|validity)              │
              │  ←──────────── max_string_length ──────────────────────────────┘
              ▼
   ┌────────────────────────────────────────────────────────┐
   │  duckdb_native_metadata                                │
   │   row_groups[]                                         │
   │     ├ row_group_index, row_count                       │
   │     ├ row_group_start          ◄── rowid synthesis     │
   │     ├ decoded_bytes_budget     ◄── split-batch sizing  │
   │     │  + _is_lower_bound       ◄── unknown-VARCHAR flag│
   │     └ columns[]   (parallel to projected_cols)         │
   │         ├ data_segments[]       (sorted by start)      │
   │         │    duckdb_segment_descriptor                 │
   │         │     block_id + additional_blocks             │
   │         │     block_offset                             │
   │         │     segment_start, segment_count             │
   │         │     compression  (enum)                      │
   │         │     max_string_length                        │
   │         ├ validity_segments[]  (sorted by start)       │
   │         └ is_rowid     ◄── synth BIGINT range, no segs │
   │   viable + viability_failure_reason                    │
   └────────────────────────────────────────────────────────┘
```

Two DuckDB-internal contracts the walker relies on:

- **column_path comma rule** — `StandardColumnData::GetColumnSegmentInfo` recurses into the validity child with `col_path.push_back(0)`, so `"[c, 0]"` is the validity column and `"[c]"` is data. A single byte scan distinguishes the two.
- **PartitionStats iteration order** — matches `RowGroupCollection::SegmentNodes()` at v1.5.2; relied on for indexing the handles vector by `row_group_index`. The explicit per-column sort on `segment_start` guards against future ordering drift.

## Viability gates (first-violation bail)

| Bucket | Accepted | Rejected |
|---|---|---|
| **Data compression** | Uncompressed, Constant, RLE, Dictionary, BitPacking, FSST, DICT_FSST, ALP, ALPRD | Everything else (ZSTD-via-PRAGMA, PFOR, Chimp, Patas, …) |
| **Validity compression** | Uncompressed, Empty, Constant (all-valid), Roaring (host-decoded later) | Everything else |
| **Logical type** | All scalar `sirius::type_id` enumerators except those listed | HUGEINT/UHUGEINT (no 128-bit decode), DECIMAL precision >18 (DECIMAL128 storage), STRUCT/LIST (nested), INVALID/SQLNULL (sentinels) |
| **VARCHAR extra** | Uncompressed VARCHAR with no max-length stat → falls back to 256B/row and flips `_is_lower_bound` | Dictionary/FSST/DICT_FSST VARCHAR with no max-length stat → refused (host-side pre-decode buffer sizing would OOB) |

## Out of scope (deferred to follow-up PRs)

- Roaring validity host-decode (needs the sirius_io substrate to read the bytes)
- `table_filter` pruning (lives in the upcoming split provider)
- Operator-level escape gates — `dynamic_filters`, sample options, virtual columns — are caller-side checks on the `LogicalGet`, applied by the wiring PR
- Block-id → byte-range conversion is not in this PR; that happens at the sirius_io boundary in a follow-up.

## Tests

Add 13 cases with 100 assertions.

Refusal coverage: empty projection, parallel-vector mismatch, HUGEINT, DECIMAL128, STRUCT, LIST, ZSTD-via-PRAGMA force_compression.

Plus: rowid sentinel emission, validity-segment separation by `column_path`, and a `CompressionTypeToString` reverse-map round-trip drift guard — if upstream renames a codec string, the test breaks loudly rather than the walker silently mis-mapping.

### DIFF
--- a/src/op/scan/duckdb_native_metadata.cpp
+++ b/src/op/scan/duckdb_native_metadata.cpp
@@ -35,8 +35,8 @@ namespace sirius::op::scan {
 
 namespace {
 
-// Strings sourced from `duckdb/src/common/enums/compression_type.cpp` at v1.5.2.
-// Note: "Empty Validity" (NOT "Empty") is the canonical spelling.
+// Keys sourced from `duckdb/common/enums/compression_type.cpp` at v1.5.2.
+// "Empty Validity" (not "Empty") is the canonical spelling for COMPRESSION_EMPTY.
 const std::unordered_map<std::string, duckdb::CompressionType>& compression_string_to_enum()
 {
   static const std::unordered_map<std::string, duckdb::CompressionType> map = {
@@ -60,24 +60,25 @@ const std::unordered_map<std::string, duckdb::CompressionType>& compression_stri
   return map;
 }
 
+// Unrecognized strings map to COMPRESSION_COUNT, DuckDB's trailing sentinel
+// — `is_supported_*` rejects it via the default arm.
 duckdb::CompressionType parse_compression_string(const std::string& s)
 {
   const auto& map = compression_string_to_enum();
   auto it         = map.find(s);
-  if (it == map.end()) { return duckdb::CompressionType::COMPRESSION_AUTO; }
+  if (it == map.end()) { return duckdb::CompressionType::COMPRESSION_COUNT; }
   return it->second;
 }
 
-// `column_path` formatter:
 // `StandardColumnData::GetColumnSegmentInfo` recurses into the validity child
-// with `col_path.push_back(0)`, producing "[col, 0]" vs "[col]" for data.
+// with `col_path.push_back(0)` — "[col, 0]" for validity, "[col]" for data.
 bool is_validity_path(const std::string& column_path)
 {
   return column_path.find(',') != std::string::npos;
 }
 
-// Complete switch — every `sirius::type_id` must appear so a future
-// enumerator becomes a compile error rather than a silent accept.
+// Exhaustive switch: a new `sirius::type_id` enumerator should compile-fail
+// here rather than be silently accepted.
 bool is_supported_logical_type(const sirius::logical_type& type, std::string& reason_out)
 {
   switch (type.id()) {
@@ -142,9 +143,8 @@ bool is_supported_data_compression(duckdb::CompressionType c)
 bool is_supported_validity_compression(duckdb::CompressionType c)
 {
   switch (c) {
-    // CONSTANT here is in practice the all-valid case (genuinely all-null
-    // columns get EMPTY); the decoder's all-valid pre-fill covers it.
-    // ROARING is host-decoded to a regular bitmap before the GPU sees it.
+    // CONSTANT is the all-valid case (all-null columns land in EMPTY).
+    // ROARING is host-decoded to a plain bitmap before the GPU sees it.
     case duckdb::CompressionType::COMPRESSION_UNCOMPRESSED:
     case duckdb::CompressionType::COMPRESSION_EMPTY:
     case duckdb::CompressionType::COMPRESSION_CONSTANT:
@@ -189,8 +189,7 @@ class varchar_max_length_resolver {
   }
 
  private:
-  // col_id is narrowed to uint32 — a schema with >2^32 columns would alias
-  // in the cache. Realistic schemas are far below that.
+  // col_id is narrowed to uint32; >2^32 columns would alias.
   static std::uint64_t key(duckdb::idx_t rg, duckdb::idx_t col)
   {
     return (static_cast<std::uint64_t>(rg) << 32) | static_cast<std::uint32_t>(col);
@@ -216,8 +215,8 @@ duckdb_segment_descriptor build_segment_descriptor(const duckdb::ColumnSegmentIn
   return desc;
 }
 
-// PartitionStatistics::count is the source of truth: a rowid-only
-// projection has no data segments to sum, so segment-sum would land at 0.
+// Use PartitionStatistics::count as the source of truth so rowid-only
+// projections (no data segments to sum) still get a real row_count.
 void compute_row_counts(duckdb_native_metadata& md,
                         const std::vector<duckdb::PartitionStatistics>& partition_stats)
 {
@@ -242,11 +241,9 @@ void compute_row_counts(duckdb_native_metadata& md,
   }
 }
 
-// VARCHAR with no advertised max-string-length falls back to
-// VARCHAR_UNKNOWN_LENGTH_FALLBACK_BYTES per row and flips the row group's
-// `decoded_bytes_budget_is_lower_bound`. Without the flag, downstream
-// split-batch sizing would silently under-count when the actual strings
-// exceed the fallback.
+// VARCHAR columns whose row group did not advertise a max-string-length
+// stat fall back to VARCHAR_UNKNOWN_LENGTH_FALLBACK_BYTES per row and the
+// row group's `decoded_bytes_budget_is_lower_bound` is set.
 void compute_decoded_byte_budgets(duckdb_native_metadata& md,
                                   const std::vector<sirius::logical_type>& projected_types)
 {
@@ -279,9 +276,9 @@ void compute_decoded_byte_budgets(duckdb_native_metadata& md,
   }
 }
 
-// The walker over-allocates entries up to `max_row_group_index + 1`, so
-// row groups with all projected segments filtered out arrive here at 0.
-// Rowid-only entries are legitimately 0-data and are kept.
+// The walker over-allocates to `max_row_group_index + 1`; row groups
+// whose every projected segment was filtered out arrive here with
+// row_count == 0. Rowid-only entries are kept.
 void drop_empty_trailing_row_groups(duckdb_native_metadata& md)
 {
   while (!md.row_groups.empty() && md.row_groups.back().row_count == 0) {
@@ -308,12 +305,17 @@ duckdb_native_metadata walk_duckdb_native_metadata(
   duckdb_native_metadata md;
   md.viable = false;
 
+  auto refuse = [&md](std::string reason) {
+    md.viability_failure_reason = std::move(reason);
+    SIRIUS_LOG_DEBUG("[duckdb_native_metadata] refused: {}", md.viability_failure_reason);
+  };
+
   if (projected_cols.empty()) {
-    md.viability_failure_reason = "no projected columns";
+    refuse("no projected columns");
     return md;
   }
   if (projected_cols.size() != projected_types.size()) {
-    md.viability_failure_reason = "projected_cols and projected_types size mismatch";
+    refuse("projected_cols and projected_types size mismatch");
     return md;
   }
 
@@ -321,9 +323,8 @@ duckdb_native_metadata walk_duckdb_native_metadata(
     if (projected_cols[ci].is_rowid) { continue; }
     std::string reason;
     if (!is_supported_logical_type(projected_types[ci], reason)) {
-      md.viability_failure_reason =
-        "column " + std::to_string(projected_cols[ci].storage_idx.GetPrimaryIndex()) + ": " +
-        reason;
+      refuse("column " + std::to_string(projected_cols[ci].storage_idx.GetPrimaryIndex()) + ": " +
+             reason);
       return md;
     }
   }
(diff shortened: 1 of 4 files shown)


src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -0,0 +1,442 @@
...
+      if (col_md.is_rowid) { continue; }
+      duckdb::idx_t col_count = 0;
+      for (const auto& d : col_md.data_segments) {
+        col_count += d.segment_count;
+      }
+      if (col_count > 0) {
+        row_count = col_count;
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
ducndh: That is correct fix it with a follow up commit with an added test

test/cpp/scan/test_duckdb_native_walker.cpp
```diff
@@ -0,0 +1,344 @@
...
+duckdb::DataTable& get_storage(duckdb::Connection& con, const std::string& table_name)
+{
+  REQUIRE(con.Query("BEGIN TRANSACTION"));
+  auto& ctx     = *con.context;
+  auto& catalog = duckdb::Catalog::GetCatalog(ctx, "");
+  duckdb::CatalogTransaction txn(catalog, ctx);
+  auto& schema = catalog.GetSchema(txn, "main");
+  auto entry   = schema.GetEntry(txn, duckdb::CatalogType::TABLE_ENTRY, table_name);
+  REQUIRE(entry);
+  return entry->Cast<duckdb::DuckTableEntry>().GetStorage();
```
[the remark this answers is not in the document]
ducndh: change the test set up but the naming could use some helps

test/cpp/scan/test_duckdb_native_walker.cpp
```diff
@@ -0,0 +1,344 @@
...
+    REQUIRE_FALSE(rg.columns[0].is_rowid);
+    REQUIRE(rg.columns[1].is_rowid);
+    REQUIRE(rg.columns[1].data_segments.empty());
+    REQUIRE(rg.columns[1].validity_segments.empty());
+    // Rowid contributes 8 bytes per row to the budget.
+    REQUIRE(rg.decoded_bytes_budget >=
+            static_cast<std::size_t>(rg.row_count) * sizeof(std::int64_t));
+  }
+}
+
```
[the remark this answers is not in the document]
ducndh: Yep fixed this as per previous comment

src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -0,0 +1,447 @@
...
+    {"PFOR", duckdb::CompressionType::COMPRESSION_PFOR_DELTA},
+    {"BitPacking", duckdb::CompressionType::COMPRESSION_BITPACKING},
+    {"FSST", duckdb::CompressionType::COMPRESSION_FSST},
+    {"Chimp", duckdb::CompressionType::COMPRESSION_CHIMP},
+    {"Patas", duckdb::CompressionType::COMPRESSION_PATAS},
+    {"ZSTD", duckdb::CompressionType::COMPRESSION_ZSTD},
+    {"ALP", duckdb::CompressionType::COMPRESSION_ALP},
+    {"ALPRD", duckdb::CompressionType::COMPRESSION_ALPRD},
+    {"Roaring", duckdb::CompressionType::COMPRESSION_ROARING},
+    {"DICT_FSST", duckdb::CompressionType::COMPRESSION_DICT_FSST},
```
joosthooz, line 57: What is the difference between FSST, DICT_FSST and Dictionary? Are these all duckdb internal enum values?

src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -0,0 +1,447 @@
...
+    auto& col_md = md.row_groups[rg_idx].columns[ci];
+    if (validity_seg) {
+      col_md.validity_segments.push_back(std::move(desc));
+    } else {
+      col_md.data_segments.push_back(std::move(desc));
+    }
+  }
+
+  // DuckDB's iteration is already monotonic at v1.5.2 — the explicit sort
+  // guards against future ordering drift.
```
joosthooz: What does this mean?

src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -0,0 +1,447 @@
...
+    {"Empty Validity", duckdb::CompressionType::COMPRESSION_EMPTY},
+  };
+  return map;
+}
+
+duckdb::CompressionType parse_compression_string(const std::string& s)
+{
+  const auto& map = compression_string_to_enum();
+  auto it         = map.find(s);
+  if (it == map.end()) { return duckdb::CompressionType::COMPRESSION_AUTO; }
```
joosthooz: Why return AUTO here, isn't there an UNKNOWN or something like that?

src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -0,0 +1,447 @@
...
+    }
+  }
+
+  // PartitionStatistics order matches `RowGroupCollection::SegmentNodes()`
+  // iteration order at v1.5.2 — relied on for indexing by row_group_index.
+  auto partition_stats = storage.GetPartitionStats(context);
+  row_group_handles handles;
+  handles.reserve(partition_stats.size());
+  for (auto& ps : partition_stats) {
+    duckdb::idx_t row_start = ps.row_start.IsValid() ? ps.row_start.GetIndex() : 0;
```
joosthooz: Consider checking if there is a value in `ps.row_start.GetIndex()` while `IsValid()` is false, and then failing/warning

src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -0,0 +1,447 @@
...
+    }
+  }
+
+  compute_row_counts(md, partition_stats);
+  compute_decoded_byte_budgets(md, projected_types);
+  drop_empty_trailing_row_groups(md);
+
+  md.viable = true;
+  SIRIUS_LOG_DEBUG(
+    "[duckdb_native_metadata] walked {} row groups across {} projected columns; viable=true",
```
joosthooz, line 449: Is the `viability_failure_reason` being logged somewhere higher in the call stack? Otherwise consider adding a debug here at the places where the viability check fails.

test/cpp/scan/test_duckdb_native_walker.cpp
```diff
@@ -0,0 +1,387 @@
...
+  auto md = walk_duckdb_native_metadata(storage, *con.context, cols, ts);
+  REQUIRE_FALSE(md.viable);
+  REQUIRE(md.viability_failure_reason.find("LIST") != std::string::npos);
+}
+
+TEST_CASE("walker refuses unsupported data compression (force ZSTD)",
+          "[scan][duckdb_native_walker]")
+{
+  auto [db_owner, con] = sirius::make_test_db_and_connection();
+  exec_ok(con, "PRAGMA force_compression='zstd'");
```
joosthooz: Not sure how to do it, but it would be nice if this test could catch it if this pragma would be ignored

joosthooz (commented):
This looks good, would just like to hear your thoughts on a couple of small comments, but ready to approve

src/op/scan/duckdb_native_metadata.cpp
ducndh: All three are DuckDB's CompressionType enum values (duckdb/common/enums/compression_type.hpp) with DICT_FSST is just the DICT and FSST stacked: dictionary indices, with FSST-encoded dictionary entries.

Each needs its own decode path, so all three appear in the supported-data list

src/op/scan/duckdb_native_metadata.cpp
ducndh: Future proof that the segment_id list will always be sorted with passing to down stream. Add a comment to make it more explicit or I can delete this

src/op/scan/duckdb_native_metadata.cpp
ducndh: DuckDB has no UNKNOWN. I am going to switch to COMPRESSION_COUNT, DuckDB's documented trailing sentinel ("This has to stay the last entry of the type!"). It makes more sense than AUTO and is_supported_* still rejects it via the default arm.

src/op/scan/duckdb_native_metadata.cpp
ducndh: Yah add a log warn here

src/op/scan/duckdb_native_metadata.cpp
ducndh: It wasn't logged. I take the change to refactor every refusal site through a small refuse(reason) lambda that sets viability_failure_reason and emits SIRIUS_LOG_DEBUG and covers all 5 bail-out points too

test/cpp/scan/test_duckdb_native_walker.cpp
ducndh: Forcing ZSTD was more complicated than I thought so I choose another route to test for unknown compressoin in the future

joosthooz (approved):

PR #744, opened 2026-05-04
Patch use of cudaMemcpyBatchAsync for CUDA < 12.8

mike-wendt:
Fixes our CUDA 12 distribution pipelines that have been broken since the merge of #675 that introduced the usage of `cudaMemcpyBatchAsync`. `cudaMemcpyBatchAsync` was introduced in CUDA 12.8 and this patch ensures that for CUDA 12.X compatibility we fall back to sequential stream-ordered copies on CUDA < 12.8

### DIFF
--- a/src/io/sirius_datasource.cpp
+++ b/src/io/sirius_datasource.cpp
@@ -99,6 +99,8 @@ size_t copy_pinned_slices_to_device(
     copied += n;
   }
 
+#if CUDA_VERSION >= 12080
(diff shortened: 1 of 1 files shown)


mike-wendt:
A CUDA 12 [pipeline](https://github.com/sirius-db/sirius/actions/runs/25225667519/job/73968542458#step:26:3838) that is impacted

bwyogatama (approved):

PR #746, opened 2026-05-05
s3datasource foundation - (1/3) of S3 to parquet scan path 

ran-yuan-rui:
## Context

This is **PR1 of 3** that splits the work in #742 , per the review feedback @aminaramoon 

The split:
- **PR1 (this PR)** — backend-neutral foundation: URI parser + datasource_factory + datasource_registry + helper. 
- **PR2** — `credential_provider` abstraction (S3-scoped, presigned-URL boundary) + sirius's hand-rolled SigV4 as the default impl. Independent of PR1.
- **PR3** — S3 reactor (`s3_io_object`, signing-blind `s3_ioctx`, MinIO fixtures). Depends on PR1 + PR2.
- **Integration** — registry placement on `SiriusContext`, SET handlers, scan-path dispatch, and connector error handling are explicitly out of these 3 PRs here; will wait for @aminaramoon changes later to handle.

## What's in this PR

`namespace sirius::io`:

- **`parse(uri)`** — RFC-3986-aligned URI parser. Bare absolute paths route to `file` scheme; `file://`-qualified URIs accept absolute paths only; object-store URIs split into `host` (bucket) + `path` (key). Rejects bare relative paths, malformed input, and bad percent-escapes.
- **`datasource_registry`** — thread-safe (`shared_mutex`) `scheme → shared_ptr<sirius_ioctx>` map. Populated externally by integration code; consumed read-only by the factory.
- **`datasource_factory::create(uri, registry, config)`** — strict entry point. Local files (bare absolute, `file://`) go to `cudf::io::datasource::create` — pre-PR3 baseline preserved. Object-store schemes look up the registered ioctx but throw `"object construction is not yet implemented"` until a backend PR fills in the construction step.
- **`datasource_factory::create_for_parquet_scan(uri, registry, config)`** — lenient entry point for the parquet-scan path. Bare relative paths (DuckDB iceberg / hive fixtures still hand these out) bypass the strict factory and go straight to cudf default; everything else delegates to `create()`.
- **`object_store_config`** — POD config carrier (endpoint / region / access_key / secret_key / `s3_transport` enum). Inert in this PR; first consumer arrives with the S3 backend / integration PR.
- **`extract_scheme()` / `extract_path()`** — thin compatibility shims over `parse()`.

## Reviewable-boundary 

- **No scheme-aware branches in the factory.** Dispatch is pure registry lookup; the factory never says `if (scheme == "s3") …`.
- **The factory never instantiates `s3_ioctx`** (or any backend ioctx). Backends are pre-registered by integration code; the factory only consults the registry.
- **No backend-specific includes** (no `io/s3/*.hpp` from PR1 sources).
- **Locked by a regression test** (`datasource_factory PR1 does not construct object-store datasources` in `test/cpp/io/test_datasource_factory.cpp`): registers a `mock_ioctx` against `s3`, calls the factory, asserts it throws *and* `mock_ioctx::make_datasource_calls == 0`.

## What's deliberately NOT here

- S3 backend (`s3_io_object`, `s3_ioctx`, libcurl + SigV4 plumbing) — **PR3**.
- `credential_provider` abstraction + default SigV4 impl + `presign_url` — **PR2**.
- `sirius_engine` mutations, registry placement on `SiriusContext`, SET handlers (`s3_endpoint` / `s3_region` / `s3_access_key` / `s3_secret_key` / `s3_transport`), `sirius_read_parquet` SQL function + bind-time materializer, scan-path dispatch through `parquet_scan_info`, and `split_connector::set_error` connector error API — **integration PR** (reviewer-owned per the comment thread).

## Test plan

Targeted suite passing on Linux CI host:

```bash
build/release/extension/sirius/test/cpp/sirius_unittest \
  "[uri_parser],[datasource_factory],[object_store_config]"
Result: 20/20 test cases, 101 assertions, all passing.

### CODE REVIEW

**Comment by ran-yuan-rui** [comment]
@mike-wendt Today I saw 3 time hang for the testing on different places and I downloaded one failure which has current finding: the failure does not look related to my S3 datasource foundation tests.   The new PR tests all ran and passed before the job got stuck:

- object_store_config tests ran at [18/1026] - [21/1026]
- datasource_factory tests ran at [298/1026] - [306/1026]
- uri_parser tests ran at [307/1026] - [313/1026]

The C++ unit test step continued past those and later stalled in an existing GPU integration test area. The one printed Catch2 test was:
[396/1026] gpu_execution - basic left join 1 parquet
After that there was no further test output for almost an hour. The GitHub Actions test job has timeout-minutes: 60, and the runner eventually canceled the step:
The operation was canceled.
Terminate orphan process: ... sirius_unittest

So my current judgment from the logs is: this is a timeout / hang in the existing gpu_execution parquet left join integration test, not an assertion failure and not a failure in the my PR.

Could you help rerun the testing job or check whether the GPU T4 runner had a transient issue? If it reproduces, the likely isolation command is:
./build/release/extension/sirius/test/cpp/sirius_unittest "gpu_execution - basic left join 1 parquet" --abort

File: src/io/uri_parser.cpp [hunk 1]
```diff
@@ -0,0 +1,203 @@
...
+    auto host_sv = (slash == std::string_view::npos) ? authpath : authpath.substr(0, slash);
+    auto key_sv  = (slash == std::string_view::npos) ? std::string_view{} : authpath.substr(slash);
+    if (host_sv.empty()) fail("empty host", uri);
+    // Why: do not percent-decode host; decoding could smuggle a '/' via %2F and
+    // change the authority/path boundary after parsing.
+    out.host = std::string{host_sv};
+
+    // Collapse any run of leading slashes to 0 slashes.
+    // Why: s3://bucket/key and s3://bucket//key must both yield key="key";
+    // object stores treat them as identical.
```
bwyogatama: Double checking this. Is this always true? I thought in S3-style object store, they treat / as character in the object key

src/io/uri_parser.cpp
```diff
@@ -0,0 +1,203 @@
...
+  auto delim = uri.find(kSchemeDelim);
+  if (delim == std::string_view::npos)
+    fail("relative path not allowed; use absolute or scheme://", uri);
+  if (delim == 0) fail("empty scheme", uri);
+
+  auto scheme_sv = uri.substr(0, delim);
+  if (!is_valid_scheme(scheme_sv)) fail("invalid scheme syntax", uri);
+  // Why: RFC 3986 §3.1 — scheme is case-insensitive; normalize early so registry
+  // lookup can use exact string match.
+  out.scheme = to_lower(scheme_sv);
```
bwyogatama, line 161: We lowercase the scheme here but when i looked at `datasource_factory`, `register_ioctx` doesn't lowercase the scheme, it just stores the scheme verbatim. Will this cause inconsistencies?

bwyogatama (approved):

src/io/uri_parser.cpp
ran-yuan-rui: Yes - and I will do the change

src/io/uri_parser.cpp
ran-yuan-rui: Good catch. You're right that S3 treats `/` as part of the object key. I’ll change the parser to strip only the single URI separator after the bucket/authority and preserve the remaining leading slashes in the key, with tests for `s3://bucket//key` and `s3://bucket///key`.
s3://bucket/key      -> key
s3://bucket//key     -> /key
s3://bucket///key    -> //key
s3://bucket/a//b     -> a//b
s3://bucket/         -> throw empty object key

ran-yuan-rui:
> @mike-wendt Today I saw 3 time hang for the testing on different places and I downloaded one failure which has current finding: the failure does not look related to my S3 datasource foundation tests. The new PR tests all ran and passed before the job got stuck:
> 
> * object_store_config tests ran at [18/1026] - [21/1026]
> * datasource_factory tests ran at [298/1026] - [306/1026]
> * uri_parser tests ran at [307/1026] - [313/1026]
> 
> The C++ unit test step continued past those and later stalled in an existing GPU integration test area. The one printed Catch2 test was: [396/1026] gpu_execution - basic left join 1 parquet After that there was no further test output for almost an hour. The GitHub Actions test job has timeout-minutes: 60, and the runner eventually canceled the step: The operation was canceled. Terminate orphan process: ... sirius_unittest
> 
> So my current judgment from the logs is: this is a timeout / hang in the existing gpu_execution parquet left join integration test, not an assertion failure and not a failure in the my PR.
> 
> Could you help rerun the testing job or check whether the GPU T4 runner had a transient issue? If it reproduces, the likely isolation command is: ./build/release/extension/sirius/test/cpp/sirius_unittest "gpu_execution - basic left join 1 parquet" --abort

After rebase with PR757 - the hang fixed

src/io/uri_parser.cpp
```diff
@@ -0,0 +1,203 @@
...
+  out.reserve(s.size());
+  for (char c : s)
+    out.push_back(static_cast<char>(std::tolower(static_cast<unsigned char>(c))));
+  return out;
+}
+
+}  // namespace
+
+parsed_uri parse(std::string_view uri)
+{
```
aminaramoon, line 125: man I wish we had boost available, would've made it much easier to do this

src/include/io/datasource_factory.hpp
```diff
@@ -0,0 +1,193 @@
...
+   *
+   * The strict @c create keeps its parser-strict contract — prefer it for
+   * callers that should reject unscheme'd input as a real bug.
+   *
+   * @throw std::invalid_argument if the URI is empty (preserves strict
+   *                              @c create's empty-URI rejection).
+   * @throw std::runtime_error    on object-store dispatch failure (same
+   *                              as strict @c create).
+   */
+  static std::unique_ptr<cudf::io::datasource> create_for_parquet_scan(
```
aminaramoon, line 177: I am not sure why we have create_for_parquet_scan here ? can you give me an example why it is needed, especailly given that the signiture is the same as the main factory function

aminaramoon (commented):
a few minor comments. overall I think this is a workable solution. Thanks

src/io/uri_parser.cpp
ran-yuan-rui: Same — `boost::url` would handle it easier

src/include/io/datasource_factory.hpp
ran-yuan-rui: Two reasons for the separate entry point:

1. **Bare relative paths.** DuckDB's iceberg and hive test fixtures hand out
   paths like `test/cpp/integration/data/lineitem.parquet` — no leading `/`,
   no `scheme://`. The strict URI parser deliberately rejects those per
   RFC 3986 §3 (relative references need an explicit base, and the scan path
   has no base to plug in). `create_for_parquet_scan` bypasses the parser
   for that one case and hits cudf's default datasource directly; everything
   else delegates to strict `create()`.

2. **Strict callers shouldn't lose the validation safety net.** Other call
   sites — e.g. the `sirius_read_parquet` SQL function in your integration
   PR's territory — should reject malformed input as a real bug. 

Concrete behavior:

| Input                           | `create`              | `create_for_parquet_scan`      |
|---------------------------------|-----------------------|--------------------------------|
| `/abs/foo.parquet`              | parser → cudf default | delegates to `create`          |
| `file:///abs/foo.parquet`       | parser → cudf default | delegates to `create`          |
| `s3://bucket/key`               | parser → registry     | delegates to `create`          |
| `foo.parquet` (relative bare)   | throws                | bypasses parser → cudf default |

Identical signature is intentional — callers can swap one for the other
without changing arguments. The only behavior delta is the relative-path
shortcut.

PR #749, opened 2026-05-05
scan_manager: polymorphic scan_info + cache-extracted factory (DuckDB format GPU-native Part 8a)

ducndh:
Delivers the design proposed in PR #731 review (a polymorphic scan_info base with virtual make_provider, replacing the parquet-typed factory body) at the moment it stays cheap — the manager still has one consumer (parquet), so this PR is purely structural.

Three coordinated changes:

1. New op::scan::scan_info polymorphic base (src/include/op/scan/scan_info.hpp). Carries the fields common to every file-format scan (file_paths, column_ids, projection_ids, names, returned_types, partition_indices, table_filters, approximate_batch_size, scan_output_arity). Subclasses add format-specific fields and implement make_provider() to construct their split_provider.

2. parquet_scan_info derives from scan_info; its make_provider() builds a parquet_split_provider (logic moved out of sirius_scan_manager into the new src/op/scan/parquet_scan_info.cpp).

3. sirius_scan_manager::create_provider_for splits the cache short-circuit into a separate try_cached_for(scan_info&, op_id) pre-step that uses only the common base fields (so it stays format-agnostic) and dispatches format-specific construction through info->make_provider(*this).

Behavior should be unchanged on the parquet path. Operator stores unique_ptr<scan_info> instead of unique_ptr<parquet_scan_info>; pipeline converter sets the new scan_output_arity field from scan_op.types.size() so the manager can build scan_plan without reaching into the operator.

This unblocks adding the iceberg path as a sibling subclass — iceberg's follow-up PR can add iceberg_scan_info : scan_info + iceberg_split_provider without touching the factory and also preparation for the DuckDb subclass to do the same thing

### DIFF
--- a/src/include/op/scan/scan_info.hpp
+++ b/src/include/op/scan/scan_info.hpp
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2025, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "helper/logical_type.hpp"
+#include "sirius_config.hpp"
+
+#include <duckdb/common/column_index.hpp>
+#include <duckdb/common/multi_file/multi_file_data.hpp>
+#include <duckdb/common/types.hpp>
+#include <duckdb/common/vector.hpp>
+#include <duckdb/planner/table_filter.hpp>
+
+#include <cstddef>
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace sirius::scan_manager {
+class sirius_scan_manager;
+class split_provider;
+}  // namespace sirius::scan_manager
+
+namespace sirius::op::scan {
+
+/**
+ * @brief Polymorphic base for per-format bind-data parked on a gpu scan operator.
+ *
+ * Carries the scan fields common to every file-format scan (file paths,
+ * projection, filters, partitions, output arity). Subclasses add format-specific
+ * fields (e.g. iceberg snapshot id, manifest paths) and implement @ref make_provider
+ * to construct the format's split_provider.
+ *
+ * The scan_manager reads scan_info during prepare_for_query to either short-circuit
+ * to a cached_split_provider (using only base fields, so format-agnostic) or
+ * dispatch through @ref make_provider to the format-specific provider.
+ *
+ * Populated once by the pipeline converter, then moved into the operator. Consumed
+ * once by the scan_manager via the operator's take_scan_info().
+ */
+struct scan_info {
+  /// Logical types of the scan source's columns, in source schema order.
+  duckdb::vector<sirius::logical_type> returned_types;
+  std::vector<std::string> file_paths;
+  /// Column ids exposed by the table function.
+  duckdb::vector<duckdb::ColumnIndex> column_ids;
+  /// Indices into @ref column_ids that the planner projected (empty = read all columns).
+  duckdb::vector<duckdb::idx_t> projection_ids;
+  /// Column names in source schema order.
+  duckdb::vector<std::string> names;
+  /// Filter set for row-group pruning / pushdown. May be null.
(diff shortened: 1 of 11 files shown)


kevkrist (approved):
Looks good to me. Thanks!

src/include/op/scan/scan_info.hpp
```diff
@@ -0,0 +1,95 @@
...
+  /// Column names in source schema order.
+  duckdb::vector<std::string> names;
+  /// Filter set for row-group pruning / pushdown. May be null.
+  duckdb::unique_ptr<duckdb::TableFilterSet> table_filters;
+  duckdb::vector<duckdb::HivePartitioningIndex> partition_indices;
+  /// Target uncompressed bytes per row-group partition.
+  std::size_t approximate_batch_size = sirius::config::DEFAULT_SCAN_TASK_BATCH_SIZE;
+  /// Number of output columns the gpu scan operator returns.
+  /// Set by the pipeline converter from the scan operator's `types.size()`.
+  std::size_t scan_output_arity = 0;
```
kevkrist: Consider deleting this default initialization. Is there a reason it's there? Seems like it could cause some headaches if not explicitly initialized during construction.

src/include/scan_manager/sirius_scan_manager.hpp
```diff
@@ -221,14 +222,33 @@ class sirius_scan_manager {
...
+  /// \brief Try to short-circuit @p info to a cached_split_provider when a
+  ///        pinned entry matches the scan's file paths. Returns nullptr on
+  ///        miss so the caller can fall through to per-format dispatch.
+  ///        Format-agnostic: uses only fields on the scan_info base.
+  ///
+  /// @param info   Scan bind-data — file_paths, column_ids, names, types,
+  ///               projection_ids, partition_indices, table_filters,
+  ///               scan_output_arity. Read-only; not consumed.
+  /// @param op_id  Operator id for diagnostic logging.
+  std::unique_ptr<split_provider> try_cached_for(op::scan::scan_info const& info,
```
kevkrist: This is kind of a weird name to me. Try cached for ... what?

src/include/op/scan/sirius_gpu_parquet_scan_operator.hpp
```diff
@@ -66,7 +67,7 @@ class sirius_gpu_parquet_scan_operator : public sirius_physical_operator {
    */
   sirius_gpu_parquet_scan_operator(duckdb::vector<sirius::logical_type> types,
                                    duckdb::idx_t estimated_cardinality,
-                                   std::unique_ptr<parquet_scan_info> scan_info);
+                                   std::unique_ptr<scan_info> scan_info);
```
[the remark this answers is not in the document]
kevkrist: I don't think the 2-arg constructor is used anywhere -- consider deleting.

src/include/op/scan/scan_info.hpp
```diff
@@ -0,0 +1,95 @@
...
+   * @brief Build the format-specific split_provider this scan_info dispatches to.
+   *
+   * Called once by the scan_manager after the cache short-circuit has been
+   * checked and missed. May consume mutable fields on @c *this (e.g. moving
+   * @ref table_filters into the constructed provider).
+   *
+   * @param manager  The scan_manager driving this query (provides the thread
+   *                 pool and any cross-format context the provider needs).
+   */
+  virtual std::unique_ptr<scan_manager::split_provider> make_provider(
```
kevkrist, line 98: It looks like all this method currently uses is the io context. Do you envision needing more than that for this function?

kevkrist (commented):
Some small requests/questions -- we are currently freezing merging until multi-gpu lands in dev.

src/include/op/scan/scan_info.hpp
ducndh: Yeah that was a bit sloppy from my side. This actually would cause silent failure instead of letting the error scream, which is a better default.

src/include/scan_manager/sirius_scan_manager.hpp
ducndh: Change the name hopefully it clearer now. it is just trying for the cached provider

src/include/op/scan/sirius_gpu_parquet_scan_operator.hpp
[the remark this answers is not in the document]
ducndh: Yah passing in null ptr for the third arg is just less headache for now in the test files

src/include/op/scan/scan_info.hpp
ducndh: not at the moment and you are right that this could be over engineering. the cost of expanding this in the future is cheap anyway so I will just delete this

ducndh:
@kevkrist  since the multi GPU changes scan task, do you think it is necessary to do another round of review after I resolve the conflict and do we need to ask other for review too?

kevkrist:
@ducndh Looks good to me

PR #750, opened 2026-05-05
added version distinction for cudaMemcopyBatchAsync usage

wmalpica:

PR #752, opened 2026-05-05
CI: Add CUDA 12 PR testing and use CUDA 12 runners for CUDA 12 distribution workflows

mike-wendt:
This week there have been two PRs merged that break `dev` and the Distribution workflows as they are the only workflow to compile and build CUDA 12 versions of Sirius. Both of these PRs used a function introduced to CUDA 12.8 and required version gating to support CUDA 12.x comptatibility

This PR makes two important changes to ensure we catch these issues before they are merged to `dev`

1. New PR testing with a parallel Test workflow that builds and tests CUDA 12
2. Updates the Distribution workflows to use a CUDA 12 Ubuntu 22.04 runner for building CUDA 12 Sirius
  - The pixi environment `cuda12` ensures we have a CUDA 12 environment even when the base image is CUDA 13, this is just to match the testing done in `1` for consistency
  - Also worth noting that CUDA 12 images do exist for Ubunutu 24.04, so we use 22.04 instead; our CUDA 13 build/test uses 24.04

### DIFF
--- a/.github/workflows/distribution.yml
+++ b/.github/workflows/distribution.yml
@@ -64,4 +64,4 @@ jobs:
       cuda_archs: '75-real;80-real;86-real;90a-real'
       cuda_version: '12'
       artifact_postfix: '-cuda12'
-      runners: '{"linux_x64": ["self-hosted", "ubuntu-24.04", "cuda-13", "cpu-xl"], "linux_arm64": ["self-hosted", "ubuntu-24.04", "cuda-13", "arm-xl"]}'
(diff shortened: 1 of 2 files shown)


mbrobbel (approved):

mike-wendt:
Looks like we need #750 to merge first so we can continue test. However, the current failure highlights this would stop us from merging a broken CUDA 12 build. That said, I do need to continue the test if the build is successful

mike-wendt:
#754 merged applying the same fix as in #750

mike-wendt:
Will retest and merge once #757 is merged

PR #757, opened 2026-05-06
Fix data batch refactor deadlock

wmalpica:
I am adding a new notification of downstream pipelines to trigger when a pipeline is updated to check that its finished even tough its already finished. 
The fact this is happening means we have a race condition which needs to be properly fixed, but this patches it for now


mike-wendt (approved):
Patch confirmed working, GPU testing working again

PR #758, opened 2026-05-06
s3datasource credential providers - (2/3) of S3 to parquet scan path 

ran-yuan-rui:
## Context

This is **PR2 of 3** that splits the work in #742.

The split:
- [PR1 (#746)](https://github.com/sirius-db/sirius/pull/746) — backend-neutral foundation (factory + registry + URI parser + inert config). **Independent of this PR; can be reviewed in parallel.**
- **PR2 (this PR)** — `credential_provider` abstraction (S3-scoped, presigned-URL seam) + Sirius's hand-rolled SigV4 as the default impl.
- **PR3** — S3 reactor (`s3_io_object`, signing-blind `s3_ioctx`, MinIO fixtures). Depends on PR1 + PR2.
- **Integration** — registry placement on `SiriusContext`, SET handlers, scan-path dispatch, connector error handling are out of scope here per the comment thread.

This branch is cut from `upstream/dev` directly (no PR1 ancestry); the two PRs share zero code-level dependencies.

## What's in this PR

`namespace sirius::io::s3`:

- **`credential_provider`** ([credential_provider.hpp](src/include/io/s3/credential_provider.hpp)) — abstract interface, single virtual `get_presigned_url(s3_object_ref, presign_method)`. Method-specific (presign_method::GET / HEAD only). Public surface is presign-only — no `get_credentials()` — so impls without raw key material (signed-URL services, broker chains) compose cleanly. Lets downstream projects plug in their own credential / signer implementation without forcing Sirius to depend on `aws-sdk-cpp`.
- **`static_credentials`** ([static_credentials.hpp](src/include/io/s3/static_credentials.hpp)) — POD: access_key_id / secret_access_key / session_token / optional expires_at. Consumed by the default impl's ctor; not part of the abstract interface.
- **`sirius_sigv4_credential_provider`** ([.hpp](src/include/io/s3/sirius_sigv4_credential_provider.hpp) / [.cpp](src/io/s3/sirius_sigv4_credential_provider.cpp)) — default implementation. Parses endpoint (scheme + host[:port]) at construction, validates creds / region / TTL, builds path-style canonical URI per call, delegates to `sigv4::presign_url`. Default TTL 5 min — inline-at-request-time presigning means short TTL is safe and limits leaked-URL blast radius.
- **`mock_credential_provider`** ([mock_credential_provider.hpp](src/include/io/s3/mock_credential_provider.hpp)) — header-only test fixture. Returns canned URL or throws on demand; thread-safe counters; used by the credential_provider tests, will be reused by PR3 reactor tests and downstream test suites.

`namespace sirius::io`:

- **`credential_error`** ([io_errors.hpp](src/include/io/io_errors.hpp)) — exception type for credential acquisition / signing failures. Future IO-layer error types share this header.

`namespace sirius::io::s3` (modifications to existing sigv4):

- **`presign_url(method, scheme, host, canonical_uri, creds, ts, ttl)`** — new public function on [sigv4.hpp](src/include/io/s3/sigv4.hpp). Builds a SigV4 query-string-signed URL with X-Amz-* params (Algorithm, Credential, Date, Expires, SignedHeaders, optional Security-Token, Signature). Uses `UNSIGNED-PAYLOAD`; signs only `host` so callers can freely add `Range` / `Accept` / etc.
- **`sigv4_signer_config`** — new `session_token` field (default empty). Both `sign_request` (header-based) and `presign_url` (query-string-based) auto-inject the security token when non-empty (covers temporary STS / IMDS / SSO credentials).

## Public seam: presigned URL, not raw credentials

Boundary by design. The `credential_provider` interface yields fully-qualified URLs; the reactor (PR3 onwards) issues HTTP requests against those URLs and never sees the underlying access key. Downstream forks can plug in:

- An AWS-SDK-backed provider (no extra dep on Sirius)
- An internal auth broker that issues signed URLs centrally
- An IMDS / STS chain
- Any future SSO scheme

without touching this PR's source.

## What's deliberately NOT here

- S3 reactor (`s3_io_object`, signing-blind `s3_ioctx`, MinIO fixtures) — **PR3**.
- `sirius_engine` / `SiriusContext` mutations, SET handlers, `sirius_read_parquet`, scan-path dispatch, connector error API — **integration PR** (reviewer-owned).
- AWS-SDK-cpp adapter — downstream's responsibility per the design.
- libcurl — only OpenSSL is added as a build dep here (already required by SigV4).

## Test plan

Targeted suite passing on Linux CI host:

```bash
build/release/extension/sirius/test/cpp/sirius_unittest \
  "[s3][sigv4],[s3][credential_provider]"
Result: 27 test cases / 893 assertions, all passing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -377,6 +377,9 @@ set(TEST_SOURCES
     test/cpp/expression_executor/test_gpu_expression_executor.cpp
     test/cpp/expression_executor/test_gpu_expression_translator.cpp
     test/cpp/helper/test_logical_type.cpp
+    test/cpp/io/s3/test_sigv4.cpp
+    test/cpp/io/s3/test_sirius_sigv4_credential_provider.cpp
+    test/cpp/io/s3/test_static_credentials.cpp
     test/cpp/io/test_datasource_factory.cpp
     test/cpp/io/test_uri_parser.cpp
     test/cpp/integration/test_gpu_execution_multi_format.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,7 @@ find_package(spdlog REQUIRED CONFIG)
 find_package(yaml-cpp REQUIRED CONFIG)
 find_package(absl REQUIRED CONFIG)
 find_package(PkgConfig REQUIRED)
+find_package(OpenSSL REQUIRED)
 pkg_check_modules(NUMA REQUIRED IMPORTED_TARGET numa)
 pkg_check_modules(LIBURING REQUIRED IMPORTED_TARGET liburing)
 
@@ -164,6 +165,8 @@ set(EXTENSION_SOURCES
     src/io/admission_control.cpp
     src/io/datasource_factory.cpp
     src/io/prefetching_cache.cpp
+    src/io/s3/sigv4.cpp
+    src/io/s3/sirius_sigv4_credential_provider.cpp
     src/io/sirius_datasource.cpp
     src/io/uri_parser.cpp
     src/io/uring/uring_ioctx.cpp
@@ -330,9 +333,10 @@ endforeach()
 
 # Additional libraries only needed by the static extension
 target_link_libraries(sirius_extension PkgConfig::NUMA PkgConfig::LIBURING
-                      yaml-cpp::yaml-cpp absl::any_invocable)
+                      OpenSSL::Crypto yaml-cpp::yaml-cpp absl::any_invocable)
 
-target_link_libraries(sirius_loadable_extension PkgConfig::LIBURING)
+target_link_libraries(sirius_loadable_extension PkgConfig::LIBURING
+                      OpenSSL::Crypto)
 
 # Required by DuckDB's export set (duckdb_static -> sirius_extension ->
 # cucascade_static / telemetry_bridge)
--- a/src/include/io/io_errors.hpp
+++ b/src/include/io/io_errors.hpp
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2025, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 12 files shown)


aminaramoon (approved):
This looks great! I like the credential provider abstraction and it should allow other folks to hook into this datasource in engines besides sirius.

aminaramoon:
just a question, I wonder if we need to make some of these errors retriable in the engine! like with S3, often you get errors due to server being temporarily down and unresponsive or rate limiting causing some request to bounce. we definitely don't want to fail on these failures without retrying the,. I wonder if we need to create a sirius_error that has a retriability flag in it, I'd love to hear your take on this ?

ran-yuan-rui:
> just a question, I wonder if we need to make some of these errors retriable in the engine! like with S3, often you get errors due to server being temporarily down and unresponsive or rate limiting causing some request to bounce. we definitely don't want to fail on these failures without retrying the,. I wonder if we need to create a sirius_error that has a retriability flag in it, I'd love to hear your take on this ?

Thanks, and agreed. I think we do want a retriability-aware error path.

My take is that retry classification should live at the IO / connector layer — the S3 reactor / connector is where we see the request context, HTTP status, curl error code, idempotency, and timeout / rate-limit details.

For S3 reads, GET / HEAD are idempotent, so I'd expect something like:

- **retriable**: connect timeout, request timeout, temporary DNS / network failure, HTTP 408 / 429 / 500 / 502 / 503 / 504, S3 `SlowDown`
- **non-retriable**: malformed endpoint / config, invalid credentials,`SignatureDoesNotMatch`, `AccessDenied`, `NoSuchBucket` / `NoSuchKey` (unless the caller explicitly wants eventual-consistency retry semantics)

A `sirius_error` / `io_error` carrying an error kind plus `is_retriable()`sounds like the right shape — then the engine / connector can apply bounded retries with backoff without every backend inventing its own policy.

I will try address this in PR3 which will need wait for your 740 merge first

PR #759, opened 2026-05-06
feature-S3datasource-s3reactor - (3/3) of S3 to parquet scan path

ran-yuan-rui:
## Context

PR3 of 3 splitting #742 per the original review thread on that PR. Adds the libcurl-backed S3 reactor that consumes PR2's `credential_provider` abstraction. The reactor is **signing-blind** — never sees raw keys, never computes a SigV4 signature, never injects an Authorization header.

The split:
- [PR1 #746](https://github.com/sirius-db/sirius/pull/746) — backend-neutral foundation (factory + registry + URI parser).
- [PR2 #758](https://github.com/sirius-db/sirius/pull/758) — `credential_provider` abstraction + SigV4 default impl.
- **PR3 (this PR)** — S3 reactor + MinIO integration scaffolding.
- **Integration PR** (reviewer-owned) — wires these three onto `SiriusContext`. See "Integration contract" below for the concrete handoff.

This PR is **stacked on PR2 (#758)**. Independent of PR1 — no compile-time dependency on `datasource_factory` / `datasource_registry` / `object_store_config`.

## What's in this PR

`namespace sirius::io::s3`:

- **`s3_io_object`** ([.hpp](src/include/io/s3/s3_io_object.hpp)) — thin metadata holder (bucket / key / size). Inherits from `sirius_io_object` (PR #675). Cherry-picked verbatim from `merge-s3datasource @ 600e4ddb`.
- **`s3_ioctx`** ([.hpp](src/include/io/s3/s3_ioctx.hpp) / [.cpp](src/io/s3/s3_ioctx.cpp)) — libcurl HTTP Range GET reactor implementing the `sirius_ioctx` contract (PR #675). Per request: `auto url = _cfg.creds->get_presigned_url(obj, method); libcurl_GET(url)`. The only header it emits is `Range:` (unsigned by design — the URL's `SignedHeaders=host` deliberately leaves it free so callers can attach Range / Accept / etc. without invalidating the signature).

**Build wiring**:
- CMakeLists: `pkg_check_modules(CURL REQUIRED IMPORTED_TARGET libcurl)`; `s3_ioctx.cpp` registered under `EXTENSION_SOURCES`; `PkgConfig::CURL` linked into both `sirius_extension` and `sirius_loadable_extension`.
- pixi.toml: `openssl` and `libcurl` declared explicitly under `[feature.dev-libs.dependencies]` (previously satisfied transitively through cudf / spdlog; declaring them makes the dep contract hermetic).
- vcpkg.json: `openssl` and `curl` added to `dependencies` so vcpkg-hermetic builds resolve them rather than falling through to system libs.
- Makefile: `s3-up` / `s3-down` / `s3-test` / `s3-cpp-test` targets cherry-picked from `merge-s3datasource` (no other Makefile changes).
- `test/cpp/integration/s3/`: MinIO docker-compose + env / fixtures scripts + sirius.yaml + README, cherry-picked verbatim.

## Integration contract — what the integration PR needs from this stack

The detailed recipe (credentials → provider → reactor → `initialize_cache` → registry, plus symmetric teardown with the `shutdown()` must precede `clear()` discipline) lives in the `s3_ioctx` class doxygen at [`src/include/io/s3/s3_ioctx.hpp`](src/include/io/s3/s3_ioctx.hpp) and a runnable version in [`make_live_ioctx`](test/cpp/io/s3/test_s3_ioctx.cpp). **See the doxygen for code, lifetime contract, and pseudocode caveats.** The bullets below are the handoff checklist:

1. **Place `datasource_registry` on `SiriusContext`** (not `sirius_engine`); bootstrap in `initialize`, drain + clear in `terminate`.
2. **Materialize PR2 types** from `object_store_config`: `static_credentials` → `sirius_sigv4_credential_provider` → `s3_ioctx_config` → register `s3_ioctx` under `"s3"`.
3. **Restore the factory's S3 dispatch branch.** PR1's `datasource_factory::create` currently throws `"object construction is not yet implemented"`; integration PR adds the branch that calls `s3_ioctx::head_object_size` + constructs `s3_io_object` + `make_datasource`. (Alternative: virtual `make_io_object(parsed_uri)` on `sirius_ioctx` for scheme-agnostic dispatch.)
4. **Restore SET handlers** (`SetS3Endpoint` / `Region` / `AccessKey` / `SecretKey` / `Transport`) and optionally the `sirius_read_parquet` SQL function from `merge-s3datasource`.
5. **Connector error propagation** (`split_connector::set_error` + scan-manager) — reviewer signaled in #742 they'd redo this; PR3 doesn't touch it.
6. **Enable prefetching** by calling `initialize_cache(buffer_pool, ...)` once after `s3_ioctx` construction; cache + pinned chunks then inherit `SiriusContext`-scoped lifetime.
7. **`shared_ptr` discipline at call sites** — `s3_ioctx` and `s3_io_object` must always be held by `shared_ptr`; the reactor's async UAF safety depends on `shared_from_this()` working.
8. **Downstream credential providers** are out of scope for this stack — forks plug in their own `credential_provider` subclass (AWS SDK / IMDS / STS / SSO) with no source changes to Sirius.

---

## Tests

`[s3][ioctx]` (5 unit cases — no MinIO required):
- `s3_io_object` identity + `raw_file_cache_id` (`s3://bucket/key`).
- Config validation; **EOF clip happens before presigning** (`provider->call_count() == 0` after past-EOF reads, proving the clip short-circuits the round trip).
- Method-specific presigned URL routing — HEAD / GET each hit the provider with the right `presign_method`; mock provider throws to short-circuit libcurl.
- `compute_physical_range` clipping.
- **Async lifetime regression** — uses a `blocking_throwing_provider` that signals when the worker thread enters and waits to be released. Test schedules `host_read_async`, waits until the worker is mid-flight, drops the test scope's `shared_ptr<s3_ioctx>` and `shared_ptr<s3_io_object>`, then releases the provider. Verifies the handler is invoked with `credential_error` rather than UAF'ing. This case fails without the `shared_from_this()` capture fix.

`[s3][ioctx][integration]` (2 MinIO cases — gated on `SIRIUS_TEST_S3_*` env, `make s3-up` brings them up):
- Single-object full read / tail read / past-EOF return-zero against a real MinIO instance.
- Multi-range read with 4 ranges (including a small one and a tail one) compared against the local fixture.

---

## Build dependencies

- **libcurl** (new, declared via `pkg_check_modules(CURL REQUIRED IMPORTED_TARGET libcurl)`). Used by `s3_ioctx` only.
- **OpenSSL** — already required by sigv4 in PR2; this PR adds it to the explicit dep manifests (pixi.toml, vcpkg.json) since PR2 had relied on transitive resolution.

---


src/io/s3/s3_ioctx.cpp
```diff
@@ -0,0 +1,536 @@
...
+                               std::uint8_t* dst,
+                               io_completion_handler handler)
+{
+  // Capture shared_ptrs to keep both the ioctx and the io_object alive until
+  // the detached worker completes; otherwise a caller dropping the datasource
+  // (and its shared_ptr<sirius_io_object> / shared_ptr<sirius_ioctx>) before
+  // the future resolves would leave us with dangling `this` / `&obj`.
+  auto self      = shared_from_this();
+  auto obj_owner = obj.shared_from_this();
+  dispatch_async(std::move(handler), [self, obj_owner, offset, size, dst]() {
```
bwyogatama, line 410: So why do we use detached thread instead of bounded thread pool like we what we use in the local disk path?

src/io/s3/s3_ioctx.cpp
```diff
@@ -0,0 +1,536 @@
...
+}
+
+void s3_ioctx::host_read_ranges_async(sirius_io_object& obj,
+                                      std::vector<cudf::io::text::byte_range_info> const& ranges,
+                                      std::span<cudf::host_span<std::byte>> dst,
+                                      io_completion_handler handler)
+{
+  auto self      = shared_from_this();
+  auto obj_owner = obj.shared_from_this();
+  dispatch_async(std::move(handler), [self, obj_owner, ranges, dst]() {
```
bwyogatama: `dst` is being passed into the detached thread and is a `std::span<cudf::host_span<std::byte>>`.
Is it possible that`cudf::host_span<std::byte>` might have gone out of scope in the caller when the detached thread try to access `dst`?

src/io/s3/s3_ioctx.cpp
```diff
@@ -0,0 +1,536 @@
...
+                                       std::vector<cudf::io::text::byte_range_info> const& ranges,
+                                       std::span<cudf::host_span<std::byte>> dst)
+{
+  if (ranges.size() != dst.size())
+    throw std::invalid_argument("s3_ioctx::host_read_ranges: ranges/dst size mismatch");
+  std::size_t total = 0;
+  for (std::size_t i = 0; i < ranges.size(); ++i) {
+    auto offset = static_cast<std::size_t>(ranges[i].offset());
+    auto size   = static_cast<std::size_t>(ranges[i].size());
+    if (dst[i].size() < size)
```
bwyogatama: The clipping of the size happens in `host_read`, shouldn't we clip first before checking if we should `throw`?

src/io/s3/s3_ioctx.cpp
ran-yuan-rui: the reason was that PR3 is still a standalone S3 reactor PR and intentionally does not wire into Sirius runtime resources yet, so there was no scan-manager / context-owned bounded executor available on this branch. That said, once this is rebased on #740, I will fix is to route async S3 work through the scan-manager / context-owned bounded dispatcher (or inject an executor into `s3_ioctx`) rather than keeping detached threads.

src/io/s3/s3_ioctx.cpp
ran-yuan-rui: Agreed — did not check when I cherry-pick the whole file from old branch

src/io/s3/s3_ioctx.cpp
ran-yuan-rui: Agreed. will add coverage for the EOF-crossing range case alongside it.

src/io/s3/s3_ioctx.cpp
ran-yuan-rui: fixed at 89def0f

src/io/s3/s3_ioctx.cpp
ran-yuan-rui: fixed at 89def0f

src/io/s3/s3_ioctx.cpp
ran-yuan-rui: Side note: Just saw Amin's #740's commit `0974bb3d` just added internal retry-with-bytes- tracking inside `uring_reactor::host_read` for short-reads / EINTR. The S3 follow-up (whichever shape we land on for thread pool injection) is anatural place to add the parallel HTTP retry path (5xx / 429 / 408 +
libcurl transient errors), since the structural template is now in tree.

aminaramoon:
1. why cannot we make the s3 ioctx a reactor based ioctx ? that way it can take advantage of the cache as well.
do we need added functionality in the reactor concept for this to work ?
2. do you want to merge this PR, then I rebase it with mine, or do you want to rebase this PR with mine and prepare it for merge.

ran-yuan-rui:
> 1. why cannot we make the s3 ioctx a reactor based ioctx ? that way it can take advantage of the cache as well.
>    do we need added functionality in the reactor concept for this to work ?
> 2. do you want to merge this PR, then I rebase it with mine, or do you want to rebase this PR with mine and prepare it for merge.

On (1): The Reactor concept assumes a persistent native handle + batched submission semantics that fit local-FD backends (uring, and future GDS/cuFile, RDMA) well, but HTTPS-style backends have a few mismatches: no persistent handle per request, `enqueue_bulk` doesn't help IOPS for independent high-latency GETs, and chunking across reactors actively hurts TLS session reuse + connection pool affinity. So `s3_ioctx`
inheriting `sirius_ioctx` directly felt cleaner than forcing it through the Reactor concept.

That said, worth noting: cache wiring lives on the `sirius_ioctx` base itself (via `initialize_cache(buffer_pool&, …)`), not on `templated_ioctx`. So s3_ioctx can attach the cache today — it justneeds `_cache.reset()` in its dtor for safe teardown (same fix shape needed in `~templated_ioctx()`, since cache lives in base but workers dispatch through derived members).

sirius_ioctx (contract)
│
├── templated_ioctx<Reactor> ← "FD family"
│ ├── uring_ioctx (CPU bounce buffer)
│ ├── gds_ioctx (GPU direct DMA via cuFile)
│ └── S3rdma_ioctx (future)
│
└── (direct inherit) ← "HTTP/object-store family"
├── s3_ioctx
├── gcs_ioctx (future)
└── azure_ioctx (future)

On (2): Let's go with rebasing #759 onto #740. I've actually been keeping a local branch tracking what that adaptation looks like against your current `e706cecc` — mostly mechanical (rename overrides to
`*_io`, add `create_io_object` + `supports`, add `object_path()` on`s3_io_object`, drop the buffer-returning + sync `host_read_ranges` overloads). Happy to push it as a force-update on #759 once #740 is closer to merge. Does that work?

aminaramoon (changes requested):
I do not believe we can merge this pull request in its current state. The error handling is very limited, and the asynchronous model is broken and does not scale with cuDF.

1. Reading hundreds/thousands of pages in each parquet file with cuDF can create numerous asynchronous reads, hence the overhead of thread creation and detachment, which is not a scalable form of asynchronous I/O here.
2. Error handling for retriable errors in cURL—errors that require immediate retries or wait-and-retry mechanisms—occurs frequently when reading from S3. I believe the current io_ctx only fetches data from S3 on a very narrow, successful path and doesn;t handle any of the errors.
3. The range APIs are merely a sequence of synchronous APIs, which fails to mask the latency incurred by each cURL call.
4. The lack of cache support makes achieving reasonable performance from S3 very difficult.

ran-yuan-rui:
> I do not believe we can merge this pull request in its current state. The error handling is very limited, and the asynchronous model is broken and does not scale with cuDF.
> 
> 1. Reading hundreds/thousands of pages in each parquet file with cuDF can create numerous asynchronous reads, hence the overhead of thread creation and detachment, which is not a scalable form of asynchronous I/O here.
> 2. Error handling for retriable errors in cURL—errors that require immediate retries or wait-and-retry mechanisms—occurs frequently when reading from S3. I believe the current io_ctx only fetches data from S3 on a very narrow, successful path and doesn;t handle any of the errors.
> 3. The range APIs are merely a sequence of synchronous APIs, which fails to mask the latency incurred by each cURL call.
> 4. The lack of cache support makes achieving reasonable performance from S3 very difficult.

Yes - you are right. As we discussed on on slack, this is because I am chasing 740 so the missing part can not push yet, the code were on feature-S3datasource-s3reactor-on-740 and feature-S3datasource-runtime-on-740 already.
Please check the proposal on how to split 740 into 2 parts, once we align and you push the fist part of 740 - I will close 759 and replace a new PR to address all of your concerns here

ran-yuan-rui:
close it and replace by https://github.com/sirius-db/sirius/pull/784

PR #760, opened 2026-05-06
CI: Force vcpkg cache to S3 for self-hosted runners on Check workflow

mike-wendt:
This is to address issues where the current GHA cache limited to 10GB has evictions that trigger a full build that can take 1h30m or more. Switching to S3 should reduce these cache misses and when we do miss the penalty should be 5-20mins instead


mike-wendt:
Examples of recent cache misses in the last ~12hrs, all over 1.5hrs:
1. https://github.com/sirius-db/sirius/actions/runs/25437847534/job/74620563001
2. https://github.com/sirius-db/sirius/actions/runs/25414840056/job/74544083566

mike-wendt:
Changing the approach to use a new S3 bucket instead of sharing this with `sccache`, need to create that bucket first in the cluster then come back here for testing

mike-wendt:
Ready to merge after @mbrobbel is able to review

mbrobbel (approved):

PR #762, opened 2026-05-06
docs: add communication subsystem design proposal

9prady9:
TODO (pradeep)

Will add PR description once final design is ratified.


bwyogatama:
using #914 instead

PR #764, opened 2026-05-06
scan: DuckDB format GPU-native decode kernel (Part 3: RLE)

ducndh:
# Summary

GPU decode kernel for DuckDB's RLE codec. Two device kernels per call: a per-segment build kernel (cooperative inclusive scan over run lengths) and an expand kernel (per-row `upper_bound` on the cumsum, gather, write). Sibling of #737, stacks on #736.

## DuckDB RLE on-disk format

```
byte 0    8                                  rle_count_offset    seg_end
│         │                                  │                   │
├─uint64──┼──── values: T[entry_count] ─────┼──── padding ──────┼── counts: u16[entry_count] ──
│  header │                                  │                   │
└─offset where counts begin                  └─0-padded to 8B    └─per-entry run lengths
```

- **`entry_count` is implicit** — walk `counts[]` summing until the total hits `descriptor.row_count`.
- **Padding is real** (DuckDB's `AlignValue<8>`); trust the header offset, don't compute counts from `8 + values_size`.
- **`rle_count_t = uint16_t`** so a single run is ≤ 65535 rows; longer runs split into multiple consecutive same-value entries. DuckDB never emits zero counts.

## The algorithm in one picture

Tiny example with 4 entries:

```
counts  = [ 3,  2,  4,  1]          ← uint16 run lengths
values  = [ A,  B,  C,  D]          ← T-typed values
cumsum  = [ 3,  5,  9, 10]          ← inclusive prefix sum (we build this on device)

expanded output:
  row    0  1  2  3  4  5  6  7  8  9
  value  A  A  A  B  B  C  C  C  C  D
                          ▲
                          │  for row 6:  upper_bound(cumsum, 6) = 2
                          │              (first idx where cumsum > 6)
                          └─ output[6] = values[2] = C
```

That's the whole expand kernel: each thread does `upper_bound` for its
output row, gathers `values[entry]`, writes.

## Two-kernel pipeline — zero D2H syncs

```
HOST                              STREAM (one CUDA stream, ordered)
─────                             ────────────────────────────────────
build_desc[N]  ────────H2D───────▶ kernel_build_cumsum
                                     │   (1 CTA per segment)
                                     │   reads rle_count_offset
                                     │   cooperatively scans counts
                                     │   via cub::BlockScan
                                     ▼
                                   cumsum[N × MAX_ENTRIES]   on device
                                   entry_count[N]            on device
                                     │
chunk_desc[K]  ────────H2D───────▶ kernel_decode_rle<T>
                                     │   (1 CTA per 2048-row chunk;
                                     │    K total chunks across N segs)
                                     │   reads cumsum, gathers values,
                                     │   writes output
                                     ▼
                                   d_output (the column)
```

## Build kernel — multi-tile BlockScan

`cub::BlockScan` handles `BLOCK_DIM × VPT_BUILD = 256 × 16 = 4096` items per call. DuckDB's worst-case ec is ~87K (`T=int8`, 256 KiB segment), so we iterate tiles within one CTA and propagate the running sum:

```
counts: [c₀ c₁ … c₄₀₉₅ | c₄₀₉₆ … c₈₁₉₁ | c₈₁₉₂ … ]
        └── tile 0 ────┘└── tile 1 ────┘└── tile 2 ─ …

per tile:
  ─ each of 256 threads loads 16 counts (blocked layout)
  ─ in-register inclusive prefix per thread
  ─ cub::BlockScan over per-thread aggregates
  ─ write tile's cumsum to gmem with `s_running_total` baked in
  ─ thread 255 updates s_running_total for next tile

bail out when:
  ─ s_first_match != ~0u           (an entry's cumsum hit row_count)
  ─ s_running_total >= row_count   (overshot)
```

`RLE_BUILD_MAX_ENTRIES = 90112` (22 tiles) covers the worst case; oversized segments surface as `entry_count = 0` (zero-fill).

## Expand kernel — narrow-window binary search + warp-vote fast path

Each CTA owns `RLE_ROWS_PER_CHUNK = 2048` consecutive rows of one segment.
Cumsum is in shmem when `ec ≤ 4096`; gmem-resident otherwise.

### Narrow-window

A chunk usually covers far fewer entries than the segment has total. Thread 0 finds the band once, broadcasts via shmem:

```
segment cumsum:  [································································]
                                       ▲                  ▲
                                       │  this chunk's    │
                                       │  2048 rows map   │
                                       │  to this band    │
                          band_lo ─────┘                  └────── band_hi
                                          ◀── 256 threads search ──▶
                                              inside [band_lo, band_hi]
                                              log₂(band) ≪ log₂(ec)
```

Drops per-thread depth from `log₂(ec)` to `log₂(band_width)` — typically 12 → 4-6.

### Warp-vote fast path (long-run shapes)

When the segment's average run is ≥ warp width:

```
warp's 32 output rows:  1024  1025  1026  ...  1055
                          └────── all map to entry #7 ──────┘

  lane 0:    search band once → entry = 7
  shfl:      broadcast entry=7 to all 32 lanes
  shfl:      broadcast cumsum[7] to all 32 lanes
  if cumsum[7] > row 1055   (covers all 32 rows):
      ──▶ one __ldg(values[7]) + __stwt per thread
  else (warp straddles entry boundary):
      ──▶ fall back to per-thread band search
```

Gated by `chunk_rows / 32 ≥ ec` so short-run shapes (where the bound check would fail almost every iteration) skip the path entirely.

### Address-space duplication (compiler hint)

Shmem and gmem cumsum paths are **textually duplicated** instead of unified through one `cumsum` pointer:

```cpp
if (ec ≤ RLE_SMEM_MAX_ENTRIES) {
    /* path A: search s_cumsum  → compiler emits ld.shared */
} else {
    /* path B: search desc.d_cumsum → compiler emits ld.global */
}
```

Merging the two paths through a `uint32_t const* cumsum = ...` would force `ld.generic` (slower, runtime address-space check).

Output stores: `__stwt` (write-once, never reread). Value gathers: `__ldg` (broadcast-heavy on long runs, hits the read-only cache).

## Malformed input → zero-fill

`entry_count == 0` is the universal sentinel collapsing every malformed class. Output buffer is uninitialised by the dispatcher, so this is the only thing between a malformed segment and leaking prior device contents.

| Detected by    | Condition                                       |
|----------------|-------------------------------------------------|
| build kernel   | `bytes_size < 8`                                |
| build kernel   | `rle_count_offset` outside `[8, bytes_size)`    |
| build kernel   | any `count == 0`                                |
| build kernel   | sum of counts ≠ `row_count`                     |
| build kernel   | values area too small for `ec × type_size`      |
| build kernel   | `ec > RLE_BUILD_MAX_ENTRIES = 90112`            |
| expand kernel  | sees `ec == 0` → `__stwt(out, T(0))` for chunk  |


src/include/cuda/scan/gpu_decode_rle.cuh
```diff
@@ -0,0 +1,50 @@
...
+
+#include <cuda/scan/gpu_native_decode.cuh>
+
+#include <cstdint>
+
+namespace sirius::cuda::scan {
+
+/// 8-byte rle_count_offset header at the start of every RLE segment.
+/// See `duckdb/src/storage/compression/rle.cpp::RLEConstants::RLE_HEADER_SIZE`.
+inline constexpr uint32_t RLE_HEADER_SIZE = 8;
```
kevkrist:
```suggestion
static constexpr uint32_t RLE_HEADER_SIZE = sizeof(uint64_t);
```

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+
+namespace {
+
+// 256-thread CTA matches BITPACKING for cross-codec consistency. 2048-row
+// chunk gives VPT=8, enough to amortise per-CTA setup without bloating
+// per-thread register state.
+constexpr uint32_t BLOCK_DIM          = 256;
+constexpr uint32_t RLE_ROWS_PER_CHUNK = 2048;
+constexpr uint32_t VPT                = RLE_ROWS_PER_CHUNK / BLOCK_DIM;
+static_assert(BLOCK_DIM * VPT == RLE_ROWS_PER_CHUNK);
```
kevkrist: Is this not true by definition?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+#include <string>
+#include <vector>
+
+namespace sirius::cuda::scan {
+
+namespace {
+
+// 256-thread CTA matches BITPACKING for cross-codec consistency. 2048-row
+// chunk gives VPT=8, enough to amortise per-CTA setup without bloating
+// per-thread register state.
```
kevkrist: Can we get rid of this AI comment bloat about the amortization?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+constexpr uint32_t BLOCK_DIM          = 256;
+constexpr uint32_t RLE_ROWS_PER_CHUNK = 2048;
+constexpr uint32_t VPT                = RLE_ROWS_PER_CHUNK / BLOCK_DIM;
+static_assert(BLOCK_DIM * VPT == RLE_ROWS_PER_CHUNK);
+
+// Build kernel processes counts in tiles of BUILD_TILE_ENTRIES; running
+// total propagates across tiles, so total entry count is unbounded.
+// VPT_BUILD=16 keeps per-thread register state at ~64 B (16 × uint32 cumsum
```
kevkrist: Can we get rid of AI explanatory comments from 68-71.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
+/*
```
kevkrist, line 1: This took me a while to understand. It would be nice to have a description of the segment layout for RLE at the top, maybe an ASCII diagram (use Claude), and how the header stores the offset at which the counts buffer starts, etc. I understand that this information is scattered throughout the comments, but having it in a consolidated place makes it easier to understand.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+constexpr uint32_t RLE_SMEM_BYTES       = RLE_SMEM_MAX_ENTRIES * sizeof(uint32_t);
+
+// DuckDB's worst-case ec for an RLE segment is bounded by
+// (block_size - header) / (sizeof(T) + sizeof(rle_count_t)). For block_size
+// = 256 KiB and the narrowest type (T=int8 with rle_count_t=uint16), that's
+// (262144-8)/3 = 87376 entries. Round up to a tile multiple (22 × 4096 =
+// 90112) so the per-segment cumsum slice is tile-aligned.
+constexpr uint32_t RLE_BUILD_MAX_ENTRIES = 22u * BUILD_TILE_ENTRIES;  // 90112
+static_assert(RLE_BUILD_MAX_ENTRIES % BUILD_TILE_ENTRIES == 0);
+static_assert(RLE_BUILD_MAX_ENTRIES >= 87376, "must cover DuckDB's max ec for T=int8 segments");
```
kevkrist: Why are we static asserting these when they are true by definition?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+  __shared__ uint32_t s_malformed;
+  __shared__ uint32_t s_first_match;
+  __shared__ uint32_t s_running_total;
+
+  uint32_t const sid = blockIdx.x;
+  if (sid >= num_segs) return;
+  auto const desc = descs[sid];
+
+  if (threadIdx.x == 0) {
+    s_malformed     = 0;
```
kevkrist: There is a lot of malformed checking here. I don't know that it's ever actually detected downstream. Is this planned for later? I see that DuckDB itself doesn't make any of these runtime checks. Are they really necessary?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+  __syncthreads();
+
+  if (s_malformed) {
+    if (threadIdx.x == 0) d_entry_counts[sid] = MALFORMED_FLAG;
+    return;
+  }
+
+  uint32_t const row_count = desc.row_count;
+  uint16_t const* counts   = reinterpret_cast<uint16_t const*>(desc.d_bytes + s_offset);
+  uint32_t const capacity = s_capacity < RLE_BUILD_MAX_ENTRIES ? s_capacity : RLE_BUILD_MAX_ENTRIES;
```
kevkrist: Can we rename this to something more descriptive like `max_entries`?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+    BlockScan(scan_temp).ExclusiveSum(my_aggregate, my_prefix);
+
+    if (local_zero_seen) atomicOr(&s_malformed, 1u);
+
+#pragma unroll
+    for (uint32_t v = 0; v < VPT_BUILD; ++v) {
+      uint32_t const i  = tile + threadIdx.x * VPT_BUILD + v;
+      uint32_t const cs = s_running_total + my_prefix + my_local_cumsum[v];
+      if (i < capacity) cumsum_out[i] = cs;
+      if (i < capacity && cs == row_count) atomicMin(&s_first_match, i + 1);
```
kevkrist: This took me a while to understand. Please document that the purpose of s_first_match is to capture the index at which we first reach the total number of rows for this segment.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+constexpr uint32_t BUILD_TILE_ENTRIES   = BLOCK_DIM * VPT_BUILD;  // 4096
+constexpr uint32_t RLE_SMEM_MAX_ENTRIES = BUILD_TILE_ENTRIES;
+constexpr uint32_t RLE_SMEM_BYTES       = RLE_SMEM_MAX_ENTRIES * sizeof(uint32_t);
+
+// DuckDB's worst-case ec for an RLE segment is bounded by
+// (block_size - header) / (sizeof(T) + sizeof(rle_count_t)). For block_size
+// = 256 KiB and the narrowest type (T=int8 with rle_count_t=uint16), that's
+// (262144-8)/3 = 87376 entries. Round up to a tile multiple (22 × 4096 =
+// 90112) so the per-segment cumsum slice is tile-aligned.
+constexpr uint32_t RLE_BUILD_MAX_ENTRIES = 22u * BUILD_TILE_ENTRIES;  // 90112
```
kevkrist: Can we constexpr calculate the value 22 here using something like ::cuda::ceil_div?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+#include <vector>
+
+namespace sirius::cuda::scan {
+
+namespace {
+
+// 256-thread CTA matches BITPACKING for cross-codec consistency. 2048-row
+// chunk gives VPT=8, enough to amortise per-CTA setup without bloating
+// per-thread register state.
+constexpr uint32_t BLOCK_DIM          = 256;
```
kevkrist, line 64: Don't need to change now, but in the future, we might want to make BLOCK_DIM a template value parameter to the kernel, so that we can sweep different values to find the optimal one. Same for VPT below.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+constexpr uint32_t RLE_ROWS_PER_CHUNK = 2048;
+constexpr uint32_t VPT                = RLE_ROWS_PER_CHUNK / BLOCK_DIM;
```
kevkrist: Can you please make this `VALUES_PER_THREAD` to make it more descriptive? This applies to the other decompression files as well.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+      lo = mid + 1;
+    } else {
+      hi = mid;
+    }
+  }
+  return lo;
+}
+
+/// Body of the expand kernel, parameterised on the cumsum pointer source.
+/// `__forceinline__` so each call site keeps its own statically-known
```
kevkrist: No need to document the reason for __forceinline__

src/cuda/scan/gpu_decode_rle.cu
kevkrist: The main reason I ask is because these checks clutter the code somewhat and obscure its focus, and some variable & parameters could go away (e.g., type_size).

src/include/cuda/scan/gpu_decode_rle.cuh
```diff
@@ -0,0 +1,50 @@
...
+
+/// Decode an RLE codec run into `d_output`. Two device kernels per call:
+/// `kernel_build_cumsum` (one CTA per segment) reads each segment's header
+/// + counts and writes an inclusive prefix sum to a device-resident buffer;
+/// `kernel_decode_rle` (one CTA per RLE_ROWS_PER_CHUNK-row slice) expands
+/// rows via per-row upper_bound + value gather. Both kernels are launched on
```
kevkrist: No need to document that the kernels are launched on the passed stream.

src/include/cuda/scan/gpu_decode_rle.cuh
```diff
@@ -0,0 +1,50 @@
...
+
+namespace sirius::cuda::scan {
+
+/// 8-byte rle_count_offset header at the start of every RLE segment.
+/// See `duckdb/src/storage/compression/rle.cpp::RLEConstants::RLE_HEADER_SIZE`.
+inline constexpr uint32_t RLE_HEADER_SIZE = 8;
+
+/// Decode an RLE codec run into `d_output`. Two device kernels per call:
+/// `kernel_build_cumsum` (one CTA per segment) reads each segment's header
+/// + counts and writes an inclusive prefix sum to a device-resident buffer;
```
kevkrist: Recommended `...and calculates an inclusive prefix sum over the counts per segment.`

src/include/cuda/scan/gpu_decode_rle.cuh
kevkrist: gather or scatter?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+
+  std::vector<gpu_segment_desc const*> live;
+  live.reserve(run.segments.size());
+  for (auto const& seg : run.segments) {
+    if (seg.row_count == 0) continue;
+    live.push_back(&seg);
+  }
+  if (live.empty()) return;
+
+  size_t const n = live.size();
```
kevkrist: Can we make `n` more descriptvie: like `num_live_segments`?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+    RLE_BUILD_MAX_ENTRIES,
+    d_entry_counts.data(),
+    type_size,
+    static_cast<uint32_t>(n));
+
+  // Build per-CTA chunk descriptors. Cumsum slice + values pointer can be
+  // computed on host without observing the build kernel's output.
+  size_t total_chunks = 0;
+  for (auto const* seg : live) {
+    total_chunks += (seg->row_count + RLE_ROWS_PER_CHUNK - 1) / RLE_ROWS_PER_CHUNK;
```
kevkrist: Use ::cuda::ceil_div

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+  }
+  std::vector<rle_chunk_desc> h_descs;
+  h_descs.reserve(total_chunks);
+  for (size_t i = 0; i < n; ++i) {
+    auto const& seg          = *live[i];
+    uint32_t const rc        = seg.row_count;
+    uint32_t const num_cnks  = (rc + RLE_ROWS_PER_CHUNK - 1) / RLE_ROWS_PER_CHUNK;
```
kevkrist: Use ::cuda::ceil_div

src/cuda/scan/gpu_decode_rle.cu
kevkrist: If you have to clamp here, is this not indicative of malformation? Either clamp and document the necessity of the clamp with the malformation flag (or some other error flag) or drop the clamp at RLE_BUILD_MAX_ENTRIES.

kevkrist (commented):
Very clever kernel pattern!

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+/// pointer would force ld.generic.
+template <typename T>
+__device__ __forceinline__ void rle_decode_chunk_body(uint32_t const* __restrict__ cumsum,
+                                                      T const* __restrict__ values,
+                                                      T* __restrict__ out_chunk,
+                                                      uint32_t ec,
+                                                      uint32_t lr_start,
+                                                      uint32_t rc)
+{
+  // Narrow the search window to the entry band this chunk actually covers.
```
kevkrist: Can we call entry band something like 'entry range' in the documentation?
Also, can you spell out entry_count and row_count, and define what lr_start is? I assume this means segment-local row start? Also, please make it clear that these are per-segment values.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+constexpr uint32_t VPT                = RLE_ROWS_PER_CHUNK / BLOCK_DIM;
+static_assert(BLOCK_DIM * VPT == RLE_ROWS_PER_CHUNK);
+
+// Build kernel processes counts in tiles of BUILD_TILE_ENTRIES; running
+// total propagates across tiles, so total entry count is unbounded.
+// VPT_BUILD=16 keeps per-thread register state at ~64 B (16 × uint32 cumsum
+// + 16 × uint16 counts). The shmem cumsum cache in the expand kernel is
+// sized to one tile so chunks with ec ≤ tile_size get the shmem fast path;
+// larger ec falls through to a gmem-resident cumsum read.
+constexpr uint32_t VPT_BUILD            = 16;
```
kevkrist: I think this value is high-machine dependent. On machines with a smaller register file this will reduce occupancy due to register pressure. Consider either a) documenting that this parameter should be HW-tuned or b) decreasing this value to 8 if it doesn't sacrifice performance.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+template <typename T>
+__device__ __forceinline__ void rle_decode_chunk_body(uint32_t const* __restrict__ cumsum,
+                                                      T const* __restrict__ values,
+                                                      T* __restrict__ out_chunk,
+                                                      uint32_t ec,
+                                                      uint32_t lr_start,
+                                                      uint32_t rc)
+{
+  // Narrow the search window to the entry band this chunk actually covers.
+  // For typical-shape chunks ec_band << ec, dropping per-thread search
```
kevkrist: What is ec_band?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+  // depth from log2(ec) to log2(ec_band).
+  __shared__ uint32_t s_band_lo;
+  __shared__ uint32_t s_band_hi;
+  if (threadIdx.x == 0) {
```
kevkrist, line 347: Can you add a comment about what this is doing?
You are using binary search for the first and last row in this 'band' of a segment to narrow the search space for the interior rows.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+    if (hi >= ec) hi = ec - 1;
+    s_band_lo = lo;
+    s_band_hi = hi + 1u;
+  }
+  __syncthreads();
+  uint32_t const band_lo = s_band_lo;
+  uint32_t const band_hi = s_band_hi;
+
+  bool const long_runs_heuristic = rc / 32u >= ec;
+  uint32_t const lane            = threadIdx.x & 31u;
```
kevkrist: The compiler can resolve % 32 to this bit instruction by strength reduction. This is just a preference (immediate clarity) but leave as-is if you wish. Also, consider this elsewhere as well, and consider making the meaning of 32 clear here -- its the number of threads in a warp. It means something else in other places.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+    uint32_t hi             = rle_upper_bound(cumsum, 0, ec, last_row);
+    if (hi >= ec) hi = ec - 1;
+    s_band_lo = lo;
+    s_band_hi = hi + 1u;
+  }
+  __syncthreads();
+  uint32_t const band_lo = s_band_lo;
+  uint32_t const band_hi = s_band_hi;
+
+  bool const long_runs_heuristic = rc / 32u >= ec;
```
kevkrist: Can you explan this heuristic? Is it, 'if the average run length in this segment is >= 32, use warp-broadcast path'? If so, please document that, and then explain why it makes sense.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+      entry_warp = __shfl_sync(0xFFFFFFFFu, entry_warp, 0);
+
+      uint32_t cumsum_at_entry = 0;
+      if (lane == 0) cumsum_at_entry = cumsum[entry_warp];
+      cumsum_at_entry = __shfl_sync(0xFFFFFFFFu, cumsum_at_entry, 0);
+
+      uint32_t const warp_last = warp_first + 31u;
+
+      if (warp_last < cumsum_at_entry) {
+        __stwt(out_chunk + i, __ldg(values + entry_warp));
```
kevkrist: So, in other PRs you've used different store intrinsics. Why __stwt() here and not the streaming store from before or vis-versa? I assume this is because you don't want these stores to steal cache entries. Can you state that purpose? Also, I'm guessing it's important because you want the search space for binary search to stay in cache? Can you state that as well?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+    }
+    return;
+  }
+
+  T const* values         = reinterpret_cast<T const*>(desc.d_values);
+  uint32_t const lr_start = desc.local_row_start;
+
+  extern __shared__ uint32_t s_cumsum[];
+
+  // Two textually-distinct branches: each call site passes a pointer with a
```
kevkrist: You can get rid of this AI-generated explanation.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+    for (uint32_t i = threadIdx.x; i < rc; i += blockDim.x) {
+      __stwt(out_chunk + i, T(0));
+    }
+    return;
+  }
+
+  T const* values         = reinterpret_cast<T const*>(desc.d_values);
+  uint32_t const lr_start = desc.local_row_start;
+
+  extern __shared__ uint32_t s_cumsum[];
```
kevkrist: Why do you extern this and lift it from kernel launch? If it's static, you can just declare the shared static array here:
`__shared__ uint32_t s_cumsum[RLE_SMEM_MAX_ENTRIES]`.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+
+      uint32_t const warp_last = warp_first + 31u;
+
+      if (warp_last < cumsum_at_entry) {
+        __stwt(out_chunk + i, __ldg(values + entry_warp));
+      } else {
+        uint32_t const local_row = lr_start + i;
+        uint32_t entry           = rle_upper_bound(cumsum, band_lo, band_hi, local_row);
+        if (entry >= ec) entry = ec - 1;
+        __stwt(out_chunk + i, __ldg(values + entry));
```
kevkrist: For __ldg(), can you also justify the use of low-level intrinsics?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+      uint32_t const i = v * blockDim.x + threadIdx.x;
+      if (i >= rc) break;
+      uint32_t const warp_first = lr_start + (i & ~31u);
+
+      uint32_t entry_warp = 0;
+      if (lane == 0) {
+        entry_warp = rle_upper_bound(cumsum, band_lo, band_hi, warp_first);
+        if (entry_warp >= ec) entry_warp = ec - 1;
+      }
+      entry_warp = __shfl_sync(0xFFFFFFFFu, entry_warp, 0);
```
kevkrist: Please consider using the more modern `cuda::device::warp_shuffle_` family of APIs.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+      cumsum_at_entry = __shfl_sync(0xFFFFFFFFu, cumsum_at_entry, 0);
+
+      uint32_t const warp_last = warp_first + 31u;
+
+      if (warp_last < cumsum_at_entry) {
+        __stwt(out_chunk + i, __ldg(values + entry_warp));
+      } else {
+        uint32_t const local_row = lr_start + i;
+        uint32_t entry           = rle_upper_bound(cumsum, band_lo, band_hi, local_row);
+        if (entry >= ec) entry = ec - 1;
```
kevkrist: I don't understand this. Can you explain this?

src/cuda/scan/gpu_decode_rle.cu
kevkrist: Can you tighten this to something like `band_row_count / 32 > (band_hi - band_lo)`?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+#pragma unroll
+  for (uint32_t v = 0; v < VPT; ++v) {
+    uint32_t const i = v * blockDim.x + threadIdx.x;
+    if (i >= rc) break;
+    uint32_t const local_row = lr_start + i;
+    uint32_t entry           = rle_upper_bound(cumsum, band_lo, band_hi, local_row);
+    if (entry >= ec) entry = ec - 1;
+    __stwt(out_chunk + i, __ldg(values + entry));
+  }
+  (void)lane;
```
kevkrist: Why is this here?

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+
+  // Build per-segment build descriptors.
+  std::vector<rle_build_desc> h_build(n);
+  for (size_t i = 0; i < n; ++i) {
+    h_build[i] = {live[i]->d_bytes, live[i]->bytes_size, live[i]->row_count};
+  }
+
+  // Allocate device-side cumsum buffer (one MAX_ENTRIES slot per segment)
+  // and entry-count array.
+  rmm::device_uvector<uint32_t> d_cumsums(n * RLE_BUILD_MAX_ENTRIES, stream, mr);
```
kevkrist: Probably fine for now, but this is probably a large over-allocation. I see it's necessity with the current pattern, but it's worth flagging with a comment.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,474 @@
...
+  }
+
+  if (threadIdx.x == 0) {
+    uint32_t ec = s_first_match;
+    if (s_malformed || ec == ~0u) {
+      d_entry_counts[sid] = MALFORMED_FLAG;
+      return;
+    }
+    uint32_t values_bytes = s_offset - RLE_HEADER_SIZE;
+    if (values_bytes < ec * type_size) {
```
kevkrist: Is this check really necessary? If not, you can get rid of type_size parameter.

test/cpp/scan/test_gpu_decode_rle.cpp
```diff
@@ -0,0 +1,579 @@
+/*
```
kevkrist, line 1: claude says: no targeted test for the warp-broadcast boundary case in `rle_decode_chunk_body`. Please add one that exercises that branch.

kevkrist (commented):
Lots of nits to help this be more understandable.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Not strictly — If someone tunes BLOCK_DIM and it stops dividing RLE_ROWS_PER_CHUNK, the division silently truncates and this catches it. However, we can divide more clever and obvious heuristic/default to make it less painful to maintain

src/cuda/scan/gpu_decode_rle.cu
ducndh: The dispatcher upstream doesn't zero-init d_output, so without these checks a malformed segment leaves output rows uninitialized → leaks whatever device contents were there before. The MALFORMED_FLAG → zero-fill in kernel_decode_rle is the only barrier.                   

I don't know if we ever need to care about security but here are the 3 options I can think of:
1. Status quo (keep all).
2. Drop checks; require the dispatcher to zero-init d_output upstream.                      
3. Drop checks AND zero-fill — accept the leak and matches DuckDB's OLAP model but they allocate their own buffer.

What do you think?

src/cuda/scan/gpu_decode_rle.cu
ducndh: The clamp does flow into the malformed flag — when real ec exceeds the cap, the loop never satisfies s_first_match != ~0u, exits on the sentinel, and sets MALFORMED_FLAG at the function tail. I will set malformed directly when the clamp triggers to make it more obvious

src/include/cuda/scan/gpu_decode_rle.cuh
ducndh: Gather. Each thread does an `upper_bound` on the cumsum, reads `values[entry]` (irregular reads), then writes to a sequential output index.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Both removed in the latest push — `RLE_BUILD_MAX_ENTRIES` is now built via `::cuda::ceil_div(RLE_DUCKDB_MAX_EC, BUILD_TILE_ENTRIES) * BUILD_TILE_ENTRIES`, so the tile-multiple property and the `>= 87376` coverage are baked into the construction.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Stale name from an earlier draft — the actual variables are `band_lo`/`band_hi`. Comment fixed.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Done — comment now states thread 0 binary-searches the chunk's first/last rows once to find `[band_lo, band_hi)`, dropping per-thread search depth from `log2(entry_count)` to `log2(band_hi - band_lo)`.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Done. Comment now reads: "Avg run ≥ warpSize ⇒ most warps cover one entry; lane 0 searches once and broadcasts (cheaper than 32 searches). Short-run shapes skip — the bound check would miss most iterations."

ducndh: Good catch — segment-level under-fires the warp path on chunks falling in dense regions of an otherwise-sparse segment. Will bench `band_row_count / 32 > (band_hi - band_lo)` against the current form on A100 + Turing (pareto + short_runs are the relevant shapes).

src/cuda/scan/gpu_decode_rle.cu
ducndh: Output is write-once / never reread; cumsum + values are hot for every chunk's binary search. `__stwt` streams stores past L1 to keep cumsum in cache. Inline comment now says exactly that. Difference vs other codecs: their outputs may be reused downstream, ours aren't.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Defensive clamp. `rle_upper_bound` returns `entry_count` only if the row exceeds `cumsum[entry_count-1]` (= row_count); well-formed inputs never trigger this, but the clamp keeps the `values[entry]` read in-bounds if it does. Inline comment now says so.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Read-only cache. On the warp-broadcast path one load services all 32 lanes (broadcast); on the per-lane fallback values are heavily reused across rows. Either way wins over plain `ld.global`.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Stale `(void)lane` suppression — `lane` was only used inside the `long_runs_heuristic` branch but declared above it. Moved the declaration inside the branch and removed the suppression.

src/cuda/scan/gpu_decode_rle.cu
ducndh: No good reason — size is a compile-time constant. Switched to `__shared__ uint32_t s_cumsum[RLE_SMEM_MAX_ENTRIES]` and dropped the dynamic-shmem (`RLE_SMEM_BYTES`) argument from all four launches.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Will bench 8 vs 16 on Turing (smaller register file) + A100. If 8 is within noise on both, drop. If 16 wins on A100, keep with HW-tunable note. Reporting back here.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Same thread as `.cu:132` — folding into that decision. If you accept option 2 there (drop checks + cudaMemsetAsync upstream), the `type_size` parameter goes away with it.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Filed as follow-up.

src/cuda/scan/gpu_decode_rle.cu
ducndh: Eh I will have to think about how IO can do some sidecar loading for the meta data for this instead of over allocate this so I will file it as an issue will doing the IO

src/cuda/scan/gpu_decode_rle.cu
ducndh: A100 bench results (median of 3 runs, each = 10 timed iters after 3 warmup):

| Shape | BASELINE (VPT_BUILD=16) | E (VPT_BUILD=8) |
|---|---|---|
| long_runs | 800–1122 (bimodal) | 1122 |
| medium_runs | 1014 | 1030 (+1.6%) |
| pareto_runs | 1108 | 1104 (−0.4%) |
| short_runs | 242 | 243 (~) |

On A100: net neutral, all within noise. Side-effect: VPT_BUILD=8 halves BUILD_TILE_ENTRIES (4096→2048) and therefore RLE_SMEM_MAX_ENTRIES — chunks with 2048 < ec ≤ 4096 fall through to the gmem-cumsum path. short_runs (ec=4096) hits this in our bench and the A100 absorbs it (L2 caches the cumsum well across CTAs of the same segment).

Without a clear A100 win to motivate the change, keeping VPT_BUILD=16 for now. Will rerun on Turing when next on the cc box (smaller register file should be where any occupancy gain shows up); if its a clean win there, will flip + add HW-tunable note.

src/cuda/scan/gpu_decode_rle.cu
ducndh: A100 bench (median 3×10 iters):

| Shape | BASELINE | D `rc/32 > (band_hi-band_lo)` |
|---|---|---|
| long_runs | 800–1122 | 1125 |
| medium_runs | 1014 | **876 (−14%)** |
| pareto_runs | 1108 | 1107 |
| short_runs | 242 | 250 |

D regresses medium_runs. Mechanism: flips the path from per-lane to warp-broadcast. Run len 120 → ~27% of warps straddle a run boundary; straddle warps pay BOTH broadcast machinery AND per-lane fallback. The 73% non-straddle warps win only marginally because per-lane on a 17-entry band is already cheap (~4 iter). Reverted.

src/cuda/scan/gpu_decode_rle.cu
ducndh: A100 bench (median 3×10 iters): VPT=8 is neutral across all four shapes (within noise). Going with 8 anyway as the safer default — older archs have the same register file (65536 regs/SM) but tighter per-thread budgets, so VPT=8 is at-worst neutral and possibly better there. Side effect: `BUILD_TILE_ENTRIES` halves to 2048 → `RLE_SMEM_MAX_ENTRIES` also 2048; chunks with 2048 < ec ≤ 4096 fall to the gmem-cumsum path (L2 absorbs it on A100).

src/cuda/scan/gpu_decode_rle.cu
ducndh: Quick mechanism explanation for why band-local regressed:

Each output row needs to know "which entry does this row belong to?" There are two paths the kernel can take:

- **Fast path (warp-broadcast)**: if every row in a 32-thread warp falls within the same entry, only lane 0 does the binary search and broadcasts the answer to the other 31 lanes. Big win — 1 search instead of 32.
- **Slow path (per-lane)**: every thread does its own search. Cheap per thread (~4 iterations on a small band), but 32 of them.

The original heuristic ("segment avg run ≥ 32") is conservative — only fires the fast path when runs are reliably long.

Your band-local form fires the fast path **more often**, which is correct in spirit. The problem is the fast path itself isn't free when warps don't fit cleanly inside one run. With run length 120 in a 2048-row chunk, the chunk crosses ~17 run boundaries, so ~17 of 64 warps straddle two entries. Those straddle warps pay **twice**: once for the broadcast machinery (lead-thread search + shuffles + bound check), then again for the per-lane fallback they get forced into anyway. The other 47 warps win with broadcast, but their win is small because each per-lane search they avoided was only ~4 iterations on a small band. Net: small_win × 47 + big_loss × 17 = negative on shapes like medium_runs.

The fix isn't band-local vs segment-level — it's making the broadcast path cheap for marginal cases (e.g., pack `entry_warp + fast/slow bit` into one shuffle instead of two, have straddle warps do an incremental forward search from `entry_warp` instead of restarting).

test/data/decode_baselines/sm80.json
```diff
@@ -0,0 +1,73 @@
+{
```
kevkrist: This needs to be removed from dev. Per Joost's offline conversation, this needs to get documented in the PR description, if at all.

src/cuda/scan/gpu_decode_rle.cu
kevkrist: Yeah, I know what __ldg() means, and this doesn't really have anything to do with the broadcast -- the memory subsystem will broadcast no matter what cache hint you use. On the per-lane fallback, sure, but __ldg means skip L1, so really the default load (caching also in L1) would be most beneficial in this case. So what I'm guessing is true is that the AI is trying to prevent eviction of the cumsum array from L1 when it's not in shared memory. This is consistent with why it's using __stwt for stores.

src/cuda/scan/gpu_decode_rle.cu
```diff
@@ -0,0 +1,462 @@
...
+    s_band_hi = hi + 1u;
+  }
+  __syncthreads();
+  uint32_t const band_lo = s_band_lo;
+  uint32_t const band_hi = s_band_hi;
+
+  // Avg run ≥ warpSize ⇒ most warps cover one entry; lane 0 searches once
+  // and broadcasts (cheaper than 32 searches). Short-run shapes skip — the
+  // bound check would miss most iterations.
+  bool const long_runs_heuristic = rc / 32u >= entry_count;
```
kevkrist: Does this really have a measurable performance benefit? It kinda feels like a speculative optimization. Also, you are adding instruction overhead when warp_last < cumsum_at_entry, so there is a cost-benefit/amortization balance to be struck here.

Also the comment 'Short-run shapes skip...' doesn't make any sense to me.

kevkrist (approved):
Please rebase and merge.

src/cuda/scan/gpu_decode_rle.cu
kevkrist: Leaving as-is, with a corrected average using the chunk count rather than the segment count.

kevkrist (approved):

PR #766, opened 2026-05-07
Fix for race condition between task creation and finalizing pipelines

wmalpica:
This PR fixes a race condition which was primarily caused by the fact that we can be checking if a pipeline is completed at the same time as tasks are being created. 

This PR adds a mutex to the sirius_pipeline class. Now task creation (including pulling data from a port to create the task) is done within this lock and updating the pipeline status (to set it to complete)  is also done within the same lock, removing the race condition.

This PR also fixes other minor issues related to pipeline and operator completion, such as ensuring all operators are finalized and checking at the end of a query to ensure that is the case.
It also fixes CPU_SOURCE_TASKS so that memory reservation is done in the duckdb_scan_executor instead of in its compute_task function.

### DIFF
--- a/src/pipeline/gpu_pipeline_task.cpp
+++ b/src/pipeline/gpu_pipeline_task.cpp
@@ -184,6 +184,9 @@ gpu_pipeline_task::gpu_pipeline_task(
       }
     }
   }
+  if (auto* pipeline = _global_state->cast<gpu_pipeline_task_global_state>().get_pipeline()) {
+    pipeline->mark_task_created();
+  }
 }
 
 gpu_pipeline_task::~gpu_pipeline_task()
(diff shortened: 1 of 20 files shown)


src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -184,6 +184,9 @@ gpu_pipeline_task::gpu_pipeline_task(
       }
     }
   }
+  if (auto* pipeline = _global_state->cast<gpu_pipeline_task_global_state>().get_pipeline()) {
+    pipeline->mark_task_created();
```
bwyogatama, line 188: So if we do this then we have to update the task retry mechanism as well.

Because right now, when the task is retried, the it will skip the `mark_task_completed` in the task destructor. But when it creates the new task it will increment `mark_task_created` cause this now lives in the constructor

So during retry our `task_created > task_completed`

bwyogatama (approved):

src/pipeline/gpu_pipeline_task.cpp
wmalpica: Good catch. I have removed the concept of marking a task as recheduled so that now the task destructor will always increment the task_completed flag

PR #767, opened 2026-05-07
Add per-query pin_table benchmark mode to TPC-H runner

bwyogatama:
## Summary

- Adds `--pinning-mode {none,per-query}` to `benchmark_and_validate.sh` and `run_tpch_parquet.sh`. In `per-query` mode the Sirius engine wraps each query block with `CALL pin_table(<glob>, tier='gpu', name=<table>, cols=[...])` for every table the query reads, runs the iterations back-to-back, then `CALL unpin_table(...)` — isolating per-query pinning cost from query execution and exercising the `cached_split_provider` scan path introduced in #721.
- New helper `test/tpch_performance/tpch_pin_columns.py` keeps the per-query column→table mapping (Q1..Q22), detects parquet layout (single file, `_*` partitioned, or subdirectory), and emits a glob whose `FileSystem::GlobFiles` expansion matches the corresponding `CREATE VIEW … read_parquet([…])` so the path-equality check in `sirius_scan_manager.cpp:92-99` lines up.
- Marker pairs (`__TPCH_PIN_BEGIN__`, `__TPCH_UNPIN_BEGIN__`, …) keep pin/unpin `Run Time` lines outside the iteration-time parser window so `timings.csv` reflects query-only time.
- **Per-query Sirius log split now works under Super Sirius transparent execution.** The splitter previously matched `QueryBegin: call gpu_execution(...)` (the legacy wrapper). Transparent execution logs queries as their raw SQL (`QueryBegin: SELECT ...`), so no per-query `q<N>/sirius.log` files were being produced. The splitter now matches any `QueryBegin:` and excludes the session prologue (`CREATE VIEW`) plus pinning-mode CALLs (`pin_table` / `unpin_table`), producing `runs/.../sirius/q<N>/sirius.log` slices for both `--pinning-mode none` and `--pinning-mode per-query`. Multi-session runs already get isolated per-query logs via `SIRIUS_LOG_DIR=$Q_DIR` and don't go through the splitter.
- DuckDB engine ignores the `--pinning-mode` flag and runs as the unchanged baseline.

Tier default is `gpu` (the only tier currently implemented in `sirius_extension.cpp:681-683`). Once host-tier support lands, set `SIRIUS_PIN_TIER=host` to flip the emitted SQL — no other change needed.

## Test plan

- [x] `bash -n` syntax-clean on both scripts; `pre-commit run` clean (black-formatted)
- [x] SF10 run with `--pinning-mode per-query`: all 22 queries match DuckDB (`validation.csv` all `success`), 168 `using cached_split_provider` hits, 0 fall-through misses, 72 pin / 72 unpin calls balanced
- [x] Per-query `sirius/q<N>/sirius.log` slices produced under `runs/.../`
- [x] Backward compatibility: omitting `--pinning-mode` produces a run identical in structure to today's runs
- [ ] Larger-scale run (SF100+) once available — pin amortization is more visible at I/O-bound scales

### DIFF
--- a/test/tpch_performance/CLAUDE.md
+++ b/test/tpch_performance/CLAUDE.md
@@ -70,13 +70,29 @@ export SIRIUS_CONFIG_FILE=$(pwd)/test/cpp/integration/integration.yaml
 ```
 
 Each run creates a directory under `runs/<timestamp>_sf<SF>_2iter/` containing:
-- `run_info.txt` — git branch/revision, tree clean/dirty, build freshness, hostname, memory, CPUs, GPUs, filesystem read benchmark
+- `run_info.txt` — git branch/revision, tree clean/dirty, build freshness, hostname, memory, CPUs, GPUs, filesystem read benchmark, pinning_mode setting
 - `run_info.patch` — full git diff when tree is dirty
 - `sirius_config.yaml` — copy of the Sirius config used
 - `sirius/` and `duckdb/` — per-engine logs, per-query results and timings
 - `validation.csv` — per-query match/error status
 - `comparison.txt` — cold/warm timing table with speedup ratios
 - `timings.csv` — long-format iteration runtimes (engine,query,iteration,runtime_s)
+
+#### `--pinning-mode per-query` (PR #721 pin_table)
+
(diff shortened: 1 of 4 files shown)


aminaramoon (approved):
Thanks Bobbi !

PR #773, opened 2026-05-08
scan: DuckDB format GPU-native decode kernel (Part 4: Float ALP/ALPRD)

ducndh:
# Summary

GPU decode kernels for DuckDB's ALP (Adaptive Lossless floating-Point) and ALPRD (Real Doubles) codecs. ALP encodes float values as bitpacked integer mantissas with a per-vector factor / exponent / frame-of-reference plus an exception list; ALPRD (the fallback for floats ALP can't compress) splits the raw float bit-pattern into a small dictionary-coded high part and a variable-width low part. One CTA decodes one 1024-row vector. Sibling of #737, #764; stacks on #736.

## DuckDB ALP on-disk format

```
ALP segment (one segment holds N 1024-row "vectors"):

byte 0   4                                 metadata_end          seg_end
│        │                                 │                     │
├─uint32─┼─── per-vector data (forward) ───┼── trailer table ────┤
│ end    │                                 │  uint32[N], V at    │
│        │                                 │  metadata_end-(V+1)*4
└─offset of metadata_end
```

```
ALP per-vector (compressed) — 13-byte header + bitstream + exceptions:

┌───┬───┬─────┬────────────────┬──┬────────────────┬───────┬─────────┐
│exp│fac│ ec  │ FOR (uint64)   │bw│ packed uint64  │ exc   │ exc pos │
│ u8│ u8│ u16 │                │u8│ mantissas      │ T[ec] │ u16[ec] │
└───┴───┴─────┴────────────────┴──┴────────────────┴───────┴─────────┘
0   1   2     4               12  13              ↑       ↑
                                                  │       │
                            bitstream = round_up(rows,32)*bw/8 bytes
```

- `exp == 0xFF` → uncompressed-sentinel: 1-byte `0xFF` then raw `T[rows]`.
- `bw ≤ sizeof(T)*8`. Exception positions bound-checked against `fill_rows`.
- Bitstream is padded to 32-row groups (`BitpackingPrimitives::GetRequiredSize`).

## DuckDB ALPRD on-disk format

```
ALPRD segment (small dict header lives between metadata_end and trailer):

byte 0   4   5   6   7                                 metadata_end ↘
│        │   │   │   │                                 │             │
├─uint32─┼─u8┼─u8┼─u8┼── uint16 dict[dict_size] ───────┼── per-vec ──┤
│ end    │ rb│ lb│ds │ (≤ 8 entries, lb ≤ 3)           │  data       │
└─offset
```

```
ALPRD per-vector (compressed) — 2-byte header + two bitstreams + exceptions:

┌─────┬───────────────────┬─────────────────────┬───────┬─────────┐
│ ec  │ left dict-idx     │ right parts         │ exc   │ exc pos │
│ u16 │ (lb-bit packed)   │ (rb-bit packed,     │ left  │ u16[ec] │
│     │                   │  EXACT_T = u32/u64) │ u16[] │         │
└─────┴───────────────────┴─────────────────────┴───────┴─────────┘
0     2
```

- `ec == 0xFFFF` → uncompressed-sentinel: 2-byte `0xFFFF` then raw `EXACT_T[rows]`.
- Reconstructed bits = `(uint64_t)dict[left_idx] << rb | right_part`, bit-cast to `T`.

## The decode in one picture

ALP, vector with `bw=11`, `exp=2`, `fac=0`, `FOR=100`:

```
packed mantissas (11-bit each):  [m₀ m₁ m₂ ... m₁₀₂₃]
                                          │
                                          ▼  unpack(idx, bw=11)
                                       (uint64 mantissa)
                                          │
                                          ▼  + FOR   (unsigned wrap; cast to int64)
                                       enc_int64
                                          │
                                          ▼  × FACT[fac]   × FRAC<T>[exp]
                                       T plain
                                          │
                                          ▼  __stwt
                                     out[row_offset + idx]

then, for each (pos, exc_value) in exception list:
    __stwt out[row_offset + pos] := exc_value     (bound-checked)
```

ALPRD, vector with `rb=48`, `lb=3`, `dict_size=8`:

```
left bitstream (3-bit):  [l₀ l₁ l₂ ...]   right bitstream (48-bit):  [r₀ r₁ r₂ ...]
                            │                                            │
                            ▼  unpack(idx, lb)                            ▼  unpack(idx, rb)
                          left_idx (0..7)                               right_part
                            │                                            │
                            ▼  bound-check < sh_dict_size                │
                          dict[left_idx]   (uint16, zero-extended)       │
                            │                                            │
                            └────────────────┬───────────────────────────┘
                                             ▼  (left_val << rb) | right_part
                                          uint64_t bits
                                             │
                                             ▼  memcpy → double / float
                                          plain T
                                             │
                                             ▼  __stwt  out[row_offset + idx]
```

## Per-vector pipeline — single kernel, one CTA per vector

```
HOST                          STREAM (one CUDA stream)
─────                         ─────────────────────────
build_vector_descs(run)
  expands segments → vectors
descs[N_VECS]   ────H2D────▶ kernel_decode_alp<T>      (or kernel_decode_alprd)
                               │  blockIdx.x = vid
                               │  thread 0:  parse header → shmem broadcast
                               │  __syncthreads
                               │  cooperative gmem→shmem stage of bitstream(s)
                               │  __syncthreads
                               │  every thread:  unpack → arith → __stwt
                               │  __syncthreads (only if exceptions exist)
                               │  exception scatter (bound-checked positions)
                               ▼
                             d_output (the column)
```

`build_vector_descs` runs on the host: walks the segment trailer table, expands `[seg, V]` into a flat list of per-vector launch descriptors. The kernel sees one CTA per vector.

## Decode kernel — header parse, cooperative stage, parallel decode

Each CTA owns one 1024-row vector. `BLOCK_DIM = 256` threads per CTA.

### Single-thread metadata parse

Vector header is 13 bytes (ALP) or 2 bytes (ALPRD-vector) starting at a runtime byte offset. Thread 0 walks the header, validates every field against the trusted `desc.vec_row_count`, publishes results through static-shmem scalars (`sh_invalid`, `sh_data_off`, `sh_exp`, `sh_fac`, `sh_for`, `sh_bw`, `sh_packed_bytes`). All other threads wait at the `__syncthreads` that follows. Cheap on this workload — the header is ~50 reads per CTA and the kernel is occupancy-saturated, so the SM hides the parse behind other CTAs' compute.

### Cooperative gmem→shmem stage

```
gmem packed bitstream:  [···bytes···][···bytes···][···bytes···]
                              │  (not 4-byte aligned in general)
                              ▼  cooperative copy
shmem packed_words[]:   [w₀ w₁ w₂ … w_live]│[guard][guard]
                                            └─ 2 guard words for the
                                               64-bit straddle case
                                               of unpack_from_shmem
```

Each thread reads one `uint32` per iteration via `memcpy(&v, src + off, 4)` (the canonical cross-arch idiom for unaligned uint32 load — nvcc lowers to `ld.u32`). `cg::memcpy_async` with `cuda::aligned_size_t<N>` would be the more portable primitive but isn't usable here: `packed_bytes = vp + 13u` is provably only 1-byte aligned (the per-vector header pushes the bitstream off any wider boundary), so the alignment hint can't be passed and the API falls back to byte-granularity loads on sm_75. Comment in `stage_bytes_to_shmem` explains the choice.

### Parallel decode loop

```cpp
for (i = threadIdx.x; i < fill_rows; i += blockDim.x) {
    u    = unpack_from_shmem<uint64_t>(shmem, i, bw);   // 3-word read + shift + mask
    enc  = (int64_t)(u + sh_for);                       // unsigned wrap, then sign
    plain = (T)enc * d_alp_fact[sh_fac] * alp_frac<T>[sh_exp];
    __stwt(out + i, plain);
}
```

ALPRD's loop is the same shape but unpacks two streams (`right_words` and `left_words`), gathers `sh_dict[left_idx]`, and combines the bits before bit-casting to `T`.

`__stwt` is "store with no L1 cache" — output is write-once, never reread by this kernel. ALPRD's exception scatter does re-read `out[pos]` to recover the right-bits, which is correct because `__stwt` invalidates the issuing SM's L1 line, so the post-`__syncthreads` read refills from L2 with the freshly written value.

## Malformed input → zero-fill

`sh_invalid == true` is the universal sentinel collapsing every malformed class. The kernel zero-fills `desc.vec_row_count` rows using the host-supplied (trusted) row count and returns. Output buffer is uninitialised by the dispatcher, so this is the only thing between a corrupt segment and leaking prior device contents.

| Codec | Detected by | Condition |
|---|---|---|
| ALP / ALPRD | header parse | `seg_bytes < 4` |
| ALP / ALPRD | header parse | `metadata_end > seg_bytes` or `metadata_end < trailer_need` |
| ALP / ALPRD | header parse | `data_off < 4` or `data_off >= metadata_end` |
| ALP | per-vector | `bw > sizeof(T)*8` |
| ALP | per-vector | exception payload past `metadata_end` |
| ALP | per-vector | `factor > 18` or `exponent > MAX_EXPONENT` (10 / f32, 18 / f64) — keeps `d_alp_fact` / `alp_frac<T>` in-bounds |
| ALPRD | header parse | `right_bw > sizeof(T)*8` |
| ALPRD | header parse | `left_bw > 3` |
| ALPRD | header parse | `dict_size > MAX_DICTIONARY_SIZE` (= 8) |
| ALP / ALPRD | uncompressed sentinel | raw values would overrun `metadata_end` |
| ALPRD | dict gather | corrupt `left_idx >= sh_dict_size` → reads `EXACT_T(0)` (sh_dict is zero-init) |

## Test coverage

`test/cpp/scan/test_gpu_decode_alp.cpp`:
- ALP happy-path: single compressed vector (f64, factor/exponent ≠ 0); single compressed vector with exceptions (f32); uncompressed-sentinel; multi-vector segment crossing the 1024-row boundary
- ALPRD happy-path: canonical layout (f64); exception left-part patches; uncompressed-sentinel (f32)
- 8 canary defensive tests (pre-fill output 0xCC, REQUIRE every byte zero post-decode) covering each malformed-input class above
- 7 bench-scale verify cases (one per microbench shape) that reproduce the bench segment layout and assert every output row against a host model — guards against silent miscompiles since the bench discards its decoded output

`test/cpp/scan/bench_decode_codecs.cpp` adds 7 microbenches:
- ALP f64 bw=8 / 11 / 16, ALP f32 bw=20 / 32, ALPRD f64 r48/l3 + r56/l2

`test/data/decode_baselines/sm75.json` populated with median-of-3 throughputs as a regression gate.

### DIFF
--- a/src/cuda/scan/gpu_decode_alp.cu
+++ b/src/cuda/scan/gpu_decode_alp.cu
@@ -14,42 +14,47 @@
  * limitations under the License.
  */
 
-// ALP / ALPRD decode kernels. One CTA = one 1024-row vector. On any bounds
-// check failure the kernel zero-fills `desc.vec_row_count` rows using the
-// trusted host descriptor row count, never parsed metadata.
+//===----------------------------------------------------------------------===//
+// clang-format off
+// ALP / ALPRD decode kernels. The work partitioning is 1 CTA per 1024-row vector.
+// On any bounds check failure the kernel zero-fills `desc.vec_row_count` rows.
 //
-// On-disk layouts (verified against
-// duckdb/src/include/duckdb/storage/compression/{alp,alprd}/):
+// On-disk layouts (see: duckdb/src/include/duckdb/storage/compression/{alp,alprd}/):
 //
 //   ALP segment (one segment holds N 1024-row "vectors"):
 //
-//   byte 0   4                                 metadata_end          seg_end
-//   │        │                                 │                     │
-//   ├─uint32─┼─── per-vector data (forward) ───┼── trailer table ────┤
-//   │ end    │                                 │  uint32[N], V at    │
-//   │        │                                 │  metadata_end-(V+1)*4
-//   └─offset of metadata_end
+//   byte 0   4                                                                    metadata_end
+//   │        │                                 │                                  │
+//   ├─uint32─┼─── per-vector data (forward) ───┼── trailer table (backward) ──────┤
+//   │ end    │                                 │  uint32[N], V at                 │
+//   │        │                                 │  metadata_end-(V+1)*4            │
+//     └─offset of metadata_end
 //
 //   ALP per-vector (compressed) — 13-byte header + bitstream + exceptions:
 //
 //   ┌───┬───┬─────┬────────────────┬──┬────────────────┬───────┬─────────┐
 //   │exp│fac│ ec  │ FOR (uint64)   │bw│ packed uint64  │ exc   │ exc pos │
-//   │ u8│ u8│ u16 │                │u8│ mantissas      │ T[ec] │ u16[ec] │
+//   │ u8│ u8│ u16 │                │u8│ codes          │ T[ec] │ u16[ec] │
 //   └───┴───┴─────┴────────────────┴──┴────────────────┴───────┴─────────┘
-//   0   1   2     4               12  13              ↑       ↑
-//                                                     │       │
-//                                bitstream = round_up(rows,32)*bw/8 bytes
+//   0   1   2     4               12  13              [A]     [B]      [C]
 //
-//   exp == 0xFF → uncompressed-sentinel: 1-byte 0xFF then raw T[rows].
-//   bw ≤ sizeof(T)*8. Exception position bound-checked against fill_rows.
+//   [A] = 13 + round_up(rows, 32) * bw / 8   (end of bitstream)
+//   [B] = [A] + ec * sizeof(T)               (end of exception values)
+//   [C] = [B] + ec * sizeof(u16)             (end of vector)
+//
+//   - Decoded value = code * 10^fac * 10^(-exp).
+//   - Exception positions are walked after decoding to install the exception values at the
+//     corresponding vector positions.
+//   - exp == 0xFF → uncompressed-sentinel: 1-byte 0xFF then raw T[rows].
 //
 //   ALPRD segment (small dict header lives between metadata_end and trailer):
 //
-//   byte 0   4   5   6   7                                 metadata_end ↘
-//   │        │   │   │   │                                 │             │
-//   ├─uint32─┼─u8┼─u8┼─u8┼── uint16 dict[dict_size] ───────┼── per-vec ──┤
-//   │ end    │ rb│ lb│ds │ (≤ 8 entries, lb ≤ 3)           │  data       │
-//   └─offset
+//   byte 0   4   5   6   7                                                                       metadata_end
+//   │        │   │   │   │                                 │             │                       │
+//   ├─uint32─┼─u8┼─u8┼─u8┼── uint16 dict[dict_size] ───────┼── per-vec ──┼── trailer (backward) ─┤
+//   │ end    │ rb│ lb│ds │   (≤ 8 entries, lb ≤ 3)         │   data      │  uint32[N], V at      │
+//   │        │   │   │   │                                 │             │  metadata_end-(V+1)*4 │
+//     └─offset of metadata_end
 //
 //   ALPRD per-vector (compressed) — 2-byte header + two bitstreams + exceptions:
 //
@@ -60,9 +65,12 @@
 //   └─────┴───────────────────┴─────────────────────┴───────┴─────────┘
 //   0     2
 //
-//   ec == 0xFFFF → uncompressed-sentinel: 2-byte 0xFFFF then raw EXACT_T[rows].
-//   Reconstructed bits = (uint64_t)dict[left_idx] << rb | right_part, then
-//   bit-cast to T (double / float).
+//   - Decoded value = dict[left_idx] << rb | right_part
+//   - Exception positions are walked after decoding to install the exception values at the
+//     corresponding vector positions.
+//   - ec == 0xFFFF → uncompressed-sentinel: 2-byte 0xFFFF then raw EXACT_T[rows].
+//===----------------------------------------------------------------------===//
+// clang-format on
 
 #include "cuda/scan/gpu_decode_alp.cuh"
 
@@ -70,8 +78,15 @@
 #include <rmm/device_uvector.hpp>
 
 #include <cuda/cmath>
+#include <cuda/std/limits>
+#include <cuda/std/numeric>
+#include <cuda/std/type_traits>
 #include <cuda_runtime.h>
 
+#include <duckdb/storage/compression/alp/alp_constants.hpp>
+#include <duckdb/storage/compression/alprd/alprd_constants.hpp>
+
+#include <climits>
 #include <cstdint>
 #include <cstring>
 #include <stdexcept>
@@ -84,14 +99,12 @@ namespace sirius::cuda::scan {
 namespace {
 
 constexpr uint32_t BLOCK_DIM = 256;
-
-// Max factor index — the encoder couples (factor, exponent), so this bound
-// is shared across f32 and f64. `MAX_EXPONENT` lives on AlpTypedConstants<T>
-// because exponent IS type-specific.
 constexpr uint8_t ALP_MAX_FACTOR =
   static_cast<uint8_t>(std::extent_v<decltype(duckdb::AlpConstants::FACT_ARR)> - 1);
 
-/// One CTA's unit of work for ALP / ALPRD: one vector within one segment.
+/**
+ * @brief Vector-level descriptor for a CTA
+ */
 struct alp_vector_desc {
   uint8_t const* d_segment;    ///< Device pointer to the segment's first byte.
   uint32_t segment_bytes;      ///< Size of the staged segment buffer.
@@ -100,61 +113,115 @@ struct alp_vector_desc {
   uint32_t global_row_offset;  ///< Output offset, in rows, for this vector.
 };
 
-// Mirrors of `AlpConstants::FACT_ARR` and `AlpTypedConstants<T>::FRAC_ARR`.
-__device__ __constant__ int64_t d_alp_fact[19] = {1LL,
-                                                  10LL,
-                                                  100LL,
-                                                  1000LL,
-                                                  10000LL,
-                                                  100000LL,
-                                                  1000000LL,
-                                                  10000000LL,
-                                                  100000000LL,
-                                                  1000000000LL,
-                                                  10000000000LL,
-                                                  100000000000LL,
-                                                  1000000000000LL,
-                                                  10000000000000LL,
-                                                  100000000000000LL,
-                                                  1000000000000000LL,
-                                                  10000000000000000LL,
-                                                  100000000000000000LL,
-                                                  1000000000000000000LL};
+/**
+ * @brief Per-vector parse outcome. Routes the CTA to one of three paths:
+ * zero-fill (invalid), raw bit-cast (uncompressed sentinel), or bitpacked
+ * decode (compressed).
+ */
+enum class alp_parse_status : uint8_t { invalid, uncompressed, compressed };
 
-__device__ __constant__ float d_alp_frac_f32[11] = {1.0F,
-                                                    0.1F,
-                                                    0.01F,
-                                                    0.001F,
-                                                    0.0001F,
-                                                    0.00001F,
-                                                    0.000001F,
-                                                    0.0000001F,
-                                                    0.00000001F,
-                                                    0.000000001F,
-                                                    0.0000000001F};
+/**
+ * @brief Parsed per-vector metadata. Thread 0 populates this from the
+ * segment header + trailer; the rest of the CTA reads it through shmem.
+ */
+struct alp_vector_meta {
+  alp_parse_status status;
+  uint32_t data_off;              ///< Byte offset of this vector's data within the segment.
+  uint8_t exp;                    ///< Decode exponent index.
+  uint8_t fac;                    ///< Decode factor index.
+  uint16_t exc_count;             ///< Number of exception positions.
+  uint64_t for_val;               ///< Frame-of-reference (added to each unpacked code).
+  uint8_t bw;                     ///< Bitpack width.
+  uint8_t const* packed_bytes_p;  ///< Pointer to the byte-stream of packed codes.
+  uint32_t packed_bytes_count;    ///< Size of the bitpacked stream in bytes.
+};
 
-__device__ __constant__ double d_alp_frac_f64[21] = {1.0,
-                                                     0.1,
-                                                     0.01,
-                                                     0.001,
-                                                     0.0001,
-                                                     0.00001,
-                                                     0.000001,
-                                                     0.0000001,
-                                                     0.00000001,
-                                                     0.000000001,
-                                                     0.0000000001,
-                                                     0.00000000001,
-                                                     0.000000000001,
-                                                     0.0000000000001,
-                                                     0.00000000000001,
-                                                     0.000000000000001,
-                                                     0.0000000000000001,
-                                                     0.00000000000000001,
-                                                     0.000000000000000001,
-                                                     0.0000000000000000001,
-                                                     0.00000000000000000001};
+/**
+ * @brief Per-vector ALPRD parse outcome — see `alp_parse_status` for the
+ * three routing paths.
+ */
+enum class alprd_parse_status : uint8_t { invalid, uncompressed, compressed };
 
+/**
+ * @brief Parsed per-vector ALPRD metadata. Thread 0 populates this from the
+ * segment header + trailer; the rest of the CTA reads it through shmem. The
+ * segment dictionary is loaded separately by a cooperative dict-load.
+ */
+struct alprd_vector_meta {
+  alprd_parse_status status;
+  uint32_t data_off;                    ///< Byte offset of this vector's data within the segment.
+  uint16_t exc_count;                   ///< Number of exception positions.
+  uint8_t right_bw;                     ///< Right-bits packed width.
+  uint8_t left_bw;                      ///< Left-bits packed width.
+  uint8_t dict_size;                    ///< Number of entries in the segment dictionary.
+  uint8_t const* left_packed_bytes_p;   ///< Pointer to the byte-stream of left dict-indices.
+  uint32_t left_packed_bytes_count;     ///< Size of the left bitpacked stream in bytes.
+  uint8_t const* right_packed_bytes_p;  ///< Pointer to the byte-stream of right parts.
+  uint32_t right_packed_bytes_count;    ///< Size of the right bitpacked stream in bytes.
+};
+
+/**
+ * @brief Factor and exponent tables for ALP decoding: decoded val = code * fact[fac] * frac[exp].
+ * Placed in constant memory. Mirrors of `AlpConstants::FACT_ARR` and
+ * `AlpTypedConstants<T>::FRAC_ARR`.
+ */
+__device__ __constant__ int64_t d_alp_fact[] = {1LL,
+                                                10LL,
+                                                100LL,
+                                                1000LL,
+                                                10000LL,
+                                                100000LL,
+                                                1000000LL,
+                                                10000000LL,
+                                                100000000LL,
+                                                1000000000LL,
+                                                10000000000LL,
+                                                100000000000LL,
+                                                1000000000000LL,
+                                                10000000000000LL,
+                                                100000000000000LL,
+                                                1000000000000000LL,
+                                                10000000000000000LL,
+                                                100000000000000000LL,
+                                                1000000000000000000LL};
+
+__device__ __constant__ float d_alp_frac_f32[] = {1.0F,
+                                                  0.1F,
+                                                  0.01F,
+                                                  0.001F,
+                                                  0.0001F,
+                                                  0.00001F,
+                                                  0.000001F,
+                                                  0.0000001F,
+                                                  0.00000001F,
+                                                  0.000000001F,
+                                                  0.0000000001F};
+
+__device__ __constant__ double d_alp_frac_f64[] = {1.0,
+                                                   0.1,
+                                                   0.01,
+                                                   0.001,
+                                                   0.0001,
+                                                   0.00001,
+                                                   0.000001,
+                                                   0.0000001,
+                                                   0.00000001,
+                                                   0.000000001,
+                                                   0.0000000001,
+                                                   0.00000000001,
+                                                   0.000000000001,
+                                                   0.0000000000001,
+                                                   0.00000000000001,
+                                                   0.000000000000001,
+                                                   0.0000000000000001,
+                                                   0.00000000000000001,
+                                                   0.000000000000000001,
+                                                   0.0000000000000000001,
+                                                   0.00000000000000000001};
+
+/**
+ * @brief Get the pointer to the fraction array in constant memory for float/double
+ */
 template <typename T>
 __device__ __forceinline__ T const* alp_frac_ptr();
 template <>
@@ -168,107 +235,107 @@ __device__ __forceinline__ double const* alp_frac_ptr<double>()
   return d_alp_frac_f64;
 }
 
-// Per-vector data starts at arbitrary byte offsets — metadata reads cannot
-// assume 2/4/8-byte alignment, so go through `memcpy`.
-__device__ __forceinline__ uint16_t ld16(uint8_t const* p)
+/**
+ * @brief Perform an unaligned load of type @p T from buffer @p p.
+ *
+ * Per-vector data starts at arbitrary byte offsets, so metadata reads cannot make alignment
+ * assumptions.
+ */
+template <typename T>
+__device__ __forceinline__ T ld_unaligned(uint8_t const* p)
 {
-  uint16_t v;
+  T v;
   memcpy(&v, p, sizeof(v));
   return v;
 }
 
-__device__ __forceinline__ uint32_t ld32(uint8_t const* p)
+/**
+ * @brief Calculate the number of bytes for a bitpacked buffer given @p row_count rows and @p width
+ * bits per row.
+ *
+ * Mirror of DuckDB's `BitpackingPrimitives::GetRequiredBytes`.
+ */
+__device__ __host__ __forceinline__ uint32_t bp_required_bytes(uint32_t row_count, uint32_t width)
 {
-  uint32_t v;
-  memcpy(&v, p, sizeof(v));
-  return v;
+  // DuckDB packs values in groups of 32
+  constexpr uint32_t BITPACK_GROUP = 32;
+
+  if (width == 0) return 0;
+  // Round up to the nearest multiple of 32
+  auto const n_rounded = ::cuda::ceil_div(row_count, BITPACK_GROUP) * BITPACK_GROUP;
+  // Count the number of bytes needed (n_rounded is a multiple of 8 by definition)
+  return (n_rounded * width) / 8;
 }
 
-__device__ __forceinline__ uint64_t ld64(uint8_t const* p)
-{
-  uint64_t v;
-  memcpy(&v, p, sizeof(v));
-  return v;
-}
-
-// Mirror of `BitpackingPrimitives::GetRequiredSize`.
-__device__ __host__ __forceinline__ uint32_t bp_required_bytes(uint32_t count, uint32_t width)
-{
-  if (width == 0u) return 0u;
-  uint32_t const rounded = ::cuda::ceil_div(count, 32u) * 32u;
-  return rounded * width / 8u;
-}
-
-// Caller must supply 2 trailing guard words past the live data (the
-// 64-bit straddle case reads up to 3 consecutive 32-bit words).
+/**
+ * @brief Unpack a value of type ResT from @p packed bitstream.
+ * @note The caller must supply 2 trailing guard words past the live data.
+ */
 template <typename ResT>
 __device__ __forceinline__ ResT unpack_from_shmem(uint32_t const* packed,
                                                   uint32_t idx,
                                                   uint32_t width)
 {
-  if (width == 0u) return ResT(0);
+  constexpr uint32_t WORD_BYTES       = sizeof(uint32_t);
+  constexpr uint32_t WORD_BITS        = ::cuda::std::numeric_limits<uint32_t>::digits;
+  constexpr uint32_t DOUBLE_WORD_BITS = ::cuda::std::numeric_limits<uint64_t>::digits;
+  constexpr uint64_t DOUBLE_WORD_MAX  = ::cuda::std::numeric_limits<uint64_t>::max();
 
-  uint64_t bit_pos  = static_cast<uint64_t>(idx) * width;
-  uint32_t word_idx = static_cast<uint32_t>(bit_pos >> 5);
-  uint32_t bit_off  = static_cast<uint32_t>(bit_pos & 31u);
+  if (width == 0) return 0;
 
-  uint64_t combined = static_cast<uint64_t>(packed[word_idx]);
-  if (bit_off + width > 32u) { combined |= static_cast<uint64_t>(packed[word_idx + 1]) << 32; }
-  uint64_t result = combined >> bit_off;
+  auto const bit_pos  = static_cast<uint64_t>(idx) * width;
+  auto const word_idx = static_cast<uint32_t>(bit_pos / WORD_BITS);
+  auto const bit_off  = static_cast<uint32_t>(bit_pos % WORD_BITS);
 
-  if constexpr (sizeof(ResT) > 4) {
-    if (bit_off > 0 && bit_off + width > 64u) {
-      result |= static_cast<uint64_t>(packed[word_idx + 2]) << (64u - bit_off);
+  auto combined = static_cast<uint64_t>(packed[word_idx]);
+  if (bit_off + width > WORD_BITS) {
+    combined |= static_cast<uint64_t>(packed[word_idx + 1]) << WORD_BITS;
+  }
+  auto result = combined >> bit_off;
+
+  if constexpr (sizeof(ResT) > WORD_BYTES) {
+    static_assert(::cuda::std::is_same_v<ResT, uint64_t>);
+    if (bit_off > 0 && bit_off + width > DOUBLE_WORD_BITS) {
+      result |= static_cast<uint64_t>(packed[word_idx + 2]) << (DOUBLE_WORD_BITS - bit_off);
     }
   }
 
-  uint64_t mask = (width >= 64u) ? ~uint64_t{0} : ((uint64_t{1} << width) - 1u);
+  auto const mask = (width >= DOUBLE_WORD_BITS) ? DOUBLE_WORD_MAX : ((uint64_t{1} << width) - 1u);
   return static_cast<ResT>(result & mask);
 }
 
-// `src` is any byte alignment; `byte_count` need not be a multiple of 4.
-// We can't use `cg::memcpy_async` with `cuda::aligned_size_t<N>` here:
-// `src = vp + 13` where `vp = seg_base + data_off` and `data_off` varies
-// per vector at runtime, so neither 4- nor 16-byte alignment is provable.
-// Each thread doing `memcpy(&v, src + off, 4)` lowers to `ld.u32` (with
-// the unaligned-access handler on sm_75) — same primitive cg::memcpy_async
-// would use on an aligned middle, without the prefix/middle/tail split.
+/**
+ * @brief Copy bytes from @p src to @p dst_words in SMEM. memcpy is used because there is no
+ * guarantee on the byte alignment for @p src pointer.
+ */
 __device__ __forceinline__ void stage_bytes_to_shmem(uint32_t* dst_words,
                                                      uint8_t const* src,
                                                      uint32_t byte_count,
                                                      uint32_t pad_words)
 {
-  uint32_t live_words = ::cuda::ceil_div(byte_count, 4u);
+  constexpr uint32_t WORD_BYTES = sizeof(uint32_t);
+
+  auto const live_words = ::cuda::ceil_div(byte_count, WORD_BYTES);
   for (uint32_t w = threadIdx.x; w < live_words; w += blockDim.x) {
-    uint32_t off = w * 4u;
-    uint32_t v   = 0u;
-    if (off + 4u <= byte_count) {
-      memcpy(&v, src + off, 4);
+    auto const byte_offset = w * WORD_BYTES;
+    uint32_t v             = 0;
+    if (byte_offset + WORD_BYTES <= byte_count) {
+      memcpy(&v, src + byte_offset, WORD_BYTES);
     } else {
       // Tail: at most 3 leftover bytes. Zero-pad the high bytes.
-      uint32_t left = byte_count - off;
-      memcpy(&v, src + off, left);
+      auto const left = byte_count - byte_offset;
+      memcpy(&v, src + byte_offset, left);
     }
     dst_words[w] = v;
   }
   for (uint32_t w = live_words + threadIdx.x; w < live_words + pad_words; w += blockDim.x) {
-    dst_words[w] = 0u;
+    dst_words[w] = 0;
   }
 }
 
-// Read a value the kernel previously wrote with `__stwt`. Correct only after
-// a __syncthreads barrier separating the write from this read: `__stwt`
-// invalidates the issuing SM's L1 line, so the post-sync ordinary load
-// refills from L2 with the freshly written value. Any change to the main
-// loop's store hint (e.g., switching to `__stcg` or vectorized stores) must
-// revisit this helper. Used by ALPRD's exception scatter to recover
-// right-bits without a second pre-pass over `out`.
-template <typename T>
-__device__ __forceinline__ T read_back_after_stwt(T const* out, uint32_t pos)
-{
-  return out[pos];
-}
-
+/**
+ * @brief Zero-fill @p out buffer with @p row_count zeros in the case of malfomation.
+ */
 template <typename T>
 __device__ __forceinline__ void zero_fill_vector(T* out, uint32_t row_count)
 {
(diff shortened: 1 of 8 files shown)


src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,803 @@
...
+    case 4:
+      launch_alprd_typed<float>(
+        descs.data(), descs.size(), reinterpret_cast<float*>(d_output), stream, mr);
+      break;
+    case 8:
+      launch_alprd_typed<double>(
+        descs.data(), descs.size(), reinterpret_cast<double*>(d_output), stream, mr);
+      break;
+    default: throw_unsupported_type(type_size, "ALPRD");
+  }
```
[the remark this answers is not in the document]
ducndh: Pushing back on this one — adding  validation here-only would diverge from the codebase's documented convention.  (gpu_native_decode.cu:549-561) explicitly delegates DuckDB→cudf type-narrowing checks to the upstream viability walker and limits this layer to obvious cases (variable-width types). `decode_constant_data` and `decode_uncompressed_data` both follow that policy — none of them validate `type.id()`.

The gap you describe is real (`type_size==4` is true for both INT32 and FLOAT32, and the kernel would silently mis-decode), but "viability walker enforces codec↔type compatibility" is the codebase's stance. Tightening that should happen uniformly at the dispatcher (`dispatch_data_run`) in a follow-up rather than piecemeal in one codec.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+}
+
+// `src` is any byte alignment; `byte_count` need not be a multiple of 4.
+// We can't use `cg::memcpy_async` with `cuda::aligned_size_t<N>` here:
+// `src = vp + 13` where `vp = seg_base + data_off` and `data_off` varies
+// per vector at runtime, so neither 4- nor 16-byte alignment is provable.
+// Each thread doing `memcpy(&v, src + off, 4)` lowers to `ld.u32` (with
+// the unaligned-access handler on sm_75) — same primitive cg::memcpy_async
+// would use on an aligned middle, without the prefix/middle/tail split.
+__device__ __forceinline__ void stage_bytes_to_shmem(uint32_t* dst_words,
```
kevkrist, line 311: The same alignment issues are true in the bitpacking case. There, you optimized for the case where the alignment is 16B. Here, you don't. I recommend using this method to modularize the bit-unpacking kernel and to provide cross-kernel consistency and optimization propagation. It also remains an open question whether cg::memcpy_async is a cross-architecture loss. On sm80+ the async copy does in fact directly copy from L2 to SMEM, whereas memcpy goes through registers. In general, avoid small optimizations for specific architectures if it means avoiding APIs that can unlock performance on other architectures.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  uint64_t mask = (width >= 64u) ? ~uint64_t{0} : ((uint64_t{1} << width) - 1u);
+  return static_cast<ResT>(result & mask);
+}
+
+// `src` is any byte alignment; `byte_count` need not be a multiple of 4.
+// We can't use `cg::memcpy_async` with `cuda::aligned_size_t<N>` here:
+// `src = vp + 13` where `vp = seg_base + data_off` and `data_off` varies
+// per vector at runtime, so neither 4- nor 16-byte alignment is provable.
+// Each thread doing `memcpy(&v, src + off, 4)` lowers to `ld.u32` (with
+// the unaligned-access handler on sm_75) — same primitive cg::memcpy_async
```
kevkrist: Please erase architecture-specific comments from released code (here, sm_75). You are correct that you cannot use cg::memcpy_async with the alignment unless you check the 16B alignment, which you do in the bitpacking kernel. You could use the default cg::memcpy_async() and unlock the async direct L2->SMEM copy on sm_80+.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+// would use on an aligned middle, without the prefix/middle/tail split.
+__device__ __forceinline__ void stage_bytes_to_shmem(uint32_t* dst_words,
+                                                     uint8_t const* src,
+                                                     uint32_t byte_count,
+                                                     uint32_t pad_words)
+{
+  uint32_t live_words = ::cuda::ceil_div(byte_count, 4u);
+  for (uint32_t w = threadIdx.x; w < live_words; w += blockDim.x) {
+    uint32_t off = w * 4u;
```
kevkrist: Please consider using a static constant like WORD_BYTES here so that 4 feels less like a magic number.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  }
+
+  uint64_t mask = (width >= 64u) ? ~uint64_t{0} : ((uint64_t{1} << width) - 1u);
+  return static_cast<ResT>(result & mask);
+}
+
+// `src` is any byte alignment; `byte_count` need not be a multiple of 4.
+// We can't use `cg::memcpy_async` with `cuda::aligned_size_t<N>` here:
+// `src = vp + 13` where `vp = seg_base + data_off` and `data_off` varies
+// per vector at runtime, so neither 4- nor 16-byte alignment is provable.
```
kevkrist: not provable *at compile-time*

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+// Per-vector data starts at arbitrary byte offsets — metadata reads cannot
+// assume 2/4/8-byte alignment, so go through `memcpy`.
+__device__ __forceinline__ uint16_t ld16(uint8_t const* p)
+{
+  uint16_t v;
+  memcpy(&v, p, sizeof(v));
+  return v;
+}
+
+__device__ __forceinline__ uint32_t ld32(uint8_t const* p)
```
kevkrist: If you like using this API, consider naming ld32_unaligned to make clear that its used when the source pointer's alignment is not known. Also, you should use this with the metadata parsing in the other decompression modules.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+// 64-bit straddle case reads up to 3 consecutive 32-bit words).
+template <typename ResT>
+__device__ __forceinline__ ResT unpack_from_shmem(uint32_t const* packed,
+                                                  uint32_t idx,
+                                                  uint32_t width)
+{
+  if (width == 0u) return ResT(0);
+
+  uint64_t bit_pos  = static_cast<uint64_t>(idx) * width;
+  uint32_t word_idx = static_cast<uint32_t>(bit_pos >> 5);
```
kevkrist: You can use `/ 32` and `% 32` -- the compiler will resolve those to these bit manipulation instructions by strength reduction, and your intent will be clearer.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  uint64_t v;
+  memcpy(&v, p, sizeof(v));
+  return v;
+}
+
+// Mirror of `BitpackingPrimitives::GetRequiredSize`.
+__device__ __host__ __forceinline__ uint32_t bp_required_bytes(uint32_t count, uint32_t width)
+{
+  if (width == 0u) return 0u;
+  uint32_t const rounded = ::cuda::ceil_div(count, 32u) * 32u;
```
kevkrist: It's not clear what 'count' is a count of, and thus what the numbers 32 and 8 mean here. I assume count is row count, so rounded is the number of words you need? Please make that clearer, either with better names or with function documentation (please don't bloat with too much AI).

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  return rounded * width / 8u;
+}
+
+// Caller must supply 2 trailing guard words past the live data (the
+// 64-bit straddle case reads up to 3 consecutive 32-bit words).
+template <typename ResT>
+__device__ __forceinline__ ResT unpack_from_shmem(uint32_t const* packed,
```
kevkrist, line 275: This function is identical to unpack_value in gpu_decode_bitpacking.cu. Please deduplicate. Consider making a utils file with common utilities across the different decoding modules.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+    sh_uncompressed = false;
+    sh_data_off     = 0;
+    sh_exp          = 0;
+    sh_fac          = 0;
+    sh_exc_count    = 0;
+    sh_for          = 0;
+    sh_bw           = 0;
+    sh_packed_bytes = 0;
+
+    if (seg_bytes < 4u) {
```
kevkrist:
```suggestion
    if (seg_bytes < sizeof(uint32_t)) {
```
Even better, use `duckdb::AlpRDConstants::METADATA_POINTER_SIZE`

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+
+// Read a value the kernel previously wrote with `__stwt`. Correct only after
+// a __syncthreads barrier separating the write from this read: `__stwt`
+// invalidates the issuing SM's L1 line, so the post-sync ordinary load
+// refills from L2 with the freshly written value. Any change to the main
+// loop's store hint (e.g., switching to `__stcg` or vectorized stores) must
+// revisit this helper. Used by ALPRD's exception scatter to recover
+// right-bits without a second pre-pass over `out`.
+template <typename T>
+__device__ __forceinline__ T read_back_after_stwt(T const* out, uint32_t pos)
```
kevkrist: What is the point of this function?
Is the idea that there is an implicit contract? 'Before using this function you must invoke __syncthreads() after a __stwt() to the same L1 cache line'?

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+{
+  return out[pos];
+}
+
+template <typename T>
+__device__ __forceinline__ void zero_fill_vector(T* out, uint32_t row_count)
```
kevkrist, line 340: This can also be used in the other DuckDB decoding modules. If you want to use it here, you should also use it elsewhere.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+    sh_exc_count    = 0;
+    sh_for          = 0;
+    sh_bw           = 0;
+    sh_packed_bytes = 0;
+
+    if (seg_bytes < 4u) {
+      sh_invalid = true;
+    } else {
+      uint32_t metadata_end = ld32(seg_base);
+      uint64_t trailer_need = uint64_t{4u} + (uint64_t{desc.vec_idx} + 1u) * 4u;
```
kevkrist: The math for trailer_need is obscure. After thinking about it a while, it's clear that it's because we need metadata_end - (vec_idx + 1)*4 >= 4, so that 4B metadata entry at metadata_end - (vec_idx + 1) * 4 fits. But this is lost and I shouldn't have to spend too much time reconstructing.

kevkrist (changes requested):
This is a start.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+      if (metadata_end > seg_bytes || metadata_end < trailer_need) {
+        sh_invalid = true;
+      } else {
+        // Vector pointer must land in (4, metadata_end).
+        uint32_t entry_off = metadata_end - (desc.vec_idx + 1u) * 4u;
+        uint32_t data_off  = ld32(seg_base + entry_off);
+        if (data_off < 4u || data_off >= metadata_end) {
+          sh_invalid = true;
+        } else {
```
kevkrist: This metadata parsing is deeply nested. I think this can be addressed with a separate metadata parsing function that returns whenever sh_invalid is set. It would serve 2 purposes: declutter the kernel and avoid deeply nested conditional blocks.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+//   ├─uint32─┼─── per-vector data (forward) ───┼── trailer table ────┤
+//   │ end    │                                 │  uint32[N], V at    │
+//   │        │                                 │  metadata_end-(V+1)*4
+//   └─offset of metadata_end
+//
+//   ALP per-vector (compressed) — 13-byte header + bitstream + exceptions:
+//
+//   ┌───┬───┬─────┬────────────────┬──┬────────────────┬───────┬─────────┐
+//   │exp│fac│ ec  │ FOR (uint64)   │bw│ packed uint64  │ exc   │ exc pos │
+//   │ u8│ u8│ u16 │                │u8│ mantissas      │ T[ec] │ u16[ec] │
```
kevkrist: Is mantissa the right word here? This is something different from the mantissa in an IEEE floating-point standard represetnation.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+        // Vector pointer must land in (4, metadata_end).
+        uint32_t entry_off = metadata_end - (desc.vec_idx + 1u) * 4u;
+        uint32_t data_off  = ld32(seg_base + entry_off);
+        if (data_off < 4u || data_off >= metadata_end) {
+          sh_invalid = true;
+        } else {
+          sh_data_off = data_off;
+          uint8_t exp = seg_base[data_off];
+          if (exp == ALP_UNCOMPRESSED_SENTINEL) {
+            uint64_t raw_end = uint64_t{data_off} + 1u + uint64_t{fill_rows} * sizeof(T);
```
kevkrist: The +1u is a little magical to me. Can you please use something like `duckdb::AlpConstants::EXPONENT_SIZE`?

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+            if (raw_end > metadata_end) {
+              sh_invalid = true;
+            } else {
+              sh_uncompressed = true;
+            }
+          } else if (uint64_t{data_off} + 13u > metadata_end) {
```
kevkrist: The +13u is a little magical to me. Can you consider either a) documenting that this is the compressed vector header size or b) defining a static constant `COMPRESSED_VECTOR_HEADER_SIZE`? This applies also below, where 13u appears in the payload_end calculation.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+            sh_invalid = true;
+          } else {
+            uint8_t const* vp  = seg_base + data_off;
+            uint16_t exc_count = ld16(vp + 2);
+            uint64_t for_value = ld64(vp + 4);
+            uint8_t bw         = vp[12];
+
+            if (bw > sizeof(T) * 8u) {
```
kevkrist: Consider using `::cuda::std::numeric_limits<T>::digits()` for extra clarity.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+#include <stdexcept>
+#include <string>
+#include <type_traits>
+#include <vector>
+
+namespace sirius::cuda::scan {
+
+namespace {
+
+constexpr uint32_t BLOCK_DIM = 256;
```
kevkrist, line 101: If you intend to have every codec decompression module use the same value of BLOCK_DIM, consider defining this somewhere once, rather than in the source file for each module.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+              if (payload_end > metadata_end) {
+                sh_invalid = true;
+              } else {
+                // ALP encoder couples (factor, exponent): factor is bounded
+                // by FACT_ARR extent (shared across types), exponent is
+                // bounded by AlpTypedConstants<T>::MAX_EXPONENT (10/f32,
+                // 18/f64). Out-of-range corrupt values would crash the
+                // constant-memory load.
+                uint8_t fac          = vp[1];
+                uint8_t max_exponent = duckdb::AlpTypedConstants<T>::MAX_EXPONENT;
```
kevkrist: Can you just use `duckdb::AlpTypedConstants<T>::MAX_EXPONENT` in the comparison below since it's not used anywhere else (and it's constexpr)

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+        T v;
+        memcpy(&v, &bits, sizeof(T));
+        __stwt(out + i, v);
+      }
+    } else {
+      for (uint32_t i = threadIdx.x; i < fill_rows; i += blockDim.x) {
+        uint64_t bits = ld64(raw_p + i * 8u);
+        T v;
+        memcpy(&v, &bits, sizeof(T));
+        __stwt(out + i, v);
```
kevkrist: It's not clear to me why __stwt() provides a performance win here. Why not cache-streaming? Then you can skip the synchronization barrier at line 442 for example.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  if (exc_count > 0u) {
+    // Sync only on the exception path: every thread must finish its main-loop
+    // store before any thread overwrites a row with its exception value.
+    // `sh_exc_count` is broadcast-shared, so all threads take the same branch.
+    __syncthreads();
+    uint8_t const* exc_base   = packed_bytes + sh_packed_bytes;
+    uint8_t const* exc_p_base = exc_base + uint32_t{exc_count} * sizeof(T);
+    for (uint32_t e = threadIdx.x; e < exc_count; e += blockDim.x) {
+      uint16_t pos;
+      memcpy(&pos, exc_p_base + e * 2u, sizeof(pos));
```
kevkrist: 2u -> sizeof(u16_t)

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+      if (pos < fill_rows) {
+        T v;
+        if constexpr (sizeof(T) == 4) {
+          uint32_t bits = ld32(exc_base + e * 4u);
```
kevkrist: 4u -> sizeof(float_t)
8u -> sizeof(double_t)

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+// (≤ 16 bytes) plus per-vector parsed state. EXACT_T is the unsigned
+// bit-mirror of T.
+template <typename T>
+__global__ void kernel_decode_alprd(alp_vector_desc const* __restrict__ descs,
+                                    T* __restrict__ d_output,
+                                    uint32_t num_vecs)
+{
+  static_assert(sizeof(T) == 4 || sizeof(T) == 8,
+                "ALPRD only supports float / double; the dispatcher rejects other widths");
+  using EXACT_T = typename std::conditional<sizeof(T) == 4, uint32_t, uint64_t>::type;
```
kevkrist:
```suggestion
  using EXACT_T = ::cuda::std::conditional_t<sizeof(T) == 4, uint32_t, uint64_t>;
```

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+// ALP / ALPRD decode kernels. One CTA = one 1024-row vector. On any bounds
+// check failure the kernel zero-fills `desc.vec_row_count` rows using the
+// trusted host descriptor row count, never parsed metadata.
+//
+// On-disk layouts (verified against
+// duckdb/src/include/duckdb/storage/compression/{alp,alprd}/):
+//
+//   ALP segment (one segment holds N 1024-row "vectors"):
+//
+//   byte 0   4                                 metadata_end          seg_end
```
kevkrist: Hmm metadata_end looks to be misplaced here. It should be where seg_end is now I believe, no?

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+//   0   1   2     4               12  13              ↑       ↑
+//                                                     │       │
+//                                bitstream = round_up(rows,32)*bw/8 bytes
+//
+//   exp == 0xFF → uncompressed-sentinel: 1-byte 0xFF then raw T[rows].
+//   bw ≤ sizeof(T)*8. Exception position bound-checked against fill_rows.
+//
+//   ALPRD segment (small dict header lives between metadata_end and trailer):
+//
+//   byte 0   4   5   6   7                                 metadata_end ↘
```
kevkrist: It looks like this diagram is missing the metadata trailer that contains the offsets into the per-vector data.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+      uint32_t metadata_end = ld32(seg_base);
+      uint64_t trailer_need = uint64_t{ALPRD_HEADER_SIZE} + (uint64_t{desc.vec_idx} + 1u) * 4u;
+      if (metadata_end > seg_bytes || metadata_end < trailer_need) {
+        sh_invalid = true;
+      } else {
+        uint8_t right_bw  = seg_base[4];
+        uint8_t left_bw   = seg_base[5];
+        uint8_t dict_size = seg_base[6];
+        // right_bw + left_bw fits in storage; left_bw ≤ 3 ⇒ dict_size ≤ 8.
+        if (right_bw > sizeof(T) * 8u || left_bw > 3u || dict_size > ALPRD_MAX_DICT_SIZE) {
```
kevkrist: Use
::cuda::std::numeric_limits<T>::digits()
2u -> duckdb::AlpRDConstants::MAX_DICTIONARY_BIT_WIDTH

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+          sh_invalid = true;
+        } else {
+          uint64_t header_end = uint64_t{ALPRD_HEADER_SIZE} + uint64_t{dict_size} * 2u;
```
kevkrist: 2u -> duckdb::AlpRDConstants::DICTIONARY_ELEMENT_SIZE

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+          if (header_end > metadata_end) {
+            sh_invalid = true;
+          } else {
+            sh_right_bw  = right_bw;
+            sh_left_bw   = left_bw;
+            sh_dict_size = dict_size;
+
+            uint32_t entry_off = metadata_end - (desc.vec_idx + 1u) * 4u;
```
kevkrist: Here and elsewhere:
4u -> duckdb::AlpRDConstants::METADATA_POINTER_SIZE
2u -> duckdb::AlpRDConstants::EXCEPTIONS_COUNT_SIZE

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+                  sh_invalid = true;
+                } else {
+                  sh_uncompressed = true;
+                }
+              } else {
+                sh_exc_count         = exc_count;
+                uint32_t left_bytes  = bp_required_bytes(fill_rows, left_bw);
+                uint32_t right_bytes = bp_required_bytes(fill_rows, right_bw);
+                // payload = 2 + left_bytes + right_bytes + exc_count*(2+2)
+                uint64_t payload_end =
```
kevkrist: 2u -> duckdb::AlpRDConstants::EXCEPTIONS_COUNT_SIZE
exc_count * (duckdb::AlpRDConstants::EXCEPTION_SIZE + duckdb::AlpRDConstants::EXCEPTION_POSITION_SIZE)

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+    sh_uncompressed = false;
+    sh_data_off     = 0;
+    sh_exc_count    = 0;
+    sh_right_bw     = 0;
+    sh_left_bw      = 0;
+    sh_dict_size    = 0;
+    sh_left_bytes   = 0;
+    sh_right_bytes  = 0;
+
+    if (seg_bytes < ALPRD_HEADER_SIZE) {
```
kevkrist: Please consider refactoring the metadata parsing into a separate function as mentioned above.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  // Cooperatively zero-init the full 8-entry dict before reading entries
```
kevkrist: Drop everything after the first sentence in this comment (clearly AI). Also, thread 0's parse does not run concurrently.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  // [0, sh_dict_size). With left_bw ≤ 3 (validated above), left_idx ∈ [0,8),
+  // so the inner-loop `sh_dict[left_idx]` read is unconditionally safe — a
+  // corrupt left_idx ≥ sh_dict_size lands on a zero slot, no bounds check
+  // needed in the hot path. Thread 0's parse runs concurrently.
+  if (threadIdx.x < ALPRD_MAX_DICT_SIZE) { sh_dict[threadIdx.x] = 0; }
+  __syncthreads();
+  if (!sh_invalid && threadIdx.x < uint32_t{sh_dict_size}) {
+    uint16_t dv;
+    memcpy(&dv, seg_base + ALPRD_HEADER_SIZE + threadIdx.x * 2u, sizeof(dv));
```
kevkrist:
```suggestion
    memcpy(&dv, seg_base + ALPRD_HEADER_SIZE + threadIdx.x * duckdb::AlpRDConstants::DICTIONARY_ELEMENT_SIZE, duckdb::AlpRDConstants::DICTIONARY_ELEMENT_SIZE);
```

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  }
+  __syncthreads();
+
+  if (sh_invalid) {
+    zero_fill_vector<T>(out, fill_rows);
+    return;
+  }
+
+  if (sh_uncompressed) {
+    uint8_t const* raw_p = seg_base + sh_data_off + 2u;
```
kevkrist: Again, this 2u is the size of the exception count. Please use the duckdb static constexpr named constant

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  __shared__ uint16_t sh_exc_count;
+  __shared__ uint8_t sh_right_bw;
+  __shared__ uint8_t sh_left_bw;
+  __shared__ uint8_t sh_dict_size;
+  __shared__ uint16_t sh_dict[ALPRD_MAX_DICT_SIZE];
+  __shared__ uint32_t sh_left_bytes;
+  __shared__ uint32_t sh_right_bytes;
+
+  if (threadIdx.x == 0) {
+    sh_invalid      = false;
```
kevkrist: In all of your codec modules you've used a different prefix for shared memory variables. sh_, sm_, s_. Please pick one and stick to it.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  if (sh_invalid) {
+    zero_fill_vector<T>(out, fill_rows);
+    return;
+  }
+
+  if (sh_uncompressed) {
+    uint8_t const* raw_p = seg_base + sh_data_off + 2u;
+    if constexpr (sizeof(T) == 4) {
+      for (uint32_t i = threadIdx.x; i < fill_rows; i += blockDim.x) {
+        uint32_t bits = ld32(raw_p + i * 4u);
```
kevkrist: 4u -> sizeof(T)

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+        T v;
+        memcpy(&v, &bits, sizeof(T));
+        __stwt(out + i, v);
+      }
+    } else {
+      for (uint32_t i = threadIdx.x; i < fill_rows; i += blockDim.x) {
+        uint64_t bits = ld64(raw_p + i * 8u);
```
kevkrist: 8u -> sizeof(T)

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+    }
+    return;
+  }
+
+  // Stage left and right packed streams into shmem. Each region carries 2
+  // trailing guard words for the 64-bit straddle case.
+  uint8_t const* left_bp  = seg_base + sh_data_off + 2u;
+  uint8_t const* right_bp = left_bp + sh_left_bytes;
+
+  extern __shared__ uint32_t shmem[];
```
kevkrist: Since this is static constexpr in the kernel invocation, why not make this a statically sized array rather than extern?

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  uint32_t left_live    = ::cuda::ceil_div(sh_left_bytes, 4u);
```
kevkrist: 4u -> sizeof(uint32_t)

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  uint32_t* left_words  = shmem;
+  uint32_t* right_words = shmem + left_live + 2u;
```
kevkrist: Please document what this 2u is (pad words for unpacking)

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+
+  stage_bytes_to_shmem(left_words, left_bp, sh_left_bytes, /*pad_words=*/2u);
+  stage_bytes_to_shmem(right_words, right_bp, sh_right_bytes, /*pad_words=*/2u);
+  __syncthreads();
+
+  uint8_t const right_bw = sh_right_bw;
+  uint8_t const left_bw  = sh_left_bw;
+  // right_bw == sizeof(T)*8 ⇒ shifting EXACT_T by that width is UB; gate it.
+  bool const full_right = (right_bw >= sizeof(T) * 8u);
```
kevkrist: Use ::cuda::std::numeric_limits<T>::digits()

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+    if (full_right) {
+      bits = right_val;
+    } else {
+      uint16_t left_idx = unpack_from_shmem<uint16_t>(left_words, i, left_bw);
+      EXACT_T left_val  = static_cast<EXACT_T>(sh_dict[left_idx]);
+      bits              = (left_val << right_bw) | right_val;
+    }
+    T v;
+    memcpy(&v, &bits, sizeof(T));
+    __stwt(out + i, v);
```
kevkrist, line 695: Again, not sure the purpose of __stwt there. There doesn't seem to be anything that you need to keep hot in cache, which would explain the need for __stwt here.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  // Exception scatter recovers right-bits via `read_back_after_stwt` —
+  // see helper for the L1-invalidate contract.
+  uint16_t const exc_count = sh_exc_count;
+  if (exc_count > 0u) {
+    // Sync only on the exception path: read_back_after_stwt(out, pos) reads
+    // what other threads wrote in the main loop, so the barrier is required
+    // here. `sh_exc_count` is broadcast-shared — uniform branch is safe.
+    __syncthreads();
+    uint8_t const* exc_base   = right_bp + sh_right_bytes;
+    uint8_t const* exc_p_base = exc_base + uint32_t{exc_count} * 2u;
```
kevkrist: As above, for 2u here and below, please use the duckdb::AlpRDConstants:: named value.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+    T v;
+    memcpy(&v, &bits, sizeof(T));
+    __stwt(out + i, v);
+  }
+
+  // Exception scatter recovers right-bits via `read_back_after_stwt` —
+  // see helper for the L1-invalidate contract.
+  uint16_t const exc_count = sh_exc_count;
+  if (exc_count > 0u) {
+    // Sync only on the exception path: read_back_after_stwt(out, pos) reads
```
kevkrist: Yeah, I don't understand this. This is an artifact of bypassing the cache, so why bypass the cache on stores at all?

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+      __stwt(out + pos, v);
+    }
+  }
+}
+
+std::vector<alp_vector_desc> build_vector_descs(gpu_codec_run const& run)
+{
+  std::vector<alp_vector_desc> descs;
+  // Small over-reservation; tight worst case is ceil(row_count / 1024) per segment.
+  descs.reserve(run.segments.size() * 2);
```
kevkrist: Can you explain to me the *2 here?

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+                               stream.value()));
+
+  // Live shmem footprint per CTA: bitpacked stream, max width sizeof(T)*8,
+  // ALP_VECTOR_SIZE values. Two guard words past the live data for the
+  // 64-bit straddle case.
+  constexpr uint32_t max_width        = sizeof(T) * 8u;
+  constexpr uint32_t max_packed_bytes = ::cuda::ceil_div(ALP_VECTOR_SIZE * max_width, 8u);
+  constexpr uint32_t max_packed_words = ::cuda::ceil_div(max_packed_bytes, 4u) + 2u;
+  constexpr size_t shmem_bytes        = max_packed_words * sizeof(uint32_t);
+  static_assert(shmem_bytes <= 48u * 1024u,
```
kevkrist: This is an architecture-specific constraint that shouldn't be in production code.

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  RMM_CUDA_TRY(cudaMemcpyAsync(d_descs.data(),
+                               h_descs,
+                               num_vecs * sizeof(alp_vector_desc),
+                               cudaMemcpyHostToDevice,
+                               stream.value()));
+
+  // Live shmem footprint per CTA: bitpacked stream, max width sizeof(T)*8,
+  // ALP_VECTOR_SIZE values. Two guard words past the live data for the
+  // 64-bit straddle case.
+  constexpr uint32_t max_width        = sizeof(T) * 8u;
```
kevkrist: It seems to me that the shared memory bytes can be a template value parameter to the kernel, and then you don't overallocate in the float32 case, and (as noted above) you don't need the extern declaration on the SMEM in the kernel. Alternatively you can constexpr calculate this in the kernel just from T.

kevkrist: Use ::cuda::std::numeric_limits<T>::digits()

src/cuda/scan/gpu_decode_alp.cu
```diff
@@ -0,0 +1,809 @@
...
+  rmm::device_uvector<alp_vector_desc> d_descs(num_vecs, stream, mr);
+  RMM_CUDA_TRY(cudaMemcpyAsync(d_descs.data(),
+                               h_descs,
+                               num_vecs * sizeof(alp_vector_desc),
+                               cudaMemcpyHostToDevice,
+                               stream.value()));
+
+  // Live shmem: left stream (left_bw ≤ 3) + right stream (right_bw ≤
+  // sizeof(T)*8). Guard pair after each region.
+  constexpr uint32_t max_left_bytes  = ::cuda::ceil_div(ALP_VECTOR_SIZE * 3u, 8u);
```
kevkrist: Please use the named constants in the duckdb::AlpRdConstants namespace here. Regarding SMEM, see my comment above for launch_alp_typed().

kevkrist (changes requested):

src/cuda/scan/gpu_decode_alp.cu
kevkrist: Replaced with 'code'

src/cuda/scan/gpu_decode_alp.cu
kevkrist: Let's just leave it for now. Tracking as a follow-up task with issue #781.

src/cuda/scan/gpu_decode_alp.cu
kevkrist: Deferring into #781

src/cuda/scan/gpu_decode_alp.cu
kevkrist: Deferring to the refactor in #781

src/cuda/scan/gpu_decode_alp.cu
kevkrist: Switched to `s_` prefix

src/cuda/scan/gpu_decode_alp.cu
kevkrist: Removing in preference for a comment at the appropriate __syncthreads() / load site.

src/cuda/scan/gpu_decode_alp.cu
kevkrist: Leaving as-is

src/cuda/scan/gpu_decode_alp.cu
kevkrist: Leaving as-is

src/cuda/scan/gpu_decode_alp.cu
kevkrist: Misunderstanding.

src/cuda/scan/gpu_decode_alp.cu
kevkrist: Deleted this and did a quick calculation of the total number of vectors.

kevkrist (approved):

PR #774, opened 2026-05-11
feat(pin_table): support tier='host' for host-tier caching

aminaramoon:
Pin parquet tables into pinned host memory and serve scans from there, slicing the host chunks per-query and converting back to GPU only when a scan task starts.

- pin_table: convert each cudf::table to cucascade::host_data_representation via the GPU↔HOST converter; uses an explicit rmm::cuda_stream because cudaMemcpyBatchAsync rejects the default stream.
- scan_manager: pinned_entry now tracks tier and host_chunks; create_provider_for picks the host-mode cached_split_provider when matched.
- cached_split_provider: new host constructor; start() slices each chunk by the requested column indices and emits one host-resident data_batch per chunk.
- scan_cached_operator_data: prepare_for_processing converts host-resident batches to gpu_table_representation using the task's reserved GPU memory space; get_estimated_size_in_bytes uses the generic representation accessor.
- Integration test: pin_table host tier scan and aggregate over lineitem.

Cucascade-submodule fallout (constructor went private, allocation became shared_ptr): switched to host_table_allocation::create() in cpu_source_task, duckdb_scan_task, and the host_table_utils test; converted host_table_chunk_reader to shared_ptr; templatized multiple_blocks_allocation_accessor methods on the smart-pointer type so unique_ptr and shared_ptr callers both work.

### DIFF
--- a/src/include/scan_manager/cached_split_provider.hpp
+++ b/src/include/scan_manager/cached_split_provider.hpp
@@ -21,9 +21,11 @@
 
 #include <cudf/column/column.hpp>
 
+#include <cucascade/data/cpu_data_representation.hpp>
 #include <cucascade/memory/memory_space.hpp>
 #include <duckdb/planner/expression.hpp>
 
+#include <cstddef>
 #include <memory>
 #include <vector>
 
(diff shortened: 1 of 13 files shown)


src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -160,6 +147,54 @@ std::unique_ptr<split_provider> sirius_scan_manager::create_provider_for(
         }
       }
 
+      if (entry.tier == cucascade::memory::Tier::HOST) {
+        // Map each D-position to its index inside the captured host chunk. column_names
+        // is in capture order, so we look up the requested data column by name. A missing
+        // column means the user pinned a subset that doesn't cover this scan — fall back
+        // to the parquet path so the query still succeeds.
```
dhruv9vats, line 154: > fall back to the parquet path so the query still succeeds

we are currently throwing if the lookup fails?

src/include/scan_manager/cached_split_provider.hpp
```diff
@@ -53,15 +55,33 @@ namespace sirius::scan_manager {
...
+    std::shared_ptr<duckdb::Expression> filter_expression,
+    std::shared_ptr<op::scan::scan_plan const> plan);
+
   std::future<void> start(exec::thread_pool& pool, split_connector& connector) override;
 
  private:
+  // Populated when constructed via the GPU constructor; empty in HOST mode.
   std::vector<std::vector<std::shared_ptr<cudf::column>>> _columns_per_request;
+  // Populated when constructed via the HOST constructor; empty in GPU mode.
+  std::vector<std::shared_ptr<cucascade::host_data_representation>> _host_chunks;
```
dhruv9vats, line 83: nit: potential for std::variant use.

dhruv9vats (approved):

felipeblazing (approved):

src/sirius_extension.cpp
```diff
@@ -813,11 +829,43 @@ void SiriusExtension::PinTableFunction(ClientContext& context,
...
+    // is released — the pinned bytes live in pinned host memory.
+    auto& registry = sirius::converter_registry::get();
+    // The GPU↔HOST converter uses cudaMemcpyBatchAsync which requires a real,
+    // non-default stream. Use a per-call rmm::cuda_stream so we don't accidentally
+    // serialize against other concurrent work on the default stream.
+    rmm::cuda_stream pin_stream;
+    auto stream_view = pin_stream.view();
+    std::vector<std::shared_ptr<cucascade::host_data_representation>> host_chunks;
+    host_chunks.reserve(tables.size());
+    for (auto& table : tables) {
```
bwyogatama: Does this mean that we actually have to bring the whole tables to GPU first, before moving it to CPU?

src/scan_manager/sirius_scan_manager.cpp
aminaramoon: we catch the exception and fallback to reading from file.

bwyogatama (approved):

PR #776, opened 2026-05-11
Feature: Per operator memory estimate for when we don't have historical info

wmalpica:
This PR introduces the ability for operators to define custom memory estimate for when there is no history available. The estimate for a pipeline would then be the max of every operator in its pipeline.

The default estimate is still what it was before, which 2X the input.

This feature is introduced, particularly to address the fact that Parquet scans consume way more than 2X the input. Its typically observed to be  8-10X the uncompressed size of the read. 
This PR only introduces custom estimates for Parquet Scans, Concat and Partition operators. Others can be added as needed.

### DIFF
--- a/src/include/op/sirius_physical_operator.hpp
+++ b/src/include/op/sirius_physical_operator.hpp
@@ -220,6 +220,17 @@ class partitioned_operator_data : public pipelineable_operator_data {
   std::size_t _partition_idx = 0;
 };
 
+/**
+ * @brief Input statistics passed to no_history_peak_memory_estimate().
+ *
+ * Carries the two dimensions that operator overrides typically condition on:
+ * the number of input batches and the total uncompressed byte count.
+ */
+struct input_stats {
+  std::size_t num_batches = 0;
+  std::size_t bytes       = 0;
+};
+
 //! sirius_physical_operator is the base class of the physical operators present in the
 //! execution plan
 class sirius_physical_operator {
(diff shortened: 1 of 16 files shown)


src/include/op/sirius_physical_operator.hpp
```diff
@@ -278,6 +289,28 @@ class sirius_physical_operator {
...
+   *
+   * A return value of 0 means "no additional peak memory expected"; the caller will fall
+   * back to the 2× default when every operator in the pipeline returns 0.
+   *
+   * @param stats  Batch count and total input bytes for the task about to run.
+   * @return Estimated peak GPU bytes this operator will allocate.
+   */
+  [[nodiscard]] virtual std::size_t no_history_peak_memory_estimate(const input_stats& stats) const
+  {
+    return stats.bytes * 2;
```
aminaramoon, line 311: do you want to max(2Xbytes, some_reasobale_number (100MB))

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
```diff
@@ -272,4 +272,10 @@ std::unique_ptr<operator_data> sirius_gpu_parquet_scan_operator::execute(
   return std::make_unique<pipelineable_operator_data>(std::move(batches));
 }
 
+std::size_t sirius_gpu_parquet_scan_operator::no_history_peak_memory_estimate(
+  const op::input_stats& stats) const
+{
+  return stats.bytes * 8;
```
aminaramoon, line 278: same comment

aminaramoon (approved):
left a comment,

src/include/op/sirius_physical_operator.hpp
wmalpica: I dont see the need. If its small, then let it be small. If its an underestimate but its still small, then its not a problem, we should have some overhead available. If we underestimated and it turned out to be that we needed a large amount of memory, then a small minimum reasonable value wont help us either.

PR #779, opened 2026-05-13
Use tightened `data_batch`

dhruv9vats:
Migrate data_batch construction, locking, and accessor usage to cuCascade PR #120.

Adjust cached read-only accessors and pass-through paths for move-only lock handles


bwyogatama:
Using #985 instead

PR #784, opened 2026-05-17
s3datasource backend  and scan_manager integration (replace PR759)

ran-yuan-rui:
# Summary
This PR adds an S3 datasource backend to Sirius and wires it into `sirius_scan_manager`. It **supersedes PR #759** — the same work, rebased directly onto `dev`, with the PR #759 review feedback addressed.

**Scope — please read before reviewing.** This PR delivers the S3 datasource at the **runtime / C++ level only**:

- ✅ `sirius_scan_manager` routes `s3://` paths to the S3 backend and reads parquet from S3, with prefetch caching and HTTP retry — verified end-to-end against MinIO.
- ⚠️ **End-to-end SQL is not connected yet.** `gpu_execution('SELECT … read_parquet("s3://…")')` does not yet route to the S3 backend. The SQL surface needs a bind-time parquet-footer fetch, and the design of that footer handling still needs alignment with @aminaramoon  — that is a **follow-up PR**.
- ⚠️ **The S3 integration tests are not yet in the default CI unit-test run.** They run as a dedicated, MinIO-gated suite (`make s3-up` + the `[s3]` tag), invoked explicitly. Folding them into the main CI suite is deferred to the follow-up PR.

So: this PR = "the S3 runtime, C++-verified and ready"; the follow-up PR = "SQL reaches it end-to-end, plus AWS coverage and CI integration".

### Threads
- **No per-request thread creation in production.** Async reads dispatch onto caller-injected thread pool.`sirius_scan_manager` owns that pool (default 8 threads)
- **Bounded concurrency.** ≤ 8 worker threads; the libcurl easy-handle pool is capped (default 16 handles) and grown on demand. In-flight HTTP requests are thread-bounded.
- **The detached-thread path is test-only.** With no injected pool (standalone / unit-test scenarios), async reads fall back to a per-request detached `std::thread`. Production always injects the pool, so the engine never hits this path.

### Memory
- **The S3 backend holds almost no resident memory of its own.** Host range reads stream libcurl output directly into the caller-provided host buffer; the S3 backend keeps no resident object buffer.
- **Device reads use bounded host staging.** S3 has no device-native path, so a device read bounces through a host staging buffer before the H2D copy. That buffer is a single block borrowed from the scan manager's host memory resource (`fixed_size_host_memory_resource`) and reused chunk-by-chunk, so a device read's transient host memory is capped at one FSMR block regardless of the read size. (A standalone build with no injected FSMR keeps a plain `std::vector` fallback — test-only; production always injects the FSMR.)
- **The curl handle pool** is a small fixed overhead; each handle keeps libcurl's connection cache so TCP / TLS connections are reused across requests.

### Retry
- **Bounded attempts** — 3 total attempts (2 retries) per request.
- **Retriable conditions** — HTTP 408 / 429 / 5xx and transient libcurl codes (connect / timeout class). Everything else — other 4xx, signature / not-found / auth failures — fails fast with no retry.
- **Backoff** — exponential with jitter; the server's `Retry-After` header (429 / 503) is honored when present, capped at 30 s.
- **Worst-case stuck time per request is bounded** — each attempt is capped by a curl
  timeout (default 60 s); total ≈ timeout × attempts + backoff.

## Non-S3 code changes

Most of the diff is the new `src/io/s3/` module and its tests. The changes to existing engine code are **additive and behavior-preserving for non-S3 paths**: `sirius_scan_manager` now holds a vector of backends and dispatches per path, and `SiriusContext` composes the S3 config from the YAML object-store settings — the local-file path is unchanged. The one behavior change touching a non-S3 path makes a scan_manager-routed local parquet scan fall back to cudf (instead of throwing) when no Sirius backend is configured, restoring the documented behavior under `dev`'s current default.

## Testing

- **Local MinIO only, so far.** S3 integration tests run against a pinned MinIO container: `make s3-up` → `sirius_unittest "[s3]"` → `make s3-down` — 76 test cases, 0 failures.
- An optional perf benchmark (`make s3-bench`) is included; it is hidden from the default est run and reachable only via that target.

## TODO / follow-ups

- **Follow-up PR — SQL surface.** Make `gpu_execution('… read_parquet("s3://…")')` route to the S3 backend end-to-end; 
- **AWS testing.** S3 integration is currently verified only against local MinIO. Once the
  SQL surface lands and the path is end-to-end, add real-AWS testing.

## Closes

Closes PR #759.


aminaramoon (approved):
Please wait for this PR to land before merging 

https://github.com/sirius-db/sirius/pull/732

thanks,

ran-yuan-rui:
**Update — merged latest dev with multi GPU support **

- **Merged `dev` (#732 multi-GPU)** into this branch. Resolved the scan-side ioctx collision (dev keys by GPU/NUMA, this branch keyed by backend) into two-dimensional model: S3 shared by backend, local sharded by target GPU.

- **IO-backend ownership unification** : `SiriusContext` now owns the scan-side IO backends (`s3_ioctx` + S3 async pool + prefetch `buffer_pool`/cache, alongside the per-NUMA urings); `scan_manager` borrows them via `(config, vector<shared_ptr<sirius_ioctx>>)` and constructs none. Prefetch cache is initialized on the **real read-path** backends (per-NUMAurings + s3_ioctx) → revives the local cache (was dead on a vestigial uring).`terminate()` reordered so `buffer_pool` outlives every cache (no cache→pool UAF; relies on `~s3_ioctx`'s `_cache.reset(); shutdown()`)  

- **Validation**: clean-room (`rm -rf build` + fresh `make release`) + full
  `sirius_unittest` with MinIO/S3 e2e, strict — **1,397/1,397, 49.8M assertions,
  all pass**; pre-commit clean.

- **Still v1 / out of scope**: NUMA-aware S3 (multi-GPU → single shared S3, );

PR #787, opened 2026-05-19
scan: introduce duckdb_db_io_object + duckdb_block_layout helper (DuckDB format GPU-native Part 7)

ducndh:
## Overview
Identity-only foundation for the duckdb-native scan path. Two new headers+ one .cpp + tests.

- `duckdb_db_io_object` (`src/io/duckdb_db_io_object.{hpp,cpp}`): a `sirius_io_object` subclass that identifies a DuckDB `.db` file by its absolute path and reports the on-disk size captured at construction via `stat(2)`. Provides the (cache_id, size) handle the prefetching cache (#675) needs to dedupe inserts and bound reads.

- `duckdb_block_payload_offset` (`src/include/op/scan/duckdb_block_layout.hpp`): header-only helper wrapping the arithmetic from `SingleFileBlockManager::GetBlockLocation`:
    `BLOCK_START = FILE_HEADER_SIZE * 3 = 12288`
    `payload_offset = BLOCK_START + block_id * (block_size + header_size) + header_size`
    
Matches what `BufferManager::Pin(handle).Ptr()` returns. Lets callers serve `.db` block bytes from the IO substrate without a `BlockManager::RegisterBlock` round-trip.

Tests cover stat-derived size, empty file, empty-path rejection, and missing-file failure.

## Test plan
- `build/release/extension/sirius/test/cpp/sirius_unittest "[io][duckdb_db_io_object]"`

## Notes:

- This PR has zero behavioral impact on its own — both pieces are consumed only by later PRs in the duckdb-native scan stack.
- The class deletes copy + move because the inherited `enable_shared_from_this` doesn't play nicely with moves based on my opinion and this would force future callers to go through `std::make_shared<duckdb_db_io_object>(...)`.


src/include/io/duckdb_db_io_object.hpp
```diff
@@ -0,0 +1,76 @@
+/*
```
kevkrist: I understand why you used 'duckdb_db_...' but why not just duckdb_io_object? duckdb_db_io_object is a little clunky. And the duckdb_block_layout omits the 'db' part.

src/include/op/scan/duckdb_block_layout.hpp
```diff
@@ -0,0 +1,68 @@
...
+// skipping the per-block header. Reading raw bytes via sirius_ioctx::host_read
+// at the payload offset gives the same view.
+
+/// Precondition: @p block_id must be a real block id (>= 0). DuckDB uses
+/// `INVALID_BLOCK = -1` as a sentinel on `block_id_t` (a signed `int64_t`);
+/// callers must filter those out (typical pattern: CONSTANT segments hold
+/// their value in stats and never reference a block). Passing a negative
+/// id throws @c std::invalid_argument rather than underflowing to a huge
+/// offset and returning silent garbage.
+inline std::size_t duckdb_block_payload_offset(duckdb::BlockManager const& bm,
```
kevkrist: Can you document what exactly the returned value represents? I think it's clear from the code (offset to the payload bytes for that block [after the block header]), but it's better to be explicit.

src/io/duckdb_db_io_object.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include "io/duckdb_db_io_object.hpp"
+
+#include <filesystem>
+#include <stdexcept>
+#include <utility>
+
+namespace sirius::io {
+
+duckdb_db_io_object::duckdb_db_io_object(std::string path) : _absolute_path(), _size_bytes(0)
```
kevkrist:
```suggestion
duckdb_db_io_object::duckdb_db_io_object(std::string const& path) : _absolute_path(), _size_bytes(0)
```

src/io/duckdb_db_io_object.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "io/duckdb_db_io_object.hpp"
+
+#include <filesystem>
+#include <stdexcept>
+#include <utility>
```
kevkrist: Unused header.

src/include/op/scan/duckdb_block_layout.hpp
kevkrist: What about sibling subclasses of duckdb::BlockManager? They would compile but probably break stuff.
Consider either accepting SingleFileBlockManager as parameter or dynamic_casting and throwing if the wrong sibling type is passed.

src/include/io/duckdb_db_io_object.hpp
```diff
@@ -0,0 +1,76 @@
...
+ * The constructor canonicalizes the input path via @c std::filesystem::canonical
+ * so two callers passing the same file under different spellings (relative vs
+ * absolute, with intermediate symlinks) land on the same cache key.
+ *
+ * `size()` is captured at construction. A live DuckDB session may grow the
+ * file as new blocks are allocated; the read-only OLAP scan path freezes the
+ * size at scan-start, which is correct for that use case but means write-
+ * aware callers should construct fresh objects when re-reading.
+ */
+class duckdb_db_io_object : public sirius_io_object {
```
kevkrist: Mark class as `final`

src/include/op/scan/duckdb_block_layout.hpp
```diff
@@ -0,0 +1,68 @@
...
+//   then blocks start at BLOCK_START = FILE_HEADER_SIZE * 3 = 12288.
+//   For each block_id ∈ [0, num_blocks):
+//     [block_start, block_start + block_header_size)        — checksum
+//     [block_start + block_header_size, block_start + alloc) — payload
+//     (BlockManager::GetBlockSize() bytes)
+//   where:
+//     alloc        = bm.GetBlockSize() + bm.GetBlockHeaderSize()
+//     block_start  = BLOCK_START + block_id * alloc
+//
+// Matches what BufferManager::Pin(handle).Ptr() returns: the payload pointer,
```
kevkrist: It might be worth adding a comment on why we avoid BufferManager::Pin() -- to bypass DuckDB's CPU cache, I assume? -- in favor of our own API.

src/include/op/scan/duckdb_block_layout.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <duckdb/storage/storage_info.hpp>
+
+#include <cstddef>
+#include <cstdint>
+#include <stdexcept>
+#include <string>
+
+namespace sirius::op::scan {
+
+// DuckDB .db file layout (mirrors SingleFileBlockManager::GetBlockLocation,
```
kevkrist: Do you think it's a good idea to comment that we can't use this DuckDB API because it's private? It might be good to keep such a comment around to track what we might ask the DuckDB team to make public. 
With this API public we can avoid any issues with file format drift and needing to synchronize this function with DuckDB version whenever there is a format change. Up to you.

src/include/op/scan/duckdb_block_layout.hpp
```diff
@@ -0,0 +1,68 @@
...
+/// id throws @c std::invalid_argument rather than underflowing to a huge
+/// offset and returning silent garbage.
+inline std::size_t duckdb_block_payload_offset(duckdb::BlockManager const& bm,
+                                               duckdb::block_id_t block_id)
+{
+  if (block_id < 0) {
+    throw std::invalid_argument("duckdb_block_payload_offset: block_id must be >= 0 (got " +
+                                std::to_string(block_id) + ")");
+  }
+  constexpr std::size_t BLOCK_START =
```
kevkrist: Consider making static constexpr at namespace scope.

kevkrist (approved):

src/include/io/duckdb_db_io_object.hpp
ducndh: yah if duckdb uses another file handle it might be helpful but db is kinda implicit so I will just do duckdb_io_object

src/include/op/scan/duckdb_block_layout.hpp
ducndh: aight

src/include/op/scan/duckdb_block_layout.hpp
ducndh: good catch

src/include/op/scan/duckdb_block_layout.hpp
ducndh: good catch tighten this to enforce better contract

PR #789, opened 2026-05-20
[DO NOT MERGE] CI: Update GPU test workflow to use multi-GPU nodes

mike-wendt:
Testing switching from `gpu-t4` to `gpu-4xt4` runners for #732


mike-wendt:
Changes in test workflow work for existing SNSG tests. We should be good to go for SNMG introduced in #732 

Closing this PR

Issue #604, opened 2026-04-09
Phase 3: Create `sirius_plan_builder` for bind-time plan construction


mbrobbel:
Part of #601.

New `sirius_plan_builder::build()` that orchestrates: iceberg metadata prefetch → result collector wrapping → meta-pipeline build → pipeline conversion. Called from `GPUExecutionBind()` after physical plan generation. `sirius_prepared_statement_data` stores the complete `plan_result`. `initialize_internal()` reduced to runtime materialization only.

Created by Claude

Issue #545, opened 2026-04-02
Tracking issue: decouple Sirius internal from DuckDB


mbrobbel:
This is a tracking issue for decoupling Sirius internals from DuckDB. The goal is mainly to reduce the extension glue to a minimum, making it easier to update to newer DuckDB versions.

PR #790, opened 2026-05-20
Construct GPU pipelines at plan time (#604)

bwyogatama:
## Summary

**Move every GPU operator into the Physical Plan tree so pipeline conversion becomes a pure topology pass.**

### Before

```
DuckDB Logical Plan
  → Physical Plan (no GPU operators)
  → Pipelines (no GPU operators)
  → Pipelines (with GPU operators)    ← inserted at runtime by the converter
```

### After (this PR)

```
DuckDB Logical Plan
  → Physical Plan (with GPU operators)    ← inserted by the plan generator
  → Pipelines (with GPU operators)
```

GPU operators (PARTITION, CONCAT, the SORT chain, MERGE operators, scan companions, CPU_SOURCE) used to be created *after* pipeline construction, so they never appeared in the Physical Operator Tree. That blocked:

1. **`EXPLAIN`** — `EXPLAIN` prints the Physical Operator Tree, so GPU operators were invisible. Unblocks `gpu_explain` (#529).
2. **`op->parent` / `op->children`** — GPU operators had no tree linkage, so any code that walks the tree couldn't reach them. Unblocks transparent execution (#518) and DuckDB decoupling (#545).

This PR moves all GPU operator construction into `sirius_physical_plan_generator` and rewrites the pipeline build to read directly from the tree via per-operator `build_pipelines` overrides. The converter is reduced to tree-parent `compute_repository_wiring` with no operator mutations.

**Status:** `USE_TREE_BASED_PIPELINE_BUILD` is now **ON by default**. Differential dump gates compare the legacy and tree-based paths byte-for-byte on every TPC-H SF1 query (DuckDB + parquet scan sources); **22/22 queries pass byte-identical** on both gates. Legacy code paths still coexist behind the flag for one-PR revert safety — deletion is split into follow-ups (see below).

## Design

Follows DuckDB's `PhysicalPlanGenerator::Plan` → frozen plan → `PhysicalOperator::BuildPipelines` model:

- **`_parent_op`** is a base-class field on `sirius_physical_operator` populated by a recursive `set_parent_ops` post-pass at the end of `Plan()`. Plan-tree mutation uses a `wrap_child` helper with move-semantics on `children[]` — no raw pointers held across mutations.
- **`build_pipelines` produces post-finalize state directly.** `is_ready` absorbs the finalize work (reverse + set source/sink). `create_child_meta_pipeline` pre-populates new pipeline's operators with `[sink_op]`. Source-leaves use `add_pipeline_operator`.
- **Custom overrides** for PARTITION (promote tree child to deeper child_meta's sink when child has its own children), HJ/NLJ/RIGHT_DELIM_JOIN (assert `build_child.is_sink()`; build_meta sink is CONCAT), and the four sink-leaf scans (DUCKDB_SCAN, ICEBERG_SCAN, GPU_PARQUET_SCAN, CPU_SOURCE).
- **Repository wiring** switches from source-pointer matching to tree-parent lookup. The 10-case dispatch in `compute_repository_wiring` collapses into one uniform pattern plus `resolve_port_id` / `resolve_barrier` helpers.

## What's NOT in this PR (follow-ups)

Legacy code is still in the tree behind the flag for revert safety. Deletion is split across follow-ups:

- **Engine cleanup (-~240 LOC):** delete `prefetch_iceberg_delete_data`, `iceberg_delete_data_cache_`, `construct_sirius_specific_operator` engine method, `construct_iceberg_scan_operator`, `resolve_iceberg_table_path`, `new_pipeline_breakers`. Plumb the planner's iceberg cache through to the converter.
- **Converter cleanup (-~800 LOC):** delete all `split_*` helpers (incl. `split_delim_join_sink`), the free-function `construct_sirius_specific_operator`, `inserted_operators_`, `iceberg_cache_`, legacy `compute_repository_wiring` branch, `finalize_pipeline_structure`'s operators-push half. Rename `compute_repository_wiring_tree_based` → `compute_repository_wiring`.
- **Focused unit tests:** `test_plan_tree_shape.cpp` (tree-rewrite invariants) + `test_partition_build_pipelines.cpp` (PARTITION's custom override).
- **Remove the flag (-~50 LOC):** delete `Config::USE_TREE_BASED_PIPELINE_BUILD`, the SQL option, the planner's flag check, and the differential test. Single-path final state.
- **DISTINCT_MERGE / PARTITION_DISTINCT chain inside DELIM JOIN** — open gap. Resolution: move HGB ownership from `delim_join.distinct` (unique_ptr) into `delim_join.children[N]` and make `distinct` a non-owning raw pointer.

## Test plan

- [x] `pre-commit run -a` clean on every touched file
- [x] `pixi run -e default make` clean on every commit
- [x] `sirius_unittest` green at every commit
- [x] Differential dump gate (DuckDB scans): 22/22 TPC-H SF1 queries byte-identical, `kKnownFailing` empty
- [x] Differential dump gate (parquet scans): all queries byte-identical
- [x] TPC-H integration regression (`[TPC-H]` tag): 36320/36320 assertions, 49 test cases
- [x] TPC-H SF10 22/22 vs DuckDB CPU baseline

## Known unrelated flakes

Two GPU-state flakes that pre-date this PR's work and pass on retry:

1. `gpu_execution - order by lineitem with long varchar column` (test ~703) — SIGSEGVs in `transparent_execution_test_utils.hpp`. Reproduces in isolation against `dev`.
2. `gpu_execution iceberg - V2 positional deletes order by desc` (test ~424) — SIGSEGVs in the same Catch2 reporter macro. Distinct test from #1.

Both are GPU-state dependent (~hundreds of preceding tests pollute), pass on retry, and warrant separate cleanup issues.


wmalpica (approved):

mbrobbel (approved):

src/pipeline/sirius_pipeline_converter.cpp
```diff
@@ -1115,6 +1144,269 @@ void sirius_pipeline_converter::compute_repository_wiring()
...
+    }
+
+    // RIGHT_DELIM_JOIN: emit partition_join + distinct sibling references.
+    if (sink_op->type == T::RIGHT_DELIM_JOIN) {
+      auto& right_delim    = sink_op->Cast<op::sirius_physical_right_delim_join>();
+      auto* partition_join = right_delim.partition_join;
+      auto* distinct_op    = right_delim.distinct;
+      if (partition_join) {
+        auto it = dest_for_op.find(partition_join);
+        // B.3+B.4+B.6 (#604): partition_join is owned by the DELIM_JOIN and executed
```
dhruv9vats:
```suggestion
        // partition_join is owned by the DELIM_JOIN and executed
```
we should get rid of LLM bread crumbs

mbrobbel (approved):
Let's get this in

PR #791, opened 2026-05-20
scan: duckdb_native_split_provider (DuckDB format GPU-native Part 8b)

ducndh:
Stacks on #749 (polymorphic scan_info). Implements the duckdb-native `split_provider`: a metadata-only walk over `.duckdb` storage that emits one split per row-group batch to the scan_manager.

- `op::scan::duckdb_native_scan_info` — bind-data subclass of the polymorphic `scan_info`. Carries `DataTable*`, `ClientContext*`, projected columns + types, and the source `db_path`. Its `make_provider(io_ctx)` override forwards the ioctx to the provider.

- `scan_manager::duckdb_native_split_provider` — runs the metadata walker in its ctor; throws `std::runtime_error` on non-viable metadata so the extension's transparent fallback routes the query to DuckDB CPU (empty splits would hang the pipeline on the FULL barrier). Partitions row groups into batches honoring `approximate_batch_size` and a per-column varchar byte cap. Parks the `sirius_ioctx` as `_io_ctx`; the dedicated scan task in a later PR uses it to route `.db` block reads through `sirius_ioctx::host_read`. A single oversized row group is preserved as its own singleton batch.

- Unit tests cover the null-arg / size-mismatch invariants, the HUGEINT-rejection path, single / multi / singleton-batch partitioning, and the shared `scan_info` invariant across batches.

### Per-segment `max_string_length`

DuckDB tracks `MaxStringLength` at two granularities: per-storage-segment (in `ColumnSegmentInfo::segment_stats`, a text blob) and per-row-group (in `PartitionRowGroup::GetColumnStatistics`, computed as `max` over segments via `StringStats::Merge`). The walker parses the per-segment value from the blob and stores it on each `duckdb_segment_descriptor`:

- Every VARCHAR storage path in DuckDB calls `UpdateStringStats` — `Uncompressed`, `Dictionary`, `FSST`, `DICT_FSST`, `Zstd` — so the stat is always written. The walker refuses any row group where a varchar segment's blob lacks the field, surfacing as a CPU fallback.
- `Some(0)` is the legal all-empty-row-group case (e.g. clickbench `SocialAction`, `ParamOrderID`, `Title` row groups with every value `""`); the dict decoder produces a zero-byte chars region and the partitioner contributes 0 bytes for it. No 256-byte fallback.
- A DEBUG-only assert verifies `max(per-segment) == typed_API_rg_max` for every (rg, col) so any drift in DuckDB's `StringStats::Merge` semantics fires loudly in test builds.

### Per-column varchar byte cap

`VARCHAR_BYTE_CAP = INT32_MAX`. cudf strings columns use int32 chars-offsets, so a single column's decoded bytes per scan output must stay within int32 range. The bound enforced per row group is `Σ_segments seg.segment_count × seg.max_string_length` — exact, since the walker stores per-segment values. 

## Test plan
- `build/release/extension/sirius/test/cpp/sirius_unittest "[scan][duckdb_native_walker]"`
- `build/release/extension/sirius/test/cpp/sirius_unittest "[scan][duckdb_native_split_provider]"`

### DIFF
--- a/src/include/op/scan/duckdb_native_scan_info.hpp
+++ b/src/include/op/scan/duckdb_native_scan_info.hpp
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2026, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "helper/logical_type.hpp"
+#include "op/scan/duckdb_native_metadata.hpp"
+
+#include <duckdb/main/client_context.hpp>
+#include <duckdb/storage/data_table.hpp>
+
+#include <cstddef>
+#include <string>
+#include <vector>
+
+namespace sirius::op::scan {
+
+struct duckdb_native_scan_info {
+  duckdb::DataTable* storage = nullptr;
+  duckdb::ClientContext* context = nullptr;
+
+  std::vector<projected_column> projected_cols;
+  std::vector<sirius::logical_type> projected_types;
+
+  std::string db_path;
+
+  std::size_t approximate_batch_size = 0;
+};
+
+}  // namespace sirius::op::scan
--- a/src/include/op/scan/duckdb_native_scan_info.hpp
+++ b/src/include/op/scan/duckdb_native_scan_info.hpp
@@ -39,6 +39,10 @@ class split_provider;
 
 namespace sirius::op::scan {
 
+/// DuckDB-native bind-data. Single-use: `make_provider()` moves the whole
+/// object into the split_provider's owned scan_info, leaving `storage` /
+/// `context` null on the source. A second call would trip the non-null
+/// preconditions in `duckdb_native_split_provider`'s ctor.
 struct duckdb_native_scan_info : public scan_info {
   duckdb::DataTable* storage     = nullptr;
   duckdb::ClientContext* context = nullptr;
@@ -48,6 +52,7 @@ struct duckdb_native_scan_info : public scan_info {
 
   std::string db_path;
 
+  /// Single call only — moves *this into the returned split_provider.
   std::unique_ptr<scan_manager::split_provider> make_provider(
     std::unordered_map<int, std::shared_ptr<sirius::io::sirius_ioctx>> const& gpu_ioctxs) override;
 };
--- a/src/include/op/scan/duckdb_native_scan_info.hpp
+++ b/src/include/op/scan/duckdb_native_scan_info.hpp
@@ -18,18 +18,29 @@
 
 #include "helper/logical_type.hpp"
 #include "op/scan/duckdb_native_metadata.hpp"
+#include "op/scan/scan_info.hpp"
 
 #include <duckdb/main/client_context.hpp>
 #include <duckdb/storage/data_table.hpp>
 
 #include <cstddef>
+#include <memory>
 #include <string>
+#include <unordered_map>
 #include <vector>
+
+namespace sirius::io {
+class sirius_ioctx;
+}  // namespace sirius::io
+
+namespace sirius::scan_manager {
+class split_provider;
+}  // namespace sirius::scan_manager
 
 namespace sirius::op::scan {
 
-struct duckdb_native_scan_info {
-  duckdb::DataTable* storage = nullptr;
+struct duckdb_native_scan_info : public scan_info {
+  duckdb::DataTable* storage     = nullptr;
   duckdb::ClientContext* context = nullptr;
 
   std::vector<projected_column> projected_cols;
@@ -37,7 +48,8 @@ struct duckdb_native_scan_info {
 
   std::string db_path;
 
-  std::size_t approximate_batch_size = 0;
+  std::unique_ptr<scan_manager::split_provider> make_provider(
+    std::unordered_map<int, std::shared_ptr<sirius::io::sirius_ioctx>> const& gpu_ioctxs) override;
 };
 
 }  // namespace sirius::op::scan
--- a/src/include/scan_manager/duckdb_native_split_provider.hpp
+++ b/src/include/scan_manager/duckdb_native_split_provider.hpp
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2026, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "op/scan/duckdb_native_metadata.hpp"
+#include "op/scan/duckdb_native_scan_info.hpp"
+#include "op/sirius_physical_operator.hpp"
+#include "scan_manager/split_provider.hpp"
+
+#include <atomic>
+#include <cstddef>
+#include <functional>
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace sirius::scan_manager {
+
+class duckdb_native_split_provider : public split_provider {
+ public:
+  struct split_payload : public op::operator_data {
+    std::vector<op::scan::duckdb_row_group_metadata> row_groups;
+    std::shared_ptr<op::scan::duckdb_native_scan_info const> scan_info;
+  };
+
+  explicit duckdb_native_split_provider(op::scan::duckdb_native_scan_info info);
+
+  ~duckdb_native_split_provider() override;
+
+  duckdb_native_split_provider(const duckdb_native_split_provider&)            = delete;
+  duckdb_native_split_provider& operator=(const duckdb_native_split_provider&) = delete;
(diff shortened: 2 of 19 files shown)


src/scan_manager/duckdb_native_split_provider.cpp
```diff
@@ -0,0 +1,187 @@
...
+    const std::size_t this_rg_bytes = rg.decoded_bytes_budget;
+    if (any_varchar) {
+      std::fill(this_rg_col_bytes.begin(), this_rg_col_bytes.end(), 0);
+      for (std::size_t c = 0; c < num_cols; ++c) {
+        if (is_varchar[c]) this_rg_col_bytes[c] = rg_varchar_bytes_for_col(rg, c);
+      }
+    }
+
+    // Only close a non-empty batch; an over-cap row group always lands in
+    // the current batch as its first member (singleton if it's alone).
```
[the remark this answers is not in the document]
ducndh: Very rare but sure add a guard to avoid this (you genuinely need a 17Kb string per row to throw this)

src/scan_manager/duckdb_native_split_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include <stdexcept>
+#include <utility>
+
+namespace sirius::scan_manager {
+
+namespace {
+
+// cudf strings columns use int32 chars-offsets; one column's bytes per
+// scan output must stay <= INT32_MAX.
+constexpr std::size_t VARCHAR_BYTE_CAP =
```
kevkrist: Please use `cudf::size_type` instead of `int32_t`.

src/scan_manager/duckdb_native_split_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "scan_manager/duckdb_native_split_provider.hpp"
+
+#include "log/logging.hpp"
```
kevkrist, line 19: Seems to be unused

src/scan_manager/duckdb_native_split_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include <cstdint>
+#include <limits>
+#include <stdexcept>
+#include <utility>
+
+namespace sirius::scan_manager {
+
+namespace {
+
+// cudf strings columns use int32 chars-offsets; one column's bytes per
```
kevkrist: I don't think this is true. The offsets column can use int64_t. It's just that the number of rows is capped by int32_t max. See the legacy sirius code. But I think this is fine.

src/scan_manager/duckdb_native_split_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+  // No caps in play → single batch.
+  if (approximate_batch_size == 0 && !any_varchar) {
+    batches.push_back({0, row_groups.size()});
+    return batches;
+  }
+
+  std::size_t batch_first = 0;
+  std::size_t batch_bytes = 0;
+  std::vector<std::size_t> col_bytes(num_cols, 0);
+  // Hoisted scratch — reset per iteration to skip the allocator on tables
```
kevkrist: Please get rid of this AI comment.

src/scan_manager/duckdb_native_split_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+{
+  std::vector<duckdb_native_split_provider::row_group_batch> batches;
+  if (row_groups.empty()) return batches;
+
+  const std::size_t num_cols = projected_types.size();
+  std::vector<bool> is_varchar(num_cols, false);
+  for (std::size_t c = 0; c < num_cols; ++c) {
+    is_varchar[c] = projected_types[c].is_varchar();
+  }
+  const bool any_varchar =
```
kevkrist: Why don't you just set any_varchar in the loop above?

src/include/scan_manager/duckdb_native_split_provider.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include "op/scan/duckdb_native_metadata.hpp"
+#include "op/scan/duckdb_native_scan_info.hpp"
+#include "op/sirius_physical_operator.hpp"
+#include "scan_manager/split_provider.hpp"
+
+#include <atomic>
+#include <cstddef>
+#include <functional>
+#include <memory>
+#include <string>
```
kevkrist: Unused header

kevkrist: Move header into source file.

src/op/scan/duckdb_native_metadata.cpp
kevkrist, line 458: This can get marked true even if there are no row groups. Consider marking false when there are no row groups (may never happen?) to avoid a hang in the scheduler (no tasks scheduled -> mark_task_completed() never called -> etc. -- I believe you fixed this for another case already).

src/include/op/scan/duckdb_native_scan_info.hpp
```diff
@@ -0,0 +1,54 @@
...
+struct duckdb_native_scan_info : public scan_info {
+  duckdb::DataTable* storage     = nullptr;
+  duckdb::ClientContext* context = nullptr;
+
+  std::vector<projected_column> projected_cols;
+  std::vector<sirius::logical_type> projected_types;
+
+  std::string db_path;
+
+  std::unique_ptr<scan_manager::split_provider> make_provider(
```
kevkrist, line 56: I think it's worth adding some documentation to this file.
For example, make_provider moves the entire object (default move will move all fields) which the superclass acknowledges is called once, but still worth clarifying here I think that the single call is a hard contract -- a second call will throw in the split provider constructor (that scan_info.storage must be non-null).

src/scan_manager/duckdb_native_split_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+  std::size_t idx = _next_batch_idx.fetch_add(1, std::memory_order_relaxed);
+  if (idx >= _batches.size()) { return {}; }
+
+  row_group_batch batch = _batches[idx];
+  return [this, batch]() -> std::vector<std::unique_ptr<op::operator_data>> {
+    auto payload       = std::make_unique<split_payload>();
+    payload->scan_info = _scan_info;
+    payload->row_groups.reserve(batch.count);
+    for (std::size_t i = batch.first_idx; i < batch.first_idx + batch.count; ++i) {
+      payload->row_groups.push_back(_metadata.row_groups[i]);
```
kevkrist: Move these row groups rather than deep-copy?

src/scan_manager/duckdb_native_split_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+    // Empty splits would hang the pipeline on the FULL barrier. The throw
+    // surfaces as a query error and the extension's transparent fallback
+    // routes to DuckDB CPU.
+    SPDLOG_DEBUG("[duckdb_native_split_provider] non-viable: {}",
+                 _metadata.viability_failure_reason);
+    throw std::runtime_error("duckdb-native scan rejected query: " +
+                             _metadata.viability_failure_reason);
+  }
+
+  _batches = partition_row_groups_into_batches(
```
kevkrist, line 134: There is some divergence here with the parquet provider framework, which is fine I think, but worth documenting explictly. For parquet, a batch is a set of files, which is a unit of work for a scan manager worker thread, which parses the metadata and provides splits, which are units of work for scan operator. It seems that here batch = split, right? All the work is consolidated in the constructor of the split provider parsing the metadata and partitioning the batches, executed by the query executor thread, and then the scan manager threads don't do any work other than atomically claim batches from the predefined list of batches in the split provider. This makes sense since the metadata is in memory and file I/O doesn't need to be spread out, it's just that there is an asymmetry in who is doing the work. For a single file in parquet, a scan manager thread does all the work, for duckdb native, the query executor thread does all the work in constructing the split provider. To be sure, everything is in memory and fast, so probably not a big performance issue, just worth mentioning, and there's something to be said for structural alignment across formats. Or am I missing something here?

src/scan_manager/duckdb_native_split_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+  }
+  if (_scan_info->context == nullptr) {
+    throw std::invalid_argument("duckdb_native_split_provider: scan_info.context must be non-null");
+  }
+  if (_scan_info->projected_cols.size() != _scan_info->projected_types.size()) {
+    throw std::invalid_argument(
+      "duckdb_native_split_provider: projected_cols and projected_types must be parallel");
+  }
+
+  _metadata = op::scan::walk_duckdb_native_metadata(*_scan_info->storage,
```
kevkrist, line 120: It seems to me that the metadata walking and the row group partitioning can be fused so that there is a single loop over all row groups no? There seems to be some duplication, e.g., in computing the varchar bytes for a segment. I understand that these can be thought of as logically separate steps, so not fusing makes sense, but it feels tighter to me. Leave it up to you. Could alternatively have different loops, but have second loop read per-column chars stored in the row_group_metadata by the walker. Just some thoughts.

src/op/scan/duckdb_native_metadata.cpp
kevkrist: Why not throw rather than produce wrong results? Or set viable=false?

kevkrist (approved):

kevkrist (approved):

src/op/scan/duckdb_native_metadata.cpp
ducndh: Great catch was not thinking about this one while fixing the other one at that point

src/scan_manager/duckdb_native_split_provider.cpp
ducndh: Great catch. Did not think too hard when reading this the frist time but the cost is not samll

src/scan_manager/duckdb_native_split_provider.cpp
ducndh: It is a bit messier than that but I would just do the safetest thing

src/scan_manager/duckdb_native_split_provider.cpp
ducndh: Add a comment for this

src/scan_manager/duckdb_native_split_provider.cpp
ducndh: Yah this was created during 2 different passes so I was sloppy with it. The second loop should do something less bad now

PR #792, opened 2026-05-20
scan: GPU-native DuckDB scan operator (DuckDB format GPU-native Part 10)

ducndh:
Stacks on PR 8b (duckdb_native_split_provider). Adds the source operator that decodes a duckdb-native split into a cudf::table, plus the seq_scan→GPU pipeline-converter hook and the opt-in SET flag.

- `sirius_gpu_duckdb_native_scan_operator` — physical operator, sibling to the parquet scan operator. Stores the scan_info, takes splits from the scan_manager via a split_connector, and dispatches each split to `decode_duckdb_native_split` on its CUDA stream.

- `duckdb_native_scan_task.cpp` — split decoder. Coverage in this PR: fixed-width data codecs UNCOMPRESSED / RLE / BITPACKING, rowid synthesis via `cudf::sequence`, UNCOMPRESSED / EMPTY / CONSTANT validity. Variable-width codecs (VARCHAR variants), CONSTANT data segments (block_id=-1 value-in-stats), ROARING validity, and additional_blocks staging throw at the dispatch boundary — those are the next PR.

Per-segment host bytes come from `BlockManager::RegisterBlock` + `BufferManager::Pin`; each segment is `cudaMemcpyAsync`'d into one device staging buffer per split.

- `sirius_pipeline_converter::insert_duckdb_native_scan_operator` — detects `seq_scan` on a `DuckTableEntry` and inserts the GPU operator in place of the CPU fallback, gated on `op_params.enable_gpu_duckdb_native_scan`.

- `scan_manager::create_provider_for(duckdb_native_op*)` — cache-probes via `try_cached_for` and otherwise dispatches through `scan_info::make_provider()`. Same shape as the parquet overload.

- `enable_gpu_duckdb_native_scan` SET/YAML hook — opt-in flag, off by default. Mirrors existing operator_params plumbing.

## Things we can fold into this but kinda stylistic choice for third-paty

1. **`const_cast<memory_space*>` in `duckdb_native_scan_task.cpp:pick_gpu_memory_space_for_duckdb_native_scan`**

2. **Parallel registration storage in `sirius_scan_manager`** — `_duckdb_native_providers_by_op` + `_duckdb_native_scan_op_order` mirror the parquet pair, with parallel branches in `prepare_for_query` and `start_metadata_processing`. `scan_info::make_provider` already covers format-specific dispatch; collapsing the storage to a single base-typed map would require a virtual `take_scan_info()` on the operator base class. 

## Test plan
- `build/release/extension/sirius/test/cpp/sirius_unittest "[scan]"`
- SET `enable_gpu_duckdb_native_scan=true` and run TPC-H SF=10 against a `.duckdb` fixture; q1-q22 should match with no GPU-side fallback on any fixed-width-data + simple-validity column.

## Test plan
- `build/release/extension/sirius/test/cpp/sirius_unittest "[scan]"`
- SET `enable_gpu_duckdb_native_scan=true` and run TPC-H SF=10 against a `.duckdb` fixture; q1-q22 should match with no GPU-side fallback on any fixed-width-data + simple-validity column.


src/include/op/scan/scan_info.hpp
```diff
@@ -67,7 +67,7 @@ struct scan_info {
   std::size_t approximate_batch_size = sirius::config::DEFAULT_SCAN_TASK_BATCH_SIZE;
   /// Number of output columns the gpu scan operator returns.
   /// Set by the pipeline converter from the scan operator's `types.size()`.
-  std::size_t scan_output_arity;
+  std::size_t scan_output_arity = 0;
```
kevkrist, line 70: Why was this default arity re-added?

src/op/scan/duckdb_native_decoder.cpp
```diff
@@ -0,0 +1,878 @@
...
+    auto c = v.compression;
+    if (c == duckdb::CompressionType::COMPRESSION_UNCOMPRESSED ||
+        c == duckdb::CompressionType::COMPRESSION_ROARING) {
+      return true;
+    }
+  }
+  return false;
+}
+
+cudf::data_type sirius_to_cudf_type(sirius::logical_type const& t)
```
kevkrist, line 128: For this and below, why do you use the vector methods? You could here, e.g., call `to_duckdb` and then invoke `GetCudfType()` without the vector indirection.

src/op/scan/sirius_gpu_duckdb_native_scan_operator.cpp
```diff
@@ -0,0 +1,166 @@
...
+  auto batch = sirius::make_data_batch(std::move(table), *mem_space, stream);
+  std::vector<std::shared_ptr<cucascade::data_batch>> batches;
+  batches.push_back(std::move(batch));
+  return std::make_unique<pipelineable_operator_data>(std::move(batches));
+}
+
+std::size_t sirius_gpu_duckdb_native_scan_operator::no_history_peak_memory_estimate(
+  const op::input_stats& stats) const
+{
+  return stats.bytes * 4;
```
kevkrist, line 163: Is there a reason for 4? I know the parquet scan operator uses 8 without explanation. I suppose I'm curious why the asymmetry of 4 instead of 8?

src/include/op/scan/duckdb_native_scan_info.hpp
```diff
@@ -47,15 +49,19 @@ struct duckdb_native_scan_info : public scan_info {
   duckdb::DataTable* storage     = nullptr;
   duckdb::ClientContext* context = nullptr;
 
+  // projected_cols / projected_types are sized to the FULL scan output (includes filter-only
+  // columns appended by sirius_plan_get.cpp). Filter application + projection down to
+  // output_types happens inside decode_duckdb_native_split.
   std::vector<projected_column> projected_cols;
```
kevkrist, line 55: Is there a way to make this work with `scan_plan`? It would be nice to have centralized logic for pruning filter-only columns, etc. I see that there is an explanation for why scan_plan isn't used in sirius_gpu_duckdb_native_scan_operator.cpp, but maybe there's a way to make scan_plan work in both places? Worth thinking about.

src/op/scan/duckdb_native_decoder.cpp
```diff
@@ -0,0 +1,878 @@
+/*
```
kevkrist, line 1: I think this file name is likely to cause confusion with `parquet_scan_task` and the old scan task infrastructure. Maybe duckdb_native_decoder.cpp? Up to you, but it confused me at first glance.

src/op/scan/duckdb_native_scan_task.cpp
```diff
@@ -0,0 +1,878 @@
...
+  handles.push_back(std::move(main_handle));
+  std::vector<uint8_t> concat;
+  concat.resize(seg.bytes_size);
+  std::memcpy(concat.data(), handles.front().Ptr() + seg.block_offset, seg.bytes_size);
+
+  for (auto add_id : seg.additional_blocks) {
+    auto add_handle = block_manager.RegisterBlock(add_id);
+    auto h          = buffer_manager.Pin(add_handle);
+    auto offset     = concat.size();
+    concat.resize(offset + block_size);
```
kevkrist: Is there a way to do this without resizing concat() mid-loop? Probably not a big performance issue with only a handful of additional_blocks, but the pattern at read_block_via_io and elsewhere do a single resize.

kevkrist (commented):
Need to hand off to someone else

src/include/memory/numa_small_pinned_mr.hpp
```diff
@@ -111,20 +111,20 @@ class numa_small_pinned_mr {
 
   void* allocate_sync(std::size_t bytes, std::size_t alignment = alignof(std::max_align_t))
   {
-    return allocate(cuda::stream_ref{cudaStream_t{nullptr}}, bytes, alignment);
+    return allocate(::cuda::stream_ref{cudaStream_t{nullptr}}, bytes, alignment);
```
aminaramoon: this is allocating on legacyDefault stream, legacy default strem has some weird propertoies like it waits for all other streams to finish before doing it work, it is blocking.
I think it is better to use ptds and syncrhnoize it here

aminaramoon (approved):
just left a question that needs asnwering

src/op/scan/duckdb_native_decoder.cpp
ducndh: I port a fully loaded version that worked since age ago so this might be remnant from that time

src/op/scan/sirius_gpu_duckdb_native_scan_operator.cpp
ducndh: Duckdb advertise light weight compression so I just divide 8 by 2 but feel free to bump it to 8.

src/include/op/scan/duckdb_native_scan_info.hpp
ducndh: My reading is that scan_plan needs a bit of refacter to accept the duckdb_native plan as it is very centralized around parquet style reading right now. Do you want me to do a follow up PR to address this refactor? @kevkrist

PR #794, opened 2026-05-20
feat: dynamic table-filter pushdown (Phase 1)

### LINKED ISSUE #645: Support dynamic filters in scan
These are produced for joins and can be a row-group pruning opportunity for large probe scans.

### LINKED ISSUE #795: Dynamic table filters stripped by plan->Copy() in transparent execution mode
## Symptom

In `gpu_execution` (transparent execution) mode, **every `LogicalComparisonJoin::filter_pushdown` and `LogicalGet::dynamic_filters` is null** by the time Sirius's planner walks the captured logical plan. 

## Root cause

The transparent optimizer hook captures the optimized logical plan via:

https://github.com/sirius-db/sirius/blob/dev/src/transparent/sirius_optimizer_extension.cpp#L82

```cpp
auto plan_copy = plan->Copy(context);
ctx->set_captured_logical_plan(std::move(plan_copy));
```

`LogicalOperator::Copy` is implemented as a serialize → deserialize round-trip (DuckDB `src/planner/logical_operator.cpp:231-250`):

```cpp
unique_ptr<LogicalOperator> LogicalOperator::Copy(ClientContext &context) const {
    MemoryStream stream(Allocator::Get(context));
    BinarySerializer serializer(stream, options);
    serializer.Begin(); this->Serialize(serializer); serializer.End();
    stream.Rewind();
    return BinaryDeserializer::Deserialize<LogicalOperator>(stream, context, parameters);
}
```

So what survives Copy is exactly what each operator's `Serialize` writes.

**`LogicalComparisonJoin::Serialize`** (DuckDB `src/storage/serialization/serialize_logical_operator.cpp:326-337`) writes 9 properties (IDs 200-208): `join_type, mark_index, left_projection_map, right_projection_map, conditions, mark_types, duplicate_eliminated_columns, delim_flipped, predicate`. **`filter_pushdown` is not among them.**

**`LogicalGet::Serialize`** (DuckDB `src/planner/operator/logical_get.cpp:239-262`) writes 14 properties (IDs 200-214): `table_index, returned_types, names, projection_ids, table_filters, function/bind, parameters, named_parameters, input_table_types, input_table_names, projected_input, column_indexes, extra_info, ordinality_idx, row_group_order_options`. **`dynamic_filters` is not among them.**

Both fields are default-constructed (null) on the deserialized copy. 

## Why this matters

The `JoinFilterPushdownOptimizer` (DuckDB) *does* run in transparent mode (Sirius's pre-optimizer hook doesn't disable it), and `EXPLAIN ANALYZE` on TPC-H Q14 shows DuckDB generating `Dynamic Filters:` in the physical plan. The optimizer attaches `filter_pushdown` to `LogicalComparisonJoin` and `dynamic_filters` to `LogicalGet` correctly — they're just dropped on Sirius's `Copy()` step.

Phase 1.1 (#794) wires the entire producer → channel → consumer path, but it can never activate until this is fixed.

## Verification once fixed

After the fix, running a single iteration of TPC-H Q14 (or Q19) via `gpu_execution` on #794 should produce log lines like:

```
[sirius_plan_get] LogicalGet has dynamic_filters attached (channel key=0x...)
[sirius_plan_comparison_join] Wired hash join with 1 dynamic-filter probe target(s).
[sirius_physical_hash_join] Pushed 1 dynamic filter(s) across 1 target(s) (... build rows, 1 keys).
```

## Related

- #794 — Phase 1.1 (the wiring that's currently dormant).
- #664 — original dynamic-filter scaffolding.

kevkrist:
# Dynamic Table-Filter Pushdown (Dynamic Filters - Phase 1)

This PR introduces dynamic table filters for parquet scans. 

For multi-gpu, since the probe scan distribution cannot be known at filter construction time, the filter is replicated across ALL devices. Reservations are not made for zone-map device scalars but are for cuco `static_set`s and Bloom filters.

Filter selection is based on the following heuristic: if the cuco `static_set` fits in the device's L2 cache, the exact `static_set` is preferred. Otherwise, Bloom filter is selected. Zone-maps are turned on/off by configuration flag. They are only applied at scan-time and not in the separate post-scan operator (as are the membership filters `static_set` and Bloom filter). 

Filter application is adaptive: filter operators update operator-level statistics to record the selectivity of the filter, and a) reject filters that don't prune, and b) reorder filters (if there are more than one) so that the most selective filter is applied first. There is also pruning on the build side: a build that would keep most probe rows (an unfiltered build subtree, or a build whose key range covers most of the domain) is detected and never emits a filter.

Fixes #645 

### Configuration:

- Scale factor: `SF300`
- Machine: `GB200`
- Pinning: `--pinning-mode per-query`
- Zone maps: off in both runs
- Timing metric: median runtime across all 5 iterations for each query

### GPU-Tier Per-Query Median

| query | df off | df on | delta | gain |
|---|---:|---:|---:|---:|
| Q1 | 0.473s | 0.483s | +0.010s | +2.11% |
| Q2 | 0.151s | 0.091s | -0.060s | -39.74% |
| Q3 | 0.613s | 0.634s | +0.021s | +3.43% |
| Q4 | 0.232s | 0.242s | +0.010s | +4.31% |
| Q5 | 0.332s | 0.111s | -0.221s | -66.57% |
| Q6 | 0.150s | 0.150s | +0.000s | +0.00% |
| Q7 | 0.383s | 0.292s | -0.091s | -23.76% |
| Q8 | 0.745s | 0.755s | +0.010s | +1.34% |
| Q9 | 0.704s | 0.483s | -0.221s | -31.39% |
| Q10 | 0.463s | 0.373s | -0.090s | -19.44% |
| Q11 | 0.101s | 0.101s | +0.000s | +0.00% |
| Q12 | 0.232s | 0.251s | +0.019s | +8.19% |
| Q13 | 0.493s | 0.493s | +0.000s | +0.00% |
| Q14 | 0.121s | 0.131s | +0.010s | +8.26% |
| Q15 | 0.121s | 0.131s | +0.010s | +8.26% |
| Q16 | 0.211s | 0.191s | -0.020s | -9.48% |
| Q17 | 0.463s | 0.332s | -0.131s | -28.29% |
| Q18 | 0.574s | 0.413s | -0.161s | -28.05% |
| Q19 | 0.353s | 0.352s | -0.001s | -0.28% |
| Q20 | 0.161s | 0.161s | +0.000s | +0.00% |
| Q21 | 0.926s | 0.804s | -0.122s | -13.17% |
| Q22 | 0.151s | 0.121s | -0.030s | -19.87% |
| **Total** | **8.153s** | **7.095s** | **-1.058s** | **-12.98%** |

### HOST-Tier
| **Total** | **12.832s** | **11.373s** | **-1.459s** | **-11.37%** |

### HOST-Tier — 2 GPUs
| **Total** | **13.769s** | **10.846s** | **-2.923s** | **-21.23%** |

### Notes

Small per-query regressions are observable, but keep in mind that the per-run standard deviation is ~10ms (observe that Q1 'regresses' even though none of the dynamic filter infrastructure is activated). When a filter *is* wired, dynamic filters add some build-side work to reduce/build the filter, push it through the filter set, and check/apply it on the probe side; if a query has little or late pruning, that overhead can show itself (e.g., Q3: filters are wired but never applied because the probe scans drain before the filters publish). Q14/Q15 are different: no producer was wired and no filter was pushed, so those small deltas are best read as harness variance OR scan-side dynamic-filter metadata bookkeeping before the converter elides the producer-less channel. The new build-side task priority (only activated when dynamic filters are planned) can also shift GPU scheduling and thus alter performance.

In this run, useful dynamic-filter publish events appeared for Q2, Q5, Q7, Q8, Q9, Q10, Q16, Q17, Q18, Q20, and Q21. The largest wins came from Q5, Q9, Q17, Q18, Q2, Q7, Q10, and Q21.

## Limitations and Follow-ups

**Tracked on GitHub:** umbrella #1015. Items with a dedicated issue are tagged inline below; the remaining limitations are roadmap items captured in the umbrella.

 - [HIGHEST] Currently, only dynamic _table_ filters are supported. Joins higher up in the join graph (whose probe source operator is not `SCAN`) do not enjoy the benefits of dynamic filtering at the moment. (#1010)
 - Currently, only `BUILD_PROBE` joins are supported for simplicity. Dynamic table filters should be extended to `STANDARD` joins. The additional complexity there arises from partition-awareness and multi-batch partitions.
 - Only equality join keys produce filters. Inequality conditions (handled by the mixed-join predicate) get none. As long as the inequality conditions are ANDed with equality join conditions, dynamic filters tied to the equality condition can still be pushed.
 - Dynamic filters are only serving parquet scans right now.
 - The dynamic zone-map is applied only on the parquet path (the reader's `set_filter`); the pinned/cached path applies membership filters only, so a zone-map there does nothing.

Follow-ups:
 - **Small IN-list direct kernel.** (#1011) For a small IN-list (say `<= 10` values), skip the `cuco::static_set` build and probe with a kernel that just loops over the list values per probe key — cheaper than hashing for tiny sets.
 - **Parameter tuning.** (#1012) The gate's skip threshold, the L2-fit membership sizing, and the build-side domain-coverage cutoff are heuristic constants; they should be tuned empirically rather than guessed.
 - **Better build-side pruning.** (#1013) The publish decision is currently coarse (unfiltered subtree / domain coverage). Per-key cardinality or NDV estimates would let the build skip more low-value filters and keep more high-value ones.
 - **Bound build-task prioritization.** (#1014) Prioritizing every build subtree front-loads their hash tables. A cap on concurrently-prioritized subtrees — or prioritizing each build only relative to its own consumer, once that consumer is live — would bound the resident-memory peak on many-join or large-build plans.

## Components

- **Producer** — `sirius_physical_hash_join`. Per equality build key, computes a filter and publishes it.
- **Channel** — `sirius_dynamic_filter_set`. Thread-safe, append-only, keyed by *consumer* column index; co-owned by both ends.
- **Router** — `sirius_physical_plan_generator::get_or_create_dynamic_filter_channel`. At plan time, finds-or-creates one channel per producer/consumer pair and attaches both ends.
- **Consumer** — parquet / table scan. Reads the channel opportunistically and applies whatever is present.
- **Filter zoo** — `sirius_dynamic_zone_map_filter`, `sirius_dynamic_in_list_filter`, `sirius_dynamic_bloom_filter`, behind capability mixins (`sirius_ast_lowerable`, `sirius_mask_applicable`).

## How they fit together

```
PLAN TIME  (gated by enable_dynamic_filter_pushdown)
─────────────────────────────────────────────────────
  DuckDB pairs a join with a scan via a shared
  DynamicTableFilterSet*  ──────── route key ────────┐
                                                      ▼
            get_or_create_dynamic_filter_channel(key)  ⇒  one shared sirius_dynamic_filter_set
                 │                                                  │
       attach producer end                                attach consumer end
   probe_targets[{channel, probe_col_idx}]                  sirius_dynamic_filters
                 │                                                  │
                 ▼                push_filter(col, f)               ▼
        ┌──────────────────┐  ───────────────────────▶   ┌────────────────────┐
        │  HASH JOIN BUILD │      sirius_dynamic_         │  parquet / table   │
        │   (producer)     │  ◀───  filter_set  ───▶      │  SCAN (consumer)   │
        └──────────────────┘     filters_for_column(col)  └────────────────────┘


RUN TIME
─────────────────────────────────────────────────────
  build side completes
        │   push_data_batch_partitioned (build-port hook)
        │     • per equality key: reduce min/max  +  build IN-list / Bloom
        │     • drain the build stream            ← filters valid before they are visible
        ▼
   push_filter(consumer_col, filter)  ───▶  channel  (publish once, idempotent)
                                              │
                                              │  opportunistic, lock-free
  each scan split ──▶ has_filters()? ──▶ apply whatever is present at that moment
        (a filter that lands late simply prunes the splits read after it — never a correctness issue)
```

## Two consumer paths

A filter is applied by the path its capability supports; the two compose without coordination.

```
                       sirius_dynamic_filter (from the channel)
                                     │
            ┌────────────────────────┴─────────────────────────┐
   AST-lowerable (zone-map)                        mask-applicable (IN-list / Bloom)
            │                                                   │
   merge_dynamic_filters_into_ast                  sirius_physical_dynamic_filter operator
   → reader_options::set_filter                    → apply_dynamic_filters_to_view (post-decode)
            │                                          behind a selectivity gate
   cuDF prunes row groups by statistics                       │
   and evaluates it during decode                  drops rows the join would reject anyway
            │                                                   │
            └─────────────────── fewer rows ───────────────────┘
                        join re-checks every survivor ⇒ results unchanged
```
Note that zone maps are ONLY applied to the scan operator, whereas the mask-applicable filters are inserted as an independent post-scan operator.

## Configuration

| Setting | Default | Effect |
|---|---|---|
| `enable_dynamic_filter_pushdown` | `true` | Master switch — wires the membership (IN-list / Bloom) filters. |
| `enable_dynamic_zone_map_filter` | `false` | Additionally emit a zone-map for read-time row-group pruning (requires the master). |

## Where to read the code

| Concern | File |
|---|---|
| Filter base, channel, zone-map | `src/op/sirius_dynamic_filter.{hpp,cpp}` |
| Membership filters (GPU) | `src/cuda/sirius_dynamic_{in_list,bloom}_filter.cu` |
| Producer (build → publish) | `src/op/sirius_physical_hash_join.cpp` |
| Router & plan-gen wiring | `src/planner/sirius_physical_plan_generator.cpp`, `src/planner/sirius_plan_{get,comparison_join}.cpp` |
| Build-task prioritization | `src/pipeline/task_scheduler.cpp` |
| Consumer merge / apply / gate | `src/op/scan/dynamic_filter_merge.cpp`, `src/include/op/scan/dynamic_filter_gate.hpp` |
| Scan consumer integration | `src/op/scan/parquet_gpu_ingestible.cpp`, `src/op/scan/sirius_physical_dynamic_filter.cpp` |
| Design notes | `docs/super-sirius/dynamic-filters.md` |


src/include/op/sirius_dynamic_filter.hpp
```diff
@@ -167,6 +175,163 @@ class sirius_dynamic_zone_map_filter final : public sirius_dynamic_filter,
...
+ * exactly the rows the join would discard. It cannot be expressed as a cheap AST or row-group stats
+ * predicate, so it rides the @ref sirius_mask_applicable path: a row-level persistent-set probe
+ * applied post-decode. Intended for small-to-moderate key sets; larger builds use a bloom filter.
+ *
+ * @note A probe key equal to the key type's minimum (INT32_MIN / INT64_MIN) is treated as an
+ *       empty-slot sentinel by the underlying @c cuco::static_set and will not match, even if that
+ *       value was inserted into the build set. This is safe — the join remains authoritative, so a
+ *       missed singleton only costs a pruning opportunity, never a false negative — but it requires
+ *       that build keys never legitimately use that minimum (a precondition satisfied by
+ * TPC-H-style surrogate-key domains).
```
Jedi18: This seems wrong, it's not safe? It would be a false negative not a missed pruning opportunity.

Example: Say the join's build side has the key INT32_MIN (along with other keys), and the probe side has a row with that same key. Those two rows should match and appear in the output. But because INT32_MIN is the value cuco reserves to mark empty slots, the build key never actually lands in the set, so when the probe row's INT32_MIN is checked, the set says "not present," the mask comes back false, and the scan drops that row before it ever reaches the join. The matching row silently vanishes from the result.

Possible solution could be to keep track of whether we see this sentinel value while constructing the build side?

src/planner/sirius_plan_comparison_join.cpp
```diff
@@ -331,6 +388,46 @@ sirius_physical_plan_generator::plan_comparison_join(duckdb::LogicalComparisonJo
...
     hj.join_stats = std::move(op.join_stats);
 
+    // --- Wire dynamic-filter producer targets ---
+    // For each downstream scan DuckDB has paired with this join, look up the shared channel by
+    // the DynamicTableFilterSet pointer (the route key) and stash it along with the per-key
+    // consumer column indices on the join.
+    //
+    // Plan-time selectivity gate: an UNFILTERED build subtree is — for FK-shaped joins — the
+    // whole key domain, so its membership filter keeps every probe row by construction.
+    if (hj.filter_pushdown) {
```
Jedi18: We should maybe also gate this on `join_type == INNER`?

For LEFT/OUTER the probe side is preserved, every probe row must be emitted, NULL-padded if it doesn't match. So pruning non-matching probe rows here would silently drop rows that should be in the result. Not an issue today since DuckDB's optimizer refuses to emit filter_pushdown for LEFT/OUTER, but nothing enforces that on our side

src/op/sirius_physical_hash_join.cpp
```diff
@@ -1214,9 +1234,308 @@ std::unique_ptr<operator_data> sirius_physical_hash_join::execute(const operator
...
+      default: return std::nullopt;
+    }
+  };
+
+  for (std::size_t k = 0; k < filter_pushdown->join_condition.size(); ++k) {
+    auto const cond_idx = filter_pushdown->join_condition[k];
+    // Skip cast keys: the build-side scalar / membership set is built in the build column's type
+    // (col.type() below), but the consumer lowers it against the probe column's storage type;
+    // casting it across would need a separate code path.
+    if (cond_idx < key_casts.size() && key_casts[cond_idx].cast_right) {
```
Jedi18: Add a similar check for the cast_left case? We can hit issues if the probe key is cast too.

e.g. take `CAST(probe_bigint AS INTEGER) = build_int`, where build_int only contains the value 0:

* the zone-map is built from the build keys, so its range is [0, 0], and it gets pushed down to the raw BIGINT probe column
* a probe row with probe_bigint = 2³² casts to 0 — cuDF wraps on overflow instead of erroring — so the join matches it against build_int = 0
* but at the scan, the zone-map already tested the raw 2³² against [0, 0] and pruned it

So a row that should match is silently dropped. The membership filters dodge this (they self-skip when the probe type doesn't match the build keys); only the zone-map path is exposed.

src/include/op/sirius_dynamic_filter.hpp
kevkrist: You are right this documentation is bad and wrong, and I was throwing this issue under the rug. I want to avoid the runtime calculations, so here is what I'm going to do instead to add zero overhead:
 1) check the statistics on the column, and if the global min is greater than the sentinel, we're good
 2) if the global min trips the sentinel, just use Bloom, which doesn't need a key sentinel.

src/op/sirius_physical_hash_join.cpp
kevkrist: Good catch!

src/planner/sirius_plan_comparison_join.cpp
kevkrist: This is actually too restrictive. E.g., a RIGHT join or SEMI join should be allowed, too. We rely on duckdb at join_filter_pushdown_optimizer.cpp to decide whether to apply a dynamic table filter at all, so we count on duckdb to make the right decision for us. I added a comment to make this clear.

src/include/op/sirius_dynamic_filter.hpp
Jedi18: Thanks for the fix, though Claude flagged an issue with it. Seems like a genuine issue but I haven't done a deeper analysis into the code to confirm

```
⚠️ join_stats can be misaligned for multi-condition joins
The two indices used in stats_exclude_inlist_sentinel come from different points in the optimizer pipeline:

join_stats is populated at STATISTICS_PROPAGATION (duckdb/src/optimizer/optimizer.cpp:289), in the condition order at that time.
cond_idx comes from filter_pushdown->join_condition, built later at JOIN_FILTER_PUSHDOWN (duckdb/src/optimizer/optimizer.cpp:320), which first calls ReorderConditions (equal-conditions-first).
ReorderConditions (duckdb/src/execution/operator/join/physical_comparison_join.cpp:34-73) reorders conditions but does not touch join_stats, and nothing reorders it afterward. So for a mixed join like a.x < b.x AND a.y = b.y:

stats order: [less_L, less_R, eq_L, eq_R]
after the reorder, the equality key becomes cond_idx = 0, so the code reads join_stats[1] = less_R — the wrong condition's build stat.
The physical-width guard catches this when the two columns differ in width (→ returns false → Bloom, safe). But if both build keys are the same width (e.g. two BIGINTs), it passes the guard and we validate the wrong column's min — which can return sentinel_safe = true for a column that actually contains INT_MIN, reintroducing the original bug in that scenario.

Pure multi-equality joins are fine (no relative reorder), and single-condition joins are fine — so this is a narrow conjunction, but it defeats the fix in precisely the multi-condition case it's meant to cover.

Suggested fix: look the stat up by the build column index rather than by condition position, or assert that the two index spaces align.
```

src/planner/sirius_plan_comparison_join.cpp
Jedi18: Ok sounds good, thanks

src/include/op/sirius_dynamic_filter.hpp
kevkrist: Argh yeah. It looks like it's OK for now since only BUILD_PROBE joins are supported. But I also wonder about transaction-safety with this statistic. TO be safe I am adding an explicit check on the build keys. It introduces a synchronization though, to get the bool back to HOST, which I don't like. I will push this safe path and evaluate the impact on performance.

Jedi18: Ok great, hopefully it doesn't have much of a performance impact

src/op/sirius_physical_hash_join.cpp
```diff
@@ -1288,9 +1359,432 @@ std::unique_ptr<operator_data> sirius_physical_hash_join::execute(const operator
...
+      return 0;
+    }
+    minimum = std::min(minimum, static_cast<std::size_t>(l2));
+  }
+  return minimum == std::numeric_limits<std::size_t>::max() ? 0 : minimum;
+}
+
+/// Builds, replicates, and fans out one immutable filter snapshot from a complete join build batch.
+/// It borrows plan/key metadata; its caller owns source readiness and exactly-once arbitration.
+class dynamic_filter_publisher final {
```
wmalpica: dynamic_filter_publisher and dynamic_filter_publish_plan should go into their own file

src/op/sirius_physical_hash_join.cpp
```diff
@@ -1288,9 +1359,432 @@ std::unique_ptr<operator_data> sirius_physical_hash_join::execute(const operator
...
+        !_dynamic_filter_plan.enabled()) {
+      return;
+    }
+  }
+
+  // Leave the state OPEN when the delivered batch is not GPU-resident; execute may later claim it
+  // from the migrated/execution-ready GPU representation. This is defense-in-depth: normal
+  // build-side CONCAT delivery is GPU-resident and completes publication here before this join's
+  // immediate probe data scan is scheduled. A scan target below an intervening join can run sooner.
+  auto ro  = batch->to_read_only();
```
wmalpica: We probably want to get this read only lock before we call push_data_batch_partitioned, otherwise there is a potential race where the data might get downgraded as soon as its deposited into a data_repository, in which case we would not be able to get the read only lock

src/op/scan/dynamic_filter_merge.cpp
```diff
@@ -0,0 +1,246 @@
...
+  // If it grew beyond the snapshot that disabled the gate, the combined mask deserves one fresh
+  // measurement. An immediate Phase 1 probe normally starts after the complete fan-out.
+  return filters.filter_count() > _decided_filter_count.load(std::memory_order_relaxed);
+}
+
+void dynamic_filter_gate::record_keep_ratio(std::size_t rows_before,
+                                            std::size_t rows_after,
+                                            std::size_t observed_filter_count)
+{
+  constexpr double keep_threshold = 0.25;  // disable if a batch keeps > 25% of its rows
```
wmalpica: This seems incredibly low. This means you can filter out 70% of the data and  still consider the filter not good enough. That does not make sense. Why so low? Heck I would make it more like 0.85. We might want to make this a configurable parameter.

src/include/op/scan/dynamic_filter_gate.hpp
```diff
@@ -0,0 +1,108 @@
...
+  {
+    return kept > k_filter_skip_keep_threshold;
+  }
+
+ private:
+  enum class state { unknown, active, disabled };
+
+  /// A filter that keeps more than this fraction of the rows it sees prunes too little to repay
+  /// its per-split mask kernel, so it is dropped from later splits.
+  static constexpr double k_filter_skip_keep_threshold = 0.5;
```
wmalpica, line 101: This threshold also seems very low.

wmalpica (changes requested):

src/op/sirius_physical_hash_join.cpp
kevkrist: Good call -- thanks for this. I believe this then makes the second publication window essentially dead, which simplifies the code.

src/op/scan/dynamic_filter_merge.cpp
kevkrist: Yes, switching to 0.9 and configurable.

src/include/op/scan/dynamic_filter_gate.hpp
kevkrist: Switching to 0.9 here also, and making it configurable.

wmalpica (approved):

CMakeLists.txt
```diff
@@ -431,6 +455,7 @@ foreach(_target sirius_extension sirius_loadable_extension)
   # (cuCascade::cucascade) and cudf::cudf.
   target_link_libraries(
     ${_target}
+    CUDA::cuda_driver
```
mbrobbel: I don't think we need this?
```suggestion
```

CMakeLists.txt
```diff
@@ -327,8 +342,16 @@ set(EXTENSION_SOURCES
 # cmake-format: on
 
 # Legacy Sirius — all legacy-specific config lives in src/legacy/CMakeLists.txt.
+# cuCollections (cuco) device code — used by the dynamic Bloom filter — requires
+# nvcc's extended device lambda and relaxed constexpr support.
+string(APPEND CMAKE_CUDA_FLAGS
+       " --expt-extended-lambda --expt-relaxed-constexpr")
```
mbrobbel: We should set this via `target_compile_options`:
```
foreach(_target sirius_extension sirius_loadable_extension)
  target_compile_options(
    ${_target}
    PRIVATE
      $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:--extended-lambda>)
endforeach()
```
And do we need `--expt-relaxed-constexpr`?

CMakeLists.txt
```diff
@@ -93,6 +93,19 @@ find_package(yaml-cpp REQUIRED CONFIG)
 find_package(absl REQUIRED CONFIG)
 find_package(PkgConfig REQUIRED)
 find_package(OpenSSL REQUIRED)
+
+# cuCollections (cuco): header-only, used by the dynamic Bloom filter. libcudf
+# no longer ships its bundled copy, so fetch the same commit cudf is built
+# against. Header-only — populate the sources without add_subdirectory (CCCL
+# comes from cudf); only the include dir is consumed.
+include(FetchContent)
+FetchContent_Declare(
```
mbrobbel: We need a `vcpkg` port for this dependency: https://github.com/sirius-db/sirius/blob/a562339a5c0544cfb743efe3a0c73502b91f952e/vcpkg_ports/cudf/portfile.cmake#L63-L66

kevkrist: I think I've done this. Can you please double-check? This is unfamiliar stuff for me.

CMakeLists.txt
kevkrist: Looking at cuco's own builds, it doesn't look like we need that. Removed.

kevkrist (commented):
Thanks @mbrobbel can you double check the vcpkg stuff?

aminaramoon (approved):

PR #802, opened 2026-05-22
switch parquet metadata reader to use stream pool

wmalpica:


wmalpica:
too much of this code has changed. CLosing.

PR #803, opened 2026-05-22
CI: Change to gpu-2xt4 instead of gpu-4xt4 to improve GPU testing concurrency

mike-wendt:
Realized there is a misconfiguration with `gpu-4xt4` that is reducing our testing throughput. This switches to new `gpu-2xt4` runners that can share a single GPU node with 4xT4s instead of using two 4xT4 nodes per test. This will increase the concurrency for GPU testing while allowing for SNMG testing required after the merge of #732

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -114,7 +114,7 @@ jobs:
           retention-days: 1
 
   test-cuda-13:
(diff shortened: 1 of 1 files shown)


mike-wendt:
CI backed is finally fixed to isolate GPUs on the shared node so there is no overlap or errors. Running CI again to see the full output and timings to compare to the `gpu-4xt4` runners

mbrobbel (approved):

PR #804, opened 2026-05-22
Fix SEGFAULT at query end 2

wmalpica:
The previous PR https://github.com/sirius-db/sirius/pull/788 did not fully get rid of the SEGFAULT. This PR makes more improvements which appear to make the issue fully go away.

One SEGFAULT was due to unnecessary task creation requests still being alive in the queue when the query was completed and the query graph was torn down. So, that was fixed by adding some some checks which prevent adding task creation on completed operators. It also prevents a lot of instances where tasks creation requests were getting created twice (on different parts of the code).

I also removed unnecessary drain on error function at query end. But this still resulted in SEGFAULTS. So I needed to still leave code which waits for tasks to complete, even after the query has been set to complete. This is because there is a possible race condition where for the last operator (RESULT_COLLECTOR) multiple tasks can be finishing at the same time. In that case, one can set the task to completed and the query finishes. Then the other RESULT_COLLECTED task can reach the part of the code which sets the completion handler to completed and in a race condition, it can actually set the next query's completion handler to completed, causing it to segfault.

Additionally, this PR fixes some issues which we had with our signal handlers, which really were not working. Now they should be working and they will now flush the logs when a segfault is caught.

### DIFF
--- a/src/parallel/task_executor.cpp
+++ b/src/parallel/task_executor.cpp
@@ -16,6 +16,9 @@
 
 #include "parallel/task_executor.hpp"
 
+#include <stdexcept>
+#include <string>
+
 namespace sirius {
 namespace parallel {
 
@@ -93,6 +96,37 @@ void itask_executor::drain_and_wait()
   _bounded_pool->resume();
   _task_queue.reactivate();
   _manager_thread = std::thread([this] { manager_loop(); });
+}
+
+void itask_executor::wait_and_validate_empty()
+{
+  // Same quiescing as drain_and_wait(): interrupt the pool + queue so the manager
+  // exits, join it, and wait for all in-flight thread-pool tasks to finish.
+  _bounded_pool->interrupt();
+  _task_queue.interrupt();
+  if (_manager_thread.joinable()) { _manager_thread.join(); }
+  _bounded_pool->wait_all();
+
+  // Instead of draining, VALIDATE the queue is empty. A non-empty queue here means
+  // tasks were still scheduled on this executor when the query was declared complete.
+  const std::size_t remaining = _task_queue.size();
+
+  // Re-enable the pool/queue and restart the manager so the executor is left in a
+  // usable state for the next query, whether or not validation passes.
+  _bounded_pool->resume();
+  _task_queue.reactivate();
+  _manager_thread = std::thread([this] { manager_loop(); });
+
+  if (remaining != 0) {
+    SIRIUS_LOG_ERROR(
+      "itask_executor::wait_and_validate_empty: task queue NOT empty at query completion — "
+      "{} task(s) still queued; tasks were still being scheduled when the query was marked "
+      "complete",
+      remaining);
+    throw std::runtime_error(
+      "task_executor: task queue not empty at query completion (" + std::to_string(remaining) +
+      " task(s) remaining) — premature completion while work was still scheduled");
+  }
 }
 
 }  // namespace parallel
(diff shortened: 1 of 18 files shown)


bwyogatama (approved):

src/util/segfault_backtrace_handler.cpp
```diff
@@ -68,21 +76,61 @@ static void write_backtrace_line(int fd, int frame_no, const char* raw_line)
...
 }
 
+// Best-effort flush of the spdlog logger so the most recent log lines reach
+// disk before we terminate. This is NOT async-signal-safe: the _mt sink takes
+// a mutex and may allocate, so if the crash happened while a thread held the
+// logging mutex (or inside the allocator) the flush could deadlock. We bound
+// that risk with alarm(): SIGALRM's default disposition terminates the
+// process, so a hung flush kills us within the deadline instead of hanging the
+// crash handler forever. alarm() and signal() are themselves signal-safe.
+static void flush_logs_best_effort()
```
mbrobbel, line 86: In my opinion we should avoid this and only use async-signal-safe operations in the handler

src/pipeline/sirius_pipeline.cpp
```diff
@@ -325,18 +325,33 @@ bool sirius_pipeline::is_pipeline_finished() const
...
+  // If this is the original pipeline, we dont want to schedule tasks for its consumers, that will
+  // be done later.
+  if (_task_creator && !original_pipeline) {
     for (auto* consumer : get_output_consumers()) {
-      _task_creator->schedule(consumer);
+      if (!consumer->finalized) {
+        SIRIUS_LOG_WARN(
+          "sirius_pipeline::notify_downstream_pipelines: scheduling consumer op_id={}",
+          consumer->get_operator_id());
+        _task_creator->schedule(consumer);
```
mbrobbel: There is a data race. What happens if we finalize after checking and before scheduling?
And shouldn't we protect `finalized` for concurrent task access?

src/pipeline/task_scheduler.cpp
```diff
@@ -271,6 +274,42 @@ void task_scheduler::drain_after_error()
...
+  // throw if not — a non-empty queue means tasks were still being scheduled when we
+  // declared the query done.
+  //
+  // Halt the producer (task_creator) first so the checks are not racing new task
+  // creation. The task_creator is always restarted afterwards (even on throw) so the
+  // next query can run.
+  if (_task_creator) { _task_creator->stop_thread_pool(); }
+  try {
+    // The task_scheduler's pipeline task queue must be empty.
+    if (const std::size_t remaining = _task_queue.size(); remaining != 0) {
```
mbrobbel, line 290: This also needs synchronization

src/parallel/task_executor.cpp
```diff
@@ -95,5 +98,36 @@ void itask_executor::drain_and_wait()
   _manager_thread = std::thread([this] { manager_loop(); });
 }
 
+void itask_executor::wait_and_validate_empty()
```
mbrobbel, line 101: Do we want this to be thread safe?

src/pipeline/task_scheduler.cpp
wmalpica: What do you mean? wait_for_completion() is to be called at the end of the query. What I am doing here is double checking that the task queue is indeed empty. _task_queue.size() is thread safe.

src/parallel/task_executor.cpp
wmalpica: Its called at the end of the query, it should be on a single thread. We will have to revisit this when we try to support concurrent queries

src/pipeline/task_scheduler.cpp
mbrobbel: If a task was popped from _task_queue but it was not yet scheduled yet, wouldn't that cause issues when the executor restarts?

PR #806, opened 2026-05-25
Add window functions to Sirius (1/3): ROW_NUMBER / RANK / DENSE_RANK

ran-yuan-rui:
Talked with @wmalpica and he point me that this uses hash partitioning + in-window cuDF sort/rank for single-GPU in-core ranking. It is not yet the sort-partition architecture proposed for scalable/out-of-core windows. 
So I will  redo the PR.

## Summary

Adds GPU support for the ranking window functions **`ROW_NUMBER` / `RANK` / `DENSE_RANK`** to Sirius, so **supported ranking-window shapes** run on GPU via `gpu_execution` instead of forcing the whole query back to DuckDB CPU (out-of-scope shapes go through the existing fallback/error path — see Design).

This is **Phase 1 of a staged roadmap toward broader window support** (see below) — the planned phases cover the main *useful* window families, **not** full DuckDB/SQL window parity. It also includes an engine-wide DESC NULLS FIRST/LAST correctness fix in sort/top-n that the window pre-sort depends on. The logical areas are split across commits and can be reviewed independently.

## Design

**Operator.** New `sirius_physical_window`, a *partition-consumer* (same base as hash-join build / aggregate-merge — both a sink and a source). Ranking is not decomposable, so it consumes already hash-partitioned data rather than streaming batch-by-batch.

**Pipeline shape** — reuses the existing partition framework, no new machinery:

- The converter inserts a `sirius_physical_partition` (HASH by the `PARTITION BY` columns; one global partition when there's no `PARTITION BY`).
- `PARTITION → WINDOW` is a **FULL barrier**; the window operator drains one whole partition per task. This is HASH_GROUP_BY's `PARTITION → MERGE_GROUP_BY` shape minus the local partial stage, and **no `CONCAT`** (that's join-only).

**Compute is pure cuDF.** Per partition: stable-sort by `(PARTITION BY, ORDER BY dirs)` → grouped rank scan via `cudf::groupby(…, sorted::YES).scan(make_rank_aggregation(...))` (`FIRST`=row_number, `MIN`=rank, `DENSE`=dense_rank). Under the presorted (`sorted::YES`) path the rank order comes entirely from the pre-sort, so mixed ASC/DESC, multi-column keys, and all three methods are a single code path.

**Why this matters:** we reuse DuckDB's bound expressions + the existing partition/task framework and do *not* port DuckDB's CPU window machinery. As a result, the design covers multi-batch / multi-partition / NULL groups / ASC·DESC·NULLS with the existing framework plus the window operator's per-partition drain/concat path (validated GPU == DuckDB CPU). Out-of-scope shapes (aggregate-over-window, FILTER, EXCLUDE, non-column keys, heterogeneous windows, HUGEINT child cols) throw `NotImplementedException` and use the existing DuckDB fallback path when fallback is enabled.

## Roadmap

A **staged roadmap toward broader window support — not a claim of full DuckDB/SQL window parity.** P1 is ranking; P2/P3 target the main remaining *useful* window families:

| Phase | Scope | Status |
|---|---|---|
| **P1** | `ROW_NUMBER` / `RANK` / `DENSE_RANK` | **this PR** |
| P2 | aggregate-over-window (`SUM/AVG/COUNT/MIN/MAX OVER`) + default & `ROWS` frames | planned |
| P3 | `RANGE` frame, `LEAD`/`LAG`, `FIRST/LAST/NTH_VALUE`, `NTILE`, heterogeneous-window splitting | planned |

cuDF already has the primitives for almost all of P2/P3 (`grouped_rolling_window`, `grouped_range_rolling_window`, `nth_element` / `lead` / `lag`); only `NTILE` has no direct kernel (composable from `row_number` + partition count). So later phases are mostly Sirius-side orchestration.

**Beyond this roadmap** (follow-up, not yet planned): `FILTER` / `EXCLUDE`, `GROUPS` frame, `DISTINCT` aggregate windows, `PERCENT_RANK` / `CUME_DIST`, expression `PARTITION BY` / `ORDER BY` keys, richer types (`DECIMAL` / `HUGEINT` / interval range bounds), collation / string-ordering details, multi-frame splitting & reuse optimization, and out-of-core / spill / multi-GPU semantics. Full DuckDB corner-case parity is explicitly out of scope here.

## Also fixed (engine-level, surfaced by this work)

**`DESC NULLS FIRST/LAST` placement (correctness).** cuDF applies `null_order` in the ascending frame and *reverses* it for a DESCENDING column, so the naive `NULLS_FIRST?BEFORE:AFTER` mapping mis-placed NULLs for every descending sort key — e.g. `ORDER BY x DESC NULLS LAST` put NULLs *first* on GPU, disagreeing with DuckDB CPU. The mapping was duplicated across `order` / `sort_sample` / `sort_partition` / `merge_sort` / `top_n`. Centralized into `op/cudf_sort_order.hpp` (`to_cudf_null_order`, which flips `BEFORE`/`AFTER` for descending keys) and applied everywhere, including the window pre-sort. Also fixes single-key `TOP-N` on a nullable key: `cudf::top_k_order` takes no `null_order` and ignored NULLS placement when *selecting* the top-k, so a nullable key now falls back to a full sort. Pre-existing and undetected because benchmark sort keys are non-null / NULL-filtered (it needs `DESC` ∧ NULLs present ∧ order-sensitive comparison).

## Tests

All Catch2, one `sirius_unittest` binary (tag-filtered). Default regression includes window semantics; TPC-DS + perf are hidden `[.]` and run on demand.

- `test_physical_window.cpp` — plan/guard (`[window][plan]`)
- `test_gpu_execution_window.cpp` — GPU semantics incl. multi-batch/multi-partition and non-integer (VARCHAR/DATE/TIMESTAMP/DOUBLE) keys (`[integration][window][gpu]`)
- `test_gpu_execution_order_nulls.cpp` — the 4 ASC/DESC × NULLS FIRST/LAST quadrants for ORDER BY + TOP-N (`[integration][gpu_execution][nulls]`)
- COALESCE GPU==CPU coverage case in test_gpu_execution_tpch.cpp (the COALESCE impl itself now comes from upstream dev)
- `test_gpu_execution_tpcds.cpp` (Q44), `test_tpcds_plan_translation.cpp` (Q44 + Q67 translation), `test_window_benchmark.cpp` — hidden `[.]`

## Known limitations

- **Single-GPU validation only.** All testing is on one GPU. Multi-GPU is *not* exercised: partitions are independent so the framework could in principle schedule them across GPUs, but that path is **untested** here and not claimed.
- **No out-of-core / spill for blocking operators.** Window, `ORDER BY`, hash-join build and aggregate-merge **OOM instead of spilling** once the working set exceeds the GPU `usage_limit` (even at ~1.04× overflow with host memory free). Root cause: the downgrade path only spills `idle` batches, but partition-consume batches sit in `task_created`/`processing`, so it finds 0 spill candidates → OOM after retries. **not introduced or worsened by this PR** (plain `ORDER BY` OOMs identically). 
Input welcome now in this PR we just fallback.

## Notes

- Adds a committed 12 MB TPC-DS fixture `test/cpp/integration/data/duckdb/tpcds.duckdb` (same precedent as the existing TPC-H `integration.duckdb`) + a `generate_tpcds_duckdb.sh` to reproduce it; `.pre-commit-config.yaml` large-file check widened to cover `data/duckdb/*.duckdb`.


wmalpica:
@ran-yuan-rui Since this PR introduces a bug fix and a 12 MB TPC-DS fixture , could you split those two off into a separate PR, that way we get that merged sooner and the larger piece of work involved here for the WIndow function support does not delay it.

src/planner/sirius_plan_window.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+bool is_bound_ref(const duckdb::Expression& expr)
+{
+  return expr.GetExpressionClass() == duckdb::ExpressionClass::BOUND_REF;
+}
+
+// Phase 1 supports only ROW_NUMBER / RANK / DENSE_RANK, sharing one PARTITION BY / ORDER BY, with
+// column-reference keys, no FILTER / EXCLUDE, and no HUGEINT child columns. Anything else throws
+// NotImplementedException so the whole query falls back to DuckDB CPU.
+void guard_phase1(duckdb::LogicalWindow& op)
```
wmalpica, line 37: i am assuming that we will only support one window in the query for now, correct? Like, you can have multiple aggregators in the same window, but if you have different windows, then the partitioning and ordering would need to be different, and that should be outside the scope of phase 1

ran-yuan-rui:
> @ran-yuan-rui Since this PR introduces a bug fix and a 12 MB TPC-DS fixture , could you split those two off into a separate PR, that way we get that merged sooner and the larger piece of work involved here for the WIndow function support does not delay it.

Thx - will do

mbrobbel:
Marking as draft (there is a conflict).

PR #808, opened 2026-05-27
Restore some scan behavior to pre-multi-gpu

wmalpica:
I had observed that after the multi-gpu PR we were getting more scan tasks and i also noticed that setting the use_sirius_data_source flag did not really change performance.  So this PR fixes a couple things:
1. It makes the  use_sirius_data_source be useful again. In the post multi-gpu the flag was effectively being ignored
2. The multi-gpu PR made it so that we were always splitting scans based on files, so that they could be uniformly distributed between GPUs. This was resulting in more scan tasks and not always following the target scan task size. Now, that behavior will only apply if we actually have multiple GPUs. 
3. The config option scan_task_batch_size is now actually being used in parquet reads from the parquet provider (this, turns out was not working from even before the multi gpu PR)

### DIFF
--- a/src/include/io/datasource_factory.hpp
+++ b/src/include/io/datasource_factory.hpp
@@ -122,19 +122,24 @@ class datasource_factory {
   /**
    * @brief Create a @c cudf::io::datasource for @p uri.
    *
-   * Dispatch (kvikio-free invariant):
+   * Dispatch:
    *   - Every URI scheme MUST resolve through the registry to a registered
    *     @c sirius_ioctx whose @c make_datasource builds a @c sirius_datasource.
    *     The factory NEVER falls back to cudf's bundled @c file_source factory
-   *     (kvikio-backed) — that path is forbidden because kvikio's per-FileHandle
-   *     CUDA-context binding breaks multi-GPU residency.
+   *     (kvikio-backed) from inside @c create — single-GPU users that opt out
+   *     of @c sirius_datasource bypass this factory entirely instead of routing
+   *     a kvikio fallback through it. In multi-GPU mode kvikio is forbidden
+   *     because its per-FileHandle CUDA-context binding breaks multi-GPU
+   *     residency; @c sirius_config::enforce_sirius_datasource_for_multi_gpu()
+   *     ensures @c use_sirius_datasource is true whenever >1 GPU is configured.
    *   - Local file paths (@c "/data/foo.parquet", @c "file:///data/foo.parquet")
    *     route through @c kFileScheme, which is registered in
-   *     @c SiriusContext::initialize() with a uring-backed @c sirius_ioctx.
+   *     @c SiriusContext::initialize() with a uring-backed @c sirius_ioctx
+   *     (registration is skipped when @c use_sirius_datasource is false in a
+   *     single-GPU configuration).
    *   - Object-store schemes (@c s3://, etc.) require their backend to register
    *     a @c sirius_ioctx for that scheme at startup. If no ioctx is registered,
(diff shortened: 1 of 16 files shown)


src/sirius_config.cpp
```diff
@@ -432,12 +433,30 @@ void sirius_config::load_from_file(const std::filesystem::path& config_path)
...
                              e.what());
   }
 }
 
+void sirius_config::enforce_sirius_datasource_for_multi_gpu()
+{
+  size_t num_gpus = 0;
+  for (auto const& space : _memory_space_configs) {
+    if (std::holds_alternative<cucascade::memory::gpu_memory_space_config>(space)) { ++num_gpus; }
+  }
```
9prady9: nit
```suggestion
size_t num_gpus = std::ranges::count_if(_memory_space_configs, [](auto const& space) {
  return std::holds_alternative<cucascade::memory::gpu_memory_space_config>(space);
});
```

src/include/io/datasource_factory.hpp
```diff
@@ -122,19 +122,24 @@ class datasource_factory {
...
    * @brief Create a @c cudf::io::datasource for @p uri.
    *
-   * Dispatch (kvikio-free invariant):
+   * Dispatch:
    *   - Every URI scheme MUST resolve through the registry to a registered
    *     @c sirius_ioctx whose @c make_datasource builds a @c sirius_datasource.
    *     The factory NEVER falls back to cudf's bundled @c file_source factory
-   *     (kvikio-backed) — that path is forbidden because kvikio's per-FileHandle
-   *     CUDA-context binding breaks multi-GPU residency.
+   *     (kvikio-backed) from inside @c create — single-GPU users that opt out
```
9prady9, line 129: curious Q: what are the "@" references ?

src/include/scan_manager/parquet_split_provider.hpp
9prady9, line 1: the docs formatting is off in most of the headers. Or I don't know what the "@" does, never seen it before.

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
```diff
@@ -145,52 +151,65 @@ std::unique_ptr<cudf::table> sirius_gpu_parquet_scan_operator::read_table_from_m
...
-    auto const is_per_gpu_local = [&] {
-      for (auto const& [dev, ctx] : _gpu_ioctxs) {
-        if (ctx == slice.io_ctx) { return true; }
-      }
-      return false;
-    }();
-    auto const ds_ioctx = (slice.io_ctx && !is_per_gpu_local) ? slice.io_ctx : ioctx_it->second;
-    if (slice.io_object) {
-      sources.push_back(ds_ioctx->make_datasource(slice.io_object));
+    if (kvikio_fallback_mode || (!slice.io_ctx && !slice.io_object)) {
```
9prady9: Can the second arm of the OR be true when first arm is false (i.e. when `_gpu_ioctxs.empty()` is `false`)

src/include/io/datasource_factory.hpp
wmalpica: Its Doxigen formatting. They apply formatting to the next word. @p for parameters and makes them italics and the other one @c for code which make it monospacing. Claude did it

src/op/scan/sirius_gpu_parquet_scan_operator.cpp
wmalpica: good point. i simplified it

9prady9 (approved):

PR #809, opened 2026-05-27
Quent Task Instrumentation

### LINKED ISSUE #720: [Quent] Integrate Quent into Sirius
This is intended to be the parent issue for all sub-issues in the Quest - Sirius integration effort

dhruv9vats:
Defines a task model based on the Sirius' execution semantics, and adds instrumentation to capture information about task events.

Also add basic analyzers that [quent](github.com/rapidsai/quent) can use to generate query timelines for tasks.

Notes

  - This PR instruments the task lifecycle, not the new scan split lifecycle.
  - The new upstream scan path is covered because `sirius_gpu_scan_operator::execute()` still runs under `gpu_pipeline_task`.
  - Split production, footer reads, and row-group pruning happen before task creation and
    remain outside the task FSM.


docs/super-sirius/configuration.md
```diff
@@ -214,6 +214,51 @@ sirius:
...
+
+Start the Quent analyzer server over the same telemetry directory to view the
+captured telemetry:
+
+```bash
+pixi run cargo run --manifest-path rust/Cargo.toml \
+  -p sirius-telemetry-server \
+  --features ui \
+  -- \
+  --output-dir telemetry_data
```
mbrobbel: Maybe we can add a `pixi run quent`?

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -0,0 +1,1614 @@
...
+use crate::{
+    model::{SiriusModel, SiriusModelBuilder},
+    task::{Task, TaskExt},
+};
+
+pub mod model;
+pub mod task;
+pub mod view;
+
+pub struct SiriusUiAnalyzer {
```
mbrobbel, line 54: Can you please add docs to types and methods (to help reviewers)?

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -0,0 +1,1614 @@
...
+            });
+
+        Ok(QueryBundle {
+            query_id,
+            entities,
+            plan_tree,
+            resource_tree,
+            unique_operator_names,
+            quantity_specs: [
+                ("capacity_entries".to_string(), QuantitySpec::unit()),
```
mbrobbel: Should this use `QUEUE_ENTRY_CAPACITY.name`? And the same in other locations.

rust/.gitignore
```diff
@@ -1 +1,2 @@
 target
+crates/telemetry/server/ts-bindings
```
mbrobbel: Nit: move this to `crates/telemetry/server/`, to make it survive moves/renames

mbrobbel (commented):
I started looking at this but I think it would help me and other reviewers if there are some hints/instructions/details for reviewers in the PR description. Can you please add those?

docs/super-sirius/configuration.md
9prady9: +1

src/include/telemetry/telemetry_context.hpp
```diff
@@ -35,7 +45,9 @@ namespace sirius::telemetry {
 /// Owns the top-level telemetry states for a single SiriusContext.
 class telemetry_context {
  public:
-  explicit telemetry_context(const sirius::telemetry_config& config);
+  [[nodiscard]] static std::shared_ptr<const telemetry_context> create(
+    const sirius::telemetry_config& config);
```
9prady9, line 49: nice, factories gives us more control 👍🏾

src/include/telemetry/telemetry_context.hpp
```diff
@@ -70,4 +84,96 @@ void emit_plan_telemetry(
...
+inline thread_local std::optional<ExecutorThreadHandleWrapper> executor_thread_telemetry_handle{
+  std::nullopt};
+
+// Initialize the thread local ExecutorThreadHandleWrapper for this worker thread.
+inline void thread_local_executor_thread_telemtry_init(const telemetry_context& context,
+                                                       const std::string& thread_name)
+{
+  if (executor_thread_telemetry_handle.has_value()) {
+    SIRIUS_LOG_WARN("ExecutorThreadHandleWrapper was already initialized; overriding.");
+  }
```
9prady9, line 175: can't we use it if it already exists ? didn't quite get why we need to override - may be it will be clear once I finish reviewing all the files.

rust/crates/telemetry/model/src/task.rs
```diff
@@ -0,0 +1,172 @@
...
+            reserving: Reserving,
+            downgrading: Downgrading,
+            preparing: Preparing,
+            computing: Computing,
+            finalizing: Finalizing,
+        },
+        entry: created,
+        exit_from: { finalizing },
+        transitions: {
+            // Task object exists and has entered its first scheduling queue.
```
johanpel, line 121: Does this first scheduling queue have distinctly different semantics from the routing => queued queue? If so would a distinct state help understand the task lifecycle?

rust/crates/telemetry/model/src/task.rs
```diff
@@ -0,0 +1,172 @@
...
+state! {
+    Finalizing {
+        attributes: {
+            success: bool,
+        },
+    }
+}
+
+fsm! {
+    Task {
```
johanpel, line 107: Kudos for the modeling effort here

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -0,0 +1,1614 @@
...
+    type TimelineGlobalParams = QueryFilter;
+    type TimelineParams = TaskFilter;
+
+    fn extract_engine(
+        engine_id: Uuid,
+        events: impl Iterator<Item = Event<SiriusEvent>>,
+    ) -> AnalyzerResult<quent_query_engine_ui::Engine> {
+        use quent_query_engine_model::engine::EngineEvent;
+        for event in events {
+            if let SiriusEvent::Engine(EngineEvent::Init(init)) = event.data {
```
johanpel, line 86: Two clarifying comments about this, no change needed:
- In the current PoC state of Quent this seems inefficient as it implies we're iterating over all events twice, but in practice this event is always one of the (typically just the) first one.
- Quent will soon start partitioning events across one or multiple files per entity (each observer will export to its own (set of) file(s)) and into file formats with statistics that allow scanning them faster (e.g. Parquet), so even if there was a concern this will all be way more efficient, even without building some indexing service.

rust/crates/telemetry/analyzer/src/view.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        result.tasks = model
+            .tasks
+            .values()
+            .map(|task| (task.id(), task))
+            .filter(|(_, task)| {
+                task.usages()
+                    .any(|usage| result.resource(usage.resource_id()).is_ok())
+            })
+            .collect();
```
johanpel, line 103: Might need to provide a different way to filter because IIUC a task can go from created -> finalizing without using any resources, so it would not appear in the view.

johanpel (commented):
Awesome, this is a good example of how to further customize a domain-specific model. Reviewed the Rust side.

Not something you can do about in this PR, but this is also a good example of how much _more_ boilerplate we could generate from a single model definition.

9prady9 (approved):
cpp changes look fine. I have some follow up queries:

- The `rust::Box<>` inside task, whats the cost of variable ? Is it trivial enough when a large query runs ? Do we have comparison of `quent=true` vs `quent=false` run anywhere to confirm there is only noise overhead. Do we dump events even when we don't want telemetry (what happens when telemetry is false) is not 100% clear to me yet.
- perhaps a table with info on what level handle handles which component's telemetry.

I will take a stab at rust tomorrow, but not a blocker from my end.

docs/super-sirius/configuration.md
dhruv9vats: done, thanks.

rust/crates/telemetry/analyzer/src/view.rs
dhruv9vats: tasks are now selected based on matching pipeline_uuids.

src/include/telemetry/telemetry_context.hpp
dhruv9vats: we can, but this is supposed to be set only once logically, and surfacing a warning is useful as initially set contextual information will be over written.

rust/Cargo.toml
```diff
@@ -3,6 +3,10 @@ resolver = "3"
 members = [
     "crates/telemetry/bridge",
     "crates/telemetry/model",
+    "crates/telemetry/ui",
+    "crates/telemetry/analyzer",
+    "crates/telemetry/server",
+    "crates/telemetry/model",
```
dhruv9vats, line 9: The `ui`, `analyzer` and `server` crates are adapted from quent's simulator analyzer for the most part. I have tagged parts that are modified to aid reviews.

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -0,0 +1,847 @@
...
+    }
+
+    fn bulk_chunked_resource_timeline(
+        &self,
+        request: BulkChunkedTimelineRequest<Self::TimelineGlobalParams, Self::TimelineParams>,
+    ) -> AnalyzerResult<BulkChunkedTimelinesResponse> {
+        let epoch = self
+            .query_engine_model()
+            .query_epoch(request.app_params.query_id)?;
+        let view = self.model.query_view(request.app_params.query_id)?;
```
dhruv9vats, line 534: We use a "query_view" to obtain tasks and other entities, as this allows us to return appropriated filtered tasks (ones that belong a pipelines in this query), as opposed to directly for the tasks via the model.

rust/crates/telemetry/analyzer/src/view.rs
```diff
@@ -0,0 +1,306 @@
...
+
+        result.tasks = model
+            .tasks
+            .values()
+            .filter(|task| {
+                task.pipeline_uuid()
+                    .is_some_and(|pipeline_uuid| pipeline_ids.contains(&pipeline_uuid))
+            })
+            .map(|task| (task.id(), task))
+            .collect();
```
dhruv9vats, line 103: here we select tasks that belong to pipelines that are a part of this query.

rust/crates/telemetry/analyzer/src/view.rs
```diff
@@ -0,0 +1,306 @@
...
+        Ok(result)
+    }
+
+    fn collect_runtime_resource_ancestors(
```
dhruv9vats, line 107: this is to collect all resources between a specific resource and the quent-defined semantic query resource in the model resource tree. this ensures if there are a use of a application specific resource, all resource that connect it to a query resource are visible to the UI.

rust/crates/telemetry/analyzer/src/lib.rs
dhruv9vats: While this is mostly adapted from the quent simulator analyzer, as I was looking to add explanatory doc strings for easy comprehension now and moving forward, it was realized much of this will be restructured and moved into the codegen step of quent itself once the update in quent lands, such that the analyzers themselves also become quite composable. So I am hoping slide this over to next slice of work as we use this stop gap.

rust/crates/telemetry/ui/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+        #[derive(Deserialize)]
+        struct TaskFilterWire {
+            #[serde(default)]
+            pipeline_uuid: Option<Uuid>,
+            #[serde(default)]
+            operator_id: Option<Uuid>,
+            #[serde(default)]
+            physical_operator_id: Option<u32>,
+            #[serde(default)]
+            current_operator_id: Option<u32>,
```
dhruv9vats, line 68: to maintain compatibility.

dhruv9vats (commented):
Merged dev, accounting for the recent change in how we execute scans.

bwyogatama (approved):

bwyogatama:
I only looked at the cpp changes inside sirius execution, and those look good to me

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -167,6 +202,14 @@ void gpu_pipeline_executor::manager_loop()
         break;
       }
 
+      gpu_task->telemetry_handle().reserving({
```
wmalpica: This telemetry information regarding `reserving` how do we want to use it?
For this task, we already captured what we were reserving, so this would be capturing again.
Also, this would be logging AFTER it was already reserved, since the reservation in this case is happening effectively during the downgrade process, in the predicate lambda passed to the downgrade executor. 
So i think we want to remove this

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -328,6 +371,14 @@ void gpu_pipeline_executor::manager_loop()
 
           // Schedule the rescheduled task. It goes back through manager_loop()
           // to acquire a fresh reservation before execution.
+          if (auto* pipeline_task = dynamic_cast<sirius_pipeline_itask*>(task.get())) {
+            pipeline_task->telemetry_handle().finalizing({
+              .instance_name = "",
+              .success       = false,
```
wmalpica, line 369: should we capture that it failed, but we are re-trying?
And then, in a couple other places capture when it fails and does not retry?

wmalpica (changes requested):

src/pipeline/gpu_pipeline_executor.cpp
dhruv9vats: >This telemetry information regarding reserving how do we want to use it?

The idea is to surface it to the user via the UI. This is something that will have to be added to the UI.

>For this task, we already captured what we were reserving, so this would be capturing again.

True.

> Also, this would be logging AFTER it was already reserved, since the reservation in this case is happening effectively during the downgrade process, in the predicate lambda passed to the downgrade executor. So i think we want to remove this

That makes sense, this was added to make the FSM always go through reserving before preparing. But its better to think of the downgrading state as accounting for both the downgrade part and clubbed reserving part (I considered a async callback for that, but felt it was overkill for this, for now), and allow going to preparing directly from downgrading, that will be more representative. Thanks!

src/pipeline/gpu_pipeline_executor.cpp
dhruv9vats: That is a great suggestion. I will add something with which we can create a link between the original and the retried tasks in the analyzer, and highlight all of them at once in the UI. Let's do this as a follow up.

dhruv9vats: Added: https://github.com/sirius-db/sirius/issues/943

src/pipeline/gpu_pipeline_executor.cpp
dhruv9vats: [Allow going from downgrading to preparing](https://github.com/sirius-db/sirius/pull/809/commits/4a513eb0e6874fea08a3a141d2f8e761cbf5f2b2)

wmalpica (approved):

src/pipeline/sirius_pipeline_itask.cpp
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "pipeline/sirius_pipeline_itask.hpp"
+
+#include "telemetry-bridge/gen/uuid.rs.h"
+#include "telemetry/telemetry_context.hpp"
+
+#include <spdlog/fmt/bundled/core.h>
```
mbrobbel: This fails in the [vcpkg build](https://github.com/sirius-db/sirius/actions/runs/27562982426/job/81479604062):
```
/duckdb_build_dir/src/pipeline/sirius_pipeline_itask.cpp:22:10: fatal error: spdlog/fmt/bundled/core.h: No such file or directory
   22 | #include <spdlog/fmt/bundled/core.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

mbrobbel:
```suggestion
#include <spdlog/fmt/fmt.h>
```

PR #810, opened 2026-05-27
Bugfix desc null fix + add tpcds fixture

ran-yuan-rui:
## Summary

Split out from #806 per @wmalpica review comment.

This PR contains two independent pieces:

1. Fix GPU `ORDER BY` / `TOP-N` correctness for `DESC ... NULLS FIRST/LAST`.
2. Add a reusable SF0.01 TPC-DS DuckDB fixture for integration / plan-translation tests.

## RCA: DESC NULL ordering bug

DuckDB SQL treats `NULLS FIRST` / `NULLS LAST` as the final, absolute placement of NULLs in the result:

- `DESC NULLS FIRST` => NULLs first, then values descending
- `DESC NULLS LAST` => values descending, then NULLs last

Sirius previously mapped SQL null ordering directly to cuDF:

```cpp
NULLS_FIRST -> cudf::null_order::BEFORE
NULLS_LAST  -> cudf::null_order::AFTER
```

That is correct for `ASC`, but wrong for `DESC`: cuDF applies `null_order` in its sort-direction frame, so DESC effectively reverses the NULL placement. As a result, GPU `ORDER BY k DESC NULLS LAST` could place NULLs first, disagreeing with DuckDB CPU.

The same naive mapping was duplicated across the distributed sort path:

- `ORDER_BY`
- `SORT_SAMPLE`
- `SORT_PARTITION`
- `MERGE_SORT`
- `TOP_N`

There was one additional TOP-N issue: `cudf::top_k_order` has no `null_order` parameter, so single-key TOP-N on a nullable key could select the wrong top-k rows before any final sorting step.

## Fix

- Added `src/include/op/cudf_sort_order.hpp` as the single source of truth:
  - `to_cudf_order(...)`
  - `to_cudf_null_order(...)`
- `to_cudf_null_order(...)` flips `BEFORE` / `AFTER` for descending keys so SQL `NULLS FIRST/LAST` matches DuckDB semantics.
- Routed all sort/top-n operators through the helper.
- For single-key TOP-N:
  - keep `cudf::top_k_order` fast path when the key has no NULLs
  - use `sort_by_key + slice` when the key is nullable, so NULL placement is honored

## TPC-DS fixture

Update (reworked per @mbrobbel's review): Dropped the committed 12 MB tpcds.duckdb binary. The fixture is now generated at build time via an opt-in CMake target tpcds-fixture (runs generate_tpcds_duckdb.sh into ${CMAKE_BINARY_DIR}/test-fixtures/, SF 0.01 via DuckDB's dsdgen). Run cmake --build . --target tpcds-fixture once before the TPC-DS integration tests. 

## Tests

Added `test_gpu_execution_order_nulls.cpp`:

- `ORDER BY`
- `TOP-N single-key`
- `TOP-N multi-key`
- `ASC/DESC × NULLS FIRST/LAST`

The test materializes GPU output once, adds `observed_position` to make ordering differences visible, and compares against DuckDB CPU with bidirectional `EXCEPT ALL`.

Validation:

```bash
pre-commit run --all-files
pixi run -e default make release
build/release/extension/sirius/test/cpp/sirius_unittest "[nulls]"
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -524,6 +524,24 @@ endif()
 
 add_executable(sirius_unittest ${TEST_SOURCES})
 
+# TPC-DS integration test fixture, generated at build time (opt-in target). The
+# database is not committed; run `cmake --build . --target tpcds-fixture` once
+# before the TPC-DS integration tests. Requires the duckdb CLI (or python3 with
+# the duckdb module) on PATH; both are provided by the pixi environment.
+set(TPCDS_FIXTURE_DIR ${CMAKE_BINARY_DIR}/test-fixtures)
+set(TPCDS_FIXTURE ${TPCDS_FIXTURE_DIR}/tpcds.duckdb)
+set(TPCDS_FIXTURE_SCRIPT
+    ${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/integration/data/duckdb/generate_tpcds_duckdb.sh
+)
+add_custom_command(
+  OUTPUT ${TPCDS_FIXTURE}
+  COMMAND ${CMAKE_COMMAND} -E make_directory ${TPCDS_FIXTURE_DIR}
+  COMMAND bash ${TPCDS_FIXTURE_SCRIPT} ${TPCDS_FIXTURE}
+  DEPENDS ${TPCDS_FIXTURE_SCRIPT}
+  COMMENT "Generating TPC-DS test fixture (SF 0.01) at ${TPCDS_FIXTURE}"
+  VERBATIM)
+add_custom_target(tpcds-fixture DEPENDS ${TPCDS_FIXTURE})
+
 if(VCPKG_BUILD)
   set_target_properties(sirius_unittest PROPERTIES NO_SYSTEM_FROM_IMPORTED ON)
   target_include_directories(sirius_unittest BEFORE PRIVATE ${_VCPKG_INC})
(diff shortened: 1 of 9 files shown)


test/cpp/integration/data/duckdb/tpcds.duckdb
mbrobbel, line 1: This generated file should not be committed.

ran-yuan-rui: As the TPC-H test fixture follows the same pattern today — `test/cpp/integration/data/
duckdb/integration.duckdb` is a committed ~26 MB DuckDB file on `dev` right now (added in #324). This PR mirrors that pattern at a smaller scale (12 MB, SF 0.01) for TPC-DS, that's why I was think this should be fine.

Happy to switch direction if you'd prefer, but want to avoid ending up with the two fixtures handled inconsistently.

wmalpica (approved):

test/cpp/integration/data/duckdb/tpcds.duckdb
ran-yuan-rui: @mbrobbel could you check my comment on what will next step here

ran-yuan-rui: Done — dropped the committed binary and added an opt-in CMake target tpcds-fixture that regenerates it from generate_tpcds_duckdb.sh.

PR #815, opened 2026-05-28
Move `telemetry_context` into `SiriusContext`

dhruv9vats:
We want to generate telemetry from `SiriusContext` wide resources, like task queue and executors. So the `telemetry_context` needs to outlive them, and should be moved to the `SiriusContext` class. 

Also moves the start-up configuration of the telemetry to the yaml based workflow which is a more natural place for such one-time configurations. 

Retains the ability to set run-time per-query telemetry identifier.


src/include/telemetry/telemetry_context.hpp
```diff
@@ -22,19 +22,22 @@
...
+
 namespace sirius::telemetry {
 
-/// Owns the top-level telemetry states for a single Sirius session (sirius_interface level).
+/// Owns the top-level telemetry states for a single SiriusContext.
 class telemetry_context {
  public:
-  telemetry_context(std::optional<std::string> query_label);
+  explicit telemetry_context(const sirius::telemetry_config& config);
   ~telemetry_context();
```
9prady9, line 39: good change, now config controls if this gets passed I suppose ?

src/sirius_engine.cpp
```diff
@@ -110,7 +130,7 @@ duckdb::unique_ptr<duckdb::QueryResult> sirius_engine::get_result()
 {
   D_ASSERT(has_result_collector());
   if (!sirius_physical_plan) throw invalid_input_exception("sirius_physical_plan is NULL");
-  if (sirius_physical_plan.get() == NULL)
+  if (sirius_physical_plan.get() == nullptr)
```
9prady9: if that was a smart pointer, no need to even write that comparison.

src/sirius_interface.cpp
```diff
@@ -217,18 +217,27 @@ duckdb::unique_ptr<duckdb::QueryResult> sirius_interface::sirius_execute_query(
...
+    D_ASSERT(sirius_active_query->is_open_result(*pending_query));
+    duckdb::unique_ptr<duckdb::QueryResult> current_result;
+
+    auto result = sirius_execute_pending_query_result(*pending_query);
+    SIRIUS_LOG_DEBUG("Done sirius_execute_query");
+
+    return result;
+  } catch (std::exception& e) {
+    if (sirius_active_query) { cleanup_internal(nullptr, false); }
+    return sirius_error_result<duckdb::MaterializedQueryResult>(duckdb::ErrorData(e));
```
9prady9, line 239: 👍🏾

src/telemetry/telemetry_context.cpp
```diff
@@ -33,27 +32,23 @@
...
 
-telemetry_context::telemetry_context(std::optional<std::string> query_label)
+telemetry_context::telemetry_context(const sirius::telemetry_config& config)
   : engine_uuid_(uuid::now_v7()),
     worker_uuid_(uuid::now_v7()),
-    context_(quent::create_context(uuid::now_v7(),
-                                   duckdb::Config::ENABLE_QUENT ? "ndjson" : "noop",
-                                   duckdb::Config::QUENT_OUTPUT_DIRECTORY)),
+    context_(quent::create_context(
+      uuid::now_v7(), config.enable_quent ? "ndjson" : "noop", config.output_directory)),
```
9prady9, line 39: curious Q: why is this ternary operator sending in a string ?

9prady9 (approved):
a nit or so, but looks good.

src/include/telemetry/telemetry_context.hpp
dhruv9vats: yes, the start-up telemetry config that logically gets set only once is configured via the yaml based config path.

src/sirius_engine.cpp
dhruv9vats: true.

src/telemetry/telemetry_context.cpp
dhruv9vats: just an internal stop gap thing until we switch passing an enum for this as it should be; that would follow after the current codegen refactor is done.

src/sirius_engine.cpp
dhruv9vats: seem like this was a duplicate check anyways -- removed.

9prady9 (approved):

PR #825, opened 2026-05-29
Use s3 for sccache in check workflow for all jobs

mbrobbel:
This should help us to better utilize the 10GB GitHub actions cache for other cache entries.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -39,7 +39,6 @@ jobs:
 
   build:
     env:
-      SCCACHE_GHA_ENABLED: true
       CARGO_INCREMENTAL: "0"
       RUSTC_WRAPPER: sccache
     strategy:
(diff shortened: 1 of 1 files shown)


mike-wendt:
We need to make structural changes for this to work. Actions secrets are not shared on PR forks which is the majority of this repo. We can switch to self-hosted runners where all this is easier because IRSA is used so no secrets are needed. Investigating how to do this with our runners and will come back to this PR when ready to test

mike-wendt:
Progress update: The needed changes to support this PR are detailed in #845 and the first PR to implement this work is #842. Once #845 is complete, I will open an issue to discuss the sizing of the self-hosted runners to get your input. After agreeing on the size we need, I can add the runner scale set to Flux and we can start testing in this rebased PR. My goal is to get this finished by the end of the week. I'll tag you in the issue about self-hosted runner size discussion.

mike-wendt (approved):

PR #827, opened 2026-05-29
Feature/optimizing multi gpu

felipeblazing:
## Restore pull-signal backpressure in multi-GPU task scheduler

  ### Problem

  PR #732 introduced multi-GPU support by switching the task scheduler from a pull
  model (executors request work when ready) to a push model (scheduler dispatches
  tasks greedily based on data locality). This caused tasks to pile up inside
  per-GPU executor queues where the **downgrade executor cannot reach them** —
  defeating the memory-pressure relief mechanism that keeps queries from OOMing
  under heavy workloads.

  Concretely, `management_eventloop` was popping tasks from `_task_queue`
  unconditionally and pushing them to GPU executors regardless of worker-thread
  availability. Once a task sat in an executor's internal queue it held a GPU
  memory reservation that downgrade could no longer reclaim.

  ### Solution

  Restore the pre-#732 pull-signal pattern, adapted for multi-GPU data locality:

  - **GPU executors signal readiness.** After `_bounded_pool->reserve()` secures
    a worker thread, the executor publishes a `DEVICE_READY` event on the existing
    `_task_request_channel`. Tasks are only dispatched to executors that have a
    thread ready *right now*.
  - **Preference-aware matcher.** The scheduler uses two `inspectable_mpsc::pop_if()`
    passes: first a strict match against `preferred_device_id`, then a relaxed
    pass for tasks with no preference (or with a preference for a device that
    doesn't exist in the current config). Tasks preferring a *different* existing
    device stay in the queue until that device is ready — required for correctness
    when data is physically resident on a specific GPU (e.g. `cache=table_gpu`).
  - **Tasks stay downgrade-visible.** Because tasks remain in the scheduler's
    `_task_queue` until a device is ready, the downgrade executor's TIER 2 scan
    continues to find them. Steady-state per-executor queue depth is ~1.
  - **TASK_AVAILABLE sentinel.** A new `task_request_kind` enum distinguishes
    `device_ready` from `task_available`. `schedule()` publishes `task_available`
    via `_self_publisher` to wake the matcher when new work arrives while devices
    are already idle.

  ### Additional fixes included

  **SF1000 FLBA-decimal filter-pushdown crash.** Parquet files written by
  tpchgen-rs at SF1000 store decimals as `FIXED_LEN_BYTE_ARRAY`. cuDF's
  row-group stats filter then crashes with *"Invalid type and stats combination"*.
  `parquet_split_provider` now probes the first file's schema and disables filter
  pushdown when FLBA/BYTE_ARRAY decimal columns are present; the predicate is
  applied post-decode instead.

  **SF1000 null-deref in `drain_pending_tasks`.** `task_creator::drain_pending_tasks()`
  dereferenced a null `_bounded_pool` after `stop_thread_pool()` reset it on the
  error path. Added a null guard mirroring `task_executor::drain_and_wait()`.

  **cucascade bump for host-pool cross-GPU staging.** Picks up the cucascade fix
  that routes cross-GPU host-staging through the pre-pinned per-NUMA HOST pool
  instead of per-transfer `cudaHostAlloc` — root cause of the SF100 multi-GPU
  regression (1-GPU 47.7s vs 2-GPU 53.4s on join-heavy queries).

  ### Key design decisions

  | Decision | Rationale |
  |---|---|
  | Single `_task_queue` (not per-device split) | Preserves FIFO order; leaves room for future priority semantics |
  | All preferences treated as binding (step i) | Correctness-first: prevents cross-GPU CUDA errors. A follow-up will add an explicit `strict` vs `prefer` bit so scan tasks can load-balance while partitioned 
  ops stay pinned |
  | `channel::reopen()` added | Scheduler goes through stop/start cycles between queries (`drain_after_error`, shared test envs); the channel must be revivable after `close()` |
  | Filter `!evt->is_scan` on `DEVICE_READY` | The scan executor shares the same channel and emits `task_request(0, true)`; without filtering these were injected as spurious GPU readiness signals |
  | Non-existent preferred device → unpreferred | Tasks carrying a preference from a previous 2-GPU env must still run in a 1-GPU env |

  ### Files changed (14 files, +548/-75)

  - `src/pipeline/task_scheduler.{hpp,cpp}` — pull-signal matcher, `_ready_devices`, `_self_publisher`
  - `src/pipeline/gpu_pipeline_executor.cpp` — `DEVICE_READY` signal after `reserve()`
  - `src/include/pipeline/task_request.hpp` — `task_request_kind` enum
  - `src/include/exec/channel.hpp` — `reopen()` for stop/start lifecycle
  - `src/parallel/task_executor.cpp`, `src/creator/task_creator.cpp` — null guard on drain
  - `src/scan_manager/parquet_split_provider.cpp` + scan operator data — FLBA pushdown skip
  - `src/op/scan/duckdb_scan_executor.cpp`, `sirius_gpu_parquet_scan_operator.cpp` — wiring
  - `cucascade` submodule bump
  - `test/cpp/pipeline/test_task_scheduler.cpp` — 3 new tests
  - `test/cpp/scan/test_parquet_scan_task.cpp` — 2 new tests

  ### Test plan

  - [x] `[scan]` — 165 tests, 154,819 assertions (includes FLBA-decimal pushdown skip)
  - [x] `[mgpu]` — 16 tests, 74,321 assertions (includes `cache=table_gpu` cross-device)
  - [x] `[task_scheduler]` — 5 tests: stop/start cycle, multi-cycle, 2-GPU dispatch
  - [x] `[flba_decimal]` — 2 tests: FLBA decimal skips pushdown, INT64 decimal allows it
  - [x] `pre-commit run` clean on all modified files
  - [ ] TPC-H SF100/SF1000 spot-check on 1-GPU and 2-GPU configs (requires driver + dataset)

  ### Future work

  Step (ii): add an explicit `strict` vs `prefer` affinity bit on
  `gpu_pipeline_task`. Today every preference is treated as binding, which is
  correct but leaves load-balancing on the table for scans. A follow-up PR will
  let partitioned operators (joins, group-by, cached scans) declare strict
  pinning while non-partitioned scans float to any ready device.

  ### Commits

  - `d1c8755b` feat: restore pull-signal backpressure in multi-GPU task scheduler
  - `28d49f14` test: add FLBA-decimal pushdown skip and pull-signal scheduler tests
  - `e2e7fc3e` chore: bump cucascade (host-pool cross-GPU staging)
  - `af889fae` style: apply clang-format to modified files

### DIFF
--- a/src/include/pipeline/task_scheduler.hpp
+++ b/src/include/pipeline/task_scheduler.hpp
@@ -35,7 +35,9 @@
 #include <atomic>
 #include <future>
 #include <map>
+#include <optional>
 #include <queue>
+#include <set>
 
 namespace sirius::op::scan {
 class duckdb_scan_executor;
@@ -217,13 +219,26 @@ class task_scheduler {
 
   exec::inspectable_mpsc<sirius::parallel::itask> _task_queue;  ///< Queue for GPU pipeline tasks
   exec::channel<std::unique_ptr<task_request>> _task_request_channel;
+  /// Publisher used by schedule() to wake the management event loop when a new
+  /// task is pushed into _task_queue. The event loop blocks on _task_request_channel
+  /// for two event kinds: device_ready (sent by gpu_pipeline_executors when a worker
+  /// thread is reserved) and task_available (sent by schedule()).
+  std::optional<exec::publisher<std::unique_ptr<task_request>>> _self_publisher;
   std::thread _management_thread;
   std::atomic<bool> _running{false};
+
+  /// Set of GPU device_ids that have a reserved worker thread waiting for a task.
+  /// Only mutated by the management thread (matches device_ready signals from
+  /// _task_request_channel and erases on dispatch), so no synchronization needed.
+  std::set<int> _ready_devices;
 
   /// device_id -> GPU executor. std::map (not unordered_map) so iteration
   /// order is deterministic (ascending by device_id) — keeps preference-less
   /// task dispatch reproducible across runs.
   std::map<int, std::unique_ptr<gpu_pipeline_executor>> _gpu_executors;
+  /// Deprecated as of pull-signal restoration: no-preference distribution now
+  /// arises from which executor sends a device_ready signal first, not from a
+  /// counter. Field retained for source compat with set_no_pref_rr_counter_for_testing.
   std::atomic<size_t> _no_pref_rr_counter{0};
 
   sirius::creator::task_creator* _task_creator{nullptr};
--- a/src/pipeline/task_scheduler.cpp
+++ b/src/pipeline/task_scheduler.cpp
@@ -99,8 +99,8 @@ void task_scheduler::schedule(std::unique_ptr<sirius::parallel::itask> task)
   } else {
     [[maybe_unused]] auto _ = _task_queue.push(std::move(task));
     if (_self_publisher) {
-      auto wake  = std::make_unique<task_request>();
-      wake->kind = task_request_kind::task_available;
+      auto wake                 = std::make_unique<task_request>();
+      wake->kind                = task_request_kind::task_available;
       [[maybe_unused]] auto _ok = _self_publisher->send(std::move(wake));
     }
   }
--- a/src/pipeline/task_scheduler.cpp
+++ b/src/pipeline/task_scheduler.cpp
@@ -46,6 +46,11 @@ task_scheduler::task_scheduler(
   _scan_executor = std::make_unique<sirius::op::scan::duckdb_scan_executor>(
     scan_executor_config, &mem_mgr, _task_request_channel.make_publisher());
 
+  // Self-publisher: schedule() uses this to wake management_eventloop when a
+  // new task is pushed, so the loop can re-run the matcher against any device
+  // that is already in _ready_devices.
+  _self_publisher.emplace(_task_request_channel.make_publisher());
+
   auto gpu_spaces = mem_mgr.get_memory_spaces_for_tier(cucascade::memory::Tier::GPU);
   // Initialize GPU pipeline executors for each available GPU
   for (auto* space : gpu_spaces) {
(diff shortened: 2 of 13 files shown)


src/pipeline/task_scheduler.cpp
```diff
@@ -228,58 +255,85 @@ void task_scheduler::drain_after_error()
...
+        [device_id](const sirius::parallel::itask& t) -> bool {
+          const auto* gpu_task = dynamic_cast<const pipeline::gpu_pipeline_task*>(&t);
+          if (!gpu_task) { return false; }
+          auto pref = gpu_task->get_preferred_device_id();
+          return pref.has_value() && pref.value() == device_id;
+        },
+        /*front_to_back=*/true);
+      if (!task) {
+        // Fallback: take the first task with NO preference, or whose preferred
+        // device does not exist in _gpu_executors (a stale preference from a
```
wmalpica, line 326: I feel like if the preferred device does not exist, that should be an error, no? is there any legitimate situation where that would happen? just returning could be masking some sort of error

src/include/pipeline/task_scheduler.hpp
```diff
@@ -217,13 +219,26 @@ class task_scheduler {
...
 
+  /// Set of GPU device_ids that have a reserved worker thread waiting for a task.
+  /// Only mutated by the management thread (matches device_ready signals from
+  /// _task_request_channel and erases on dispatch), so no synchronization needed.
+  std::set<int> _ready_devices;
+
   /// device_id -> GPU executor. std::map (not unordered_map) so iteration
   /// order is deterministic (ascending by device_id) — keeps preference-less
   /// task dispatch reproducible across runs.
   std::map<int, std::unique_ptr<gpu_pipeline_executor>> _gpu_executors;
```
aminaramoon, line 238: why is this map ? it should be unordered_map, right ?

aminaramoon (approved):
LGTM!

PR #830, opened 2026-05-29
ENH Update Pixi CUDAARCHS to include 89 for Ada Lovelace (L4)

mike-wendt:
Currently we don't build for Ada Lovelace or any other 8.9 compute capability. This adds the missing CUDA architecture so a Sirius build can run on L4s in the cloud. This couples with our recent change to L4s for GPU testing where developers may want to reproduce the work on an L4 instance.

### DIFF
--- a/.github/workflows/distribution.yml
+++ b/.github/workflows/distribution.yml
@@ -44,7 +44,7 @@ jobs:
       build_duckdb_shell: false
       vcpkg_overlay_ports: 'vcpkg_ports'
       vcpkg_overlay_triplets: 'vcpkg_triplets'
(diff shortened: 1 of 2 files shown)


mbrobbel (commented):
We may also want to update
- https://github.com/sirius-db/sirius/blob/a6e701d628f0549b15847b693f6f887bdf6e58f9/.github/workflows/distribution.yml#L47
- https://github.com/sirius-db/sirius/blob/a6e701d628f0549b15847b693f6f887bdf6e58f9/.github/workflows/distribution.yml#L70

mbrobbel (approved):

PR #832, opened 2026-05-29
Make StarRocks shim CN-only

mbrobbel:
Instead of shared-nothing, this will be shared-data, so register as compute node.


dhruv9vats (approved):
We indeed want to be just be a compute node. Thanks.

9prady9 (commented):
audit (to track code-base updates for starrocks)

PR #849, opened 2026-06-02
DuckDB Native Scan: use `host_read_ranges_async_io()` and coalesce proximal ranges

kevkrist:
Previously, the native duckdb reader invoked `host_read()` APIs, which a) are synchronous and b) bypass our datasource backend in favor of direct `pread()`, which is problematic in part because the duckdb reader issues many small reads. This PR changes the reads to `host_read_ranges_async_io()` asynchronous calls into pinned allocations, coalescing file-adjacent segment reads, and issues bulk asynchronous memcpy to properly aligned device memory. 

Read path (TPC-H SF=30, q6, full 627 MB (compressed) lineitem split — average over trials):

| | read requests / split | read throughput — cold | read throughput — warm |
|---|---|---|---|
| `dev` (`host_read`, serial) | ~3,700 | 12.6 GB/s | 12.6 GB/s |
| this PR (`host_read_ranges`) | **~870** | **13.4 GB/s** | **16.3 GB/s** |

- ~4× fewer read requests
- ~1.06× throughput cold, ~1.3× warm on full splits

End-to-end (TPC-H SF=30, 22 queries, GPU native scan)

| | warm (median Σ of 22 q) |
|---|---|
| `dev` | 17.8 s |
| this PR | **14.7 s — 1.20×** |

Follow-up: `pick_gpu_memory_space_for_duckdb_native_scan` still returns `gpu_spaces[0]`; make it use the scheduler-assigned GPU so host + device placement are both correct under multi-GPU.

### DIFF
--- a/src/op/scan/duckdb_native_decoder.cpp
+++ b/src/op/scan/duckdb_native_decoder.cpp
@@ -626,14 +626,56 @@ void submit_and_await(rmm::device_buffer& device_buf,
     }
   }
 
-  // Bulk H2D: one async copy per host block (pinned -> contiguous device range).
+  // Bulk H2D: one copy per host block (pinned -> contiguous device range), issued
+  // as a single batched submission to collapse the per-block launch overhead.
   // Alignment-gap bytes between segments are copied as-is and never read by the
   // decode kernels (the device layout is identical to the staging pass).
+  std::vector<void*> h2d_dst;
+  std::vector<void const*> h2d_src;
+  std::vector<std::size_t> h2d_size;
+  std::size_t const n_blocks = (total + bsz - 1) / bsz;
+  h2d_dst.reserve(n_blocks);
+  h2d_src.reserve(n_blocks);
+  h2d_size.reserve(n_blocks);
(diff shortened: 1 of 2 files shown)


src/op/scan/duckdb_native_decoder.cpp
```diff
@@ -598,26 +502,209 @@ staged_column stage_one_varchar_column(staging_state& s,
...
-                           rmm::cuda_stream_view stream)
+using multiple_blocks_allocation =
+  cucascade::memory::fixed_size_host_memory_resource::multiple_blocks_allocation;
+
+/// @brief Batched pinned->device H2D (one cudaMemcpyBatchAsync launch); per-entry
+/// fallback on toolkits without the batch API.
+void batched_h2d(std::vector<void*> const& dst,
+                 std::vector<void const*> const& src,
+                 std::vector<std::size_t> const& size,
+                 rmm::cuda_stream_view stream)
```
aminaramoon, line 521: I think we have this implemtation in so many palces, we need to create a function for it somehwere accisible to others as well

aminaramoon (approved):

PR #864, opened 2026-06-03
ASAN / TSAN builds and core dump facilities

wmalpica:
This PR adds a ASAN and TSAN build profile, as well as instructions in a new debugging.md file for how to use them.
It also allows for disabling the segfault handler so that its easier to capture core dumps and instructions on how to do so.

### DIFF
--- a/cmake/CMakePresets.json
+++ b/cmake/CMakePresets.json
@@ -25,6 +25,14 @@
       "name": "clang-relwithdebinfo"
     },
     {
+      "configurePreset": "clang-asan",
+      "name": "clang-asan"
+    },
+    {
+      "configurePreset": "clang-tsan",
+      "name": "clang-tsan"
+    },
+    {
       "configurePreset": "legacy-release",
       "name": "legacy-release"
     },
@@ -43,14 +51,6 @@
     {
       "configurePreset": "ci-release",
       "name": "ci-release"
-    },
-    {
-      "configurePreset": "clang-asan",
-      "name": "clang-asan"
-    },
-    {
-      "configurePreset": "clang-tsan",
-      "name": "clang-tsan"
     }
   ],
   "cmakeMinimumRequired": {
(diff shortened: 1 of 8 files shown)


cmake/CMakePresets.json
```diff
@@ -139,6 +147,36 @@
...
+        "CMAKE_CUDA_FLAGS": "-Xcompiler -fsanitize=address -Xcompiler -fno-omit-frame-pointer -g",
+        "CMAKE_CXX_FLAGS": "-fsanitize=address -fno-omit-frame-pointer -g",
+        "CMAKE_C_FLAGS": "-fsanitize=address -fno-omit-frame-pointer -g",
+        "CMAKE_EXE_LINKER_FLAGS": "-fsanitize=address",
+        "CMAKE_MODULE_LINKER_FLAGS": "-fsanitize=address",
+        "CMAKE_SHARED_LINKER_FLAGS": "-fsanitize=address"
+      },
+      "displayName": "Clang ASan (RelWithDebInfo)",
+      "inherits": "base-clang",
+      "name": "clang-asan"
```
9prady9: nit: usually these entries appear first for easy spotting of what preset it is related to.

docs/super-sirius/debugging.md
```diff
@@ -0,0 +1,397 @@
...
+build/clang-asan/extension/sirius/test/cpp/sirius_unittest
+```
+
+### Run
+
+CUDA reserves huge virtual-address ranges that collide with ASan's shadow
+memory, so a couple of options are **required** or ASan will fail at startup:
+
+```bash
+ASAN_OPTIONS="protect_shadow_gap=0:detect_leaks=0:halt_on_error=0:abort_on_error=1:external_symbolizer_path=/usr/bin/llvm-symbolizer-18" \
```
9prady9: we use 21 version of clang, shouldn't this be same as that ?

there are couple of locations where we use 18 instead of 21.

docs/super-sirius/debugging.md
9prady9, line 1: there are some option-list mismatch in docs and scripts, we have to fix that. TSAN at least has this variation.

src/util/segfault_backtrace_handler.cpp
```diff
@@ -179,8 +180,38 @@ static void segfault_handler(int sig)
...
+  return ieq(v, "true") || ieq(v, "yes") || ieq(v, "on");
+}
+
 void install_segfault_backtrace_handler()
 {
+  // Escape hatch: when DISABLE_SIRIUS_SIGNAL_HANDLER is set, do NOT install the
+  // handler. The handler calls _exit(1) on a crash, which prevents the OS from
+  // writing a core dump; leaving it uninstalled lets the default disposition
+  // produce a core (with `ulimit -c unlimited`). See docs/super-sirius/debugging.md.
+  if (env_flag_enabled("DISABLE_SIRIUS_SIGNAL_HANDLER")) {
```
9prady9, line 208: we should add this to docs md file

cmake/CMakePresets.json
```diff
@@ -43,6 +43,14 @@
     {
       "configurePreset": "ci-release",
       "name": "ci-release"
+    },
+    {
+      "configurePreset": "clang-asan",
```
mbrobbel: Maybe we can add a base config with the common parts?

mbrobbel (approved):

src/util/segfault_backtrace_handler.cpp
wmalpica: its mentioned in docs/super-sirius/debugging.md

cmake/CMakePresets.json
wmalpica: it was following the pattern that was already  there, but i like having the name first. Updating

docs/super-sirius/debugging.md
wmalpica: I am not sure what you are referring to here

src/util/segfault_backtrace_handler.cpp
9prady9: but is an env variable right, hence thought it should be along with others.

docs/super-sirius/debugging.md
9prady9: the doc have options that don't reflect the ones in the scripts, at least thats what I saw IIRC

Issue #424, opened 2026-03-10
Improvements and fixes to sort operators


wmalpica:
The chain of operators is:
SORT, SORT_SAMPLE, SORT_PARTITION, and MERGE_SORT

SORT_PARTITION should be PIPELINE
Because SORT_SAMPLE is the one determining how to partition in SORT_PARTITION. So by the time SORT_SAMPLE completes a task, SORT_PARTITION should be able to pick it up immediatelly
like it should be a true pipeline. This needs to be addressed in pipeline construction.

In SORT_SAMPLE we need to add an overriden version of get_next_task_input() which will pull the number of batches its expecting. Right now its only pulling one, but you can see in some of the logic that its assuming it will try to pull DEFAULT_NUM_SAMPLE_BATCHES batches then concatenate them  sort and sample from that. 
So that number of batches of its expecting, we should make that a sirius config (operator_params), but we should  make it be based on a number of bytes (more like what the concat operator does). This way if we have really big batches and we pull 5 its not overwhelming. 
Also we should not concatenate, sort and then get samples, we should just do a sort merge, since the data is already sorted. 
 
Then we modify std::atomic<bool> _boundaries_computed{false}; to instead be an enum of _boundaries_computed_status = {NOT_DONE, SCHEDULED, DONE}. Or something like that.
And get_next_task_input, get_next_task_hint and execute functions all will depend on that status.
The idea being that the first task it gets several batches or maybe just one big one and calculates the boundaries. Then all other tasks basically become no-ops. Basically the status is so that when we call get_next_task_input() the first time, we set the status to SCHEDULED, so that the next task hint just returns WAITING_FOR_INTPUT_DATA or someting like that until its DONE

Another thing, there is this constant MAX_PARTITION_MEMORY_FRACTION which should be coming, i think from sirius_config. Lets make sure that is wired properly.

PR #866, opened 2026-06-03
Improvements to sort operators #1

Jedi18:
## Summary
Partial fix for [#424](https://github.com/sirius-db/sirius/issues/424). Addresses two items, the remaining will be done in another PR

1. Coalesce SORT_SAMPLE and SORT_PARTITION into one pipeline
   - Previously the 4-phase ORDER BY sort used four pipelines (ORDER → SORT_SAMPLE → SORT_PARTITION → MERGE_SORT). That added an unnecessary repository hop between operators that are tightly coupled—SORT_PARTITION reads boundaries directly from SORT_SAMPLE via _sample_op.

    - Now it is a 3-pipeline sort: ORDER → [SORT_SAMPLE → SORT_PARTITION] → MERGE_SORT. Sample and partition run in the same gpu_pipeline_task back-to-back on each batch.

2. Configurable sort partition memory fraction
     - Instead of hardcoding `MAX_PARTITION_MEMORY_FRACTION` (which is set to 33% and is used when the `max_sort_partition_bytes` is set to 0), this PR adds a sirius config called the `max_sort_partition_memory_fraction` which can be used to configure this percentage of available GPU memory)

### DIFF
--- a/src/pipeline/sirius_pipeline_converter.cpp
+++ b/src/pipeline/sirius_pipeline_converter.cpp
@@ -736,12 +736,9 @@ void sirius_pipeline_converter::split_order_by_sink(
   scheduled_.push_back(current_pipeline);
 
   // Create SORT_SAMPLE operator
-  auto sample_op   = duckdb::make_uniq<op::sirius_physical_sort_sample>(order_ptr);
+  auto sample_op = duckdb::make_uniq<op::sirius_physical_sort_sample>(
+    order_ptr, op_params_.max_sort_partition_bytes, op_params_.max_sort_partition_memory_fraction);
   auto* sample_ptr = sample_op.get();
-  if (op_params_.max_sort_partition_bytes > 0) {
-    sample_ptr->set_max_partition_bytes(op_params_.max_sort_partition_bytes);
-  }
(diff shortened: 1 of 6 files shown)


src/pipeline/sirius_pipeline_converter.cpp
```diff
@@ -741,12 +741,7 @@ void sirius_pipeline_converter::split_order_by_sink(
   if (op_params_.max_sort_partition_bytes > 0) {
     sample_ptr->set_max_partition_bytes(op_params_.max_sort_partition_bytes);
   }
-
-  // Pipeline B: ORDER (source) -> SORT_SAMPLE (sink)
-  auto sample_pipeline    = duckdb::make_shared_ptr<sirius_pipeline>(build_ctx_);
-  sample_pipeline->source = order_op.get();
-  sample_pipeline->sink   = sample_ptr;
-  scheduled_.push_back(sample_pipeline);
+  sample_ptr->set_max_partition_memory_fraction(op_params_.max_sort_partition_memory_fraction);
```
wmalpica: I think it would be better to set this parameter in the operator constructor. max_sort_partition_bytes should be set there as well. Most other operators follow the pattern of setting these in the constructor

wmalpica (approved):

PR #867, opened 2026-06-03
Added ARRAY data type (fixed-sized lists)

yayen-lin:


yayen-lin:
Renamed branch name so it's not confusing, closing this PR

PR #868, opened 2026-06-03
Reduce `pread` calls in DuckDB metadata walk

kevkrist:
`walk_duckdb_native_metadata` called `DataTable::GetcolumnSegmentInfo` which fetched segment-tree metadata for all columns, resulting in a large number of small `pread`s. Now it walks only projected columns via `RowGrouopCollection::GetRowGroup` -> `RowGroup::GetRawColumnData` -> `ColumnData::GetColumnSegmentInfo`. This PR also adds NVTX ranges around the walk.

**Results** (cold, SF10 TPC-H, RTX 4070 Ti, local NVMe):

| metric | before | after |
|---|---|---|
| metadata walk | 1405 ms | 681 ms (−52%) |
| metadata preads | 4581 | 1951 (−57%) |
| cold Q6 query | 1579 ms | 863 ms (−45%) |

Note that the metadata walk still dominates runtime at this small scale factor, so...
Follow-up: parallelize the per-row-group parse on the scan manager pool (prototype: ~3.5x faster).

### DIFF
--- a/src/op/scan/duckdb_native_metadata.cpp
+++ b/src/op/scan/duckdb_native_metadata.cpp
@@ -18,15 +18,18 @@
 
 #include "log/logging.hpp"
 
+#include <nvtx3/nvtx3.hpp>
+
 #include <duckdb/function/partition_stats.hpp>
 #include <duckdb/main/attached_database.hpp>
 #include <duckdb/storage/block_manager.hpp>
 #include <duckdb/storage/statistics/base_statistics.hpp>
 #include <duckdb/storage/statistics/string_stats.hpp>
 #include <duckdb/storage/storage_manager.hpp>
+#include <duckdb/storage/table/column_data.hpp>
+#include <duckdb/storage/table/row_group.hpp>
+#include <duckdb/storage/table/row_group_collection.hpp>
 #include <duckdb/storage/table_storage_info.hpp>
-
-#include <nvtx3/nvtx3.hpp>
 
 #include <algorithm>
 #include <cassert>
@@ -335,7 +338,7 @@ duckdb_native_metadata walk_duckdb_native_metadata(
 
   // PartitionStatistics order matches `RowGroupCollection::SegmentNodes()`
   // iteration order at v1.5.2 — relied on for indexing by row_group_index.
-  std::vector<duckdb::PartitionStatistics> partition_stats;
+  duckdb::vector<duckdb::PartitionStatistics> partition_stats;
   {
     nvtx3::scoped_range nvtx_ps{"sirius::native_metadata_partition_stats"};
     partition_stats = storage.GetPartitionStats(context);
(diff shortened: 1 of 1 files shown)


src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -354,7 +365,22 @@ duckdb_native_metadata walk_duckdb_native_metadata(
...
 
   duckdb::QueryContext qc{context};
-  auto column_segments = storage.GetColumnSegmentInfo(qc);
+  duckdb::vector<duckdb::ColumnSegmentInfo> column_segments;
+  {
+    nvtx3::scoped_range nvtx_si{"sirius::native_metadata_segment_info"};
+    // Read segment metadata for projected columns only.
+    auto& row_groups        = *storage.GetRowGroupCollection();
+    auto const n_row_groups = row_groups.GetRowGroupCount();
+    for (std::size_t rg = 0; rg < n_row_groups; ++rg) {
```
9prady9, line 374: nit: for loop with few lines less.
`for (int i : std::views::iota(0, limit))`

src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -302,6 +307,8 @@ duckdb_native_metadata walk_duckdb_native_metadata(
   const std::vector<projected_column>& projected_cols,
   const std::vector<sirius::logical_type>& projected_types)
 {
+  nvtx3::scoped_range nvtx_walk{"sirius::native_metadata_walk"};
```
9prady9, line 310: nvtx ranges add runtime cost, perhaps add then conditionally ? I wonder if we have a compile time flag for nvtx, might be worth checking for that.

src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -364,7 +390,7 @@ duckdb_native_metadata walk_duckdb_native_metadata(
     max_rg_idx = std::max(max_rg_idx, seg.row_group_index);
   }
   const std::size_t num_row_groups =
-    column_segments.empty() ? 0 : static_cast<std::size_t>(max_rg_idx) + 1;
+    column_segments.empty() ? handles.size() : static_cast<std::size_t>(max_rg_idx) + 1;
```
9prady9, line 393: was this a bug we fixed here ?

9prady9 (commented):
minor comments, lg

src/op/scan/duckdb_native_metadata.cpp
mbrobbel: https://github.com/NVIDIA/NVTX#how-do-i-disable-all-nvtx-calls-at-compile-time

9prady9: right but it might be better to add that via a CMake option if we don't have one right ? like `ENABLE_NVTX` etc, this can be later used for other instrumentation related toggling also.

mbrobbel: I would argue we should set these things via CMake presets and not via CMake options.

9prady9: hm yeah that also will work, but it will be nice to have an option which presets use.

Any approach is fine, they just have to be conditionally included.

mbrobbel: My thinking was; we add `NVTX_DISABLE` to presets for which we want to disable e.g. the distributed release build?

9prady9: ah so you want to control it as a whole for specific build config, that also works, we have it turned off always but remove that in certain configs. Got it

mbrobbel: Or the other way around; enabled by default, explicitly disabled for specific presets.

bwyogatama (approved):

PR #870, opened 2026-06-04
feat(expr): flip sirius::expression PIMPL to hold a Sirius AST node (#701)

### LINKED ISSUE #701: Phase 7b PR 8: flip sirius::expression PIMPL to hold sirius::ast::node
Sub-issue of #666. Depends on PR 7.

## Description

Rewire the Phase 7a `sirius::expression` PIMPL (landed in #668) to hold `std::unique_ptr<sirius::ast::node>` instead of `std::unique_ptr<duckdb::Expression>`. The wrapper's external surface stays; the 35 boundary files continue to see `sirius::expression` but now with Sirius-AST semantics under the hood.

The wrapper itself will be deleted in PR 10; this PR changes what it stores, not whether it exists.

## Scope

- `src/include/expression/expression_internal.hpp` — `struct expression::impl { std::unique_ptr<sirius::ast::node> node; };`
- `src/expression/expression.cpp` — `wrap(std::unique_ptr<duckdb::Expression>)` calls `ast::from_duckdb` internally; returns `expression{}` on translation failure.
- `unwrap(...)` signature changes to return `sirius::ast::node const*` (breaking but limited: all `unwrap()` consumers at this point are in executor/translator code that already handle Sirius AST post-PR 7).
- `release(...)` returns `std::unique_ptr<sirius::ast::node>`.
- `sirius::join_condition::left` / `.right` become AST-backed via the same wrapper.
- The 35 current boundary files: compile + test parity. Any code that was still doing `sirius::unwrap(e)->Cast<duckdb::Bound*Expression>()` has been eliminated by prior phases.

## Acceptance

- [ ] `expression::impl` stores `std::unique_ptr<sirius::ast::node>`.
- [ ] `wrap()` internally calls `ast::from_duckdb`.
- [ ] All 35 boundary files compile + test green after flip.
- [ ] `sirius_unittest` passes.
- [ ] `tpch-sirius.test` passes on release.
- [ ] Builds on both `ENABLE_LEGACY_SIRIUS=OFF` and `ON`.
- [ ] No TPC-H SF1 regression across all three strategies.

## Requirements traced
- REQ-WRAP-01 (first half)

9prady9:
## Summary

Flips `sirius::expression`'s PIMPL backing from `std::unique_ptr<duckdb::Expression>` to `std::unique_ptr<sirius::ast::node>`, keeping the public surface (`wrap`/`wrap_many`/`unwrap`/`release`/`join_condition`) intact. `wrap()` becomes the `ast::from_duckdb` translation boundary, the Phase-7 `add_join_condition` double-translation collapses to direct node use, and every `unwrap()`/`release()` consumer moves to the AST-typed accessors.

Resolves #701.

## Changes

- `expression::impl` now holds `unique_ptr<sirius::ast::node>`. `unwrap()` returns a borrowed `sirius::ast::node const*`/`*`; `release()` returns the owned `unique_ptr<sirius::ast::node>` and nulls the source.
- `wrap()` returns a null expression on null/unsupported input instead of throwing at the boundary.
- Aggregate consumers (`ungrouped_aggregate`, `aggregate_op_util`) read the native `sirius::ast::aggregate` node. `copy_expressions` deep-copies via a new `sirius::ast::clone(node const&)` helper, whose visitor has a terminal `static_assert` guarding future variant additions.
- Non-aggregate join sites recover a `duckdb::Expression` via `ast::to_duckdb` on the stored node (each binds the owned pointer to a local — no dangling temporaries).
- `case_expr`/`coalesce`/`reference` emit their recorded `return_type` instead of an INTEGER placeholder, preventing a DECIMAL→INT32 down-cast in the executor AST path's `post_process`.
- Null guards on `unwrap()` derefs in the aggregate operators (controlled fallback rather than a release-mode crash).
- Wrapper unit tests migrated from DuckDB pointer-identity assertions to AST-node-structure assertions.

## Testing

- Builds in both configs (`ENABLE_LEGACY_SIRIUS=OFF` and `ON`); no `src/legacy/` edits.
- Expression AST + wrapper suites pass (`[ast_clone] [ast_to_duckdb] [ast_aggregate] [ast_scaffold] [ast_from_duckdb] [ast_function_id]`, `[expression] [join_condition]`), plus aggregate operator suites (`[physical_grouped_aggregate]`, `[physical_grouped_aggregate_count_distinct]`, `[physical_grouped_aggregate_merge]`).

### DIFF
--- a/src/expression/ast/node.cpp
+++ b/src/expression/ast/node.cpp
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2025, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Single translation unit for sirius::ast::node member definitions. Keep every
+// node:: method here so they do not spread across per-feature .cpp files. The
+// per-alternative method bodies (e.g. reference::cudf_ast_op_count) live with
+// their own concerns; only node:: members belong in this file.
+
+#include "expression/ast/node.hpp"
+
+#include "sirius/exception.hpp"
+
+#include <cstddef>
+#include <string_view>
+#include <variant>
+
+namespace sirius::ast {
+
+namespace {
+
+[[noreturn]] void throw_wrong_node_type(std::string_view context, std::string_view expected)
+{
+  throw not_implemented_exception("{}: expected {}", context, expected);
+}
+
+}  // namespace
+
+bool node::is_reference() const noexcept { return holds<reference>(); }
+
+bool node::is_aggregate() const noexcept { return holds<aggregate>(); }
+
+bool node::is_function_call() const noexcept { return holds<function_call>(); }
+
+reference const& node::as_reference() const
+{
+  if (!is_reference()) { throw std::bad_variant_access(); }
+  return get<reference>();
+}
+
+aggregate const& node::as_aggregate() const
+{
+  if (!is_aggregate()) { throw std::bad_variant_access(); }
+  return get<aggregate>();
+}
+
+function_call const& node::as_function_call() const
+{
+  if (!is_function_call()) { throw std::bad_variant_access(); }
+  return get<function_call>();
+}
+
+std::size_t node::cudf_ast_op_count() const
+{
+  return std::visit([](auto const& alt) { return alt.cudf_ast_op_count(); }, v);
+}
+
+reference const& require_reference(node const* n, std::string_view context)
+{
+  if (n == nullptr) { throw_wrong_node_type(context, "non-null reference node"); }
+  try {
+    return n->as_reference();
+  } catch (std::bad_variant_access const&) {
+    throw_wrong_node_type(context, "reference node");
+  }
+}
+
+aggregate const& require_aggregate(node const* n, std::string_view context)
+{
+  if (n == nullptr) { throw_wrong_node_type(context, "non-null aggregate node"); }
+  try {
+    return n->as_aggregate();
+  } catch (std::bad_variant_access const&) {
+    throw_wrong_node_type(context, "aggregate node");
+  }
+}
+
+}  // namespace sirius::ast
--- a/src/expression/ast/node.cpp
+++ b/src/expression/ast/node.cpp
@@ -31,6 +31,12 @@ namespace sirius::ast {
 
 namespace {
 
+// Boundary helper for the require_* accessors below, which run during GPU operator
+// construction (under create_plan). A null or wrong-type node means the GPU path cannot
+// handle this expression shape — e.g. wrap() returned null for an unsupported aggregate, or
+// a group key that is not a bound reference. not_implemented_exception is deliberate: it is
+// caught by OnFinalizePrepare's `catch (std::exception&)` and triggers graceful CPU fallback
+// rather than signalling an internal Sirius invariant violation.
 [[noreturn]] void throw_wrong_node_type(std::string_view context, std::string_view expected)
 {
   throw not_implemented_exception("{}: expected {}", context, expected);
--- a/src/op/aggregate/aggregate_op_util.cpp
+++ b/src/op/aggregate/aggregate_op_util.cpp
@@ -19,10 +19,7 @@
 #include "cudf/cudf_utils.hpp"
 #include "duckdb/common/assert.hpp"
 #include "expression/aggregate_id.hpp"
-#include "expression/ast/aggregate.hpp"
-#include "expression/ast/function_call.hpp"
 #include "expression/ast/node.hpp"
-#include "expression/ast/reference.hpp"
 #include "expression/expression_internal.hpp"
 #include "sirius/exception.hpp"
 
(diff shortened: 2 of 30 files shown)


src/op/aggregate/aggregate_op_util.cpp
```diff
@@ -38,21 +41,30 @@ CudfAggregateDefinitions convert_duckdb_aggregates_to_cudf(
...
+    std::string const fname{sirius::to_duckdb_aggregate_name(aggr.function())};
+    auto const& children = aggr.arguments();
 
     // Handle AVG specially: it expands into SUM + COUNT_VALID
-    if (aggr.function.name == "avg") {
-      D_ASSERT(aggr.children.size() == 1);
-      D_ASSERT(aggr.children[0]->type == duckdb::ExpressionType::BOUND_REF);
-      auto& bound_ref = aggr.children[0]->Cast<duckdb::BoundReferenceExpression>();
-      auto col_idx    = static_cast<int>(bound_ref.index);
+    if (fname == "avg") {
```
mbrobbel: I know this just copies what we did before, but we should probably introduce a member function for these types of checks, what do you think?

src/op/aggregate/aggregate_op_util.cpp
```diff
@@ -62,34 +74,34 @@ CudfAggregateDefinitions convert_duckdb_aggregates_to_cudf(
...
     }
 
     // Handle COUNT(DISTINCT col) and COUNT(DISTINCT (col1, col2, ...)):
     // Use COLLECT_SET locally; merge via MERGE_SETS; then count list elements.
     // For multi-column, a struct column is synthesized from the component columns.
-    if (aggr.IsDistinct() && aggr.function.name == "count") {
-      D_ASSERT(aggr.children.size() == 1);
-      auto& child     = *aggr.children[0];
-      size_t position = result.cudf_aggregates.size();
+    if (aggr.distinct() && fname == "count") {
```
mbrobbel: Same here

src/op/aggregate/aggregate_op_util.cpp
```diff
@@ -101,42 +113,41 @@ CudfAggregateDefinitions convert_duckdb_aggregates_to_cudf(
       continue;
     }
 
-    // Convert DuckDB aggregate function name to cudf::aggregation::Kind
+    // Convert the aggregate function to a cudf::aggregation::Kind
     cudf::aggregation::Kind agg_kind;
-    if (aggr.function.name == "sum" || aggr.function.name == "sum_no_overflow") {
+    if (fname == "sum" || fname == "sum_no_overflow") {
```
mbrobbel: Didn't we also have this conversion logic elsewhere?

9prady9: the new ast code paths avoid this string and use enum, I think I missed this - have to figure out if we can cleanup here as well.

src/expression/ast/node.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <string_view>
+#include <variant>
+
+namespace sirius::ast {
+
+namespace {
+
+[[noreturn]] void throw_wrong_node_type(std::string_view context, std::string_view expected)
+{
+  throw not_implemented_exception("{}: expected {}", context, expected);
```
mbrobbel, line 42: `not_implemented_exception` confuses me here?

src/op/aggregate/aggregate_op_util.cpp
```diff
@@ -32,6 +29,21 @@
...
+{
+  switch (id) {
+    case sirius::aggregate_id::sum:
+    case sirius::aggregate_id::sum_no_overflow: return cudf::aggregation::Kind::SUM;
+    case sirius::aggregate_id::count: return cudf::aggregation::Kind::COUNT_VALID;
+    case sirius::aggregate_id::count_star: return cudf::aggregation::Kind::COUNT_ALL;
+    case sirius::aggregate_id::min: return cudf::aggregation::Kind::MIN;
+    case sirius::aggregate_id::max: return cudf::aggregation::Kind::MAX;
+    case sirius::aggregate_id::avg:
+    case sirius::aggregate_id::first: return std::nullopt;
```
mbrobbel, line 59: Is `count_distinct` missing?

src/op/aggregate/aggregate_op_util.cpp
```diff
@@ -113,40 +115,33 @@ CudfAggregateDefinitions convert_duckdb_aggregates_to_cudf(
...
       // COUNT(*) has no children - use 0 as a placeholder (will be handled by COUNT_ALL)
-      if (fname == "count_star") {
+      if (fid == sirius::aggregate_id::count_star) {
         result.cudf_aggregate_idx.push_back(0);
       } else {
-        throw std::runtime_error("Unsupported aggregate function: " + fname + " with no children");
+        throw std::runtime_error(std::string{"Unsupported aggregate function: "} +
+                                 std::string{sirius::to_duckdb_aggregate_name(fid)} +
+                                 " with no children");
       }
```
mbrobbel: Maybe create a function that attempts to match a `fid` to an `aggregate_id` and throw this error there instead of repeating it

src/expression/ast/node.cpp
9prady9: ? I don't get the confusion part

mbrobbel: What is not implemented here?

src/op/aggregate/aggregate_op_util.cpp
9prady9: no its the same exact set being handled in dev, I added a note of why so in the header now.

mbrobbel: https://github.com/sirius-db/sirius/pull/870/changes#diff-e67898eeef83efcf49e466cdd63b1b8ec48ee65ba508ca8c688c20a8d55c3810R35 seems to suggest a special case for nullopt for count distinct, but maybe I'm reading that wrong

src/expression/ast/node.cpp
9prady9: hm, its not necessarily that as you are seeing - well it depends on few things - added a note for clarification.

src/op/aggregate/aggregate_op_util.cpp
9prady9: Added a note why that is the case at this point of execution - https://github.com/sirius-db/sirius/pull/870/commits/cafa6aa85eba27d7b5dc96c94e97fdfe8cd89342

mbrobbel (approved):

PR #872, opened 2026-06-04
Read nested parquet (struct/list/map) on GPU + transparent s3:// scan

ran-yuan-rui:
## Summary

This PR adds Phase-1 nested Parquet support to the Sirius GPU parquet scan path.
The main goal is to let Sirius read and return common lakehouse nested columns from Parquet on the GPU path. This covers projecting top-level `STRUCT`, `LIST` / array, and `MAP` columns, including nested combinations such as struct-of-list, list-of-struct, nested struct, list-of-list, and MAP-as-list-of-struct layouts. The same scan/result materialization path works for both local Parquet files and S3 Parquet files.
This PR intentionally supports **nested projection**, not full nested expression semantics. Filtering on nested columns, joining on nested columns, grouping/aggregating by nested columns, subfield projection such as `payload.user_id` or `items[1]`, and `UNNEST` remain out of scope for this phase. Those require planner, expression executor, and operator semantics over nested values, while this PR focuses on decoding nested Parquet layout and materializing DuckDB-compatible nested results. Unsupported nested operations return clear errors instead of silently falling back or producing partial semantics.
As a supporting change, this PR also registers a minimal Sirius-owned `s3://` DuckDB FileSystem. This is **not** a CPU fallback path. It exists so DuckDB's native `read_parquet('s3://...')` can bind the file and read the Parquet footer through Sirius's existing `s3_ioctx`, without loading DuckDB `httpfs` and without exposing `sirius_read_parquet` as the public surface.

The public S3 SQL surface becomes:

SET gpu_execution = true;
SELECT ... FROM read_parquet('s3://bucket/file.parquet');

### S3 is GPU-only — three guards prevent a silent CPU fallback
S3 data has no CPU path, so every route that could quietly read `s3://` on the CPU is closed:
- The FileSystem refuses to open `s3://` when `gpu_execution` is off (there would be no
  GPU consumer).
- The FileSystem refuses to open `s3://` while an internal CPU-replay query is active
  (covers the indirect case where a failed GPU query is replayed on the CPU, including
  through a view).
- `OnFinalizePrepare` inspects the bound plan; if it reads `s3://` and GPU translation
  fails, it raises a clear "S3 CPU fallback is not supported" error instead of handing the
  plan back to DuckDB's CPU engine.

## Nested parquet read/project

- Recursive parquet-group → DuckDB `STRUCT` / `LIST` / `MAP` schema mapping at bind time.
- Nested cuDF column → DuckDB `Vector` materialization (list offset slicing, validity
  bit-packing, recursive child handling).
- Scan planning relaxed to allow nested projected columns; operators that cannot yet
  operate *on* nested values reject them explicitly rather than producing wrong results.

## Out of scope (follow-ups)

- Operating **on** nested values — filter / join / group-by over STRUCT/LIST/MAP — and
  sub-field projection pushdown (cuDF-limited). This PR covers read + project only.
- No `httpfs`, no S3 write path (`COPY … TO 's3://'` is rejected), no S3 `LIST`/glob.

## Notes for reviewers

- The `sirius_s3_filesystem` is intentionally bind-only and read-only: it backs the footer read for the GPU path and never serves a CPU scan. The three guards above are the load-bearing part of the "S3 is GPU-only, no CPU fallback" contract.


kevkrist:
@ran-yuan-rui are you sure a lot of this isn't already supported in the sirius parquet scan operator?
See #663

ran-yuan-rui:
> @ran-yuan-rui are you sure a lot of this isn't already supported in the sirius parquet scan operator? See #663

Good question — checked, and no, it's not. The GPU scan's cuDF→DuckDB type mapping had no STRUCT/LIST/MAP cases, so a nested column hit an "unsupported type" throw. Locally that just falls back to DuckDB's CPU read_parquet (so it looks supported), but on S3 — GPU-only, no CPU fallback — nested files were unreadable.

This PR adds the missing bits so the GPU scan reads + projects nested directly. Operating on nested (filter/join/group-by) is still rejected on purpose — that's cuDF-gated follow-up, not this PR.

ran-yuan-rui:
will need redesign based on new IO framework, close 872

PR #876, opened 2026-06-04
Improvements to sort operator #2

Jedi18:
## Summary
Second PR for https://github.com/sirius-db/sirius/issues/424. After this there is one more issue left that will be addressed in the third PR.

This PR improves `SORT_SAMPLE` (partition boundary sampling for multi-partition sorts) in three related changes.

### 1. Override `get_next_task_input_data()`
**Problem:** There was a mismatch between `get_next_task_hint()` and `get_next_task_input_data()`. `get_next_task_hint()` waited until enough batches were available in the input repo before signaling READY, but `get_next_task_input_data()` still only pulled a single batch per task. So even when the hint said the operator was ready to compute boundaries, `execute()` would only ever see one batch at a time — not the multi-batch sample the rest of the logic assumed.
**Change:**
- Add an overridden `get_next_task_input_data()` on `sirius_physical_sort_sample` so it aligns with what `get_next_task_hint()` is waiting for.
- Before boundary computation: pull batches from the input repo until the sampling target is reached, then hand them to `execute()` as a single multi-batch input.
- After boundaries are computed: fall back to the default behavior and process one batch at a time (passthrough mode).
### 2. Byte-based sampling via `sort_sample_bytes ` (replaces fixed batch count)
**Problem:** The old fixed batch count (`DEFAULT_NUM_SAMPLE_BATCHES = 5`) doesn't scale with variable batch sizes. With very large batches, pulling 5 can be overwhelming; with small batches, 5 may not be enough.
**Change:**
- Replace `num_sample_batches` with `sort_sample_bytes` in `operator_params` (default **512 MB**, similar to how the concat operator sizes its output).
- Expose `sort_sample_bytes` as a DuckDB extension option and YAML config.
- Update `get_next_task_hint()` to wait until enough sample bytes are available (or upstream finishes with remaining data).
- Wire the config through `sirius_pipeline_converter` when building `SORT_SAMPLE`.
- Add unit tests for byte threshold, hint behavior, and upstream-finished below threshold.
### 3. Merge pre-sorted batches (replaces concatenate + full sort)
**Problem:** Upstream `ORDER_BY` already produces locally sorted batches, but `SORT_SAMPLE` was concatenating the sample, running a full GPU sort, and only then computing partition boundaries — redundant work.
**Change:**
- Use `gpu_merge_impl::merge_order_by` to merge pre-sorted sample batches, then compute boundaries from the merged result.
- Remove the `cudf::concatenate` + full-sort path for multi-batch samples.
- Add a unit test verifying correct partition boundaries from merged pre-sorted runs.
- Update `operators.md` to describe merge-based sampling.

### DIFF
--- a/src/op/sirius_physical_sort_sample.cpp
+++ b/src/op/sirius_physical_sort_sample.cpp
@@ -46,8 +46,10 @@ bool repo_has_enough_sample_bytes(::cucascade::shared_data_repository* repo,
                                   uint64_t sort_sample_bytes,
                                   bool upstream_finished)
 {
-  if (!repo || repo->total_size() == 0) { return false; }
+  if (!repo) { return false; }
+  // Upstream done with no rows still needs a READY signal so the pipeline can drain.
   if (upstream_finished) { return true; }
+  if (repo->total_size() == 0) { return false; }
 
   uint64_t accumulated = 0;
   for (size_t part = 0; part < repo->num_partitions(); ++part) {
--- a/src/op/sirius_physical_sort_sample.cpp
+++ b/src/op/sirius_physical_sort_sample.cpp
@@ -33,13 +33,45 @@
 namespace sirius {
 namespace op {
 
+namespace {
+
+uint64_t get_batch_bytes(const std::shared_ptr<::cucascade::data_batch>& batch)
+{
+  if (!batch) { return 0; }
+  auto ro = batch->to_read_only();
+  if (!ro.get_data()) { return 0; }
+  return ro.get_data()->get_size_in_bytes();
+}
+
+bool repo_has_enough_sample_bytes(::cucascade::shared_data_repository* repo,
+                                  uint64_t sort_sample_bytes,
+                                  bool upstream_finished)
+{
+  if (!repo || repo->total_size() == 0) { return false; }
+  if (upstream_finished) { return true; }
+
+  uint64_t accumulated = 0;
+  for (size_t part = 0; part < repo->num_partitions(); ++part) {
+    for (auto batch_id : repo->get_batch_ids(part)) {
+      auto batch = repo->get_data_batch_by_id(batch_id, part);
+      if (!batch) { continue; }
+      accumulated += get_batch_bytes(batch);
+      if (accumulated >= sort_sample_bytes) { return true; }
+    }
+  }
+  return false;
+}
+
+}  // namespace
+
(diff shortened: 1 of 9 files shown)


src/op/sirius_physical_sort_sample.cpp
```diff
@@ -90,6 +119,36 @@ std::optional<task_creation_hint> sirius_physical_sort_sample::get_next_task_hin
   return std::nullopt;
 }
 
+std::unique_ptr<operator_data> sirius_physical_sort_sample::get_next_task_input_data()
+{
+  // After boundaries are computed, process one batch at a time (passthrough mode).
+  if (_boundary_state.load(std::memory_order_acquire) == 2) {
```
wmalpica, line 127: Can you make _boundary_state and enum. That way this is easier to interpret

src/op/sirius_physical_sort_sample.cpp
```diff
@@ -21,25 +21,56 @@
...
+  if (!ro.get_data()) { return 0; }
+  return ro.get_data()->get_size_in_bytes();
+}
+
+bool repo_has_enough_sample_bytes(::cucascade::shared_data_repository* repo,
+                                  uint64_t sort_sample_bytes,
+                                  bool upstream_finished)
+{
+  if (!repo || repo->total_size() == 0) { return false; }
+  if (upstream_finished) { return true; }
```
wmalpica, line 51: This logic seems to follow well what used to be here, but if the upstream is finished but it has no data, what happens? we want to make sure it does not deadlock. Can you test this out to see what happens?
Like run a query like:
select columnA, colB, from table where [some filter which filters out everything, but  not something we can be computed from a simple min max] order by columnA, colB

wmalpica (changes requested):

src/op/sirius_physical_sort_sample.cpp
Jedi18: Thanks, I tried out a query and there was no deadlock. Regardless we should account for the case where there are no rows and upstream is finished.

I moved the `repo->total_size() == 0` check to below upstream_finished check so we return true in the case of upstream finished and no data. Also added a unit test

src/op/sirius_physical_sort_sample.cpp
Jedi18: Yes I am making that change in a followup PR (along with some related refactor to see if we can avoid redundant boundary tasks)

wmalpica (approved):

PR #886, opened 2026-06-05
Improvements to sort operator #3

Jedi18:
## Summary

Refactors `sort_sample` boundary scheduling to use an explicit state machine instead of a magic `atomic<int>` and a CAS election in `execute()`.

**Before:** The task creator could schedule multiple boundary tasks in parallel. `execute()` used a CAS (`0 → 1`) to elect one winner; losers passthrough'd, wasting GPU work when boundaries weren't ready yet.

**After:** Boundary lifecycle is `NOT_DONE → SCHEDULED → DONE`:
- `get_next_task_input_data()` claims the sample input and moves to `SCHEDULED` (at most one boundary task)
- `get_next_task_hint()` returns `nullopt` while `SCHEDULED`, blocking duplicate boundary tasks
- `execute()` computes boundaries and moves to `DONE`; OOM resets to `NOT_DONE` for retry

Remaining passthrough batches after boundary computation rely on existing scheduling: when merge is scheduled downstream, its FULL barrier on pipeline B is unfinished, so `get_operator_for_next_task()` walks back to `sort_sample` — no follow-up hook needed.

### DIFF
--- a/src/op/sirius_physical_sort_sample.cpp
+++ b/src/op/sirius_physical_sort_sample.cpp
@@ -96,12 +96,15 @@ sirius_physical_sort_sample::sirius_physical_sort_sample(
 
 std::optional<task_creation_hint> sirius_physical_sort_sample::get_next_task_hint()
 {
-  // If boundaries already computed, use default behavior (process batches as they arrive)
-  if (_boundary_state.load(std::memory_order_acquire) == 2) {
-    return sirius_physical_operator::get_next_task_hint();
-  }
+  const auto state = _boundary_state.load(std::memory_order_acquire);
 
-  // Need to wait for enough sample bytes before computing boundaries
+  // Boundaries already computed — process batches one at a time.
+  if (state == BoundaryState::DONE) { return sirius_physical_operator::get_next_task_hint(); }
+
+  // Boundary task already scheduled; wait for it to finish before creating more tasks.
+  if (state == BoundaryState::SCHEDULED) { return std::nullopt; }
+
+  // NOT_DONE: wait for enough sample bytes before scheduling the boundary task.
   auto port_ids = get_port_ids();
   if (port_ids.empty()) { return std::nullopt; }
 
@@ -123,14 +126,16 @@ std::optional<task_creation_hint> sirius_physical_sort_sample::get_next_task_hin
 
 std::unique_ptr<operator_data> sirius_physical_sort_sample::get_next_task_input_data()
 {
-  // After boundaries are computed, process one batch at a time (passthrough mode).
-  if (_boundary_state.load(std::memory_order_acquire) == 2) {
-    return sirius_physical_operator::get_next_task_input_data();
-  }
+  const auto state = _boundary_state.load(std::memory_order_acquire);
 
-  // Before boundary computation: accumulate batches up to the sample byte threshold so
-  // execute() can merge pre-sorted runs and derive partition boundaries from a representative
-  // sample without pulling a fixed batch count that may be too large for big batches.
+  // After boundaries are computed, process one batch at a time (passthrough mode).
+  if (state == BoundaryState::DONE) { return sirius_physical_operator::get_next_task_input_data(); }
+
+  // Boundary input already claimed by the in-flight boundary task.
+  if (state == BoundaryState::SCHEDULED) { return nullptr; }
+
+  // NOT_DONE: accumulate batches up to the sample byte threshold so execute() can merge
+  // pre-sorted runs and derive partition boundaries from a representative sample.
   auto port_ids = get_port_ids();
   if (port_ids.empty()) { return nullptr; }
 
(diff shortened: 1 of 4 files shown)


src/op/sirius_physical_sort_sample.cpp
```diff
@@ -159,17 +166,23 @@ std::unique_ptr<operator_data> sirius_physical_sort_sample::execute(const operat
...
-  // need the boundaries themselves; sort_partition runs next in the same pipeline task.
-  int expected = 0;
-  if (!_boundary_state.compare_exchange_strong(expected, 1, std::memory_order_acq_rel)) {
+  const auto state = _boundary_state.load(std::memory_order_acquire);
+  if (state == BoundaryState::SCHEDULED) {
+    // Normal path: input was claimed by get_next_task_input_data().
+  } else if (state == BoundaryState::NOT_DONE) {
+    // OOM retry or direct unit-test invoke — claim boundary computation.
+    auto expected = BoundaryState::NOT_DONE;
+    if (!_boundary_state.compare_exchange_strong(
```
wmalpica: I dont understand when state == BoundaryState::NOT_DONE here would ever be valid.  If anything it seems like this should throw an error. Can you explain when this would be ok?

Jedi18: So this is for the case where we get an OOM during the boundary calculation. 

1) We reset the state to NOT_DONE and rethrow.
https://github.com/sirius-db/sirius/blob/f746b15df33da74265fd8dfa2ba6bfe1a3847825/src/op/sirius_physical_sort_sample.cpp#L335

2) Pipeline catches OOM and builds a reschedule exception with the same input + operator index
https://github.com/sirius-db/sirius/blob/f746b15df33da74265fd8dfa2ba6bfe1a3847825/src/pipeline/gpu_pipeline_task.cpp#L315

3) Executor creates a new task with that input, resuming at sort_sample without going through the task creator
https://github.com/sirius-db/sirius/blob/f746b15df33da74265fd8dfa2ba6bfe1a3847825/src/pipeline/gpu_pipeline_executor.cpp#L309
https://github.com/sirius-db/sirius/blob/f746b15df33da74265fd8dfa2ba6bfe1a3847825/src/pipeline/gpu_pipeline_executor.cpp#L328

Now we hit the line you mentioned.

So on retry, get_next_task_hint() and get_next_task_input_data() are not called again for that task, instead the task is scheduled again with the same input.

To verify this, I created a fake oom and logged warnings to see if this path is being taken
https://github.com/Jedi18/sirius/commit/cf55d93f1fe6a37aabb72f037725c1bc79b0885b

```
[2026-06-07 10:01:54.210] [warning] [sirius_physical_sort_sample.cpp:220] TEMP DEBUG: throwing fake rmm::out_of_memory — expect OOM reschedule then NOT_DONE retry
[2026-06-07 10:01:54.211] [warning] [gpu_pipeline_task.cpp:284] Pipeline 7: OOM at operator SORT_SAMPLE (id=17, index 0/2), requested 0 bytes (0.00 MB), global usage 0 bytes (0.00 MB), peak allocated 0 bytes (0.00 MB), bytes to materialize input 0 bytes (0.00 MB), reservation 391579840 bytes (373.44 MB), rescheduling task 7
[2026-06-07 10:01:54.211] [warning] [gpu_pipeline_executor.cpp:291] GPU Pipeline Executor: OOM reschedule (retry 1/100) for task 7 (original task 7), resuming from operator index 0
[2026-06-07 10:01:54.261] [warning] [sirius_physical_sort_sample.cpp:182] TEMP DEBUG: sort_sample execute() NOT_DONE path (OOM retry reschedule, batches=1)
```

wmalpica: i would suggest that if there is an OOM during the boundary calculation, then you leave it as SCHEDULED. This would work, right? then you dont need that extra check. And you dont run the risk of another task creation assuming that its not scheduled and doing its own thing

Jedi18: Thanks yes leaving it at SCHEDULED would be better, I've made the change

src/op/sirius_physical_sort_sample.cpp
```diff
@@ -171,21 +171,12 @@ std::unique_ptr<operator_data> sirius_physical_sort_sample::execute(const operat
...
-    if (!_boundary_state.compare_exchange_strong(
-          expected, BoundaryState::SCHEDULED, std::memory_order_acq_rel)) {
-      SIRIUS_LOG_DEBUG("Sort sample: passthrough ({} batches)", input_batches.size());
-      return std::make_unique<pipelineable_operator_data>(input.get_read_only_batches());
-    }
-  } else {
-    SIRIUS_LOG_DEBUG("Sort sample: passthrough ({} batches)", input_batches.size());
-    return std::make_unique<pipelineable_operator_data>(input.get_read_only_batches());
-  }
+  // Otherwise this is the boundary-computation task. get_next_task_input_data() already moved the
```
wmalpica: This sort of comment only makes sense from the perspective of the creation of this PR. Its indicating why some code was removed, but without the code and the history, this commentary is confusing. Always go over the comments the AI adds and remove what is not useful past the life of the PR itself

wmalpica (changes requested):

src/op/sirius_physical_sort_sample.cpp
Jedi18: Ok I will keep that in mind thanks, I'll remove this comment

wmalpica (approved):

PR #871, opened 2026-06-04
refactor: unify GPU scan operators behind gpu_ingestible abstraction

bwyogatama:
## Summary

Collapses `sirius_gpu_parquet_scan_operator` + `sirius_gpu_duckdb_native_scan_operator` into a single `sirius_gpu_scan_operator` parameterized by a pluggable `io::gpu_ingestible` interface. Adding a new data format now means writing one header/source pair (`<format>_gpu_ingestible.{hpp,cpp}`) instead of duplicating the operator + split_provider + scan_info triple per format.

Three concrete implementations land in this PR — parquet, duckdb-native, cached-parquet — wired through the same operator + split_provider. All legacy code has been deleted (no dead code left in the tree).

## Architecture

- **`io::gpu_ingestible`** — standalone abstract class (NOT a `split_provider` subclass; composition over inheritance). `has_more_splits` / `next_split_provider` drive metadata-worker enumeration; `materialize_table` / `post_filter_and_project` produce the per-task `cudf::table`.
- **`scan_manager::split_provider`** — single concrete class composing `shared_ptr<io::gpu_ingestible>`. Delegates the two virtuals to the ingestible. `run<Scheduler>` template + `worker_state` RAII unchanged from before.
- **`sirius_gpu_scan_operator`** — type `GPU_SCAN`. `take_table_info` / `install_ingestible` let `scan_manager` build the ingestible during `prepare_for_query` (peeks `file_paths()` for a pinned-cache match; on hit builds `cached_parquet_gpu_ingestible` with the stolen `table_info`; on miss dispatches through `io::make_gpu_ingestible`).
- **`sirius_scan_manager::prepare_for_query`** — single op-map keyed on `sirius_gpu_scan_operator*`; one `start_metadata_processing` loop.

## Concrete ingestibles

- `parquet_gpu_ingestible` — ports `parquet_split_provider::run_batch` + the read/filter portion of `sirius_gpu_parquet_scan_operator::read_table_from_metadata`. `materialize_table` handles filter pushdown + post-decode fallback; `post_filter_and_project` does hive-partition assembly only.
- `duckdb_native_gpu_ingestible` — uses `duckdb_native_split_payload` (top-level struct in `op::scan` namespace, promoted out of the deleted nested `split_payload`). Pre-builds the coalesced filter expression once in the ctor.
[...]

PR #889, opened 2026-06-05
bugfix - single-GPU S3 scan routing regression introduced by the KvikIO fallback path (PR 808)

ran-yuan-rui:
## Summary

Fixes the single-GPU S3 scan routing regression introduced by the KvikIO fallback path.

When `_gpu_ioctxs` is empty, local parquet scans should fall back to cudf/KvikIO. However, S3 row-group slices can still carry a non-null `slice.io_ctx` even in that mode. The previous condition routed all slices through cudf when `_gpu_ioctxs` was empty, which made `s3://...` scans fail with KvikIO's unsupported URL scheme error.

This PR changes the fallback decision to use cudf only when `slice.io_ctx == nullptr`. Any slice that was minted with a Sirius backend continues to read through `slice.io_ctx->make_datasource(...)`.

## Behavior

- Local parquet + single-GPU fallback:
  - `slice.io_ctx == nullptr`
  - still uses cudf/KvikIO
- S3 parquet + single-GPU fallback:
  - `slice.io_ctx != nullptr`
  - now correctly uses the S3-backed Sirius datasource

## Tests

Adds compiled Catch2 guards in `test_parquet_scan_task.cpp`:

- non-null `slice.io_ctx` with an `s3://` path must call the backend `make_datasource`
- null `slice.io_ctx` with a local path still uses the cudf/KvikIO fallback

The guard was intentionally placed in `test_parquet_scan_task.cpp` because the older `test_metadata_gpu_scan_operators.cpp` is no longer part of `TEST_SOURCES` and cannot be revived as-is.

### DIFF
--- a/src/op/scan/sirius_gpu_parquet_scan_operator.cpp
+++ b/src/op/scan/sirius_gpu_parquet_scan_operator.cpp
@@ -159,10 +159,16 @@ std::unique_ptr<cudf::table> sirius_gpu_parquet_scan_operator::read_table_from_m
   // use_sirius_datasource=true whenever >1 GPU is configured).
   //
   // When use_sirius_datasource=false (single-GPU only), the scan_manager
-  // hands the operator an empty _gpu_ioctxs map and the provider emits
-  // slices with null io_ctx / io_object. In that case we use
-  // cudf::io::datasource::create for every slice — kvikio's per-FileHandle
-  // CUDA-context binding is harmless with only one GPU in play.
+  // hands the operator an empty _gpu_ioctxs map and the provider emits LOCAL
+  // slices with null io_ctx / io_object; those use cudf::io::datasource::create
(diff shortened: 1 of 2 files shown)


wmalpica (approved):

bwyogatama:
Freeze merging all PRs touching the parquet scan operator, duckdb scan operator, parquet split provider, and duckdb split provider since those will be deleted in #871

aminaramoon (approved):

ran-yuan-rui:
> Freeze merging all PRs touching the parquet scan operator, duckdb scan operator, parquet split provider, and duckdb split provider since those will be deleted in #871

I suggest we merge 889 into 871 as I saw the same bug still on 871

bwyogatama:
I'm going to merge this PR into #871

bwyogatama:
Merged into #871

Issue #888, opened 2026-06-05
CUDA 12 distribution jobs are failing


mbrobbel:
This seems to be regression that started after https://github.com/sirius-db/sirius/commit/a95bd1277f6527388c41350a56244704b70a42d2. 

```FAILED: extension/sirius/CMakeFiles/sirius_extension.dir/src/op/scan/duckdb_native_decoder.cpp.o 
[...]

PR #892, opened 2026-06-05
Always trigger `distribution` check on PRs

mbrobbel:
To support setting `distribution` as `required` check (to avoid #888) we need to also run it on PRs to unblock them (because the check is required - it's not right now but I'll set it as required when this merges).

### DIFF
--- a/.github/workflows/distribution.yml
+++ b/.github/workflows/distribution.yml
@@ -19,70 +19,67 @@ on:
       - dev
   merge_group:
   pull_request:
-    paths:
-      - .github/workflows/distribution.yml
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
   cuda-13:
+    if: github.event_name != 'pull_request'
     uses: sirius-db/extension-ci-tools/.github/workflows/_extension_distribution.yml@sirius
     with:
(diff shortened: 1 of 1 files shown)


.github/workflows/distribution.yml
```diff
@@ -19,70 +19,67 @@ on:
       - dev
   merge_group:
   pull_request:
-    paths:
-      - .github/workflows/distribution.yml
```
mike-wendt, line 23: Don't we want to keep this for testing changes to the workflow?

mike-wendt (approved):

.github/workflows/distribution.yml
mbrobbel: Ideally yes, but we would need to set a flag that we can use as skip condition for the actual jobs. That requires another job that runs before the other jobs (that sets a `bool` based on the changed paths).

My thinking was; this workflow is relatively straightforward, and if we are working on it in a PR we can easily comment the `if` condition for the jobs we want to test.

If you don't agree, I can add the job that sets the condition based on the path filter.

mike-wendt: This works for me. Thanks

mike-wendt:
Updated to make this required now that it is merged

PR #893, opened 2026-06-05
Make GPU-native DuckDB scan the default

kevkrist:
Turns on the GPU-native duckdb scan by default so seq_scan over duckdb tables runs on the GPU without a flag. Falls back to the duckdb_scan_task + scan_executor code for in-memory tables and anything the native path can't read, so nothing should break (this applies particularly to transparent execution, which currently uses in-memory database). Also points the TPC-H rigging at the duckdb native scan.


kevkrist:
Closing as solved by #913

PR #895, opened 2026-06-05
Parallelize the duckdb-native scan metadata walk

kevkrist:
Parallelizes the DuckDB-native scan's per-row-group metadata walk, which
previously ran serially in the split-provider constructor and dominated cold
query runtime (~95% at TPC-H SF30, where the walk issues roughly one cold block
read per (row group × projected column) via `GetColumnSegmentInfo`).

## Design

The walk is split into a cheap serial pre-step and a parallel per-range walk,
and the scan operator coalesces the parsed ranges into cap-sized batches that
decode while the remaining ranges are still being parsed.

```
 prepare_for_query        ── driver thread (serial, cheap) ───────────────────────────
   duckdb_native_split_provider ctor
     • prepare_duckdb_native_walk()   partition stats · type gate · row-group count   (no segment IO)
     • slice [0..N) into ranges of `chunk` row groups   (SIRIUS_METADATA_PARSE_CHUNK, default 8)
   start_metadata_processing → split_provider::run(pool, connector)   (enqueues N thunks, returns fast)

 ── scan-worker pool (parallel) ──────────────────────┐      split_connector  (mutex + cv FIFO)
   thunk(range 0): walk_..._row_group_range(plan,0,k) │push─┐
   thunk(range 1): walk_..._row_group_range(plan,k,2k)│push─┤   pushed as ranges finish
   thunk(range 2): ...                                │push─┤   the cold GetColumnSegmentInfo reads
        each walks its rows × cols; latencies overlap │     │   for different ranges run concurrently
   ─────────────────────────────────────────────────────┘   ▼
                                                     [ split_payload · split_payload · ... ]
                                                                  │  get_next_split()
                                                                  ▼
 ── scan operator: get_next_task_input_data() ── single consumer ────────────────────
   batch_coalescer  ── push each range's row groups, pack to cap ──┐
     • a batch fills → emit it immediately ───────────────────────┼──► task → decode_duckdb_native_split
       (decode runs while later ranges are still parsing)         │      (pipeline executor + GPU)
     • emits exactly one undersized tail batch per scan           │
   ───────────────────────────────────────────────────────────────┘
```

Components:
- **`prepare_duckdb_native_walk`** — serial pre-step run in the provider ctor:
  partition statistics, projected-type viability, row-group count, block size. No
  per-segment IO, so unsupported types still fall back to DuckDB CPU up front.
- **`walk_duckdb_native_row_group_range`** — the IO-latency-bound segment walk for
  a disjoint row-group range, run inside each `next_split_provider` thunk on the
  scan-worker pool. Reads only immutable post-open segment trees with a per-worker
  `QueryContext`, so disjoint ranges parse concurrently and their cold reads overlap.
- **`batch_coalescer`** (new) — packs parsed row groups into batches bounded by
  `approximate_batch_size` (decoded bytes) and the cudf int32 strings limit
  (per varchar column), emitting one tail batch per scan. The scan operator feeds
  ranges in as they arrive and emits each batch as soon as it fills, so decode
  overlaps the remaining parse. Packing order is irrelevant (rowids are absolute
  per row group).

The provider ctor requires a non-null `io_ctx` and non-empty `db_path`; the old
monolithic `walk_duckdb_native_metadata` is removed and its viability checks now
live in the two new functions.

## Performance — TPC-H SF30 (cold = page cache dropped)

| query | cold (serial walk) | cold (parallel) | speedup |
|---|---|---|---|
| q6  | 2.45 s | 0.63 s | ~3.9× |
| q19 | 3.37 s | 0.98 s | ~3.4× |
| q9  | 4.61 s | 1.55 s | ~3.0× |
| q1  | 3.99 s | 1.95 s | ~2.0× |
| q21 | 3.46 s | 2.61 s | ~1.3× |

Roughly **1.3–4× per query** across all 22 (highest on lineitem-scan-dominated
queries, lowest on join/small-table ones). Warm-cache queries are unaffected.

## Tunables
- `SIRIUS_METADATA_PARSE_CHUNK` — row groups per parse range (default 8).
- `sirius.executor.scan_manager.num_threads` — scan-worker pool size (default 8;
  the metadata reads are IO-latency-bound, so pool = core count is optimal).

## Validation
- Unit tests: `batch_coalescer`, the metadata walker, and the split provider.
- 22-query TPC-H harness: 66/66 runs, 0 fallback / 0 errors.
- GPU results match DuckDB CPU on fixed-width, varchar, and aggregate queries.

### DIFF
--- a/src/op/scan/duckdb_native_gpu_ingestible.cpp
+++ b/src/op/scan/duckdb_native_gpu_ingestible.cpp
@@ -17,14 +17,17 @@
 // sirius
 #include <expression/ast/from_duckdb.hpp>
 #include <expression_executor/gpu_expression_executor.hpp>
+#include <helper/utils.hpp>
 #include <io/io_context.hpp>
 #include <io/sirius_datasource.hpp>
 #include <log/logging.hpp>
+#include <op/scan/duckdb_native_batch_coalescer.hpp>
 #include <op/scan/duckdb_native_decoder.hpp>
 #include <op/scan/duckdb_native_gpu_ingestible.hpp>
 #include <op/scan/scan_utils.hpp>
 #include <op/scan/sirius_gpu_scan_operator_data.hpp>
 #include <scan_manager/sirius_scan_manager.hpp>
+#include <scan_manager/split_connector.hpp>
 
 // cudf
 #include <cudf/table/table.hpp>
@@ -35,8 +38,11 @@
 
 // standard library
 #include <algorithm>
+#include <cstddef>
+#include <cstdlib>
 #include <memory>
 #include <stdexcept>
+#include <string>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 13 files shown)


src/op/scan/duckdb_native_gpu_ingestible.cpp
```diff
@@ -214,50 +172,114 @@ duckdb_native_gpu_ingestible::~duckdb_native_gpu_ingestible() = default;
...
+  return std::make_unique<scan_operator_input>(std::move(metadata));
+}
+
+std::unique_ptr<op::operator_data> duckdb_native_gpu_ingestible::consume_next_input(
+  scan_manager::split_connector& connector)
+{
+  // The first consumer call drains every parsed range so segment byte sizes can
+  // be finalized with table-wide visibility. Subsequent calls serve already
+  // coalesced batches.
+  finalize_metadata(connector);
```
bwyogatama: Isn't this draining the connector completely before any batch can be returned? Wouldn't this mean that the metadata parsing and the decoding is not asynchronous anymore

src/op/scan/duckdb_native_gpu_ingestible.cpp
```diff
@@ -214,50 +172,114 @@ duckdb_native_gpu_ingestible::~duckdb_native_gpu_ingestible() = default;
...
+  if (apply_filter || _projection_required) {
+    auto pf          = std::make_unique<duckdb_native_post_filter_and_projection_info>();
+    pf->apply_filter = apply_filter;
+    pf->output_arity = _projection_required ? _output_arity : 0;
+    filter_info      = std::move(pf);
+  }
+
+  auto metadata =
+    std::make_unique<io::scan_and_filter_metadata>(std::move(split_info), std::move(filter_info));
+  return std::make_unique<scan_operator_input>(std::move(metadata));
```
bwyogatama, line 230: Don't we need to also pass the preferred device id? For multi-gpu scan

bwyogatama (approved):

src/op/scan/duckdb_native_gpu_ingestible.cpp
kevkrist: Deferring to another PR. Will update here with ref when it's started.

src/op/scan/duckdb_native_gpu_ingestible.cpp
kevkrist: You're right -- It's back to the streaming consumer: each parsed range is coalesced and handed to decode as it arrives, overlapping the remaining walks.

Why it was there: ColumnSegmentInfo doesn't expose the compressed segment size, so we infer it from the block_offset delta to the next segment in the same block (last-in-block → block end). Per range, a segment whose in-block neighbor falls in another range is sized to the block end — a slight *over*-estimate.

Issue #904, opened 2026-06-09
Add integration tests in the merge queue


ran-yuan-rui:
### Use case and motivation

https://github.com/sirius-db/sirius/pull/889 this bug could detect on single-GPU mode + S3 end-to-end through MinIO. 
Our default test tags don't run the MinIO S3 e2e suite because it is slower and consumes CI/GPU/Docker resources, which is reasonable for every PR. But I think we should add a scheduled/or merge queue full suite correctness gate so issues like this surface earlier.
Proposal:
- PR default CI stays lightweight: normal Catch2 / `make test`, no MinIO.
- Every merge: run `make test` + `make s3-test` + `make s3-test-large` on CI, including MinIO-backed `[s3][integration]~[large]`. 

### Request type

New CI workflow capability

### Is this blocking current work?

No

PR #906, opened 2026-06-09
test(s3): self-managing MinIO via testcontainers; retire manual docker setup

mbrobbel:
> **Note:** this PR was prepared by an AI agent (Claude) working with @mbrobbel.

## Why

In preparation for https://github.com/sirius-db/sirius/issues/904.

The `[s3]` integration tests (S3 I/O, SigV4, scan-manager, SQL-over-S3) need a MinIO backend, and today that backend is set up **out of band** in a way that's easy to get wrong:

- `make s3-up` runs `ensure_tls_certs.sh` → `docker compose up` (two MinIO services on fixed ports 9000/9443) → `fixtures.sh` (uploads fixtures via a `minio/mc` container).
- You then `source env.sh` and run the Catch2 binary, and must remember `make s3-down` afterwards.

So the tests depend on a container already running on fixed ports (collisions / leftover state across runs), require a manual pre-step **and** manual cleanup, and silently skip if you forget to source the env. In practice this friction means the suite is rarely run end-to-end.

## What this changes

Move the whole setup **into the test binary** so running the `[s3]` suite needs no external command and no pre-existing state:

- With `SIRIUS_TEST_S3_AUTO=1`, the binary starts HTTP + self-signed-TLS MinIO on **dynamically-mapped** ports, generates fixtures, uploads them, runs the tests, and tears the containers down on exit (testcontainers' Ryuk also cleans up on a crash).
- Fixtures are uploaded **from the host** using Sirius's own SigV4 signer (`sign_request`) + libcurl — so no `minio/mc` container and no docker network are needed.
- The harness publishes the existing `SIRIUS_TEST_S3_*` env contract, so **no test files change** — only the harness, the build, and the Makefile.
- It's opt-in (`SIRIUS_TEST_S3_AUTO`), so the default `make test` path stays Docker-free; bring-up happens eagerly in `unittest.cpp` `main()` so a strict-mode failure exits non-zero instead of silently skipping.

Net result: `make s3-test` is now one self-contained command, and `s3-up`/`s3-down`, `docker-compose.yml`, `env.sh`, `fixtures.sh`, and `ensure_tls_certs.sh` are gone.

## How MinIO is launched

C++ has no first-class Testcontainers library, so we use the official **testcontainers-native** C bridge — but its published prototype can't set a container command (MinIO takes its mode from `server /data …`, with no env-var equivalent). We **don't vendor** it: CMake fetches it at a pinned commit at configure time and applies a small in-repo patch (`third_party/testcontainers-native.patch`) that adds:

- `tc_container_with_cmd` — container command support;
- real error propagation from `tc_container_run` (upstream dropped it);
- a fix for an upstream double-append bug in the container registry.

Dropping `mc` (host-side upload instead) removed the only need for container networking, so the patch stays command-only.

## Build impact

Building the unit tests now requires a **Go toolchain** (added to pixi as `go-cgo`) and **network access at configure time** (to clone testcontainers-native). This applies to every build, including CI — there is intentionally no opt-out flag.

## Testing

Verified on a Blackwell GPU host:

- `[s3][integration]~[large]~[aws]~[gpu_execution]` → **82 cases pass**
- `[s3]~[integration]~[large]~[aws]` (incl. HTTP + TLS) → **59 cases pass**

`make s3-test` excludes `[gpu_execution]`: those small-budget S3 SQL tests hardcode a 256 MiB GPU limit with no disk spill, which loops in the downgrade executor on large-VRAM GPUs (RMM pool sizing). That is a **pre-existing** issue unrelated to this change, left excluded with an explanatory comment.

## Commits

1. **Build patched testcontainers-native from upstream** — FetchContent at a pinned commit + the in-repo patch; nothing upstream is vendored.
2. **Add the testcontainers-backed MinIO harness** — `test/cpp/utils/s3_container.*` + build wiring.
3. **Drive `[s3]` tests via the harness; retire the docker/shell glue** — Makefile targets, deleted scripts, docs.


Makefile
```diff
@@ -167,47 +172,24 @@ list-presets: $(PRESETS_LINK)
...
 	  exit 1; \
 	fi
 	@set -e; \
-	trap '$(MAKE) s3-down' EXIT; \
-	$(MAKE) s3-up; \
-	source $(S3_DIR)/env.sh; \
-	export SIRIUS_TEST_S3_STRICT=1; \
-	$(S3_TEST_BIN) "[s3][integration]~[large]~[aws]"
+	export SIRIUS_TEST_S3_AUTO=1 SIRIUS_TEST_S3_STRICT=1; \
+	$(S3_TEST_BIN) "[s3][integration]~[large]~[aws]~[gpu_execution]"
```
ran-yuan-rui: make s3-test now skips the main SQL-over-S3 path.
But the small SQL-over-S3 e2e tests are tagged with [gpu_execution], for example test_s3_sql_surface.cpp:1180/1211/1356/1395/1412/1503. This means the standard make s3-test gate no longer covers:

read_parquet('s3://...') rewrite + scan
async backend SQL parity
S3 CPU fallback guard
empty result handling
Q3-shaped query
all-row-groups scan

The PR description says this exclusion is due to the 256 MiB test config looping on large-VRAM machines, but excluding SQL from the S3 gate is a real coverage regression. I’d suggest fixing the SQL test config, or adding a separate make s3-sql-test / scheduled CI gate, rather than silently removing these cases from s3-test.

CMakeLists.txt
```diff
@@ -139,6 +139,13 @@ endif()
 # Rust telemetry instrumentation (C++ FFI via Corrosion)
 add_subdirectory(rust/crates/telemetry/bridge)
 
+# Upstream testcontainers-native, fetched + patched at configure time (see
```
ran-yuan-rui, line 150: CMakeLists.txt unconditionally includes testcontainers_native.cmake, and that file unconditionally requires go and FetchContent access to GitHub. This means even builds that do not run S3 tests, or do not need testcontainers at all, now require a Go toolchain and configure/build-time network access, which might be an issue with unstable connection with github

Since this dependency is only for the test harness, I’d recommend guarding it behind a CMake option or enabling it only when building sirius_unittest / tests. Otherwise offline CI, developer builds, and release/package builds become dependent on a test-only toolchain.

Makefile
```diff
@@ -167,47 +172,24 @@ list-presets: $(PRESETS_LINK)
...
 	@if [ ! -x $(S3_TEST_BIN) ]; then \
 	  echo "s3-test-large: $(S3_TEST_BIN) not found - run \`make release\` first" >&2; \
 	  exit 1; \
 	fi
 	@set -e; \
-	trap '$(MAKE) s3-down' EXIT; \
-	$(MAKE) s3-up-large; \
-	source $(S3_DIR)/env.sh; \
-	export SIRIUS_TEST_S3_STRICT=1; \
+	export SIRIUS_TEST_S3_AUTO=1 SIRIUS_TEST_S3_LARGE=1 SIRIUS_TEST_S3_STRICT=1; \
```
ran-yuan-rui, line 193: make s3-test-large still launches six separate sirius_unittest processes. With the new harness, each process starts fresh HTTP + TLS MinIO containers and uploads fixtures again. The generated parquet file itself is cached, but the containers and the large SF10 upload are not shared across the six invocations.

The old flow started MinIO once via s3-up-large, then ran the six filters against the same backend. I’d suggest either keeping one persistent MinIO lifecycle for the large gate or adding a harness mode that can publish/reuse env across multiple invocations.

CMakeLists.txt
mbrobbel: Fixed in [ca925ce](https://github.com/sirius-db/sirius/pull/906/commits/ca925cecdb9f88e487e9dc7f6598b24fddbb01a5)

Makefile
mbrobbel: Fixed in [8302e0e](https://github.com/sirius-db/sirius/pull/906/commits/8302e0e0c5be73417286c34a931a7a75f0f7f4a8)

Makefile
mbrobbel: Fixed in [8302e0e](https://github.com/sirius-db/sirius/pull/906/commits/8302e0e0c5be73417286c34a931a7a75f0f7f4a8)

ran-yuan-rui (approved):

PR #907, opened 2026-06-09
Enable quent by default

### LINKED ISSUE #720: [Quent] Integrate Quent into Sirius

mbrobbel:
Enables quent in the default config. The default output directory is included in `.gitignore`.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -109,7 +109,7 @@ sirius:
     concat_batch_bytes: 5Gi
     max_build_hash_table_bytes: 500Mi
   telemetry:
-    enable_quent: false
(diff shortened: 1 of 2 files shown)


9prady9:
#809 is enabling this too. Perhaps we can let that absorb this change unless it is urgent for some other debugging.

9prady9:
> #809 is enabling this too. Perhaps we can let that absorb this change unless it is urgent for some other debugging.

I was wrong, ignore it. It isn't doing this always. Doesn't quent event generation has additional runtime cost though ?

bwyogatama (approved):

PR #908, opened 2026-06-09
feat(rust): add sirius-sys + sirius crate scaffolding

### LINKED ISSUE #835: Add Rust bindings for Sirius
We'll interact with Sirius from Rust. In the Doris experiment we experimented both with SQL (via DuckDB) and [a C-FFI wrapper](https://github.com/sirius-db/sirius/tree/doris/doris/crates/sirius-ffi), but it makes sense to provide a lower-level [`cxx`](https://cxx.rs/) `sirius-sys` + higher-level `sirius` crate now, because we'll be introducing public C++ APIs to use Sirius as a stage executor which we should map carefully to Rust types.

mbrobbel:
## What

First cut of Rust bindings for Sirius — two crates in the `rust/` workspace:

- **`sirius-sys`** — low-level [`cxx`](https://cxx.rs) bindings to Sirius's public C++ surface.
- **`sirius`** — safe, idiomatic wrapper.

This is scaffolding: it establishes the crates, the cxx bridge, the build/link integration, and a proof-of-life binding so later PRs can grow the API surface against a stable foundation. Replaces the ad-hoc `libloading` FFI experiment.

Closes #835.

## How it works

- **Public C++ surface** (`src/include/sirius_ffi.hpp` + `src/sirius_ffi.cpp`): an RAII `sirius::ffi::Context` — constructing it brings up an initialized engine (an `initialize()`d `duckdb::SiriusContext`), destroying it tears it down. Exported with default visibility so the symbols survive the loadable extension's `-fvisibility=hidden`. This is the seed of the public C++ API a dedicated `libsirius` will expose; the bindings link whichever artifact carries it (the DuckDB extension today, `libsirius` later).
- **`sirius-sys`** binds only that lightweight header (no internal Sirius headers), holding `Context` via `cxx::UniquePtr`. `build.rs` compiles the glue against `src/include` and links a single self-contained Sirius artifact under `$SIRIUS_BUILD_DIR` (default `build/release`) — no hand-maintained transitive dependency list. Built purely by cargo; CMake builds no Rust.
- **`sirius`** is a thin safe wrapper: `SiriusContext::new()` brings up an initialized engine; dropping it tears it down — pure RAII via `UniquePtr`, no raw pointers, no uninitialized state.
- **Linkage** is a cargo `static` feature: default `libsirius.so` (shared), `--features static` `libsirius.a` (self-contained, no runtime deps — the vcpkg static build).

## Validation

`pixi run make` then `cargo test` brings up a real GPU-initialized `SiriusContext` and drops it (topology discovery → GPU, `sirius_ioctx`, cuDF pinned memory, pipeline executor up + cleanly down): `constructs_and_drops ... ok`. `cargo fmt`/`clippy` clean.

## CI

`check.yml` runs fmt + clippy + `cargo test --no-run` (build + link) on the primary build entry. The actual run needs a GPU (construction does GPU bring-up), so it belongs on a GPU runner (test workflow) — follow-up.

## Follow-ups

- `--features static` (fully-static `libsirius.a`) is wired but unvalidated pending a self-contained bundle.
- Runtime currently needs `LD_LIBRARY_PATH` to find the linked `.so`; a dedicated installed `libsirius.so` (with SONAME) removes that.

---
🤖 Drafted by an AI agent (Claude Code) at the direction of @mbrobbel.

### DIFF
--- a/rust/crates/sirius-sys/build.rs
+++ b/rust/crates/sirius-sys/build.rs
@@ -1,0 +1,121 @@
+//! Build script for `sirius-sys`.
+//!
+//! Two jobs:
+//!  1. Compile the cxx bridge glue against Sirius's lightweight public FFI header
+//!     (`src/include/sirius_ffi.h`) — no internal Sirius headers are needed, so
+//!     this is the only include directory.
+//!  2. Link the single Sirius artifact that exports the FFI symbols (no
+//!     hand-maintained transitive dependency list). `cargo:rustc-link-{lib,search}`
+//!     propagate to downstream binaries, e.g. the `sirius` crate's `cargo test`.
+//!
+//! The artifact lives under the CMake build tree (`$SIRIUS_BUILD_DIR`, default
+//! `build/release`); build Sirius first (`pixi run make`). Linkage is selectable:
+//!  * default          — `libsirius.so` (shared; records its deps via DT_NEEDED).
+//!  * `static` feature — `libsirius.a` (self-contained; no runtime deps — the
+//!    fully static vcpkg build).
+//!
+//! A dedicated `libsirius` does not exist yet, so by default the bindings link
+//! the DuckDB extension that carries the exported FFI symbols; `resolve_lib_dir`
+//! symlinks it to `libsirius.so` as a stopgap so `-lsirius` resolves. Once Sirius
+//! ships a real `libsirius`, that symlink path simply isn't taken.
+
+use std::path::{Path, PathBuf};
+
+fn main() {
+    let manifest = PathBuf::from(std::env::var("CARGO_MANIFEST_DIR").unwrap());
+    let repo = manifest
+        .join("../../..")
+        .canonicalize()
+        .expect("resolve repo root from CARGO_MANIFEST_DIR");
+    let build_dir = std::env::var_os("SIRIUS_BUILD_DIR")
+        .map(PathBuf::from)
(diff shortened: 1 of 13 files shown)


mike-wendt (approved):
Approved for GHA codeowners

9prady9:
I should be done in next couple of hours, sorry for the delay

rust/crates/sirius-sys/build.rs
```diff
@@ -0,0 +1,121 @@
...
+
+fn main() {
+    let manifest = PathBuf::from(std::env::var("CARGO_MANIFEST_DIR").unwrap());
+    let repo = manifest
+        .join("../../..")
+        .canonicalize()
+        .expect("resolve repo root from CARGO_MANIFEST_DIR");
+    let build_dir = std::env::var_os("SIRIUS_BUILD_DIR")
+        .map(PathBuf::from)
+        .unwrap_or_else(|| repo.join("build/release"));
```
9prady9, line 32: nit:
it would be nice if we can go both debug and release routes here - like using the PROFILE or something custom ?

9prady9 (approved):

rust/crates/sirius-sys/build.rs
mbrobbel: Yes, let's do this in a follow-up.

PR #909, opened 2026-06-09
Implement projection folding

### LINKED ISSUE #917: Fold adjacent projections into a single projection
Currently we can have two projections appear adjacent to each other in the plan. We should be able to fold projections so that adjacent projections fold into one projection operator where possible

Jedi18:
## Summary

Fold adjacent physical projections during plan generation

- Collapse stacked `PROJECTION → PROJECTION` operators in the Sirius physical plan into a single projection, so we run one GPU expression-evaluation stage instead of several.
- Add AST helpers (`substitute_references`, `visit_references`) to compose projection select lists.
- Route all planner projection creation through `push_projection`, with a final `fold_adjacent_projections` pass over the full plan tree.

## Motivation

When Sirius builds a physical plan from DuckDB's logical plan, it often inserts extra `PROJECTION` operators that DuckDB's optimizer never sees:

- **Filter** — when a filter has a `projection_map` (column reordering after filtering)
- **Aggregate** — when aggregate child/filter expressions are hoisted into a projection below the aggregate
- **Table scan** — when unsupported filter columns require a projection on top of the scan
- **Logical projection** — the user's `SELECT` expressions

Each `PROJECTION` is a full GPU pipeline stage: it runs `gpu_expression_executor` over every batch. A plan like:

```
PROJECTION  (compute x + y)
  └── PROJECTION  (reorder columns)
       └── TABLE_SCAN
```

would evaluate expressions twice and materialize an intermediate batch between stages. Those two projections can be merged into one without changing results.

## How folding works

Each projection has a **select list** — one AST expression per output column. References like `#0`, `#1` point at columns from the projection's **immediate child**.

**Example — reorder then compute:**

```
Inner projection:  [#2, #0]           (swap columns 2 and 0)
Outer projection:  [add(#0, #1)]      (add the two output columns)
Child:             TABLE_SCAN (3 cols)
```

The outer `#0` and `#1` refer to the **inner projection's output**, not the scan. To fold, we substitute each outer reference with the corresponding inner expression:

```
Outer #0  →  inner select_list[0]  =  #2   (scan col 2)
Outer #1  →  inner select_list[1]  =  #0   (scan col 0)
```

Result — one projection:

```
[add(#2, #0)]
  └── TABLE_SCAN
```

This is implemented by `substitute_references`: walk the outer expression tree and replace every `#i` with a deep clone of `inner.select_list[i]`.

**Identity elision.** If the select list is a pure passthrough — `[#0, #1, #2, …]` with no reordering — and the child already has the same number of columns, the projection is dropped entirely.

## Two-phase approach

| Phase | When it runs | What it does |
|-------|--------------|--------------|
| `push_projection` | Every time a plan builder creates a projection (get, filter, aggregate, logical projection) | Wraps the child in a projection and immediately tries to fold with an adjacent child projection |
| `fold_adjacent_projections` | Once at the end of `create_plan()` | Walks the full plan tree and folds any remaining `PROJECTION → PROJECTION` chains |

The post-pass catches chains that only become adjacent after subtrees are assembled (e.g. after building join children independently).

**Scope:** only adjacent parent-child projection pairs are folded. Projections separated by a filter, join, or other operator are left alone — that is intentional.

## Safety guards

Folding changes the plan shape but must not change semantics or hide CPU-fallback triggers. Folding is **refused** in these cases:

### 1. Unsupported expression in the select list (null slot)

When `from_duckdb` cannot translate an expression, the select list gets a **null slot** at that position. That null is a signal: "this expression is not GPU-supported, trigger CPU fallback."

If we folded through a null slot, we'd silently drop that signal and might run a partial GPU plan. **Rule:** both inner and outer select lists must have only non-null entries.

```
Inner:  [#0, null]     ← slot 1 is unsupported
Outer:  [#0]
→ Do NOT fold
```

### 2. Non-trivial inner expression referenced more than once

`substitute_references` **clones** the inner expression at every reference site — it does not deduplicate. If the inner projection computed something expensive and the outer projection uses that column twice, folding would run that computation twice.

**Trivial expressions** (bare column references and constants) are fine to duplicate — they carry no GPU work.

**Non-trivial expressions** (function calls, casts, comparisons, etc.) must be referenced at most once in the outer list.

```
Inner:  [add(#0, #1)]          ← non-trivial (GPU function call)
Outer:  [#0, #0]               ← references inner col 0 twice
→ Do NOT fold (would evaluate add twice)

Inner:  [add(#0, #1), #0]
Outer:  [#0]                   ← references inner col 0 exactly once
→ OK to fold
```

This is a conservative guard. A future executor-level CSE pass could allow more aggressive folding; for now we prefer correctness over maximum fold coverage.

## Files changed

| Area | Files |
|------|-------|
| AST utilities | `substitute.cpp/hpp`, `visit.cpp/hpp` |
| Folding logic | `sirius_plan_projection_utils.cpp/hpp` |
| Plan builders | `sirius_plan_projection.cpp`, `sirius_plan_filter.cpp`, `sirius_plan_get.cpp`, `sirius_plan_aggregate.cpp` |
| Entry point | `sirius_physical_plan_generator.cpp` (post-pass) |
| Tests | `test_projection_fold.cpp`, `test_ast_substitute.cpp` |

Closes https://github.com/sirius-db/sirius/issues/917


Jedi18:
Ran on SF100, no noticeable performance improvements but from the logs I verified that the projection folding is happening

| Query | Projections (no-fold → fold) | Example folded chain |
|-------|-------------------------------|----------------------|
| Q2 | 4 → 3 | `HASH_JOIN → FILTER → PROJ → PROJ → TOP_N` → `HASH_JOIN → FILTER → PROJ → TOP_N` |
| Q7 | 2 → 1 | `HASH_JOIN → FILTER → PROJ → PROJ → HASH_GROUP_BY` → `HASH_JOIN → FILTER → PROJ → HASH_GROUP_BY` |
| Q9 | 2 → 1 | `HASH_JOIN → PROJ → PROJ → HASH_GROUP_BY` → `HASH_JOIN → PROJ → HASH_GROUP_BY` |
| Q11 | 6 → 5 | `PROJ → PROJ → UNGROUPED_AGGREGATE` → `PROJ → UNGROUPED_AGGREGATE` |
| Q13 | 3 → 2 | `PROJ → PROJ → HASH_GROUP_BY` → `PROJ → HASH_GROUP_BY` |
| Q16 | 2 → 1 | `HASH_JOIN → FILTER → PROJ → PROJ → HASH_GROUP_BY` → `HASH_JOIN → FILTER → PROJ → HASH_GROUP_BY` |
| Q17 | 6 → 4 | `HASH_JOIN → PROJ → PROJ → HASH_GROUP_BY` → `HASH_JOIN → PROJ → HASH_GROUP_BY` (+ 1 more chain) |
| Q19 | 3 → 2 | `HASH_JOIN → FILTER → PROJ → PROJ → UNGROUPED_AGGREGATE` → `HASH_JOIN → FILTER → PROJ → UNGROUPED_AGGREGATE` |
| Q20 | 4 → 3 | `HASH_JOIN → PROJ → PROJ → HASH_GROUP_BY` → `HASH_JOIN → PROJ → HASH_GROUP_BY` |
| Q22 | 6 → 3 | `GPU_SCAN → PROJ → UNGROUPED_AGGREGATE` → `GPU_SCAN → UNGROUPED_AGGREGATE` (+ 2 more chains) |

src/expression/ast/reference_utils.cpp
```diff
@@ -0,0 +1,133 @@
...
+  node const& src, std::vector<std::unique_ptr<node>> const& inner_select_list)
+{
+  return std::visit(
+    [&](auto const& alt) -> std::unique_ptr<node> {
+      using T = std::decay_t<decltype(alt)>;
+
+      if constexpr (std::is_same_v<T, reference>) {
+        // Out-of-range or null inner slots are left unchanged — a null slot is the
+        // from_duckdb signal for an unsupported expression and must not be folded through.
+        if (alt.column_index >= inner_select_list.size() || !inner_select_list[alt.column_index]) {
```
9prady9, line 72: the second condition can cause segfault right ? even if column_index is less than size of the select list, the actual identifier from the total number of columns can be larger than `inner_select_list` size right I think.

src/include/expression/ast/substitute.hpp
9prady9, line 1: I think we can merge the visit_references and substitute references into a utils header/cpp.

src/include/planner/sirius_plan_projection_utils.hpp
```diff
@@ -0,0 +1,53 @@
...
+//! Returns true when every select-list slot is a non-null reference to column i.
+bool is_identity_ast_projection(
+  duckdb::vector<std::unique_ptr<sirius::ast::node>> const& select_list);
+
+//! Returns true when both projections have only non-null select-list entries.
+bool can_combine_projections(sirius::op::sirius_physical_projection const& outer,
+                             sirius::op::sirius_physical_projection const& inner);
+
+//! Compose @p outer over @p inner, re-homing the combined operator on inner's child.
+duckdb::unique_ptr<sirius::op::sirius_physical_operator> combine_projections(
```
9prady9: Need more docs on whats defined as outer and inner projection here.

src/include/planner/sirius_plan_projection_utils.hpp
9prady9, line 1: the function args list need doc strings. in some non-trivial cases, the helpers need proper explanation as to what these can and cannot do.

src/include/expression/ast/visit.hpp
```diff
@@ -0,0 +1,36 @@
...
+namespace sirius::ast {
+
+/**
+ * @brief Invoke @p fn on every reference node in the tree rooted at @p root
+ * (pre-order, read-only).
+ *
+ * A single reusable traversal primitive so callers (e.g. projection folding's
+ * use-count analysis) do not each re-enumerate the variant alternatives.
+ */
+void visit_references(node const& root, std::function<void(reference const&)> const& fn);
```
9prady9: Did you try passing the `fn` as templated `rvalue` like `Fn&& fn` with Fn as template arg ? in C++20 that should work without having to providing the template args at call sites.

src/include/planner/sirius_plan_projection_utils.hpp
9prady9: Also some of these helpers are not used outside of this file's cpp TU, perhaps remove this expose ?

src/planner/sirius_plan_projection_utils.cpp
```diff
@@ -0,0 +1,189 @@
...
+duckdb::unique_ptr<sirius::op::sirius_physical_operator> fold_adjacent_projections(
+  duckdb::unique_ptr<sirius::op::sirius_physical_operator> plan)
+{
+  for (auto& child : plan->children) {
+    child = fold_adjacent_projections(std::move(child));
+  }
+  if (plan->type == sirius::op::SiriusPhysicalOperatorType::PROJECTION) {
+    plan = fold_projection_with_child(std::move(plan));
+  }
+  return plan;
```
9prady9, line 179: do we have a test that verifies this post-pass case ? If not, lets add one.

9prady9 (changes requested):
🎉 welcome to Sirius development!

There is one bug, rest of the comments are doc fixes and code improvements. please have a look.

src/planner/sirius_plan_projection_utils.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+// Plan-builder entry: wrap @p child in a projection, skipping creation when the
+// select list is identity, then fold with an adjacent child projection if safe.
+duckdb::unique_ptr<sirius::op::sirius_physical_operator> push_projection(
+  duckdb::unique_ptr<sirius::op::sirius_physical_operator> child,
+  duckdb::vector<sirius::logical_type> types,
+  duckdb::vector<std::unique_ptr<sirius::ast::node>> select_list,
+  std::size_t estimated_cardinality)
+{
+  if (is_identity_ast_projection(select_list) && child->types.size() == types.size()) {
```
9prady9: also, shouldn't we check for types also here ?

src/include/planner/sirius_plan_projection_utils.hpp
9prady9: While doing that, actually, feel free to move the `clone` helper from it's own individual file to this new utilities header.  Too much unnecessary text with so many files.

src/expression/ast/reference_utils.cpp
Jedi18: Sorry I don't understand, we are doing the bounds check for the column_index of the reference here?

9prady9: My bad, I was seeing a problem but misread that line itself. You are correct, there is no segfault here.

Let me try to illustrate the problem I was trying to figure out.


Assume a table with 4 input columns:
```md
SELECT projected_price + projected_discount
FROM (
  SELECT
    price AS projected_price,
    discount AS projected_discount,
    order_id AS projected_order_id
  FROM orders
) p;

orders(order_id, customer_id, price, discount)
-- physical input columns:
-- #0 = order_id
-- #1 = customer_id
-- #2 = price
-- #3 = discount
```

Input column mapping:

```text
#0 = order_id
#1 = customer_id
#2 = price
#3 = discount
```

Inner projection:

```text
[#2, #3, #0]
```

So the inner projection outputs only 3 columns:

```text
inner #0 = price
inner #1 = discount
inner #2 = order_id
```

Now suppose the outer projection accidentally contains a reference that is valid for the original scan width

```text
[add(#0, #3)]
```

That `#3` is invalid at the outer level, because the outer projection reads from the inner projection output, and the inner projection only produced columns `#0`, `#1`, and `#2`.


I am wondering if this is effectively checked. I might need to check code once more if I missed something.

Jedi18: That would not happen because if it did, even without folding the projection would fail? The plan itself would be wrong in this case

9prady9: that's true, duckdb probably validates those. Just wanted to be sure, to be honest I hardly know how planners work. So better to be safe than sorry

src/include/expression/ast/substitute.hpp
Jedi18: merged into one reference_utils. I didn't add clone to it since it's unrelated

src/include/planner/sirius_plan_projection_utils.hpp
Jedi18: Added more docs

src/include/expression/ast/visit.hpp
Jedi18: Made the change thanks, had to move the implementation to the header since the function is now templated

src/planner/sirius_plan_projection_utils.cpp
Jedi18: We do have one that tests fold_adjacent_projections but I've added another one for robustness

src/planner/sirius_plan_projection_utils.cpp
Jedi18: Agreed, I've changed it to compare the entire vector instead of just the size

9prady9 (approved):
ty, I can take care of the clone move in few mins.

PR #912, opened 2026-06-09
Sirius vector search table functions

yayen-lin:
### Headsup

**Design**

1. Execuse the branch name, this PR contains both exact brute force search (ENN) and approximate search (ANN)
2. Remove the DuckDB SQL path, and keep only the table function path (so once everything is in place we just have to wire them into native SQL support, as per discussion with Professor Xiangyao Yu)

**Technical**

1. Currently only support DuckDB native scan path (parquet doesn't have vectors so it's unsupported for now)
2. Supported distance metrics: euclidean L2 and cosine
3. Supported ANN index type: IVF-Flat
4. A sliced vector column and row-level nulls are applied with a boolean mask to make an input compaction. The implication is that if `result size < k`, there will be less rows in the result than the specified k. Element-level nulls are rejected for now as it's rarely a case and would degrade the performance.
5. Avoided using concat.

### Summary

1. Added two table function: `sirius_create_ann_index` and `sirius_knn_search`, both of which **require** the table to be pinned on GPU.
2. Added a new cache interface for ann index.
3. Removed workaround mentioned in [#147](https://github.com/NVIDIA/cuCascade/issues/147).
4. Added missing decode budget estimation calculation for ARRAY.

### High Level Workflow

**Index Creation**

- **Memory**: Index memory is reserved upfront throught the cuVS index cache. The index and its build-time scratch are allocated through the reservation's memory resource, and it's `shrink_to_fit()`. The reservation is held until the entry is erased or the session ends.
- **Impl**: Collect the column's GPU chunks as views (zero-copy) and build IVF-Flat index incrementally (one chunk at a time). This avoids materializing the whole dataset but the tradeoff is that recall degrades at low `n_probes`.

**ENN**

- **Memory**: This is allocate-on-demand but still routed through the pinned table's GPU `memory_space` allocator, so every transient is tracked by the same memory manager and freed when the qeury ends.
- **Impl**: Local search on each pinned chunk (these chunks are zero-copy reinterpreted to RAFT's device matrix view which cuVS uses under the hood), and then a merge into global top-k result.

**ANN**

- **Memory**: Same as ENN.
- **Impl**: Search against the already built index, cuVS returns global row ids so they're mapped back onto the pinned chunks for the returned top-k result.

### Sample queries

<details>
    <summary><strong>Expand to see sample queries</strong></summary>

```bash
# start Sirius with a persistent file
SIRIUS_LOG_LEVEL=debug ./build/release/duckdb test.duckdb
```

```sql
-- creating dataset
CREATE TABLE test_arrays (id INTEGER, vec FLOAT[3]);
INSERT INTO test_arrays SELECT i, [i::float, (i+1)::float, (i+2)::float] FROM range(5000) t(i);
CHECKPOINT; -- checkpoint is needed

-- (optional) cache the table
SELECT * FROM pin_table(name => 'test_arrays', tier => 'gpu', format => 'duckdb');

-- sample queries
SELECT * FROM sirius_knn_search(
    'test_arrays',                    -- table   (positional; must be GPU-pinned)
    'vec',                            -- column  (positional)
    [4980,4981,4982]::FLOAT[3],       -- query   (positional)
    k              => 20,             -- default 10
    output_columns => ['id','vec'],   -- default all columns
    metric         => 'l2sq',         -- default 'l2sq'
    use_index      => true,           -- default true
    n_probes       => 32              -- default 0 → min(n_lists,32); ANN only
);

-- samples
SELECT * FROM sirius_knn_search(
  'test_arrays',
  'vec',
  [4980,4981,4982]::FLOAT[3],
  k => 20
);

SELECT * FROM sirius_knn_search(
  'test_arrays',
  'vec',
  [4980,4981,4982]::FLOAT[3],
  k => 20,
  use_index => false
);

SELECT * FROM sirius_knn_search(
  'test_arrays',
  'vec',
  [4980,4981,4982]::FLOAT[3],
  k => 20,
  n_probes => 1024
);
```

</details>

### Changes (38 files changes)

**Table Function (1)**

- `src/sirius_extension.cpp`: Register the two table functions with bind & execute logic

**cuVS / cuDF interop (2)**

- `src/include/vss/cudf_raft_interop.hpp`, `src/cuda/vss/cudf_raft_interop.cu`: zero-copy reinterpretation of a fixed-size FLOAT LIST column into a `[n_rows, dim]` raft device matrix view; rejects nulls, slices, and non-FLOAT children.

**ANN/ENN search (13)**

- `src/include/vss/brute_force_search.hpp`, `src/cuda/vss/brute_force_search.cu`: Exact brute-force search kernel
- `src/include/vss/ivf_flat_index.hpp`, `src/cuda/vss/ivf_flat_index.cu`: Build IVF-Flat index and approximate search kernel, plus the metric → cuVS `DistanceType` mapping helpers.
- `src/include/vss/enn_top_k.hpp`, `src/vss/enn_top_k.cpp`: Per-chunk brute-force top-k (`compute_enn_top_k`, with null/sliced-row compaction) and the k-way merge of per-chunk candidates into the global top-k (`merge_enn_top_k`).
- `src/include/vss/pinned_column.hpp`, `src/vss/pinned_column.cpp`: Borrow a pinned table column's GPU chunks as zero-copy `cudf::column_view`s in pin order; rejects missing columns and multi-GPU spread.
- `src/include/vss/vector_search.hpp`, `src/vss/vector_search.cpp`: Top-level entry (`run_vector_search`), resolves GPU/host space and pinned table, uploads the query vector, dispatches ANN/ENN, and converts the result to host.
- `src/vss/vector_search_ann.cpp`: ANN implementation; search the pinned index, drop IVF padding, remap global row ids onto the chunked storage, sort by distance.
- `src/vss/vector_search_enn.cpp`: ENN implementation; per-chunk brute-force search, then merge and slice into the global top-k.
- `src/include/vss/vector_search_internal.hpp`: Shared `vector_search_context` (resolved handles + request) and the ANN/ENN impl declarations.

**Caches (6)**

- `src/include/scan_manager/sirius_scan_manager.hpp`, `src/scan_manager/sirius_scan_manager.cpp`: Track each pinned chunk's GPU `memory_space` on the pinned entry and validate/serve it, so VSS can borrow chunks in place and reject a table spread across GPUs.
- `src/include/vss/cuvs_index_cache.hpp`, `src/vss/cuvs_index_cache.cpp`: New session cache for pinned cuVS indexes which reserves GPU memory via the reservation manager, stores the type-erased index + metadata, looks up by name or by table+column+metric, and erase/clear.
- `src/include/sirius_context.hpp`, `src/sirius_context.cpp`: Own the `cuvs_index_cache` on the Sirius context (constructed with the memory manager, exposed via accessor, torn down before the manager).

**Fixes (4)**

- `src/helper/list_offset_fixup.{hpp, cpp}`, `src/include/pipeline/batch_lock_utils.hpp`: Removed from the codebase. This workaround for the cuCascade fix ([#147](https://github.com/NVIDIA/cuCascade/issues/147)) is pulled into the submodule.
- `src/op/scan/duckdb_native_metadata.cpp`: Add the missing decode budget estimation calculation for ARRAY.

**Build (3)**

- `pixi.toml` / `pixi.lock`: add the `libcuvs` (26.06) dependency.
- `CMakeLists.txt`: add the VSS sources and test files to the build.

**Tests (9)**

unit tests

- `test/cpp/vss/test_brute_force_search.cpp`: exact nearest ids + distances (L2 & cosine), and rejection of out-of-range k / mismatched dimensions.
- `test/cpp/vss/test_cudf_raft_interop.cpp`: zero-copy view + malformed-input rejection (non-LIST, dim mismatch, nulls, slices).
- `test/cpp/vss/test_cuvs_index_cache.cpp`: insert/lookup (by name and by column+metric identity, including L2 expanded/unexpanded folding), in-place replace, erase/clear.
- `test/cpp/vss/test_enn_top_k.cpp`: per-chunk top-k (including null/sliced compaction) and the global k-way merge.
- `test/cpp/vss/test_ivf_flat_index.cpp`: incremental chunked build (global-id tagging, empty-chunk skipping) + approximate search (L2/cosine, k == n_rows) + metric mapping/rejections.
- `test/cpp/vss/test_pinned_column.cpp`: zero-copy chunk views in pin order; missing-column and different-GPU rejection.
- `test/cpp/vss/test_vector_search.cpp`: empty-output schema helper and rejections.

integration tests

- `test/cpp/integration/test_gpu_execution_vector_search.cpp`: end-to-end ANN/ENN through the SQL table function, compared against a CPU reference.
- `test/cpp/utils/gpu_execution_fixture.hpp`: shared file-backed native-scan fixture + GPU-vs-CPU comparator (added returned result order check).


ran-yuan-rui:
Question — do null / sliced vector columns fall back to CPU here, or throw?

`match_vss_top_n` gates at plan time on the column *type* (`ARRAY(FLOAT)`), but `cudf_raft_interop` enforces the row-level invariants with hard `CUDF_EXPECTS` at runtime — `null_count()==0` (rows and values), `offset()==0`, values `FLOAT32`. Since the transparent hook is all-or-nothing on the plan tree, once the pattern matches there's no CPU fallback left at runtime. So a nullable `FLOAT[dim]` column that actually has a NULL row turns a query DuckDB runs fine into a hard error:

```sql
-- vec is FLOAT[3], nullable, at least one NULL row
SELECT id FROM t ORDER BY array_distance(vec, [0,0,0]::FLOAT[3]) LIMIT 5;
-- DuckDB CPU: NULL distances sort last under ASC, dropped from the top-k → fine
-- here:      pattern matches → VSS → CUDF_EXPECTS(null_count()==0) throws at execution
```

Same for a sliced column reaching the operator. If we want to keep DuckDB semantics, either gate the pattern to decline when the column is nullable (fall back), or drop the null rows in the operator before building the matrix and remap the neighbor indices. Rejecting malformed input is fine — I'd just avoid turning a valid query into an error.

A few smaller things, non-blocking:

- The pattern needs a fixed-size `ARRAY(FLOAT, dim)` column. A plain `LIST<FLOAT>` (what parquet often gives for a repeated float field) doesn't match and silently falls back to CPU — probably worth a line in the docs on how to hit the GPU path.
- The `::FLOAT[N]` cast on the query is load-bearing: without it the literal is `DOUBLE`, `extract_float_array` declines, and it silently runs on CPU. A `DOUBLE` embedding column is gated out the same way at plan time — no error, just a silent fallback that might surprise someone.
- Only `array_distance` (L2) and `array_cosine_distance` (cosine) are accelerated; `array_inner_product` / `array_cosine_similarity` / `array_negative_inner_product` fall back. Fine to scope out, just worth calling out.

yayen-lin:
Hi Ran-Yuan, I appreciate for your feedback, those are very helpful and I agree!

I will make the error handling better so Sirius throws an error instead of crashing because there are some cases where fallback to DuckDB isn't straightforward and will be hacky. And I think the team is making Sirius more independent now. But they are a good point, I make sure to specify them in the PR description. 

I think including parquet on the vss path will be nice, but like you said parquet returns a LIST and it doesn't translate to an ARRAY well. I will make sure I mention it in the PR description as well as for the `::FLOAT[N]` cast and other distance metrics. 

Thank you again for the feedback 🙏

bwyogatama:
Hi @yayen-lin 
1. What do you mean by `Remove the DuckDB SQL path, and keep only the table function path`?
2. Also with this kind of PR, since you are a new contributor it would be very difficult to review a PR this large. I would highly recommend to break this down into several smaller PRs (the smaller the better). Claude typically can do this pretty well. Othewise, I am afraid that this PR will just be sitting forever since it's too big and people don't know how to review it
Great work btw

bwyogatama:
@yayen-lin then every time your smaller PR is ready to be reviewed, ping us on the channel and I will ask someone to review it

yayen-lin:
Hey Bobbi good morning and thanks for your feedback! 

By DuckDB SQL path I meant the implicit path where we intercepted a plain `ORDER BY array_distance(...) LIMIT k` query and rewrote it into a GPU vector search. Table function path is the explicit `sirius_knn_search(...)` call. 

And that makes sense, I will definitely break it down to smaller PRs and send a message to the channel! 

Thanks again for the heads up I really appreciate it!

yayen-lin:
This PR is split into 4 separate smaller PRs (#1203 #1204 #1205 #1206) so closing it!

PR #913, opened 2026-06-09
Cleanup scan manager

aminaramoon:
THIS PR Paves the way for new IoFramework to be merged into dev.

- it creates one io_context for all gpus, as it is meant to be but was broken in multi-gpu PR
- it doesn't send io_cotnext to operators, datasource carries io_context, so datasource should be everything a reader needs.
- removes cpud executor, duckdb cpu executor, iceberg, and all of the stuff around them
- it enables gpu reader for duckdb storage by default


bwyogatama (approved):

test/cpp/scan_manager/test_sirius_scan_manager_s3.cpp
```diff
@@ -341,22 +311,16 @@ TEST_CASE("sirius_scan_manager routes borrowed S3 backend and dispatches by path
...
 
   host_cache_memory memory;
-  auto local_ctx       = make_local_ioctx();
-  auto borrowed_s3_ctx = make_mock_s3_ioctx(&memory.host_mr);
-  sirius_scan_manager manager(
-    std::move(cfg),
-    make_borrowed_backends(local_ctx, std::static_pointer_cast<sirius_ioctx>(borrowed_s3_ctx)));
+  sirius_scan_manager manager(std::move(cfg), &memory.host_mr);
 
   auto* default_ctx = manager.io_ctx();
```
ran-yuan-rui, line 316: The current removes `sirius_scan_manager::io_ctx()`, `io_ctx_for()`, and `s3_ioctx()` from the public header, but the S3 scan-manager tests still call them here and in several later cases. This looks like a direct compile blocker for `sirius_unittest`.

If the new contract is “datasource carries the io context”, these tests need to be ported to `create_datasource(path)->io_ctx()` or to a new stable test accessor. Otherwise the PR won't build with the S3 test suite enabled.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -43,38 +43,91 @@
...
+    auto s3_cfg                 = *_config.s3_config;
+    s3_cfg.host_memory_resource = host_fsmr;
+    if (s3_cfg.async_thread_pool == nullptr) {
+      _s3_thread_pool = std::make_unique<sirius::exec::static_thread_pool>(
+        _config.s3_thread_pool.num_threads,
+        _config.s3_thread_pool.thread_name_prefix,
+        _config.s3_thread_pool.cpu_affinity_list);
+      s3_cfg.async_thread_pool = _s3_thread_pool.get();
+    }
+    auto s3_ioctx = std::make_shared<sirius::io::s3::s3_blocking_ioctx>(std::move(s3_cfg));
```
ran-yuan-rui: This unconditionally constructs `s3_blocking_ioctx`, so `object_store_config::s3_use_async_backend` is ignored. That regresses the current S3 default path: `s3_use_async_backend=true` should select the async `s3_ioctx`, while `false` should select `s3_blocking_ioctx`.

This matters for the device-read performance work: the async backend is the path that batches range GETs and H2D copies without per-chunk stream synchronization.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -43,38 +43,91 @@
...
 {
-  // S6 (NUMA) increment 1: the scan_manager no longer constructs IO backends.
-  // SiriusContext owns the uring(s) + s3_ioctx + S3 async thread pool + prefetch
-  // buffer_pool/cache and passes the routing backends in as borrowed
-  // shared_ptrs. io_ctx_for / io_ctx_shared_for dispatch over this borrowed
-  // list; stop() and the destructor must NOT shut down or destroy them.
-  SIRIUS_LOG_DEBUG("[sirius_scan_manager] constructed with {} borrowed IO backend(s)",
-                   _io_ctxs.size());
+  // if (_config.use_sirius_datasource) {
+  auto ioctx = std::make_shared<sirius::io::uring_ioctx>(
```
ran-yuan-rui, line 87: This moves IO backend ownership back into `sirius_scan_manager`: it constructs the local `uring_ioctx`, the S3 backend, and the prefetch buffer pool. That conflicts with the current S3/S6 ownership model, where `SiriusContext` owns the local/S3 IO backends and cache resources, and `scan_manager` only borrows/routes them.

The PR description says this restores one io_context per GPU, but this constructor appears to create a single local `uring_ioctx` inside the scan manager. Can we align the ownership model before merging this? Otherwise the S3 backend loses the current lifecycle/teardown ordering and NUMA/async-backend wiring.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -43,38 +43,91 @@
...
     _factory(_pinned_entries)
 {
-  // S6 (NUMA) increment 1: the scan_manager no longer constructs IO backends.
-  // SiriusContext owns the uring(s) + s3_ioctx + S3 async thread pool + prefetch
-  // buffer_pool/cache and passes the routing backends in as borrowed
-  // shared_ptrs. io_ctx_for / io_ctx_shared_for dispatch over this borrowed
-  // list; stop() and the destructor must NOT shut down or destroy them.
-  SIRIUS_LOG_DEBUG("[sirius_scan_manager] constructed with {} borrowed IO backend(s)",
-                   _io_ctxs.size());
+  // if (_config.use_sirius_datasource) {
```
ran-yuan-rui, line 86: It looks like `use_sirius_datasource` is no longer honored here: the old guard is commented out and the local `uring_ioctx` is constructed unconditionally. That changes the meaning of `scan_manager_config::use_sirius_datasource=false`, which was the escape hatch for local-path cudf/KvikIO fallback.

If the intent is to remove that fallback, the config field and tests should be updated explicitly. Otherwise this should stay gated on `_config.use_sirius_datasource`.

kevkrist (approved):

test/cpp/scan_manager/test_sirius_scan_manager_s3.cpp
aminaramoon: I will do that, thanks for the comment

src/scan_manager/sirius_scan_manager.cpp
aminaramoon: yes, in my next PR I will reenable this again. the problem is making duckdb gpu reader as default reader that needs sirius_datasource

Issue #826, opened 2026-05-29
Sirius compute node implementation for Starrocks


mbrobbel:
Tracking issue for the implementation of a Sirius [compute node](https://docs.starrocks.io/docs/introduction/Architecture/#shared-data) for [Starrocks](https://github.com/starrocks/starrocks).

mbrobbel:
Moved this design proposal to a draft PR so it can be reviewed and commented inline: sirius-db/sirius#914

  It adds `experimental/starrocks/docs/exchange-design.md` — the data-exchange / public-API design for running Sirius as a StarRocks compute node (streaming source/sink, stream session, nixl cross-node transport, and the cuCascade GPU-buffer resource-sharing options). Please leave feedback on
  the PR.

Issue #841, opened 2026-05-29
Starrocks plan fragment to substrait conversion


mbrobbel:
We need a crate that converts Starrocks plan fragments into Substrait plans, which we'll convert to DuckDB plans and then to Sirius plans.

9prady9:
What is the format in which these plan fragments are fed into sirius ? bytes (in arrow format) or something else ? How does `duckdb` do it for the current extension ? (I rarely dealt with planner area, so might be trivial but I am not aware hence asking).

The eventual goal behind such a question is: I am trying to pull out APIs for sirius that lets us use sirius from duckdb and otherwise as well. To ensure it works without any form of planner, whats the best approach to receive the plan.

One way I could think of is using arrow standard but I want to understand a bit in terms of how it works now as well.

mbrobbel:
> What is the format in which these plan fragments are fed into sirius ? bytes (in arrow format) or something else ? How does `duckdb` do it for the current extension ? (I rarely dealt with planner area, so might be trivial but I am not aware hence asking).

Currently it's DuckDB plans. We use the DuckDB Substrait extension to convert Substrait (protobuf) bytes into DuckDB plans. We convert the Starrocks plan fragment to Substrait (that's this issue). The Sirius Starrocks compute node implementation gets plan fragments in thrift from the Starrocks frontend.

- Starrock plan fragments -> Substrait plan (this task, added to Sirius Starrocks project)
- Substrait plan -> DuckDB plan (provided by DuckDB substrait extension) 
- DuckDB plan -> Sirius plan (this is the Sirius planner)

> The eventual goal behind such a question is: I am trying to pull out APIs for sirius that lets us use sirius from duckdb and otherwise as well. To ensure it works without any form of planner, whats the best approach to receive the plan.

Sirius would need a plan builder API or directly support mapping Substrait plans to Sirius plans. However, currently by going through DuckDB we get all the optimizations provided by their planner. This is most likely less relevant for small plan fragments.

Issue #837, opened 2026-05-29
Streaming sink operator


mbrobbel:
Add a sink operator that pushes output data_batches to an external bounded channel (zero-copy) per sink() call and closes it at pipeline finish.

The result path fully materializes into a `ColumnDataCollection` or `MaterializedQueryResult`. A stage must emit batches incrementally so the exchange can pull and shuffle as they're produced.

Also see #836.

Issue #836, opened 2026-05-29
Streaming source operator


mbrobbel:
Add a multi-shot source operator that pulls `cucascade::data_batches` from an external bounded channel and publishes them into the pipeline, driving task creation as batches arrive and finalizing on producer close.

Sirius can only ingest via scans or a one-shot collection. A distributed stage must accept inputs pushed over its lifetime, with backpressure and explicit end-of-stream.

Also see #837.

mbrobbel:
Closed via #1094

PR #914, opened 2026-06-10
docs(starrocks): compute-node exchange & public-API design proposal

mbrobbel:
[Rendered](https://github.com/mbrobbel/sirius/blob/starrocks-exchange-design/experimental/starrocks/docs/exchange-design.md)

Design proposal for running Sirius as a StarRocks shared-data compute node — how data moves in and out of the engine while overlapping with computation.

Covers the proposed public-API surface (streaming source/sink operators, the stream session), `nixl` cross-node transport, and the three cuCascade GPU-buffer resource-sharing options. Mermaid diagrams are tagged for what exists in the engine vs is proposed vs is prior art from the Doris experiment.

Docs-only, under `experimental/starrocks/docs/`. Opening as a draft for review and comment. Tracking issue: #826.

### DIFF
--- a/experimental/starrocks/docs/exchange-design.md
+++ b/experimental/starrocks/docs/exchange-design.md
@@ -449,6 +449,21 @@ with a bump allocator and make **no** further reservation (that memory is alread
 both), so its tier is a deployment choice — host staging can be preferable on some topologies (e.g. GB/VR-class
 nodes).
 
+For an **all-to-all** exchange the scarce resource is the fixed receive-staging arena (its **leases**), not
+just total GPU bytes, and a logical plan DAG does not prevent a physical lease cycle (see §7). Option A
+therefore needs two further constraints:
+
+1. **Lease-aware spill.** Generic spilling picks idle batches by memory *tier*, not by pool origin, so it
+   frees GPU bytes without necessarily releasing a staging lease. Releasing a lease requires the received
+   batch's GPU representation to **own** it, so that **copying the batch out of staging** (into ordinary,
+   spillable memory) returns the lease to the allocator. This makes the fast-path-vs-copy-out choice
+   explicit: a received batch can stay **staging-backed** (zero-copy, holds its lease until consumed or
+   copied out) or be **copied out on arrival** (frees the lease immediately) — and spill must be able to
+   force the copy-out.
+2. **Reserved receive-staging floor.** Hold back a **non-borrowable** slice of staging (a credit budget)
+   that compute can never consume, so there is always room to copy out / reclaim a receive lease —
+   guaranteeing forward progress even under a fully shared budget.
+
 ### Spill lifecycle of an exchange batch under option A
 
 This works **only because** the incoming batch is parked as a `data_batch` in a *registered* repository —
@@ -502,12 +517,15 @@ reservation-aware adaptor `src/memory/sirius_memory_reservation_manager.cpp:42-4
 - **Batch ownership graph.** Pin down whether a channel is a second, independent owner or only a handle into
   the repository (the intended model), and enumerate every holder — repository, channel, in-flight accessor,
   downgrade executor, `nixl` send — so that "idle ⇒ spillable" and free-on-last-drop are unambiguous.
-- **Deadlock / liveness.** Backpressure on bounded channels plus a shared GPU budget can stall progress —
-  e.g. every worker slot parked on a full output channel with none left to drain inputs, or a blocked sink
-  pinning memory that spilling needs. Mitigations: the plan is a DAG (no channel cycles), a blocked sink
-  should **yield its task** rather than block a worker thread, and a queued batch should sit *idle* in its
-  repository (still spillable) while it waits. Even so, we likely want a stall/credit **watchdog** that flags
-  "all channels full, no forward progress" so a sink (or a monitor) can surface it rather than hang silently.
+- **Deadlock / liveness.** Two hazards. *(a) Thread/queue stalls* — every worker slot parked on a full
+  output channel with none left to drain inputs, or a blocked sink pinning memory that spilling needs;
+  mitigated by a blocked sink **yielding its task** (not blocking a worker thread) and keeping queued batches
+  *idle* in their repository (still spillable). *(b) Staging-lease cycle* — a logical plan DAG does **not**
+  rule this out: CN A's receive-staging fills with batches from B so A stops granting leases, while B holds
+  pinned output waiting to send to A (and symmetrically) → mutual wait. Per-destination channels isolate slow
+  receivers at the *queue* level, but a shared budget re-couples them at the *staging* level. The fix is the
+  §6 lease-aware spill + reserved receive-staging floor, plus a stall/credit **watchdog** that flags "no
+  forward progress" rather than hanging silently.
 - **Transport retry / fallback.** On a failed `nixl` transfer, retry once over `nixl`, then fall back to the
   bRPC/CPU path, and only fail the query if that also fails (cf. the Doris `TransferHealth` fallback after N
   consecutive failures). The retry/fallback runs on the wrapper's `nixl` send path (which has its own
--- a/experimental/starrocks/docs/exchange-design.md
+++ b/experimental/starrocks/docs/exchange-design.md
@@ -173,7 +173,9 @@ the producer and the engine operator the consumer; for a sink it is reversed. Th
 are ordinary Sirius **operators** that run as tasks on the existing scheduler/worker pool — there is **no
 dedicated thread pool** for them — while `nixl`'s own threads live in the **wrapper**. So each channel is a
 hand-off between wrapper threads and engine worker threads. The bounded channel is also the source's and
-sink's **only** queue — the operators hold no separate internal buffer.
+sink's **only** queue — the operators hold no separate internal buffer. Any batching helper layered on top
+of the sink, such as the per-destination coalescer in §4, must be treated as an explicit bounded/accounted
+stage rather than as a hidden unbounded queue.
 
 ```mermaid
 flowchart LR
@@ -216,35 +218,41 @@ flowchart LR
 
 ## §4 The data path: in → compute → out
 
-A batch received by `nixl` lands on an input bounded channel; the streaming source publishes it into the
-pipeline and drives task creation as batches arrive; the GPU executor reserves memory **before** dispatching
-and runs on a per-device stream pool; the streaming sink emits results incrementally; the partitioned sink
-(#838) splits partitions across **per-destination** output channels so one slow receiver cannot
-head-of-line-block the others; `nixl` sends each output batch. Because partitioning fragments each batch into
-many small per-destination slices, the sink **coalesces** a destination's slices (a GPU concat) up to a size
-or time threshold before flushing one combined batch to its channel — otherwise we ship a flood of tiny RDMA
-transfers.
+A batch received by `nixl` is first wrapped as a repository-registered `data_batch`; the input bounded
+channel carries a reference/handle to that registered batch. The streaming source consumes those channel
+entries, publishes work into the pipeline, and drives task creation as batches arrive; the GPU executor
+reserves memory **before** dispatching and runs on a per-device stream pool; the streaming sink emits results
+incrementally; the partitioned sink (#838) splits partitions across **per-destination** output channels so one
+slow receiver cannot head-of-line-block the others; `nixl` sends each output batch. Because partitioning
+fragments each batch into many small per-destination slices, the sink **coalesces** a destination's slices up
+to a size or time threshold before flushing one combined batch to its channel — otherwise we ship a flood of
+tiny RDMA transfers. That coalescing accumulator is itself bounded and repository-visible, so accumulated
+slices remain accounted and spillable while they wait to be concatenated/flushed.
 
 The unit that flows everywhere is a `shared_ptr<cucascade::data_batch>` wrapping a `cudf::table` (GPU device
-buffers), guarded by a 3-state reader/writer lock (`idle` / `read_only` / `mutable_locked`). Hand-off across
-a channel is **zero-copy**: the `shared_ptr` moves, the device memory does not.
+buffers) or, for a host-staged receive, a host representation that will be materialized to GPU before compute,
+guarded by a 3-state reader/writer lock (`idle` / `read_only` / `mutable_locked`). Hand-off across a channel
+is **zero-copy with respect to the channel itself**: the `shared_ptr`/handle moves, the underlying
+representation does not move until an explicit tier conversion, copy-out, or coalescing concat requires it.
 
 ```mermaid
 flowchart LR
   subgraph remote["Upstream CN(s)"]
     nixlIn["nixl recv"]:::priorart
   end
-  nixlIn -->|"data_batch (zero-copy shared_ptr)"| inCh["input bounded channel"]:::proposed
   subgraph sirius["Sirius engine (this CN)"]
     direction LR
+    repo["exchange input repository<br/>registered data_batch"]:::exists
+    inCh["input bounded channel<br/>(refs / handles)"]:::proposed
     src["streaming source — #836<br/>drives task creation"]:::proposed
-    repo["shared_data_repository<br/>shared_ptr&lt;data_batch&gt;"]:::exists
+    pipeRepo["pipeline repositories<br/>shared_ptr&lt;data_batch&gt;"]:::exists
     sched["task_scheduler + GPU executor<br/>reserve-before-dispatch · stream pool"]:::exists
     sink["streaming sink — #837"]:::proposed
     part["partition — #838"]:::proposed
-    coal["coalesce per destination<br/>(concat to threshold)"]:::proposed
-    inCh --> src --> repo --> sched --> sink --> part --> coal
+    coal["bounded coalesce per destination<br/>(repo-visible concat to threshold)"]:::proposed
+    repo -->|"ref / handle"| inCh --> src --> pipeRepo --> sched --> sink --> part --> coal
   end
+  nixlIn -->|"wrap + register"| repo
   coal -->|"per-destination"| outCh["output bounded channels"]:::proposed
   outCh --> nixlOut["nixl send"]:::priorart
   classDef exists fill:#e8f5e9,stroke:#2e7d32,stroke-width:2px,color:#1a1a1a;
@@ -334,15 +342,16 @@ memory **reservations block** when the budget is exhausted, and the GPU executor
 
 ## §5 Cross-node transport with nixl
 
-`nixl` (NVIDIA Inference Xfer Library) moves GPU buffers directly between CNs. At startup each CN creates one
-agent and computes its **local metadata** (`get_local_md`, covering its registered staging memory). The first
-transfer to a new peer exchanges that agent metadata over the side-channel, and both sides **cache it per
-peer** (invalidated on restart) — so later transfers skip the handshake. For each transfer the sender then
-exchanges **buffer** metadata over the side-channel; the receiver leases a **pre-registered staging buffer**
-(GPU or host; bump-allocated, RAII lease) and returns destination addresses; the sender issues a GPUDirect RDMA
-transfer and polls for completion; a `transfer_complete` handshake lets the receiver wrap the buffer as a
-`data_batch` and push it onto the destination stream's input channel. If `nixl` is unavailable (or for a
-self-transfer), it falls back to a bRPC/CPU path.
+`nixl` (NVIDIA Inference Xfer Library) moves registered buffers between CNs. The preferred path is
+GPU-to-GPU over GPUDirect RDMA; a deployment may also use host staging as the registered receive tier. At
+startup each CN creates one agent and computes its **local metadata** (`get_local_md`, covering its registered
+staging memory). The first transfer to a new peer exchanges that agent metadata over the side-channel, and
+both sides **cache it per peer** (invalidated on restart) — so later transfers skip the handshake. For each
+transfer the sender then exchanges **buffer** metadata over the side-channel; the receiver leases a
+**pre-registered staging buffer** (GPU or host; bump-allocated, RAII lease) and returns destination addresses;
+the sender issues the transfer and polls for completion; a `transfer_complete` handshake lets the receiver
+wrap the staged bytes as a repository-registered `data_batch` and enqueue a reference on the destination
+stream's input channel. If `nixl` is unavailable (or for a self-transfer), it falls back to a bRPC/CPU path.
 
 ```mermaid
 sequenceDiagram
@@ -350,7 +359,7 @@ sequenceDiagram
   participant TX as Sender CN (streaming sink)
   participant SC as side-channel (NixlMetadataService)
   participant RX as Receiver CN
-  participant RB as Receiver GPU staging buffer
+  participant RB as Receiver staging buffer (GPU or host)
   Note over TX,RX: startup — each agent computes local metadata (get_local_md), covering registered staging memory
   opt first transfer to a new peer
     TX->>SC: send local agent metadata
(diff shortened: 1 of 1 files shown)


experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,489 @@
...
+become a **streaming source** at the bottom and a **streaming sink** at the top. A **leaf** fragment sources
+from a scan; an **intermediate** fragment sources from an exchange stream fed by an upstream fragment's sink
+on another CN.
+
+```mermaid
+flowchart TB
+  tp["TExecPlanFragmentParams (thrift)"]:::proposed
+  tr["translate_fragment() — #841"]:::merged
+  sub["Substrait Plan"]:::proposed
+  dd["DuckDB plan"]:::exists
```
johanpel, line 127: Is the DuckDB plan itself still used at some point during execution or just an intermediate artifact?

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,489 @@
...
+
+---
+
+## §3 The public streaming API — stream session
+
+The **stream session** (#839) is the public entry point. It builds a streaming plan and starts it on the
+**existing** `task_scheduler` *without blocking* the caller. The wrapper feeds inputs with
+`push(stream_id, batch)` and signals end-of-stream with `close_input(stream_id)`; it drains outputs with
+`pull()` / `wait()`. Inputs and outputs are **bounded channels** of `shared_ptr<cucascade::data_batch>`;
+channel entries are **repository-registered** `data_batch`es (not bare buffers) so batches queued under
```
johanpel: IIUC the repository (is this the cucascade repository?) shares ownership of batches with the channel. Is there anything else holding on to those batches?

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,489 @@
...
+```
+
+*Refs: operator hints `src/include/op/sirius_physical_operator.hpp:476,492`; reserve-before-dispatch +
+stream pool `src/pipeline/gpu_pipeline_executor.cpp:122,228`; data unit + lock states
+`cucascade/include/cucascade/data/data_batch.hpp:49,69-267`; port barriers PIPELINE/PARTIAL/FULL
+`sirius_physical_operator.hpp:51`.*
+
+### Overlap
+
+Because the source is multi-shot and the sink is incremental, the three stages overlap on the timeline:
```
johanpel, line 354: Could you clarify what is meant by multi-shot and incremental here? Is this because the batches are sourced from multiple CNs? What about the sink?

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,489 @@
...
+```
+
+### Backpressure
+
+The bounded channels are not just buffers — they are the flow-control mechanism, and pressure propagates
+**backwards**, against the direction of data flow. When a downstream CN is slow, the sender's output channel
+fills; the streaming sink can no longer enqueue, so that pipeline stops being scheduled; the input channel
+drains more slowly and fills; `push(stream_id, batch)` then blocks (or signals not-ready), so the wrapper
+stops draining `nixl` receives, which throttles the upstream sender. The result is end-to-end, cross-node
+flow control with a bounded in-flight working set.
```
johanpel: I suppose this is not a concern of this layer, but could a sink deadlock the system? If so, can we somehow help sinks (or other parts) detect this situation?

johanpel (commented):
Just some very basic questions here for my understanding / out of interest:

experimental/starrocks/docs/exchange-design.md
mbrobbel: Thanks for the review and great questions.

It's an intermediate artifact. See https://github.com/sirius-db/sirius/issues/841#issuecomment-4601816621.

experimental/starrocks/docs/exchange-design.md
mbrobbel: Multi-shot: see #836. Incremental: see #837.

experimental/starrocks/docs/exchange-design.md
mbrobbel: > is this the cucascade repository?
Yes

> Is there anything else holding on to those batches?

Could be, yes. The repository holds it, but other `shared_ptr` copies exist in the channel, in flight tasks (during computation), the downgrade executor and when we are sending we could hold on to it (however for sending we can probably use https://github.com/NVIDIA/cuCascade/pull/148). I'll clarify this.

experimental/starrocks/docs/exchange-design.md
mbrobbel: Yes, that's a good point. I'll add a note to open questions about it.

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,513 @@
...
+*Refs: `#841` (Substrait→DuckDB→Sirius lowering); scan operators bind a `split_connector` today; the
+streaming source/sink are net-new (#836/#837).*
+
+---
+
+## §3 The public streaming API — stream session
+
+The **stream session** (#839) is the public entry point. It builds a streaming plan and starts it on the
+**existing** `task_scheduler` *without blocking* the caller. The wrapper feeds inputs with
+`push(stream_id, batch)` and signals end-of-stream with `close_input(stream_id)`; it drains outputs with
```
bwyogatama, line 198: Where do we got this stream from?

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,513 @@
...
+
+---
+
+## §4 The data path: in → compute → out
+
+A batch received by `nixl` lands on an input bounded channel; the streaming source publishes it into the
+pipeline and drives task creation as batches arrive; the GPU executor reserves memory **before** dispatching
+and runs on a per-device stream pool; the streaming sink emits results incrementally; the partitioned sink
+(#838) splits partitions across **per-destination** output channels so one slow receiver cannot
+head-of-line-block the others; `nixl` sends each output batch.
```
bwyogatama: I am not sure I'm following the interaction between `streaming_sink` and the `task_scheduler` and `gpu_executor` here. Does the executor push data batch to a data repository and then `streaming_sink` pulls from it?

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,513 @@
...
+    GPU->>GPU: reserve + compute batch N+1
+  end
+  Note over IN,OUT: receive(N+1) overlaps compute(N) overlaps send(N-1)
+```
+
+### Backpressure
+
+The bounded channels are not just buffers — they are the flow-control mechanism, and pressure propagates
+**backwards**, against the direction of data flow. When a downstream CN is slow, the sender's output channel
+fills; the streaming sink can no longer enqueue, so that pipeline stops being scheduled; the input channel
```
bwyogatama: How do we stop pipeline from being scheduled when streaming sink can no longer enqueue? I think this is related to my previous question on the interaction between the `task_scheduler` and the streaming sink

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,513 @@
...
+
+`nixl` (NVIDIA Inference Xfer Library) moves GPU buffers directly between CNs. At startup each CN creates one
+agent and computes its **local metadata** (`get_local_md`, covering its registered staging memory). The first
+transfer to a new peer exchanges that agent metadata over the side-channel, and both sides **cache it per
+peer** (invalidated on restart) — so later transfers skip the handshake. For each transfer the sender then
+exchanges **buffer** metadata over the side-channel; the receiver leases a **pre-registered GPU staging
+buffer** (bump-allocated, RAII lease) and returns destination addresses; the sender issues a GPUDirect RDMA
+transfer and polls for completion; a `transfer_complete` handshake lets the receiver wrap the buffer as a
+`data_batch` and push it onto the destination stream's input channel. If `nixl` is unavailable (or for a
+self-transfer), it falls back to a bRPC/CPU path.
```
bwyogatama: I would like to understand better the threading model of streaming sink and streaming source. Do they have their own thread pool? And does nixl has its own internal thread pool?

experimental/starrocks/docs/exchange-design.md
bwyogatama: Also does streaming sink has its own internal queue?

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,513 @@
...
+  loop poll until done
+    TX->>TX: get_xfer_status
+  end
+  TX->>SC: transfer_complete
+  SC->>RX: transfer_complete
+  RX->>RX: wrap buffer as data_batch -> push to input channel
+  alt nixl unavailable (try_new -> None) or self-transfer
+    TX->>RX: bRPC / CPU fallback (serialize + send PBlock)
+  end
+```
```
bwyogatama: We should add retry mechanism to the open question. One idea is to retry once with nixl (assuming streaming sink has its own queue and thread pool), if failed then fallback to bRPC, and if that failed then finally we can signal query failure

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,513 @@
...
+peer-md cache `nixl_exchange.rs:207-208`; buffer-metadata RPC `NixlMetadataService::exchange_metadata`
+`doris/crates/doris-rpc/src/nixl_service.rs:60-65`; RDMA transfer (`create_xfer_req` + `post_xfer_req` +
+`get_xfer_status`) `nixl_exchange.rs:629-649`; staging buffer `gpu_staging_buffer.rs`; bRPC fallback
+`nixl_exchange.rs:226-227`.*
+
+---
+
+## §6 Resource management — three options
+
+The CN wrapper holds GPU memory for `nixl` staging and receive buffers; the Sirius context holds GPU memory
```
bwyogatama, line 504: I can see there are cases where we would want the staging buffer to be on HOST, depending on the topology, especially for GB and VR types GPU. Can nixl support this?

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,513 @@
...
+> `src/include/op/sirius_physical_operator.hpp:350,371,361,373,476,492`).
+
+---
+
+## §4 The data path: in → compute → out
+
+A batch received by `nixl` lands on an input bounded channel; the streaming source publishes it into the
+pipeline and drives task creation as batches arrive; the GPU executor reserves memory **before** dispatching
+and runs on a per-device stream pool; the streaming sink emits results incrementally; the partitioned sink
+(#838) splits partitions across **per-destination** output channels so one slow receiver cannot
```
bwyogatama: Also I think we would want to be able to concat data before sending it as well since there is a pretty good chance that the data is in tiny chunks after the partition

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,513 @@
...
+batches and spills any that are idle (it filters on `idle` state, not batch origin —
+`src/include/data/convertible_data_batch.hpp:303`; candidate sweep
+`src/downgrade/downgrade_executor.cpp:206-252`). So option A takes **two** changes, not one:
+
+1. **Borrow the manager** — the Sirius context uses the wrapper's `cucascade::memory_reservation_manager`
+   instead of constructing its own (`src/sirius_context.cpp:336`).
+2. **Make exchange memory visible to it** — queued in/out batches must be held as `data_batch`es in a
+   repository registered with that manager (not in opaque bounded channels), and `nixl`'s
+   registered/staging GPU buffers — which are RAII leases, not repository batches, so the sweep never sees
+   them — must be reserved against the same budget explicitly. Without this, a shared manager still leaves
```
bwyogatama, line 560: So correct me if I'm wrong but there are two types of reservation here:
1. cucascade reservation which will reserve the pre-registered nixl buffer during initial startup, and
2. reserving/leasing the nixl staging buffer when sending or receiving data. For this one, this doesn't go through cucascade since this whole buffer is already pre-registered, and has its own reservation mechanism.
Is this correct?

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,513 @@
...
+  classDef priorart fill:#ede7f6,stroke:#6a1b9a,stroke-width:1.5px,stroke-dasharray:2 2,color:#1a1a1a;
+  classDef merged fill:#e3f2fd,stroke:#1565c0,stroke-width:2px,color:#1a1a1a;
+```
+
+*Refs: `#841` (Substrait→DuckDB→Sirius lowering); scan operators bind a `split_connector` today; the
+streaming source/sink are net-new (#836/#837).*
+
+---
+
+## §3 The public streaming API — stream session
```
bwyogatama, line 194: Something that is a bit unclear to me here is the boundary between the wrapper and sirius engine.
For example, does streaming sink and streaming source lives in the wrapper or in sirius engine?

experimental/starrocks/docs/exchange-design.md
bwyogatama: Because if the streaming sink is an independent component and just pulls from a repository, this is trickier since the task scheduler has no information about the backpressure in the streaming sink

ran-yuan-rui:
One liveness risk I think we should make explicit is that the scarce resource may not just be total GPU memory, but NIXL receive-staging capacity / leases.

The doc says the receiver leases a pre-registered staging buffer and then wraps the received buffer as a `data_batch` before pushing it to the input channel. If the intended fast path keeps the received batch backed by that staging lease until the batch is consumed or downgraded, then generic GPU spilling is not sufficient unless it is also able to reclaim staging capacity.

The current downgrade model appears to select idle repository batches by memory space, not by the underlying pool/lease origin. That can reduce general GPU pressure, but it does not necessarily release a fixed NIXL staging lease unless the exchange batch's GPU representation owns that lease and dropping/converting the representation returns it to the staging allocator. I think the design should state this explicitly.

This also affects the deadlock argument. A logical stage DAG does not by itself rule out a physical all-to-all lease cycle: CN A can have its receive-staging pool filled by batches from CN B, stop granting receive leases, while CN B is holding pinned output waiting to send to A, and the same can happen in reverse. Per-destination output channels help isolate slow receivers at the queue level, but a fully shared memory budget can re-couple them at the budget/staging level.

My recommendation would be Option A plus two additional constraints:

- make exchange spill pool-aware, so an idle received batch can be copied out of NIXL staging and release the staging lease;
- reserve a non-borrowable receive-staging escape floor / credit budget that compute cannot consume, so the system can always make progress on copy-out or receive-lease reclamation.

In other words, I still like the shared-manager direction, but I do not think MRU/LRU is the first-order question. The first-order contract is whether backpressure and spilling can reclaim the specific staging leases needed for all-to-all progress.

experimental/starrocks/docs/exchange-design.md
mbrobbel: For intermediate fragments the stream is coming from upstream fragments (other CNs via NIXL). We wrap those buffers and push the batches in the matching input channel.

experimental/starrocks/docs/exchange-design.md
mbrobbel: The streaming source/sink are operators in Sirius. There is no special or dedicated pool for them.

For NIXL we'll use threads that live in the wrapper.

experimental/starrocks/docs/exchange-design.md
mbrobbel: The streaming sink and source are operators in Sirius, so they live inside the Sirius engine. The boundary between the wrapper and the engine is the channel that is exposed via FFI to push and pull batches.

For the streaming source, the wrapper is the producer and the operator is the consumer. For the sink it's the other way around. Batches are handed over on this boundary.

experimental/starrocks/docs/exchange-design.md
mbrobbel: Yes - I'll add this to the doc.

experimental/starrocks/docs/exchange-design.md
mbrobbel: Yes we can do this. NIXL supports both host and device memory and we can make the staging buffer topology-dependent.

experimental/starrocks/docs/exchange-design.md
mbrobbel: Sounds good - I'll add something to the doc.

experimental/starrocks/docs/exchange-design.md
mbrobbel: Streaming sink is the final operator of a pipeline (in a streaming session). Compute results are pushed to a shared repository and the sink operators pulls from this and pushes the batches in the output channel.

Pushes to the channel require a schedule condition that fails when the channel is full, so backpressure flows from there; no sink tasks are scheduled so the shared repository starts to fill up.

I'll try to clear this up in the docs.

experimental/starrocks/docs/exchange-design.md
mbrobbel: The channel is bounded, and streaming sink tasks can't be scheduled when it's full. See above.

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,613 @@
...
+
+The **stream session** (#839) is the public entry point. It builds a streaming plan and starts it on the
+**existing** `task_scheduler` *without blocking* the caller. The wrapper feeds inputs with
+`push(stream_id, batch)` and signals end-of-stream with `close_input(stream_id)`; it drains outputs with
+`pull()` / `wait()`. Inputs and outputs are **bounded channels** of *handles* (repository batch-ids) to
+repository-registered `cucascade::data_batch`es. The **owner of record** is the cuCascade
```
9prady9, line 200: what is the definition of channel here ? Is it a dedicated connection between two distinct CNs or is it meant to describe just a exchange here ?

9prady9 (commented):
very good plan, I am a bit unclear on whats channel but I guess I will get more clarity when we go deep into the implementation details.

👍🏾

experimental/starrocks/docs/exchange-design.md
mbrobbel: Message-based queues with fixed capacity. These are between the operator and the wrapper.

9prady9: As in something like ZeroMQ ?

experimental/starrocks/docs/exchange-design.md
bwyogatama: Sorry I am not following, i assume this is cuda stream right? or is it something else?
If it's cuda stream, do we have a dedicated stream pool for this or something else?

experimental/starrocks/docs/exchange-design.md
bwyogatama: So if the streaming sink is just an operator, this means that it will run inside the gpu pipeline executor? It is actually just another pipeline that is scheduled by the task scheduler?

I understand how the backpressure mechanism work, but this also means that the task scheduler has to be aware that the channel is full somehow even before scheduling the streaming sink task, otherwise it will only know when streaming sink pushes it to the bounded channel.

bwyogatama: also emphasizing it in the docs that streaming sink and streaming source are just a sirius operator would be helpful, i didn't get that part before

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,613 @@
...
+partitions across **per-destination** output channels so one slow receiver cannot head-of-line-block the
+others, and it coalesces a destination's small slices up to a size or time threshold before flushing one
+combined batch to that destination's channel — otherwise we ship a flood of tiny RDMA transfers. That
+coalescing state is bounded and repository-visible, so accumulated slices remain accounted and spillable
+while they wait to be concatenated/flushed. `nixl` sends each flushed output batch.
+
+**Partition hash compatibility (correctness-critical).** The FE plans shuffle and bucket joins assuming a
+specific partition function, so the sink must reproduce StarRocks' hash **bit-exactly, per partition type** —
+get it wrong and co-located rows land on the wrong CN, silently corrupting shuffle joins/aggregations.
+StarRocks' `exchange_sink_operator` uses **three** regimes (`be/src/exec/pipeline/exchange/exchange_sink_operator.cpp`):
```
wmalpica, line 292: I can understand why its important for the hash partitioning to have to match what StarRocks is expecting. But, do we really want the partitioning operation to execute outside of Sirius? I think we probably want the partitioning to be done by Sirius, this way GPU compute all falls under Sirius. Otherwise we are sharing not only GPU memory consumption, but also compute. Compute contention can lead to slowdowns.

But even if the rust wrapper layer does the partitioning and coalescing, what do the plan fragments on either side look like?
In Sirius today a group by looks like: sirius_physical_grouped_aggregate ->sirius_physical_partition -> sirius_physical_grouped_aggregate_merge.
would it look like:
A:
sirius_physical_grouped_aggregate ->partitioned_streaming_sync->NIXL->streaming_source->sirius_physical_grouped_aggregate_merge 
B:
sirius_physical_grouped_aggregate ->sirius_physical_partition -> sirius_physical_grouped_aggregate_merge ->partitioned_streaming_sync->NIXL->streaming_source->sirius_physical_grouped_aggregate ->sirius_physical_partition -> sirius_physical_grouped_aggregate_merge
C:
sirius_physical_grouped_aggregate ->sirius_physical_partition -> sirius_physical_grouped_aggregate_merge ->partitioned_streaming_sync->NIXL->streaming_source->starrocks_aggregate_merge

experimental/starrocks/docs/exchange-design.md
```diff
@@ -0,0 +1,613 @@
...
+
+## §4 The data path: in → compute → out
+
+A batch received by `nixl` is first wrapped as a repository-registered `data_batch`; the input bounded
+channel carries a handle to that registered batch. The streaming source consumes those channel
+entries, publishes work into the pipeline, and drives task creation as batches arrive; the GPU executor
+reserves memory **before** dispatching and runs on a per-device stream pool; the streaming sink emits results
+incrementally; the partitioned sink (#837) owns both partitioning and per-destination coalescing. It splits
+partitions across **per-destination** output channels so one slow receiver cannot head-of-line-block the
+others, and it coalesces a destination's small slices up to a size or time threshold before flushing one
```
wmalpica, line 275: Note that we are also thinking about adding coalescing as part of our partitioning for Sirius SNMG

experimental/starrocks/docs/exchange-design.md
mbrobbel: Sorry, stream is an overloaded term here. I'm talking about the batches that stream from other computes nodes into this one. The wrapper maps those incoming batches to the corresponding streaming source operator. It knows how to do this because it mapped the plan fragment, and drives the exchange for it to push the batches to the corresponding streaming source operators.

experimental/starrocks/docs/exchange-design.md
mbrobbel: > So if the streaming sink is just an operator, this means that it will run inside the gpu pipeline executor? It is actually just another pipeline that is scheduled by the task scheduler?

Yes.

> but this also means that the task scheduler has to be aware that the channel is full somehow even before scheduling the streaming sink task

Yes, this is the task-creation-hint mechanism. I'll update the docs.

> also emphasizing it in the docs that streaming sink and streaming source are just a sirius operator would be helpful, i didn't get that part before

Updating the docs. Note we also link to the corresponding issues:
- https://github.com/sirius-db/sirius/issues/836
- https://github.com/sirius-db/sirius/issues/837

experimental/starrocks/docs/exchange-design.md
mbrobbel: Great, we can re-use that here.

experimental/starrocks/docs/exchange-design.md
mbrobbel: > But, do we really want the partitioning operation to execute outside of Sirius?

No - the plan is to partition + coalesce inside the GPU in Sirius (in the streaming sink operator). There is no additional GPU compute in the wrapper.

It would look like A. I'll update the docs.

PR #918, opened 2026-06-10
Implements using rebind_stream

wmalpica:
Currently all deallocations happen on the stream on which they allocations were made. This PR adds an API which allows for changing that stream, so that deallocations can be made to happen on the active stream. This API is used when converting a data_batch from GPU to Host so that the deallocations happen on the downgrade executor's stream and not block other task execution streams. It also uses the rebind_stream API when preparing the data for a task to be executed. That way any  input data that is freed, gets deallocated there as well, minimizing  peak memory consumption.

This PR depends on https://github.com/NVIDIA/cuCascade/pull/146

This PR additionally ensures that only one downgrade executor monitor request can exist at a time. This prevents from too many back to back requests. 

It also improves logging of memory state at the beginning and end of a query so that we can monitor and check for memory leaks.

### DIFF
--- a/src/downgrade/downgrade_executor.cpp
+++ b/src/downgrade/downgrade_executor.cpp
@@ -428,7 +428,8 @@ void downgrade_executor::monitor_loop()
   bool backed_off = false;
 
   while (_running.load()) {
-    if (_memory_space && _memory_space->should_downgrade_memory()) {
+    if (_memory_space && _memory_space->should_downgrade_memory() &&
+        !_monitor_request_enqueued.load(std::memory_order_relaxed)) {
       // Stateless viability gate: only issue a downgrade request when one could plausibly free
       // memory. When idle GPU batches' only lower tier is a full HOST and no DISK is configured,
       // re-firing would just re-scan every repository and the task queue, free nothing, and spam
--- a/src/downgrade/downgrade_executor.cpp
+++ b/src/downgrade/downgrade_executor.cpp
@@ -428,7 +428,6 @@ void downgrade_executor::monitor_loop()
   bool backed_off = false;
 
   while (_running.load()) {
-<<<<<<< HEAD
     if (_memory_space && _memory_space->should_downgrade_memory()) {
       // Stateless viability gate: only issue a downgrade request when one could plausibly free
       // memory. When idle GPU batches' only lower tier is a full HOST and no DISK is configured,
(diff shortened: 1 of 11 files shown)


aminaramoon (approved):

src/sirius_context.cpp
```diff
@@ -169,22 +170,41 @@ SiriusContext::~SiriusContext() noexcept
...
-               fs_mr->get_peak_total_allocated_bytes(),
-               fs_mr->get_free_blocks());
+
+  // Host pinned pool (fixed_size_host_memory_resource).
+  for (auto const* space :
+       memory_manager_->get_memory_spaces_for_tier(cucascade::memory::Tier::HOST)) {
+    auto* fs_mr =
+      space->get_memory_resource_as<cucascade::memory::fixed_size_host_memory_resource>();
+    if (fs_mr) {
+      spdlog::info("[host_pool] HOST:{} {} allocated={} bytes peak={} bytes free_blocks={}",
```
bwyogatama, line 182: I think is not the pattern that your `pattern.py` check for query boundaries btw, they check for `[host_pool] QueryBegin allocated=, and HOST_POOL_RE`

src/downgrade/downgrade_executor.cpp
```diff
@@ -346,6 +346,9 @@ void downgrade_executor::processing_loop()
     double throughput_mbs =
       (duration_ms > 0.0) ? (total_bytes / (1024.0 * 1024.0)) / (duration_ms / 1000.0) : 0.0;
     std::string request_label = req->is_monitor_request ? "monitor " : "";
+    if (req->is_monitor_request) {
+      _monitor_request_enqueued.store(false, std::memory_order_relaxed);
```
bwyogatama, line 350: What is this for? it's just being set to true and false but never used

src/sirius_context.cpp
```diff
@@ -169,22 +170,41 @@ SiriusContext::~SiriusContext() noexcept
...
+    }
+  }
+
+  // GPU device memory pools (reservation_aware_resource_adaptor), one line per GPU.
+  for (auto const* space :
+       memory_manager_->get_memory_spaces_for_tier(cucascade::memory::Tier::GPU)) {
+    auto* ra_mr =
+      space->get_memory_resource_as<cucascade::memory::reservation_aware_resource_adaptor>();
+    if (!ra_mr) { continue; }
+    spdlog::info("[gpu_pool] GPU:{} {} allocated={} bytes peak={} bytes reserved={} bytes",
```
bwyogatama, line 197: Need to change the segmenter.py too cause it assumes the SQL line is right after the first `QueryBegin`, where this now add multiple `QueryBegin` before the first SQL line

bwyogatama (approved):

src/downgrade/downgrade_executor.cpp
wmalpica: Good catch. Some of the code disappeared when I rebased. This is intended to make it so that we only enqueue one memory request at a time. This is because it will request to downgrade to free a certain amount (to reach its target threshold). But if other downgrade requests get added, we can end up with multiple downgrade requests all requesting the same, so we can end up over downgrading.

PR #923, opened 2026-06-11
S3 change to align with scan-manager cleanup (#913) 

ran-yuan-rui:
With scan-manager cleanup (#913) merged, Running the full `[s3]` suite against the merged result surfaced a handful of regressions and coverage gaps, all fixed here. And align comments with current code.

### Crash fixes

- **`make s3-test` died at 93%**: `sirius_scan_manager::create_datasource` was `noexcept` but performs an S3 HEAD via `create_io_object`, so a missing key (HTTP 404) terminated the whole process instead of surfacing a query error. Dropped the `noexcept` and documented the throwing contract; the existing "describe_parquet surfaces S3 missing-object errors cleanly" test covers it.
- **`make s3-test-large` SIGSEGV** in the S3 async backend's CUDA copy callback: `cuda_copy_done` touched the transfer and the reactor after the `cuda_done` release-store — the store is what licenses the reactor thread to delete the transfer (and, via shutdown, tear down the curl multi handle). The release-store is now the callback's last access; the wakeup fires before it, while `_copying` membership still pins both lifetimes. The async device-read benchmark is unaffected (~10x blocking at 16 connections, `stream_syncs=0`).

### Restored behaviors

- **`use_sirius_datasource=false` (local cudf/KvikIO fallback)** was a no-op: the local uring claimed local paths unconditionally. The flag is now honored at the routing level in `create_datasource` — unclaimed local slices carry no datasource and materialize falls back to `cudf::io::datasource::create`. The uring itself is still always constructed (the DuckDB-native GPU scan needs it for host reads), and object-store paths are unaffected: **S3 remains GPU-only with no CPU fallback**. Explicit `file://` URIs behave exactly like bare local paths on every fallback site.
- **Chunk prewarm was inert**: `enable_chunk_prewarm` survived but no call site inserted ranges into the prefetch cache. `parquet_gpu_ingestible::run_batch` now prewarms the selected row groups' merged column-chunk byte ranges (plus parquet magic + footer) once projection and row-group pruning are final; `describe_parquet()` stays metadata-only.

### Test coverage

- The S3 scan-manager coverage dropped by the cleanup (two test files were edited but removed from `TEST_SOURCES`, so they never compiled) is selectively migrated onto the `create_datasource` seam: backend selection (`s3_use_async_backend` → async vs blocking), S3 datasource routing (an `s3://` slice always carries a non-null
  datasource, including with `use_sirius_datasource=false`), local fallback (bare path and `file://`), scan-side prewarm on/off, and a routing spy proving slice reads flow through the Sirius datasource rather than cudf's bundled fallback. The two orphaned files are deleted; everything bound to the removed APIs was rewritten, not ported.
- A nested struct/list/map S3-vs-local-oracle probe is included; it currently exercises the clean bind-time rejection (`[parquet_helpers] nested parquet types not yet supported in sirius_read_parquet`) and will flip to the oracle comparison when nested support lands.
- **New configure-time orphan-test guard**: a `test_*.cpp` present in the tree but missing from `TEST_SOURCES` now fails the configure (this class of silent coverage loss is what allowed the gap above). Eleven pre-existing orphans are allowlisted with reasons in `test/cpp/ORPHAN_ALLOWLIST.cmake` for their owners to triage — the list is meant to only ever shrink.

### Comment accuracy

Ownership and naming comments that still described the pre-cleanup world (SiriusContext owning the backends, `parquet_split_provider`/`cached_split_provider`/`io_ctx_for` references, "reserved" uring knobs that are consumed again) now describe the merged architecture.


CMakeLists.txt
```diff
@@ -565,6 +561,28 @@ if(SIRIUS_BUILD_S3_TESTS)
...
 endif()
 
+# Orphan-test guard: a test_*.cpp present in the tree but missing from
+# TEST_SOURCES never compiles, so its coverage silently vanishes (this happened
+# twice during the scan-manager cleanup). Fail the configure for any unlisted
+# test file; the only escape hatch is an explicit, reasoned entry in
+# test/cpp/ORPHAN_ALLOWLIST.cmake. Test files added behind a CMake option must
+# be appended to TEST_SOURCES inside their option block ABOVE this check.
+include("${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/ORPHAN_ALLOWLIST.cmake")
+file(
```
mbrobbel: There was a discussion about doing something like this before here https://github.com/sirius-db/sirius/pull/356#discussion_r2887304308

Maybe we can move a check like this to pre-commit?

ran-yuan-rui: Good call — #356's points against CMake globbing apply here too. Moved the check to a local pre-commit hook (scripts/check_orphan_tests.py + test/cpp/orphan_test_allowlist.txt): it now runs at commit time and in the CI lint job (no configure/CUDA needed), and additionally flags stale allowlist entries. The CMake-side guard is gone.

Bonus from the same audit: three hidden tests turned out to be unreachable by every gate (one had a latent defect that had never executed) — unhidden and fixed in the follow-up commit.

aminaramoon (approved):

PR #926, opened 2026-06-11
fix(accuracy): keep ungrouped decimal AVG out of long double

### LINKED ISSUE #288: Avoid copy to double for avg calculation ungrouped aggregate
```
se cudf::type_id::DECIMAL32: {
        auto& s  = static_cast<cudf::fixed_point_scalar<numeric::decimal32>&>(*sum_value);
        sum_host = static_cast<long double>(s.value(stream)) / denom;
        break;
      }
      case cudf::type_id::DECIMAL64: {
        auto& s  = static_cast<cudf::fixed_point_scalar<numeric::decimal64>&>(*sum_value);
        sum_host = static_cast<long double>(s.value(stream)) / denom;
        break;
      }
      case cudf::type_id::DECIMAL128: {
        auto& s  = static_cast<cudf::fixed_point_scalar<numeric::decimal128>&>(*sum_value);
        sum_host = static_cast<long double>(s.value(stream)) / denom;
        break;
```

This can introduce rounding errors so we need to fix it

9prady9:
## Summary

Fixes #288. The ungrouped AVG finalize (`make_avg_column`) reduced the merged sum/count to host scalars and divided through `long double`, with an extra `÷10^scale` rounding for decimal sums (and a `std::llround` trip for decimal output). It now divides the already-single-row sum/count columns on-device with `cudf::binary_operation`, mirroring the grouped merge path.

This removes the avoidable precision loss (two roundings instead of DuckDB's one) and the host round trip.

## Benchmark

Single-row `DECIMAL64` sum / `INT64` count → `DOUBLE`, latency mode, 5000 iters:

| Approach | µs/call |
|---|---|
| Host `long double` round trip (before) | ~131 |
| On-device divide (after) | ~19 |
| **Speedup** | **~6.9×** |

The win is mostly from dropping the blocking device→host sync (`scalar.value()`), not the arithmetic.

## Test plan

- [x] `sirius_unittest "[physical_ungrouped_aggregate]"` (SUM/MIN/MAX/COUNT + AVG-in-merge) pass.

### DIFF
--- a/src/op/sirius_physical_ungrouped_aggregate.cpp
+++ b/src/op/sirius_physical_ungrouped_aggregate.cpp
@@ -18,7 +18,6 @@
 
 #include "cudf/cudf_utils.hpp"
 #include "data/data_batch_utils.hpp"
-#include "duckdb/common/types/decimal.hpp"
 #include "expression/aggregate_id.hpp"
 #include "expression/ast/aggregate.hpp"
 #include "expression/ast/node.hpp"
@@ -29,13 +28,14 @@
 #include "op/sirius_physical_ungrouped_aggregate_merge.hpp"
 #include "sirius/exception.hpp"
 
+#include <cudf/binaryop.hpp>
 #include <cudf/column/column_factories.hpp>
 #include <cudf/copying.hpp>
-#include <cudf/fixed_point/fixed_point.hpp>
 #include <cudf/reduction.hpp>
 #include <cudf/scalar/scalar.hpp>
 #include <cudf/scalar/scalar_factories.hpp>
 #include <cudf/types.hpp>
+#include <cudf/unary.hpp>
 #include <cudf/utilities/memory_resource.hpp>
 
 #include <rmm/resource_ref.hpp>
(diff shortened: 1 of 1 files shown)


src/op/sirius_physical_ungrouped_aggregate.cpp
```diff
@@ -235,104 +234,32 @@ std::unique_ptr<cudf::column> make_avg_column(const cudf::column_view& sum_view,
...
+    cudf::cast(sum_view, cudf::data_type{cudf::type_id::FLOAT64}, stream, memory_resource);
+  auto count_f64 =
+    cudf::cast(count_view, cudf::data_type{cudf::type_id::FLOAT64}, stream, memory_resource);
+  auto avg_f64 = cudf::binary_operation(sum_f64->view(),
+                                        count_f64->view(),
+                                        cudf::binary_operator::DIV,
+                                        cudf::data_type{cudf::type_id::FLOAT64},
+                                        stream,
+                                        memory_resource);
+  if (out_type.id() == cudf::type_id::FLOAT64) { return avg_f64; }
```
mbrobbel, line 260: ~Why would this return something other than `f64`? That would be `cudf` bug right?~ I was not reading this correctly.

mbrobbel (approved):

Issue #510, opened 2026-03-23
Use new cudf mark_join in our MARK JOIN


wmalpica:
This was merged recently into cudf

https://github.com/rapidsai/cudf/pull/21440/changes#diff-43e3755b4136c3de9a1945cb9a276c246dcef44f5284faa68ae57800012f9da9

9prady9:
I have self assigned this as I did the mark join earlier. I believe this will join in `cudf` next update.

mbrobbel:
> I have self assigned this as I did the mark join earlier. I believe this will join in `cudf` next update.

With https://github.com/sirius-db/sirius/pull/535 you can test with nightly cudf.

9prady9:
## Decision: Retain `filtered_join`-based MARK join implementation

Evaluated replacing the current implementation with [`cudf::mark_join`](https://docs.rapids.ai/api/libcudf/nightly/classcudf_1_1mark__join.html) (new in cudf 26.04). Decided to keep the existing approach for the following reasons.

### 1. `cudf::mark_join` is designed for left-table reuse — which never happens here

Per the [API docs](https://docs.rapids.ai/api/libcudf/nightly/classcudf_1_1mark__join.html), `cudf::mark_join` is suited for scenarios where **the left (build) table is reused across multiple probe operations**. In Sirius, MARK join is explicitly excluded from `BUILD_PROBE` mode (see `update_join_exec_mode`), so both left and right tables arrive fresh on every `execute()` call. The primary benefit of `mark_join` — build once, probe many — never applies.

### 2. `cudf::mark_join` has the wrong build-side convention for our pipeline

`cudf::mark_join` builds on the **left** table and probes with right, returning left (build) row indices. Sirius's `BUILD_PROBE` mode builds on the **right** table and streams left batches — the opposite convention. Adopting `mark_join` in `BUILD_PROBE` mode would require inverting the entire build/probe infrastructure for MARK join specifically.

### 3. The MIXED_JOIN path requires `cudf::mixed_left_semi_join`

When the join has both equality and non-equality conditions, the code takes the `MIXED_JOIN` path and uses `cudf::mixed_left_semi_join`, which evaluates equality conditions via a hash table and inequality conditions via an AST predicate. `cudf::mark_join` has no equivalent mixed-predicate support, so it cannot replace this path without dropping correctness on non-equi MARK joins.

### 4. `cudf::filtered_join` is the recommended single-shot API

[...]

PR #927, opened 2026-06-11
feat(join): enable BUILD_PROBE mode for MARK joins

### LINKED ISSUE #921: Enable build probe mode for mark join

9prady9:
## Summary

Enables Sirius's `BUILD_PROBE` execution mode for `MARK` joins — closes #921 (follow-up to #510 / #924).

Previously `MARK` was explicitly excluded from `BUILD_PROBE` and always ran STANDARD single-shot. With a small build/filter side and a large probe side, a `MARK` join now builds **one** reusable `cudf::filtered_join` on the right (filter) keys and probes it with each streamed left batch, reusing the hash table across probes — the same reuse model the other join types already use in `BUILD_PROBE`.

## Why `cudf::filtered_join` (and not `cudf::mark_join`)

In `BUILD_PROBE` the persistent hash table lives on the **right** side and we need **left-row** match indices for the mark column:

| API | builds on | `semi_join` returns | fits BUILD_PROBE MARK? |
|---|---|---|---|
| `cudf::filtered_join` | right (filter) | probe-side = **left** indices | ✅ yes |
| `cudf::mark_join` | left | build-side = **left**, but build must be left | ❌ would require building on left, breaking reuse-on-right |

So `filtered_join` is the only one that keeps the reusable table on the right while still yielding left-row marks. Both branches funnel through the existing `resolve_mark_join_result` to scatter matches into the `BOOL8` mark.

## Changes

- Drop the `MARK` exclusion from `update_join_exec_mode` (SEMI / ANTI / RIGHT remain excluded).
- Persist `std::unique_ptr<cudf::filtered_join> _filtered_table`; build it once on the right keys in the BUILD_PROBE build phase.
- Probe phase: `_filtered_table->semi_join(probe_keys)` → `resolve_mark_join_result` to emit all left rows + the mark.
- Reset `_filtered_table` in `on_finalize_operator`.

## Testing

- New `[markjoin]` integration tests in `test_gpu_execution_tpch.cpp` cover both MARK-producing shapes — `OR + IN` and projected `IN` — over a 150k-row probe (`orders`) against a small build subset (`customer` where `c_nationkey < 3`); both compare GPU vs CPU.
- Debug logs confirm the new path runs: `using filtered_join (BUILD_PROBE MARK)`.
- Regression: `[physical_mark_join] [semijoin] [markjoin] [hash_join] [join]` — 127 cases pass.

## Merge order

Stacked on #924 — **merge after #924 lands**. Until then this PR's diff also shows #924's two commits; they drop out automatically once #924 is merged into `dev`.

### DIFF
--- a/src/include/op/sirius_physical_hash_join.hpp
+++ b/src/include/op/sirius_physical_hash_join.hpp
@@ -18,6 +18,7 @@
 
 #include "cudf/cudf_utils.hpp"
 #include "cudf/join/distinct_hash_join.hpp"
+#include "cudf/join/filtered_join.hpp"
 #include "duckdb/common/value_operations/value_operations.hpp"
 #include "duckdb/execution/join_hashtable.hpp"
 #include "duckdb/execution/operator/join/perfect_hash_join_executor.hpp"
@@ -179,6 +180,8 @@ class sirius_physical_hash_join : public sirius_physical_partition_consumer_oper
   std::unique_ptr<cudf::hash_join> _hash_table;  // hash object to be used in BUILD_PROBE mode
   std::unique_ptr<cudf::distinct_hash_join>
     _distinct_hash_table;  // used instead of _hash_table when build keys are proven unique
+  std::unique_ptr<cudf::filtered_join>
+    _filtered_table;  // reusable build-on-right semi-join object for MARK joins in BUILD_PROBE mode
   std::optional<::cucascade::read_only_data_batch>
     _build_table;  // owned build table for BUILD_PROBE mode, to materialize build side results
   std::vector<std::unique_ptr<cudf::column>>
--- a/src/op/sirius_physical_hash_join.cpp
+++ b/src/op/sirius_physical_hash_join.cpp
@@ -76,6 +76,19 @@ static cudf::filtered_join make_right_filtered_join(cudf::table_view const& righ
 #endif
 }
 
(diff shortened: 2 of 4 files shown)


9prady9:
## BUILD_PROBE MARK join — SF1 benchmark (NVIDIA L4)

MARK join, 6M-row `lineitem` probe vs ~36k-row `part` subset build:

```sql
SELECT count(*) FROM lineitem
WHERE l_orderkey < 0 OR l_partkey IN (SELECT p_partkey FROM part WHERE p_size < 10);
```

| Mode | Median | Min | Max | Result |
| --- | --- | --- | --- | --- |
| New — BUILD_PROBE MARK | **21.66 ms** | 21.20 ms | 22.59 ms | 1,087,125 |
| Old — STANDARD MARK | 22.38 ms | 21.71 ms | 23.03 ms | 1,087,125 |

3 warmup + 9 timed iterations; results identical and match DuckDB CPU.

**Takeaway:**
- single-query latency is at parity (~3%, within noise)
- Expect gains with streaming probe sides.

wmalpica (approved):
From here we should expand this so that it also applies to semi and anti joins, so that they can also operate in BUILD_PROBE mode

9prady9:
> From here we should expand this so that it also applies to semi and anti joins, so that they can also operate in BUILD_PROBE mode

Will create tickets for them too.

PR #936, opened 2026-06-12
perf(scan): typed segment-tree walk for native metadata

kevkrist:
# Replace string-based GetColumnSegmentInfo with a typed segment-tree walk

Walk the DuckDB segment trees directly in the native-scan metadata walk (typed `block_id`/compression/counts/validity child/max-string-length) instead of calling `GetColumnSegmentInfo`, which builds a per-segment compression string, column-path string, and a `statistics.ToString()` blob that we then re-parse. All 22 TPC-H SF300 queries validate against DuckDB CPU.

## Performance (TPC-H SF300, GB200, native DuckDB scan, same-build A/B)

- Metadata-walk wall time per query (nsys NVTX `native_metadata_segment_info`): Q6 **76 → 9 ms (8.4×)**, Q21 **268 → 22 ms (12.2×)**.
- End-to-end warm runtime, all 22 queries: **13.53 s → 12.29 s (−9.1%)**, largest per-query gains on varchar / heavy-segment queries: Q11 −55.7%, Q6 −21.5%, Q1 −12.5%, Q19 −9.8%, Q9 −9.0%.

### DIFF
--- a/src/op/scan/duckdb_native_metadata.cpp
+++ b/src/op/scan/duckdb_native_metadata.cpp
@@ -21,7 +21,9 @@
 #include <nvtx3/nvtx3.hpp>
 
 #include <duckdb/common/column_index.hpp>
+#include <duckdb/common/enums/compression_type.hpp>
 #include <duckdb/common/enums/filter_propagate_result.hpp>
+#include <duckdb/function/compression_function.hpp>
 #include <duckdb/function/partition_stats.hpp>
 #include <duckdb/main/attached_database.hpp>
 #include <duckdb/planner/table_filter.hpp>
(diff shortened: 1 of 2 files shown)


src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -140,32 +112,87 @@ bool is_supported_logical_type(const sirius::logical_type& type, std::string& re
...
-  desc.additional_blocks = seg.additional_blocks;
-  desc.block_offset      = seg.block_offset;
-  desc.segment_start     = seg.segment_start;
-  desc.segment_count     = seg.segment_count;
-  desc.compression       = compression;
-  return desc;
+  auto* std_col = dynamic_cast<duckdb::StandardColumnData*>(&col_data);
+  if (!std_col) {
+    return "column " + std::to_string(column_id) + " row group " + std::to_string(rg_idx) +
+           ": column storage is not StandardColumnData (nested/unsupported)";
```
Jedi18: nit: Does col_data have GetType() or some similar function so that we have a more descriptive error message?

Jedi18 (approved):

PR #937, opened 2026-06-12
Add Simpatico as the compression subsystem

### LINKED ISSUE #947: Allow Pinned Host Tables to be compressed using Simpatico
The compression format should be set statically when you request a column to be pinned

joosthooz:
This copies in the sources of a project we've been working on called Simpatico. 

It is a compression system aimed at tabular data, that can apply a tree of operations to the columns. Each operator can have multiple outputs (like RLE produces values and run counts), each of which can be fed into additional downstream operators. These sources are basically a full rewrite of something @felipeblazing started, so that it is a C++ module/library with clear APIs. It has a code generation backend that will fuse kernels together to provide very good performance. 

The integration into Sirius allows compressing input tables that are pinned to host or gpu memory. It needs a directory with explicit plan files, 1 for each input table. If no file is found for an input table, it will not be compressed. There is a driver binary that has benchmarking and exploration functionality (for finding good plans for a given file).

Compressing data to be spilled or distributed to other workers is left as follow-up work.

Closes #947


pixi.toml
```diff
@@ -37,6 +37,7 @@ pkg-config = "*"
 mold = "*"
 libprotobuf = "*"
 rust = "*"
+cuda-nvrtc-dev = ">=12.9.86,<14"
```
mbrobbel: This pins via `cuda-version`. Also needs `pixi.lock` update.
```suggestion
cuda-nvrtc-dev = "*"
```

substrait
mbrobbel, line 1: I think this is unintentional?

src/include/sirius_config.hpp
```diff
@@ -130,6 +130,42 @@ struct telemetry_config {
...
+  /// Minimum chunk size (uncompressed bytes) below which compression is
+  /// skipped and the chunk is stored uncompressed.  0 = no threshold.
+  std::size_t min_batch_size_bytes{1ULL * 1024 * 1024};  // 1 MiB
+
+  /// Maximum compressed footprint, as a fraction of the batch's original device
+  /// size, for the compressed form to be kept.  When the compressed header +
+  /// payload exceeds this fraction of the original (i.e. compression saved too
+  /// little), the compressed data is discarded and the batch is pinned
+  /// uncompressed.  Default 0.95 (require at least a 5% saving).
+  double max_compressed_fraction{0.95};
```
wmalpica: This seems too high. With only a 5% saving I would imagine it would always be more costly to use the compression. I would have expected this threshold to be defaulted to something more like 0.5 (50%).

wmalpica (commented):
Mainly have just one issue https://github.com/sirius-db/sirius/issues/1211 which can be addressed after this merged, but its an issue of relatively high importance IMHO

src/compression/simpatico_codegen/include/api/simpatico_codegen.hpp
```diff
@@ -0,0 +1,163 @@
...
+#include <memory>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <vector>
+
+namespace simpatico {
+
+/// A single column after compression: metadata plus the compressed plan compound.
+struct compressed_column {
```
bwyogatama, line 23: Should we add `validity_mask` to this as well (for `NULLs`)? We can do this in follow up PR

src/compression/simpatico_codegen/src/operators/nvcomp_simple_compressor.hpp
```diff
@@ -0,0 +1,81 @@
...
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr)
+{
+  if (!cudf::is_fixed_width(col.type())) {
+    throw std::runtime_error("nvcomp simple compressor: only fixed-width columns supported");
+  }
+  std::size_t const uncompressed_size =
+    static_cast<std::size_t>(col.size()) * static_cast<std::size_t>(cudf::size_of(col.type()));
+
+  return batched_compress_bytes(ops, col.head<void>(), uncompressed_size, stream, mr);
```
bwyogatama, line 49: Should we use `col.data` instead of `col.head` as the logical view of the column might have an offset?

bwyogatama: I also see this in a lot of other places in the code base

src/compression/simpatico_codegen/src/simpatico_codegen.cpp
```diff
@@ -0,0 +1,334 @@
...
+      } catch (...) {
+        std::lock_guard<std::mutex> lock(err_mu);
+        if (!failed.exchange(true)) err_msg = "column worker failed with an unknown exception";
+      }
+    });
+  }
+  for (auto& t : workers)
+    t.join();
+  pool.sync_all();
+  if (failed.load()) throw plan_error(err_msg.empty() ? "column worker failed" : err_msg);
```
bwyogatama: So we probably shouldn't change the error type here after integration with Sirius. Cause for OOM, Sirius handle `rmm::out_of_memory error` but if we change the type to some other error i am not sure if the handler still works

src/compression/simpatico_codegen/src/simpatico_codegen.cpp
```diff
@@ -0,0 +1,334 @@
...
+          body(i, stream);
+        }
+      } catch (std::exception const& e) {
+        std::lock_guard<std::mutex> lock(err_mu);
+        if (!failed.exchange(true)) err_msg = e.what();
+      } catch (...) {
+        std::lock_guard<std::mutex> lock(err_mu);
+        if (!failed.exchange(true)) err_msg = "column worker failed with an unknown exception";
+      }
+    });
```
bwyogatama, line 218: I am not sure i am following why we need multiple workers for decompression here? In Sirius we typically use static thread pool and currently we create a vector of thread every time we decompress, probably not an issue but this is not very well aligned with what we have been doing so far

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -0,0 +1,157 @@
...
+set(CMAKE_CUDA_STANDARD 20)
+set(CMAKE_CUDA_STANDARD_REQUIRED ON)
+
+if(DEFINED ENV{CONDA_PREFIX})
+  set(CONDA_PREFIX
+      "$ENV{CONDA_PREFIX}"
+      CACHE PATH "Conda/pixi prefix")
+else()
+  message(FATAL_ERROR "CONDA_PREFIX must be set (run inside pixi shell)")
+endif()
```
bwyogatama, line 18: We probably want to make Simpatico's build infrastructure aligns better with the current Sirius build infra. I am not familiar enough to comment but @mbrobbel should be able to

We don't have to do it in this PR

bwyogatama: Also, i got this comment from codex which is related

Installed extensions cannot perform runtime JIT decoding. Absolute build-machine CUDA, CCCL, and source paths are compiled into the extensio (https://github.com/sirius-db/sirius/blob/fadf8017da9348167abd7c94cb0a545fae45bee3/src/compression/simpatico_codegen/CMakeLists.txt#L84-L101), then passed to NVRTC (https://github.com/sirius-db/sirius/blob/fadf8017da9348167abd7c94cb0a545fae45bee3/src/compression/simpatico_codegen/src/jit/nvrtc_compiler.cpp#L153-L171). Generated decode code also includes the source-only codegen/decode/rle_block.cuh. I confirmed the built extension contains the worktree’s absolute paths. A relocated artifact will fail its first JIT decode. Embed/install the required headers and use relocatable discovery; also package or statically link the NVRTC runtime dependency.

src/compression/simpatico_codegen/src/operators/str_split_compressor.cu
```diff
@@ -0,0 +1,140 @@
...
+  if (null_mask) {
+    auto const* bits =
+      reinterpret_cast<cudf::bitmask_type const*>(null_mask->view().data<std::uint8_t>());
+    nc = cudf::null_count(bits, 0, num_rows, stream);
+    if (nc > 0) {
+      auto mask_contents = null_mask->release();
+      mask_buf           = std::move(*mask_contents.data);
+    }
+  }
+  return cudf::make_strings_column(
```
bwyogatama, line 48: Since we are moving everything, this means that this API is a single shot decompression API, is that correct?
What i mean is that it is not meant to be called twice, like i cannot do something like:
```
auto ct = compress_with_plan(table, plan);
auto result1 = decompress(ct); // first consumer
auto result2 = decompress(ct); // second consumer
```

I's not a big deal but just want to make sure the use case that we care about

src/compression/simpatico_codegen/src/simpatico_codegen.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+// Create a stream pool with exactly max(1, column_threads) streams.
+stream_pool make_internal_pool(int column_threads)
+{
+  stream_pool pool;
+  if (!pool.init(static_cast<size_t>(std::max(1, column_threads)))) {
+    throw plan_error("failed to initialize internal stream_pool");
+  }
+  return pool;
+}
```
bwyogatama: This is also something that we need to align, probably it's better if simpatico use the same stream pool as Sirius/cucascade

src/compression/compression_converters.cpp
```diff
@@ -0,0 +1,218 @@
...
+    // `stream`. The parallel decode runs each column on its own pool stream, so
+    // those reads are NOT ordered after the fetch on `stream`. Without a barrier
+    // a worker's D2H read of a codec frame header (e.g. nvcomp's num_chunks)
+    // races the still-in-flight H2D fetch and reads a garbage size — which then
+    // sizes a std::vector and throws length_error/bad_alloc. Order the fetch
+    // before the pool-stream reads. (The serial path below already runs on
+    // `stream`, so it is stream-ordered and needs no barrier.)
+    stream.synchronize();
+    decompressed =
+      simpatico::decompress(subset, n_threads, rmm::mr::get_current_device_resource_ref());
```
bwyogatama, line 138: So another thing is here if you are using a separate thread to decompress, it will not be detected by the reservation tracker (which will track reservation on the gpu pipeline executor thread)

bwyogatama: Our reservation tracker is attached to the thread

src/compression/simpatico_codegen/src/simpatico_codegen.cpp
```diff
@@ -0,0 +1,334 @@
...
+  return std::make_unique<cudf::table>(std::move(cols));
+}
+
+std::unique_ptr<cudf::table> decompress(const compressed_table& table,
+                                        int column_threads,
+                                        rmm::device_async_resource_ref mr)
+{
+  auto pool = make_internal_pool(column_threads);
+  return decompress_columns_parallel(table, pool, mr);
+}
```
bwyogatama, line 439: Just for API security, the stream used to allocate is destroyed here, so if someone tries to free the cudf table, the stream no longer exists. We currently won't face this issue in Sirius since you do `set_stream`, but this is an safety problem in general (I also saw this on the compress API)

src/pin_table.cpp
```diff
@@ -294,22 +308,250 @@ materialized_host_pin materialize_pin_to_host(
...
+            simpatico::stream_pool comp_pool;
+            const bool parallel =
+              n_threads > 1 && comp_pool.init(static_cast<std::size_t>(n_threads));
+            auto ct = parallel
+                        ? simpatico::compress_with_plan(tbl->view(),
+                                                        compression.plan_dsl,
+                                                        comp_pool,
+                                                        rmm::mr::get_current_device_resource_ref(),
+                                                        compression.column_names)
+                        : simpatico::compress_with_plan(tbl->view(),
```
bwyogatama: Here we are using a different stream to compress than the stream that is passed to `materialize_pin_batches`. The stream used in `materialize_pin_batches` is the stream used to produce the table, and I don't think we synchonize before compressing it on different streams?

bwyogatama:
Also codex sent me these items for the JIT part, just want to share with you @joosthooz 

- Bitpack deserialization turns eight padding bytes into logical data. Every read/write cycle adds another eight bytes and may serialize uninitialized GPU allocator contents: compressed_table_io.cpp#L241-L263 (https://github.com/sirius-db/sirius/blob/d8d64db1ade3418bf888f2fb5bb94e6d0dcb6d79/src/compression/simpatico_codegen/src/api/compressed_table_io.cpp#L241-L263).
- BOOL8 has no serialization tag. Writing can report success, but reading maps the tag to EMPTY: leaf_desc.hpp#L119-L162 (https://github.com/sirius-db/sirius/blob/d8d64db1ade3418bf888f2fb5bb94e6d0dcb6d79/src/compression/simpatico_codegen/include/codegen/plan/leaf_desc.hpp#L119-L162).
- Generated 64-bit bitpack code performs range subtraction in signed int64_t, invoking overflow for ranges such as INT64_MIN..INT64_MAX: renderer.cpp#L1113-L1172 (https://github.com/sirius-db/sirius/blob/d8d64db1ade3418bf888f2fb5bb94e6d0dcb6d79/src/compression/simpatico_codegen/src/encode/jit/renderer.cpp#L1113-L1172).
- Empty RLE/bitpack chunks can reserve a word without initializing it after explicitly disabling pre-zeroing. Serialization may expose allocator contents: renderer.cpp#L1052-L1091 (https://github.com/sirius-db/sirius/blob/d8d64db1ade3418bf888f2fb5bb94e6d0dcb6d79/src/compression/simpatico_codegen/src/encode/jit/renderer.cpp#L1052-L1091).

bwyogatama:
@joosthooz one more thing, correct me if i'm wrong, but i don't think the current simpatico pinned table works with @kevkrist pinned batch pruning optimization right? If not, we should also address this in follow up PR

joosthooz:
> one more thing, correct me if i'm wrong, but i don't think the current simpatico pinned table works with @kevkrist pinned batch pruning optimization right? If not, we should also address this in follow up PR

It didn't, and it would even break pruning with compression disabled. But I pushed a commit yesterday evening to fix that. I haven't re-run any tests/benchmarks after that though, so I will do that today.

substrait
joosthooz: Yes that shouldn't have been there, it's been reverted

src/include/sirius_config.hpp
joosthooz: During plan exploration I've ruled out plans that compressed less than 1.5x (so a 0.67 threshold), but I had to make an exception for a lineitem column at 1.3x because it wouldn't fit otherwise. So that would come down to 0.75.

What's important to realize here is that this threshold is evaluated after the data has already been compressed. That's the hardest part and we're throwing the result away to spare us the decompression work on the other side. Not a problem during pinning but wasteful for spills/comms.

src/compression/simpatico_codegen/CMakeLists.txt
joosthooz: I'll look into this, haven't considered these kinds of issues for the Sirius build yet, thanks

src/compression/simpatico_codegen/src/operators/str_split_compressor.cu
joosthooz: Only because of str_split, actually. So now that you mention it, I'm looking into modifying str_split to be non-destructive (but no longer zero-copy), so that we don't have to copy the entire compressed blob from the cache each time we want to read from it. Because we would always copy the entire thing even when reading only a subset of the columns.

src/compression/simpatico_codegen/src/operators/nvcomp_simple_compressor.hpp
joosthooz: I'm going to add a guard to the API so it gives an error when calling with a sliced view. Then we'll address this in a follow-up.

src/compression/simpatico_codegen/src/simpatico_codegen.cpp
joosthooz: I'm going to address this

src/compression/simpatico_codegen/src/simpatico_codegen.cpp
joosthooz: We should definitely improve this, but I'll create a follow-up issue for it.

src/compression/simpatico_codegen/src/simpatico_codegen.cpp
joosthooz: Also for a follow-up

src/compression/simpatico_codegen/CMakeLists.txt
mbrobbel: It looks like the scope of this PR was changed to also integrate with Sirius. In that case, Simpatico shouldn't be a separate CMake project (in that case it should be a repo, like cucascade). So I agree with Bobbi, if this PR also integrates with Sirius, it should just define the targets to be added to the Sirius CMake project.

mbrobbel: However, I would strongly prefer splitting up this PR. At least the integration should be separated into a different PR.

src/pin_table.cpp
joosthooz: I'm going to address this

src/compression/compression_converters.cpp
joosthooz: Thanks for pointing this out, this needs to be addressed but I'm not sure how to do it. We can leave it as a follow-up, in the meantime I can revert to just using 1 thread for now, it does have a slight performance penalty. Let me know your preference.

bwyogatama: i'm good with a followup PR

bwyogatama: how much is the performance penalty? if its not too much we can use 1 thread yeah

src/pin_table.cpp
```diff
@@ -294,33 +385,159 @@ materialized_host_pin materialize_pin_to_host(
...
+            *tbl,
+            compression,
+            stream,
+            "materialize_all_batches_compressed",
+            [&](simpatico::compressed_table&& ct,
+                std::vector<std::uint8_t>&& /*header*/,
+                std::vector<simpatico::payload_buffer_ref> const& /*buffers*/,
+                std::uint64_t payload_bytes,
+                std::size_t uncompressed_bytes) {
+              compressed_chunk = std::make_shared<sirius::compressed_device_representation>(
```
kevkrist, line 598: I think this moves compressed buffers into the persistent cache, and the local stream pool is destroyed while it continues to sit there, and then when the pinned entry destructor goes to free the buffer on that steam there can be leaks or invalid-resource errors.

src/compression/simpatico_codegen/src/operators/alp_compressor.cu
```diff
@@ -0,0 +1,679 @@
...
+  cudaMemcpyToSymbolAsync(
+    d_alp_combos_f64, combos_f64, sizeof(combos_f64), 0, cudaMemcpyHostToDevice, s);
+
+  // Constants must be visible before any kernel that reads them runs; the
+  // upload is host-side one-time work so a bounded sync here is acceptable.
+  cudaStreamSynchronize(s);
+}
+
+// Thread-safe one-shot init: safe under concurrent multi-stream compression.
+void ensure_constants_initialized(rmm::cuda_stream_view stream)
```
kevkrist, line 253: The constant memory used for the ALP constants is device-local, so a host-staged chunk encoded on GPU 0 and decoded on GPU 1 reads uninitialized constants.

src/compression/simpatico_codegen/src/plan/compress.cpp
```diff
@@ -0,0 +1,837 @@
...
+  std::unordered_map<std::string, cudf::column_view> output_by_name;
+  output_by_name.reserve(outputs.size());
+  for (auto const& output : outputs)
+    output_by_name.emplace(output.name, output.view);
+
+  // Guard: every channel the rep marks required must be routed by the plan, or
+  // data (e.g. a nullable column's validity via str_split's null_mask) would be
+  // silently dropped. Runs only when some outputs are declared (a bare terminal
+  // `input -> str_split` stored the whole rep above and is safe).
+  for (auto const& req : repr->required_channels()) {
```
kevkrist, line 529: I think required_channels() is only implemented by str_split, dictionary, and so columns with NULLs can compress with dropped nulls and then decompress with null-free with other compressors (e.g., ALP, ANS, etc.)

src/compression/simpatico_codegen/src/api/compressed_table_io.cpp
```diff
@@ -0,0 +1,820 @@
...
+  const std::uint8_t* payload_base = r.p;
+  const std::size_t payload_total  = r.rem;
+
+  // Bounds-check every buffer up front so a truncated file is reported here
+  // rather than faulting mid-copy, then copy each buffer straight to device.
+  for (auto const& cr : col_records) {
+    for (std::size_t li = 0; li < cr.leaf_descs.size(); ++li) {
+      for (std::size_t bi = 0; bi < cr.leaf_descs[li].buffers.size(); ++bi) {
+        std::size_t sz = static_cast<std::size_t>(cr.leaf_descs[li].buffers[bi].size_bytes);
+        if (sz > 0 && cr.buf_offsets[li][bi] + sz > payload_total)
```
kevkrist: I think it'd be wise to guard the sum against overflow here, otherwise it will pass the condtion

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -953,6 +1008,35 @@ void sirius_scan_manager::insert_pinned_entry_host(
...
+      new_num_rows += static_cast<std::size_t>(chunk.columns.front()->size());
+    }
+  }
+
+  pinned_entry entry;
+  entry.cache_info    = std::move(cache_info);
+  entry.tier          = cucascade::memory::Tier::GPU;
+  entry.memory_space  = &memory_space;
+  entry.num_rows      = new_num_rows;
+  entry.device_chunks = std::move(chunks);
```
kevkrist, line 1108: Compression-enabled DuckDB GPU pins are stored only in device_chunks, but the MVCC path still reads data_batches_by_column and chunk_memory_spaces. This makes non-empty pins appear to have zero chunks and fail validation. Please derive both the chunk count and memory spaces from device_chunks for this storage path. [Codex]

src/planner/sirius_physical_plan_generator.cpp
```diff
@@ -136,7 +137,8 @@ std::unique_ptr<sirius::op::scan::parquet_ingestible_table_info> build_parquet_t
       file_paths.push_back(file.path);
     }
     info->resolved_file_paths = std::move(file_paths);
-    info->partition_indices   = bind_data.reader_bind.hive_partitioning_indexes;
+    sirius::op::scan::canonicalize_scan_file_paths(info->resolved_file_paths);
```
kevkrist: This mutates the path DuckDB used to bind Hive partitions. If a year=202 path contains a symlink runtime parsing may see the target path instead and return the wrong value or fail casting an empty value. Please preserve the original path for Hive parsing and canonicalize a separate copy for cache identity. [Codex]

kevkrist (changes requested):

src/compression/simpatico_codegen/src/simpatico_codegen.cpp
joosthooz: This should be addressed as well. There is also a test, but it only triggers in debug builds or with a sanitizer.

Edit: I meant to say there is now a commit that addresses this and adds that test

joosthooz:
I addressed most of your comments @bwyogatama and I've kept track of a list of items to create issues for. I haven't started on @kevkrist 's comments yet. I'm working on the build problem but this code does present a challenge: it needs dynamic libraries libnvcomp and libnvrtc, and a couple of headers from CCCL and CUB. So that would add runtime dependencies where the sirius duckdb extension currently only requires the cuda driver. So that would be a bummer. The parts from the headers we need, we can render them in binary form into the .so, but the runtime dependency for those libraries would remain. Of course we can just disable compression if the environment doesn't have the requirements. 
What do you think?

bwyogatama:
> I addressed most of your comments @bwyogatama and I've kept track of a list of items to create issues for. I haven't started on @kevkrist 's comments yet. I'm working on the build problem but this code does present a challenge: it needs dynamic libraries libnvcomp and libnvrtc, and a couple of headers from CCCL and CUB. So that would add runtime dependencies where the sirius duckdb extension currently only requires the cuda driver. So that would be a bummer. The parts from the headers we need, we can render them in binary form into the .so, but the runtime dependency for those libraries would remain. Of course we can just disable compression if the environment doesn't have the requirements. What do you think?

I'm not a build expert but there should be a way to do this right? I also don't mind if we address it in future PR. Tagging @mbrobbel

ran-yuan-rui:
> I addressed most of your comments @bwyogatama and I've kept track of a list of items to create issues for. I haven't started on @kevkrist 's comments yet. I'm working on the build problem but this code does present a challenge: it needs dynamic libraries libnvcomp and libnvrtc, and a couple of headers from CCCL and CUB. So that would add runtime dependencies where the sirius duckdb extension currently only requires the cuda driver. So that would be a bummer. The parts from the headers we need, we can render them in binary form into the .so, but the runtime dependency for those libraries would remain. Of course we can just disable compression if the environment doesn't have the requirements. What do you think?

I think we should solve this in the build and packaging rather than rely on runtime fallback, to be simpler as extension.  

The target should be:

```text
nvCOMP
   └── statically linked into the Sirius extension

NVRTC + builtins + PTX compiler
   └── statically linked into the Sirius extension

CCCL/CUB/Simpatico JIT headers
   └── embedded at build time and passed to NVRTC as in-memory headers

CUDA Driver
   └── the only remaining GPU runtime dependency
```

We already have a static-only vcpkg port for nvCOMP, so Simpatico should link directly to the static CMake target instead of the generic `nvcomp` library name.

For NVRTC, we should explicitly link the static NVRTC, builtins, and PTX compiler libraries. If those are unavailable, the build should fail instead of falling back to the shared libraries.

For the CCCL, CUB, and Simpatico headers used by the generated kernels, we should package them into the extension during the build and pass them directly to `nvrtcCreateProgram()`. That removes the dependency on CUDA, CCCL, and source-tree include paths from the build machine.

This keeps the extension self-contained. The runtime environment would only need a compatible NVIDIA driver, not the CUDA toolkit, nvCOMP, CCCL, or `libnvrtc.so`.

Please ignore if I make wrong idea here.

src/include/sirius_config.hpp
wmalpica: I dont understand the argument. Let me put it into more concrete numbers to explain what I am understanding from what this threshold means:
Lets say we have a table which is 1GB uncompressed. And I compress it given a provided plan. And I get 0.94GB compressed. I have compressed under the 0.95 max_compressed_fraction and therefore I keep this. 
It seems to me that the benefit of saving 0.06GB worth of memory (having memory) and data transfer (from host to GPU) is not worth the cost of having to decompress it on the other side. Which, BTW the decompressing cost is not only computational, but at some point you have to hold in memory the compressed and decompressed bytes at the same time. If you did not do compression you always end up needing less peak memory. Which is why i am arguing that it should be lower.  (At least closer to the 0.75 you are reffering to  for lineitem). 
But I know you have run some experiments, what does the data say?
Or am I totally misunderstanding that threshold?

joosthooz:
@ran-yuan-rui thanks, I 've pushed a commit that does pretty much exactly what you describe. It renders the build-time version of those headers into the binary, and statically links using vcpkg. The code size increase is about 110MB but the sizes of our builds were already in the 200-500MB range. There's an option to link libnvrtc dynamically if we want, that reduces the build size but then the library becomes a runtime dependency.

src/compression/simpatico_codegen/src/plan/compress.cpp
joosthooz: Yes, if I'm not mistaken the string operators are now the only ones that properly support nullable columns. We should make sure that the others also support them or at least fallback/throw an error if they are encountered.

src/compression/compression_converters.cpp
joosthooz: if I remember correctly it was about 1 second on a total of about 7 seconds, so like 15%

src/compression/simpatico_codegen/src/operators/str_split_compressor.cu
joosthooz: Now I'm no longer moving anything out of the compressed representations, so that double consumer example above now works. This allowed me to make reading from tables pinned to gpu memory zero-copy.

src/include/sirius_config.hpp
joosthooz: I don't have experimental numbers for this right now, because I think we should do experiments when we also support spilling. But I'll change it to 0.75, I agree this is definitely too high especially for the current use case of pinning input tables

src/compression/simpatico_codegen/include/api/simpatico_codegen.hpp
joosthooz: I've added a guard for this to prevent silently dropping nulls, we'll properly look into it in a follow-up

src/compression/simpatico_codegen/CMakeLists.txt
joosthooz: The build problems are addressed in 4c37ed688b759e63e50c0ac1b541203c6f55420a

joosthooz:
@bwyogatama your codex comments are addressed in e893bb87fe8a956a011e52013f4b1a909701fc29
@kevkrist your comments are addressed by e4cacc59c726f3877c33565b3a3966708c89b4c8
The silently dropping nulls now has a guard: a9bb1d7efc24cba5566a4e174935f9fedfdc6f88 and we'll implement it properly in a follow-up

CMakeLists.txt
```diff
@@ -95,6 +95,20 @@ find_package(PkgConfig REQUIRED)
...
 find_package(ZLIB REQUIRED)
 
+# Static NVRTC (vcpkg overlay port). Statically linking the runtime JIT compiler
+# keeps libnvrtc.so out of the distributed extension's runtime dependencies.
+# Only available in the vcpkg build; the pixi build links the conda libnvrtc.so.
+# When the nvrtc::nvrtc_static target exists, simpatico links it instead of the
+# bare 'nvrtc' library name (see
+# src/compression/simpatico_codegen/CMakeLists.txt). Toggle off to link the
+# shared libnvrtc.so instead (smaller binary, adds a runtime dependency).
+option(SIRIUS_STATIC_NVRTC
```
mbrobbel: I would prefer not to add an option for this; we can follow what we do for all other deps, link dynamic when building with pixi and only link static when building with `vcpkg`.

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
+cmake_minimum_required(VERSION 3.24)
+
+set(CMAKE_CUDA_ARCHITECTURES
+    native
+    CACHE STRING "CUDA architectures (native = detect host GPU)")
+
+project(simpatico_codegen LANGUAGES C CXX CUDA)
```
mbrobbel, line 7: Instead of a nested project, I suggest we define the targets in the existing Sirius CMake project.

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
...
+
+set(CMAKE_CXX_STANDARD 20)
```
mbrobbel, line 9: That means we don't have to do these things

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
...
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CUDA_STANDARD 20)
+set(CMAKE_CUDA_STANDARD_REQUIRED ON)
+
+if(DEFINED ENV{CONDA_PREFIX})
+  set(CONDA_PREFIX
+      "$ENV{CONDA_PREFIX}"
+      CACHE PATH "Conda/pixi prefix")
+else()
```
mbrobbel: And these things

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
...
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|arm64")
+  set(CUDA_TARGET_TRIPLE "sbsa-linux")
+else()
+  set(CUDA_TARGET_TRIPLE "x86_64-linux")
+endif()
+set(CUDA_TOOLKIT_INCLUDE
+    "${CONDA_PREFIX}/targets/${CUDA_TARGET_TRIPLE}/include")
+set(CUDA_CCCL_INCLUDE "${CUDA_TOOLKIT_INCLUDE}/cccl")
+
+include_directories(
```
mbrobbel: Let's use targets

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
...
+  ${CMAKE_CURRENT_SOURCE_DIR}/include "${CONDA_PREFIX}/include"
+  "${CONDA_PREFIX}/include/rapids" "${CUDA_TOOLKIT_INCLUDE}")
+
+add_compile_definitions(LIBCUDACXX_ENABLE_EXPERIMENTAL_MEMORY_RESOURCE)
+add_compile_options($<$<COMPILE_LANGUAGE:CUDA>:--expt-extended-lambda>)
```
mbrobbel, line 76: These should also be per target

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
...
+    ${CMAKE_COMMAND} "-DCCCL_DIR=${CUDA_CCCL_INCLUDE}"
+    "-DOUT=${EMBED_CCCL_HEADERS_OUT}" -P
+    "${CMAKE_CURRENT_SOURCE_DIR}/cmake/embed_cccl_headers.cmake"
+  DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/cmake/embed_cccl_headers.cmake"
+  COMMENT
+    "Scanning + embedding CCCL JIT header closure into ${EMBED_CCCL_HEADERS_OUT}"
+  VERBATIM)
+
+add_library(
+  simpatico STATIC
```
mbrobbel, line 121: I guess we want `OBJECT` here?

src/compression/simpatico_codegen/plans/tpch_sf1000/customer.txt
mbrobbel, line 1: Do we want to commit these files?

src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake
mbrobbel, line 1: If this is specific to `vcpkg`, let's move it to the port.

bwyogatama (approved):

kevkrist (approved):

src/compression/simpatico_codegen/src/operators/alp_compressor.cu
joosthooz: This should be fixed too

src/pin_table.cpp
joosthooz: This should be fixed

src/compression/simpatico_codegen/src/api/compressed_table_io.cpp
joosthooz: This should be ok now

src/scan_manager/sirius_scan_manager.cpp
joosthooz: I think this is now fixed

src/planner/sirius_physical_plan_generator.cpp
joosthooz: This should be addressed in [e4cacc5](https://github.com/sirius-db/sirius/pull/937/commits/e4cacc59c726f3877c33565b3a3966708c89b4c8)

CMakeLists.txt
joosthooz: Ok this is done

src/compression/simpatico_codegen/plans/tpch_sf1000/customer.txt
joosthooz: Yes these are nice to have for now, they work pretty well for TPCH data

PR #939, opened 2026-06-15
(Doc) Design draft: S3-over-RDMA transport for the Sirius IO layer

ran-yuan-rui:
## Summary

Please read the HTML version.

This PR proposes an opt-in RDMA path for exact-key Parquet reads from S3 on a single GPU. For device reads, cuObject writes each chunk into a pre-registered GPU arena, and Sirius copies it into the caller-owned RMM buffer on the caller's CUDA stream. HEAD, footer, and other host reads continue to use the existing S3 path. This avoids Sirius-side host staging and H2D without requiring arbitrary RMM allocations to be registered with the NIC.

The RDMA path uses a dedicated, direct-connect cuObject gateway. Requests enter a bounded queue, each worker owns one RDMA session, and a read completes only after all of its chunks have been delivered. An uncertain RDMA result stops the transport and requires a process restart. Once a D2D copy has been submitted, an unprovable CUDA outcome terminates the process because the arena can no longer be released safely.

The evaluation requires hardware correctness, at least 85% of the qualified RDMA link rate for objects of 16 MiB or larger, and a zero-failure 100-million-GET soak. Smaller objects are measured against HTTP to decide whether V1 should route them by size.

This PR contains design documentation only. The six decisions that need team agreement are listed in Section 1.

### DIFF
--- a/experimental/s3-rdma-transport-design.html
+++ b/experimental/s3-rdma-transport-design.html
@@ -298,30 +298,20 @@ any other gate fails — the decision is STOP: restart the process and
 return to `s3_transport=HTTP`. The build flag stays off by default
 throughout.
 
-## 8. Current Status and Remaining Work
+## 8. Open Risks and Preconditions
 
-The transport machinery in the tree is dormant; the real cuObject
-transfer has not run yet, and the rig gates above will be its first
-exercise. Five gaps close, in order, before the gates run:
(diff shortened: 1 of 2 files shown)


bwyogatama:
all design docs are drafts

bwyogatama (approved):

PR #940, opened 2026-06-15
Add `pinned-hot` for the TPCH parquet benchmark path

mbrobbel:
Adds `pinned-hot` as additional `pinning-mode` to the tpch benchmark script. This mode will be used to replace the 3 hot runs in nightly benchmarking.


roaramburu (approved):
LGTM, let's merge it.

PR #956, opened 2026-06-16
Don't special-case Quent's UI filters

dhruv9vats:
Removes the `crates/telemetry/ui` crate which specialized the UI filters. We should instead adopt the provided ones.

Also update the `quent_context` to have the same `UUID` as a root engine entity.


rust/crates/telemetry/analyzer/src/task.rs
```diff
@@ -55,11 +55,11 @@ impl TaskExt for Task {
...
         filter
-            .pipeline_uuid
+            .operator_id
             .is_none_or(|pipeline_uuid| self.pipeline_uuid() == Some(pipeline_uuid))
-            && filter
-                .physical_operator_id
-                .is_none_or(|id| self.executes_physical_operation(id))
+        // && filter
+        //     .physical_operator_id
+        //     .is_none_or(|id| self.executes_physical_operation(id))
```
mbrobbel:
```suggestion
```

mbrobbel (approved):

PR #963, opened 2026-06-16
feat(telemetry): adopt quent build-time provenance sidecar (model.qmi)

mbrobbel:
Bump the quent dependency to the build-time provenance commit and adopt the stacked breaking changes it carries:

- Write a model.qmi provenance sidecar per context. The model! macro emits the new `ModelSource` impl for `SiriusEvent` automatically; add a build.rs calling `quent_build_info::emit_source()` so the sidecar records Sirius's own git provenance instead of quent's fallback.
- Rewrite the analyzer for the tightened `UiAnalyzer` contract: drop the `TimelineGlobalParams`/`TimelineParams` associated types and use `quent_query_engine_ui::{QueryFilter, OperatorFilter}`.
- Adopt the restructured exporter API (`FileSystem { format, root }`) and the per-context `<dir>/<id>/events.<ext>` layout: rewrite the telemetry server's discovery to scan context directories (keyed on the real `Engine` event id, never the context dir id) with a legacy flat-file fallback, and drop the removed id argument from the `create_context` FFI call.

Engine discovery requires a flushed `Engine` event; a still-running context is not yet listed (events buffer until `force_flush`, as before the bump).

~Draft because this depends on https://github.com/rapidsai/quent/pull/235.~
We're adding this to support Sirius with https://github.com/rapidsai/quent/issues/234.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1406,7 +1406,7 @@ dependencies = [
 [[package]]
 name = "quent-analyzer"
 version = "0.1.0"
-source = "git+https://github.com/rapidsai/quent.git?rev=94e592f6a0a07f02dde9a3e38f236d6eeb91c8ee#94e592f6a0a07f02dde9a3e38f236d6eeb91c8ee"
+source = "git+https://github.com/rapidsai/quent.git?rev=90e7ae0802d75737b1a855789df5389862604352#90e7ae0802d75737b1a855789df5389862604352"
 dependencies = [
  "quent-attributes",
  "quent-events",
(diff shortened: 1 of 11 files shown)


mbrobbel:
Tested this:
```
❯ cat model.qmi
{
  "quent": {
    "version": "0.1.0",
    "commit": "90e7ae0802d75737b1a855789df5389862604352",
    "branch": "master",
    "dirty": true,
    "remote": "file:///home/matthijs/.cargo/git/db/quent-515d44f958e14372",
    "built_at": "2026-06-17T07:37:04Z"
  },
  "model": {
    "name": "Sirius",
    "package": "instrumentation-model",
    "type_path": "instrumentation_model::SiriusEvent",
    "source": {
      "version": "0.1.0",
      "commit": "7b47f939d06137df5daa9c5b597b32cd45c6497f",
      "branch": "quent-model-info",
      "dirty": true,
      "remote": "git@github.com:mbrobbel/sirius.git",
      "built_at": "2026-06-17T07:41:39Z"
    }
  }
```
Tracking a fix for the quent remote in https://github.com/rapidsai/quent/issues/238

dhruv9vats (approved):
Thanks for eliminating the potentially erroneous path @mbrobbel!

Issue #952, opened 2026-06-16
SELECT 1 and SELECT COUNT(*) Failing on DuckDB Tables


roaramburu:
When trying to run either SELECT 1 or SELECT COUNT(*) FROM REGION on a DuckDB tables the engine hangs.

SELECT SUM(r_regionkey) from REGION worked fine on DuckDB tables.

Jedi18:
Fixed by https://github.com/sirius-db/sirius/pull/980

PR #980, opened 2026-06-21
fix(cpu-source): dispatch cpu_source_task from task_creator manager_loop

Jedi18:
## Problem

Queries that route through `CPU_SOURCE` can hang indefinitely. This affects sources such as `VALUES`, `EMPTY_RESULT`, `DUMMY_SCAN`, and DuckDB-materialized `ColumnDataCollection` inputs.

For example:

```sql
SELECT b FROM (VALUES (1), (2), (3)) t(b);
```

The engine builds the physical plan and logs the pipeline DAG, but then waits forever because no task ever publishes data into the downstream repository.

## Root Cause

PR #913 removed `duckdb_scan_executor`, which was the only path that dispatched `cpu_source_task`.

After that cleanup, `task_creator::manager_loop()` attempted to drive all operators through the normal GPU task creation loop:

```cpp
while (!node->all_ports_empty()) {
    // create gpu_pipeline_task from the next input batch
}
```

This does not work for source operators. `CPU_SOURCE` has no input ports, so `all_ports_empty()` returns `true` immediately. The loop never runs, no task is created, and downstream pipelines block waiting for data that is never produced.

`cpu_source_task.cpp` was also no longer included in `CMakeLists.txt`, so the task implementation was not linked.

## Fix

Add an explicit `CPU_SOURCE` path in `task_creator::manager_loop()` before the normal GPU task loop.

When the task creator sees a `sirius_physical_cpu_source`, it now:

- Creates a `cpu_source_task`.
- Acquires a HOST memory reservation.
- Runs `compute_task()` to produce host batches from the DuckDB-side source.
- Calls `publish_output()` to push those batches into downstream repositories.
- Schedules downstream consumers.
- Marks successful task telemetry before task destruction.
- Signals query completion directly for terminal CPU-only pipelines.

This bypasses the normal `while (!node->all_ports_empty())` path, which is only valid for operators driven by input ports.

## Additional Fix: DUMMY_SCAN Row Preservation

Constant-only queries such as:

```sql
SELECT 42 AS x;
```

were returning 0 rows on the GPU path instead of 1 row.

There were two related issues in the `CPU_SOURCE -> DUMMY_SCAN -> PROJECTION` path.

### DUMMY_SCAN Passthrough

After `split_cpu_source()`, the original `DUMMY_SCAN` remains in the GPU pipeline. The upstream `CPU_SOURCE` produces the single input row, but `sirius_physical_dummy_scan` did not override `execute()`, so it used the base operator implementation, which returns an empty batch.

That dropped the row before `PROJECTION`, causing constant expressions to be evaluated over 0 rows.

The fix adds a passthrough `execute()` implementation for `sirius_physical_dummy_scan`, forwarding its input batch unchanged.

### Sentinel Column for 0-Column DUMMY_SCAN

cuDF derives table row count from columns. A 0-column table has 0 rows, even if it logically represents one row.

For 0-output-column `DUMMY_SCAN`, `cpu_source_task` now emits a one-row all-null `TINYINT` sentinel column. This preserves the row count through host-to-GPU conversion so `PROJECTION` can broadcast constants to exactly one row. The projection output contains only the evaluated expression columns, so the sentinel does not reach the final result.

## Tests

Adds regression coverage for:

- `WHERE 1=0` empty-result CPU source.
- Constant-only `DUMMY_SCAN` query.
- `VALUES` query backed by `CPU_SOURCE`.

## Alternatives Considered

**Eager repository pre-population**

Pre-fill downstream repositories during repository wiring. This would avoid a runtime task, but memory reservations are runtime-only, and `_mem_res_mgr` is not available during wiring.

**Host executor in task scheduler**

Submit `cpu_source_task` through a dedicated CPU executor, symmetric with GPU pipeline tasks. This is architecturally cleaner, but requires adding a host executor concept to `task_scheduler`.

**Seed port**

Make source operators appear non-empty so the existing GPU task loop fires. This still fails because the GPU path expects GPU operator global state and eventually looks up `_gpu_operator_global_state_map.at(operator_id)`.

**GPU-resident CPU source output**

Copy CPU source batches directly to device memory and publish GPU batches. This is unnecessary for these small source cases and adds complexity without clear benefit.

## Note

`DUMMY_SCAN` may log a harmless output-type validation warning because the internal sentinel batch has one column while the logical `DUMMY_SCAN` output has zero columns. The sentinel is internal and is dropped by projection before the result is produced.

### DIFF
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -324,16 +324,20 @@ void task_creator::manager_loop()
               this->schedule(consumer);
             }
           } else {
-            _task_scheduler->signal_query_complete();
+            this->signal_query_complete();
           }
-        } catch (...) {
-          SIRIUS_LOG_ERROR("Task Creator: Exception during cpu_source_task execution");
+        } catch (const std::exception& e) {
+          SIRIUS_LOG_ERROR("Task Creator: Exception during cpu_source_task execution: {}",
+                           e.what());
           // Do NOT call stop() here: this lambda runs on a _bounded_pool worker
           // thread, and stop() -> do_stop_thread_pool() -> wait_all() would wait
           // for this very task's slot to release — a self-deadlock that also
           // holds _global_state_mutex and hangs the client's drain_after_error().
           // terminate_query() reports the error; the client thread tears the
           // task_creator down via drain_after_error() once future.get() throws.
+          _task_scheduler->terminate_query(std::current_exception());
+        } catch (...) {
+          SIRIUS_LOG_ERROR("Task Creator: Exception during cpu_source_task execution");
           _task_scheduler->terminate_query(std::current_exception());
         }
       });
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -17,7 +17,9 @@
 #include "creator/task_creator.hpp"
 
 #include "log/logging.hpp"
+#include "op/scan/cpu_source_task.hpp"
 #include "op/scan/sirius_gpu_scan_operator_data.hpp"
+#include "op/sirius_physical_cpu_source.hpp"
 #include "op/sirius_physical_delim_join.hpp"
 #include "pipeline/gpu_pipeline_task.hpp"
 #include "pipeline/task_scheduler.hpp"
@@ -114,6 +116,11 @@ void task_creator::prepare_for_query(const sirius::planner::query& query)
       std::make_shared<pipeline::gpu_pipeline_task_global_state>(pipeline, telemetry_context);
     _gpu_operator_global_state_map.emplace(operator_id, std::move(gs));
   }
+}
+
+void task_creator::signal_query_complete()
+{
+  if (_task_scheduler) { _task_scheduler->signal_query_complete(); }
 }
 
 void task_creator::drain_pending_tasks()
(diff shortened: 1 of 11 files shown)


Jedi18:
@aminaramoon I'm aware that an IO refactor is underway hence why this PR is set as draft, would that fix this hang? Is this just an intermittent bug while the IO refactor is going on?

roaramburu:
this resolves issue #952

9prady9 (commented):
a few suggestions inline.

src/creator/task_creator.cpp
```diff
@@ -231,6 +238,103 @@ void task_creator::manager_loop()
...
+
+          if (!query_complete) {
+            for (auto* consumer : consumers) {
+              this->schedule(consumer);
+            }
+          } else {
+            _task_scheduler->signal_query_complete();
+          }
+        } catch (...) {
+          SIRIUS_LOG_ERROR("Task Creator: Exception during cpu_source_task execution");
```
9prady9, line 340: **Suggestion:** `catch (...)` loses the exception detail here, while the GPU path below catches `const std::exception&` and logs `e.what()`. Since almost everything thrown in this lambda is a `std::runtime_error` with a descriptive message (no downstream repo, reservation failure, bad local-state cast), logging it would make CPU-source failures much easier to diagnose:

```cpp
} catch (const std::exception& e) {
  SIRIUS_LOG_ERROR("Task Creator: Exception during cpu_source_task execution: {}", e.what());
  ...
} catch (...) { /* existing generic path */ }
```

src/creator/task_creator.cpp
```diff
@@ -231,6 +238,103 @@ void task_creator::manager_loop()
...
+              query_complete = true;
+            }
+          }
+
+          if (!query_complete) {
+            for (auto* consumer : consumers) {
+              this->schedule(consumer);
+            }
+          } else {
+            _task_scheduler->signal_query_complete();
```
9prady9: **Nit:** this could use the new null-checked wrapper `this->signal_query_complete()` (added above) instead of calling `_task_scheduler->signal_query_complete()` directly — keeps the two completion-signal call sites uniform and gets the null check for free.

src/op/scan/cpu_source_task.cpp
```diff
@@ -362,17 +371,19 @@ std::unique_ptr<op::operator_data> cpu_source_task::compute_task(rmm::cuda_strea
...
+  // the correct row count.
   auto& pipelineable_output = dynamic_cast<op::pipelineable_operator_data&>(output_data);
   for (auto& batch : pipelineable_output.get_data_batches()) {
     if (!batch) { continue; }
     {
       auto ro = batch->to_read_only();
       if (!ro.get_data()) { continue; }
     }
-    _data_repo->add_data_batch(batch);
+    for (auto* repo : _data_repos) {
```
9prady9, line 398: **Suggestion:** when `_data_repos` is empty (the terminal RESULT_COLLECTOR case), this loop silently drops every batch. That's correct today only because the sole terminal CPU-source shape is EMPTY_RESULT, which produces zero batches — but nothing guards that invariant, and unlike `gpu_pipeline_task::publish_output` this path never routes output through the sink. If a future plan ever produces a terminal CPU-source pipeline with real rows, the query would complete "successfully" with a silently empty result.

Could we add some form of logging for that case, e.g. at the top of `publish_output`:

```cpp
if (_data_repos.empty() && !pipelineable_output.get_data_batches().empty()) {
  SIRIUS_LOG_ERROR("[cpu_source_task] {} batches produced but no downstream repository — data dropped",
                   pipelineable_output.get_data_batches().size());
}
```

so a silent wrong-result turns into a diagnosable log line.

test/cpp/integration/test_gpu_execution_tpch.cpp
```diff
@@ -4674,6 +4674,27 @@ TEST_CASE_METHOD(GPUExecutionParquetFixture,
...
+
+TEST_CASE_METHOD(GPUExecutionDuckDBFixture,
+                 "gpu_execution - dummy scan (SELECT literal)",
+                 "[integration][gpu_execution][cpu_source]")
+{
+  compare_gpu_vs_cpu("select 42 as x;");
+}
+
+TEST_CASE_METHOD(GPUExecutionDuckDBFixture,
+                 "gpu_execution - values CPU source",
```
9prady9, line 4692: **Suggestion (coverage):** these three tests cover the three CPU_SOURCE kinds nicely, but the single-repo → `std::vector` repo fan-out change has no test exercising *multiple* downstream repositories. A VALUES-backed CTE referenced twice would pin that behavior, e.g.:

```sql
with t(b) as (values (1), (2), (3)) select a.b, c.b from t a join t c using (b);
```

src/op/scan/cpu_source_task.cpp
9prady9: is this a valid concern ?

Jedi18: Yep, in fact I made it throw instead of log, better to be safe

9prady9 (approved):
lgtm

PR #982, opened 2026-06-22
Add S3 async-read perf instrumentation to the s3-bench JSON baseline

ran-yuan-rui:
## Add S3 async-read perf instrumentation to the s3-bench JSON baseline

Makes S3 read performance observable across the whole path — surfaced through the reproducible `make s3-bench` JSON baseline — without adding cost to the production happy path. A `read_parquet('s3://')` run is now decomposed into stages
(open / footer / parse / scan) with per-chunk reactor metrics, so we can see *where* an S3 read spends its time.

### What's in
- **Stage timing** in the JSON baseline: `open_ms` / `footer_fetch_ms` /`metadata_parse_ms` / `scan_ms`, with `wall_clock_ms` covering all stages (previously only `read_parquet` was timed, so footer latency was invisible).
- **Per-chunk micro counters** in the async reactor — GET latency, queue-wait, observed-H2D (`memcpy issue → worker-observed completion`, *not* a pure copy time), TTFB — **gated behind a new `s3_perf_instrumentation` config flag
  (default off)**. When off, the device path takes no `steady_clock::now()` and no extra atomics; it's a bench/test knob, not a user-facing one.
- **Always-on retry / terminal-failure counters** + a `WARN` per retried request and an `ERROR` on terminal failure (the success path stays log-free; never logged from the CUDA callback).
- A single-argument `s3_ioctx(s3_ioctx_config)` constructor that threads the flag.
- `make s3-bench` now selects the unified `[s3][bench]~[aws]` tag and forces strict MinIO bring-up (`SIRIUS_TEST_S3_STRICT=1`) so a fixture failure fails loud instead of skipping to a false green.

### Perf-safety
- Production default carries only the pre-existing cheap counters; the ns-level micro is opt-in for the benchmark.
- `device_stream_sync_total == 0` is asserted (no per-chunk sync added); a gate-off assertion proves the ns counters stay 0 in the default config.

## 📊 Perf result — where the bottleneck is today

Run: TPC-H **SF10 `lineitem`** (2.07 GiB object, ~1.01 GiB read for the projected
columns), MinIO, async backend, `max_connections=32`, 1465 chunks.

**Stage breakdown (wall clock):**

| stage | ms | % |
|---|---|---|
| open (HEAD) | 1.3 | 0.1% |
| footer fetch | 3.7 | 0.3% |
| metadata parse | 7.1 | 0.6% |
| **scan (read + H2D)** | **1093.8** | **98.9%** |
| total | 1105.9 | |

**Per-chunk micro (inside scan):**

| metric | mean |
|---|---|
| queue-wait | **479.5 ms** |
| GET latency | 17.9 ms |
| observed H2D | 3.0 ms |
| TTFB | 6.6 ms |
| peak in-flight | 32 |
| stream syncs | 0 |

**Bottleneck: the scan is connection-bound — not compute, copy, or bind.**
- The scan is **99% of wall time**; open + footer + parse are ~12 ms combined.
- Inside the scan, **`queue_wait` (480 ms) ≫ `GET` (18 ms) ≫ `H2D` (3 ms)**: each
  chunk spends ~96% of its life *waiting for a free connection*, only ~4%
  transferring. cuDF hands the reactor ~1465 chunks against a 32-wide connection
  window → backpressure dominates.
- Effective wire throughput ≈ **970 MB/s** (1.06 GiB / 1.09 s); per-connection
  ≈ 57 MB/s × 32 ≈ 1.8 GB/s theoretical, so we are not saturating — MinIO/loopback
  + scheduling overhead.
- **H2D is not the limiter** (GPU-side ~6× faster than the network), and **footer is
  negligible on MinIO**.

**Levers (future work, not this PR):** higher effective concurrency / HTTP-2
multiplexing / larger chunks to amortize per-GET overhead. H2D and bind are *not* the levers for this workload.

**Caveat:** this is loopback MinIO, so `footer_fetch_ms` is tiny (no TLS handshake).
On real S3 the footer cost grows (HEAD + 2 sync GETs, each a fresh TLS connection) and becomes a real bottleneck for the *multi-file / bind-heavy* workload (date lakes), which this single-file benchmark does not stress. The instrumentation is in
place to measure that once LIST/glob lands.

### DIFF
--- a/src/io/s3/s3_reactor.cpp
+++ b/src/io/s3/s3_reactor.cpp
@@ -38,7 +38,10 @@ namespace {
 
 using namespace std::chrono_literals;
 
-// Perf-instrumentation helpers (worker-thread only; single-writer atomics).
+// Perf-instrumentation helpers. The max/set-once updates are written only by the
+// reactor worker thread today, but use real atomic RMWs (relaxed) so they are
+// correct regardless of writer count and match the `atomic_*` naming. `fetch_max`
+// is C++26; this project is C++20, so a compare_exchange loop implements the max.
 inline std::uint64_t ns_between(std::chrono::steady_clock::time_point a,
                                 std::chrono::steady_clock::time_point b) noexcept
 {
@@ -47,7 +50,8 @@ inline std::uint64_t ns_between(std::chrono::steady_clock::time_point a,
 }
 inline void atomic_max_relaxed(std::atomic<std::uint64_t>& a, std::uint64_t v) noexcept
 {
-  if (v > a.load(std::memory_order_relaxed)) a.store(v, std::memory_order_relaxed);
+  auto cur = a.load(std::memory_order_relaxed);
+  while (v > cur && !a.compare_exchange_weak(cur, v, std::memory_order_relaxed)) {}
 }
 
 struct curl_global_init_once {
@@ -639,10 +643,10 @@ void s3_reactor::submit_pending()
     try {
       build_and_add(t);  // may throw (authorize / init / add_handle) — never escapes the worker
       if (t->is_device) {
-        // single-writer (worker thread): track the high-water mark of held blocks
-        auto n = static_cast<std::uint64_t>(device_blocks_in_flight());
-        if (n > _device_peak_inflight.load(std::memory_order_relaxed))
-          _device_peak_inflight.store(n, std::memory_order_relaxed);
+        // Track the high-water mark of held blocks (atomic max, same helper as the
+        // perf counters).
+        atomic_max_relaxed(_device_peak_inflight,
+                           static_cast<std::uint64_t>(device_blocks_in_flight()));
         if (_cfg.s3_perf_instrumentation) {
           auto const now = std::chrono::steady_clock::now();
           t->t_submit    = now;  // GET submit; chunk_get is measured to completion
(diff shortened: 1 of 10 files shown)


src/io/s3/s3_reactor.cpp
```diff
@@ -37,6 +38,18 @@ namespace {
...
+// Perf-instrumentation helpers (worker-thread only; single-writer atomics).
+inline std::uint64_t ns_between(std::chrono::steady_clock::time_point a,
+                                std::chrono::steady_clock::time_point b) noexcept
+{
+  auto const d = std::chrono::duration_cast<std::chrono::nanoseconds>(b - a).count();
+  return d > 0 ? static_cast<std::uint64_t>(d) : 0;
+}
+inline void atomic_max_relaxed(std::atomic<std::uint64_t>& a, std::uint64_t v) noexcept
+{
+  if (v > a.load(std::memory_order_relaxed)) a.store(v, std::memory_order_relaxed);
```
mbrobbel: This is not atomic, we should use `std::atomic<T>::fetch_max`.

src/io/s3/s3_reactor.cpp
```diff
@@ -680,6 +731,18 @@ void s3_reactor::cleanup_easy(transfer* t)
...
+  // GET just completed: record per-chunk GET latency + TTFB before the H2D copy.
+  if (_cfg.s3_perf_instrumentation) {
+    auto const now    = std::chrono::steady_clock::now();
+    auto const get_ns = ns_between(t->t_submit, now);
+    _chunk_get_ns_total.fetch_add(get_ns, std::memory_order_relaxed);
+    _chunk_get_ns_count.fetch_add(1, std::memory_order_relaxed);
+    atomic_max_relaxed(_chunk_get_ns_max, get_ns);
+    if (_ttfb_first_submit_set && _ttfb_ns.load(std::memory_order_relaxed) == 0) {
+      _ttfb_ns.store(ns_between(_ttfb_first_submit, now), std::memory_order_relaxed);
+    }
```
mbrobbel, line 751: This is also not atomic, we should use `std::atomic<T>::compare_exchange`.

src/io/s3/s3_reactor.cpp
ran-yuan-rui: Since std::atomic<T>::fetch_max is C++26 and the project targets C++20, I used a relaxed compare_exchange_weak loop for the max (covers both _chunk_get_ns_max and _h2d_observed_ns_max), and a compare_exchange_strong-from-zero for the TTFB set-once. Also routed the pre-existing _device_peak_inflight high-water mark through the same helper for consistency. These counters are single-writer (the reactor worker) today, so it's a correctness-hardening change rather than a live race fix

src/io/s3/s3_reactor.cpp
ran-yuan-rui: Also improved with least commit

src/io/s3/s3_reactor.cpp
mbrobbel: Oh, I missed that it's a C++26 thing.

mbrobbel (approved):

PR #986, opened 2026-06-22
Minor README update

Jedi18:
Added a few things in the README that I felt were missing for a new contributor starting with Sirius, since this information required them to dig into the super sirius docs which may not be obvious at first glance.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -6,6 +6,70 @@ this project), which then pulls in SiriusDB as an extension. We use
 - symlinked sirius-specific `CMakePresets.json` (at `cmake/CMakePresets.json`) to version control the build config.
 - [pixi](https://pixi.prefix.dev/) to manage build dependencies.
 
+## Building Sirius
+
+Clone the repository with all submodules:
+
+```bash
+git clone --recurse-submodules https://github.com/sirius-db/sirius.git
+cd sirius
+```
+
+Build with Pixi (uses all available cores). The default target is `release` (GCC Release):
+
+```bash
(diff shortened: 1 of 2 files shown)


mbrobbel (commented):
Maybe we should add these to `docs/DEVELOPMENT.md`?

wmalpica:
I agree, we should move this into `docs/DEVELOPMENT.md`
But we still need to clean up the README.md so that the section `## Building and Running Sirius` is updated and points to DEVELOPMENT.md.
Also, this new section on how to build, should indicate that you can also build in `release`, `clang-relwithdebinfo` and `clang-debug`.  We should also mention that we use precommit and that it should be installed so that commits are checked against the linters

Jedi18:
Done, moved most of it to DEVELOPMENT.md and added pre-commit and other build types

docs/DEVELOPMENT.md
```diff
@@ -6,6 +6,70 @@ this project), which then pulls in SiriusDB as an extension. We use
...
+
+```bash
+./build/release/duckdb -unsigned
+```
+
+Alternatively, load the extension into an existing DuckDB shell:
+
+```sql
+LOAD 'build/release/extension/sirius/sirius.duckdb_extension';
+```
```
mbrobbel, line 42: I guess we only need `-unsigned` for normal DuckDB (not for the one that we build with Sirius pre-loaded)?

Jedi18: True, thanks for catching that! Updated it to remove -unsigned

mbrobbel (approved):

Issue #899, opened 2026-06-08
Reduce copies caused by Projections


wmalpica:
Right now we sometimes can have two projections back to back in the query plan. These should be combined.
Additionally, in Projections, sometimes we should be able to reuse the input columns for the output if those columns are not being modified, but this will be tricky. 

1. The current expression evaluation code takes in a cudf table view. So it will force a copy even in the most trivial projections. I think we want to refactor this so that it takes in a cudf table so that it can release the columns and reuse them if possible.
2. Data batches are designed to be shareable, so that two branches of the same query plan could in theory re-use the same batch. Currently batches are mutable if you hold a mutable lock, but this mutability assumes you are just mutating the memory space. There is a function `release_table(stream)` that we sometimes use, but that function can only be used if you are the sole owner. Right now in the places we use it, we are guaranteed to be the sole owner, but we should make this safer, such that it actually checks the refcount of the shared_ptr and see if it is the sole owner and in that case just release it, otherwise make a copy.  Then we can use this for projection before passing to the expression evaluator.

PR #991, opened 2026-06-22
Avoid copies in projection

wmalpica:
Summary

sirius_physical_projection previously deep-copied every output column on device — even columns that just re-reference an input column (SELECT a, c, a). This PR avoids those copies: pure column references are exposed as zero-copy views over the input, and only genuinely-computed expressions are evaluated.

What changed

sirius_physical_projection::execute() now classifies each select_list entry as a pure passthrough (sirius::ast::reference / BOUND_REF) or an expression to evaluate, then takes one of three paths per batch:

1. All evaluated — owned cudf::table (unchanged behavior).
2. All passthrough — output is a cudf::table_view over the input columns, returned as a view-backed batch whose owner is the input's read_only_data_batch lock. Zero device allocations (was N deep copies).
3. Mixed — only the non-passthrough entries are evaluated; the output view mixes the evaluated columns with input columns, jointly owned by the evaluated table (shared_ptr<cudf::table>) and the input lock.

The read-only-lock owner keeps the source columns alive and read-only-pinned for the output's lifetime, so the view can't be freed or downgraded out from under it.

Supporting changes

- gpu_expression_executor: new std::vector<sirius::ast::node const*> valuates only the subset of expressions that need it.
- make_data_batch_from_view() helper (data_batch_utils.hpp) wrapping the gpu_table_representation view-backed (owning_table_view) ctor.

Memory accounting

View-backed outputs charge get_size_in_bytes() = full pinned input (+ evaluated columns for the mixed path) — conservative (favors spilling earlier), since the input stays pinned while referenced.

Tests & docs

- New cases in test_physical_projection.cpp: mixed path, plus lifetimandles and sync before reading to prove the output keeps the inputdata alive.
- Updated Super Sirius docs: operators.md, expression-executor.md, data-management.md, memory-management.md, optimizations.md.

🤖 Generated with Claude Code

### DIFF
--- a/src/include/data/data_batch_utils.hpp
+++ b/src/include/data/data_batch_utils.hpp
@@ -131,4 +131,37 @@ inline std::shared_ptr<cucascade::data_batch> make_data_batch(
   return std::make_shared<cucascade::data_batch>(get_next_batch_id(), std::move(gpu_repr));
 }
 
+/**
+ * @brief Create a shared_ptr<data_batch> that owns a cudf::table_view via a type-erased owner.
+ *
+ * Wraps the gpu_table_representation owning_table_view ctor: the batch holds a non-owning
+ * @p view whose underlying device memory is kept alive by @p owner (e.g. a read_only_data_batch
+ * lock on a source batch, and/or a shared_ptr<cudf::table> of freshly-evaluated columns). The
+ * owner must be copy-constructible (std::any requirement). Used by the projection operator to
+ * return columns without copying passthrough (BOUND_REF) inputs.
+ *
+ * STREAM-LINEAGE: @p writer_stream must be a stream that is ordered after every write to the
+ * memory referenced by @p view (the caller is responsible for inserting any cudaStreamWaitEvent
+ * needed to establish that ordering before calling this helper).
+ *
+ * @tparam Owner Copy-constructible type that keeps @p view's device memory alive.
+ * @param view The table view to expose (data ownership lives in @p owner).
+ * @param owner The owner keeping the viewed memory alive (moved/copied into std::any).
+ * @param alloc_size Allocation size in bytes attributed to this batch.
+ * @param memory_space The memory space where the viewed data resides.
+ * @param writer_stream Stream ordered after the writes that produced @p view's data.
+ */
+template <typename Owner>
+inline std::shared_ptr<cucascade::data_batch> make_data_batch_from_view(
+  cudf::table_view view,
+  Owner&& owner,
+  std::size_t alloc_size,
+  cucascade::memory::memory_space& memory_space,
+  rmm::cuda_stream_view writer_stream)
+{
+  auto gpu_repr = std::make_unique<cucascade::gpu_table_representation>(
+    view, std::forward<Owner>(owner), alloc_size, memory_space, writer_stream);
+  return std::make_shared<cucascade::data_batch>(get_next_batch_id(), std::move(gpu_repr));
+}
+
 }  // namespace sirius
--- a/src/op/sirius_physical_projection.cpp
+++ b/src/op/sirius_physical_projection.cpp
@@ -18,16 +18,59 @@
 
 #include "config.hpp"
 #include "data/data_batch_utils.hpp"
+#include "expression/ast/reference.hpp"
 #include "expression_executor/gpu_expression_executor.hpp"
+#include "log/logging.hpp"
 
+#include <cudf/null_mask.hpp>
+#include <cudf/utilities/traits.hpp>
+
+#include <cuda_runtime.h>
 #include <nvtx3/nvtx3.hpp>
 
 #include <cucascade/cudf/gpu_data_representation.hpp>
 #include <cucascade/data/data_batch.hpp>
 #include <duckdb/common/exception.hpp>
 
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <vector>
+
 namespace sirius {
 namespace op {
+
+namespace {
+
+/// Conservative, sync-free estimate of the device bytes referenced by a column_view, used for
+/// the owning_table_view's alloc_size (memory accounting). Counts the validity bitmask and
+/// fixed-width data buffers, and recurses into children (string/list offsets, struct fields).
+/// Variable-length leaf payloads (string chars) are intentionally NOT device-read here so the
+/// projection hot path stays free of stream synchronization; the result is an estimate.
+std::size_t estimate_column_view_bytes(cudf::column_view const& col)
+{
+  std::size_t bytes = 0;
+  if (col.nullable()) { bytes += cudf::bitmask_allocation_size_bytes(col.size()); }
+  if (cudf::is_fixed_width(col.type()) && col.size() > 0) {
+    bytes += static_cast<std::size_t>(col.size()) * cudf::size_of(col.type());
+  }
+  for (cudf::size_type i = 0; i < col.num_children(); ++i) {
+    bytes += estimate_column_view_bytes(col.child(i));
+  }
+  return bytes;
+}
+
+/// Where an output column of the projection comes from.
+enum class projection_source { passthrough, evaluated };
+
+/// Per-output-column plan: a pure BOUND_REF passthrough (index = input column index) or an
+/// evaluated expression (index = position within the evaluated output table).
+struct projection_column_plan {
+  projection_source kind;
+  std::uint32_t index;
+};
+
+}  // namespace
 
 sirius_physical_projection::sirius_physical_projection(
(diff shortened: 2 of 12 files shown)


src/op/sirius_physical_projection.cpp
```diff
@@ -43,23 +64,103 @@ std::unique_ptr<operator_data> sirius_physical_projection::execute(const operato
...
+  bool const all_evaluated   = evaluated_exprs.size() == select_list.size();
 
   /// TODO: the operator should choose the execution strategy based on statistics and a deeper
   /// understand of the trade-offs between the different strategies. See:
   /// https://github.com/sirius-db/sirius/issues/636
-  sirius::gpu_expression_executor gpu_expression_executor(
-    select_list, cudf::get_current_device_resource_ref(), stream);
+  // Construct the executor once (reused across batches; execute() resets its state each call).
+  std::optional<sirius::gpu_expression_executor> gpu_expression_executor;
+  if (!all_passthrough) {
```
kevkrist, line 103: Consider adding also the condition: `&& !select_list.empty()` to defensively guard against the possibility that the select list is empty and expression executor has not been constructed.

src/op/sirius_physical_projection.cpp
```diff
@@ -43,23 +64,103 @@ std::unique_ptr<operator_data> sirius_physical_projection::execute(const operato
...
+    if (all_passthrough) {
+      std::vector<cudf::column_view> cols;
+      cols.reserve(column_plan.size());
+      for (auto const& p : column_plan) {
+        cols.push_back(input_view.column(p.index));
+      }
+      cudf::table_view out_view(cols);
+      // We pin the whole input batch alive (its read-only lock is the owner below), so charge its
+      // full size. No new columns are allocated.
+      std::size_t const referenced_bytes = input_ro.get_data()->get_size_in_bytes();
```
kevkrist: Does this not duplicatively count the input batch bytes? Once for the backing physical memory and again for this output batch which doesn't own any physical memory of its own?

src/op/sirius_physical_projection.cpp
```diff
@@ -43,23 +64,103 @@ std::unique_ptr<operator_data> sirius_physical_projection::execute(const operato
...
+
+    // ---- Path 1: every output column is an evaluated expression ----
+    if (all_evaluated) {
+      auto projected_table = gpu_expression_executor->execute(input_view);
+      output_batches.push_back(sirius::make_data_batch(std::move(projected_table), mem, stream));
+      continue;
+    }
+
+    // ---- Path 2: every output column is a pure passthrough — zero device copies. ----
+    if (all_passthrough) {
```
kevkrist, line 123: Is this a possible situation here:
SELECT A FROM T(A, B, C, D,...)
It saves the copy on A at the cost of keeping additionally B, C, D, ... alive (assuming they are used upstream)

src/op/sirius_physical_projection.cpp
```diff
@@ -43,23 +64,103 @@ std::unique_ptr<operator_data> sirius_physical_projection::execute(const operato
...
+      std::size_t const referenced_bytes = input_ro.get_data()->get_size_in_bytes();
+      // Owner = the input read-only lock: keeps the source columns alive AND pinned read-only for
+      // the output batch's lifetime, so the view can never be freed/downgraded out from under us.
+      output_batches.push_back(sirius::make_data_batch_from_view(
+        out_view, std::move(input_ro), referenced_bytes, mem, stream));
+      continue;
+    }
+
+    // ---- Path 3: mix of evaluated columns and passthroughs. ----
+    auto evaluated_owned = gpu_expression_executor->execute(input_view);
```
kevkrist, line 143: Similar to the above, it seems possible to have:
SELECT function(A), B FROM T(A, B, C, D)
Then we will have to maintain in memory function(A) and A, B, C, D in order to avoid a copy on B. Do I understand this right? If so, not saying this is the wrong thing to do -- just trying to understand.

src/include/data/data_batch_utils.hpp
```diff
@@ -131,4 +131,37 @@ inline std::shared_ptr<cucascade::data_batch> make_data_batch(
...
+ *
+ * @tparam Owner Copy-constructible type that keeps @p view's device memory alive.
+ * @param view The table view to expose (data ownership lives in @p owner).
+ * @param owner The owner keeping the viewed memory alive (moved/copied into std::any).
+ * @param alloc_size Allocation size in bytes attributed to this batch.
+ * @param memory_space The memory space where the viewed data resides.
+ * @param writer_stream Stream ordered after the writes that produced @p view's data.
+ */
+template <typename Owner>
+inline std::shared_ptr<cucascade::data_batch> make_data_batch_from_view(
```
kevkrist, line 155: [fistbump]

roaramburu:
this resolves #899

test/cpp/operator/test_physical_projection.cpp
```diff
@@ -208,3 +210,100 @@ TEMPLATE_TEST_CASE("sirius_physical_projection can duplicate/reorder columns",
...
+  auto outputs = projection.execute(pipelineable_operator_data(inputs), cudf::get_default_stream());
+  REQUIRE(dynamic_cast<const pipelineable_operator_data&>(*outputs).get_data_batches().size() == 1);
+  auto out_batch =
+    dynamic_cast<const pipelineable_operator_data&>(*outputs).get_data_batches().at(0);
+
+  // Drop every external handle to the input batch and sync the stream before validating. The
+  // output's composite owner (input read-only lock) must keep the passthrough columns' memory
+  // alive even after the input variable disappears.
+  inputs.clear();
+  input_batch.reset();
```
kevkrist, line 254: I'm getting pushback from claude that the `reset()` here is not exercising the intended test correctly, since nothing memsets or scrubs released memory in the memory pool after `reset()`. It's *possible* that the memory was in fact freed but just not overwritten. Two recommendations from claude:
1. Assert pointer identity of the input memory and the output memory (proves zero-copy)
2. Force pool reuse after dropping the input by allocating and writing a scratch batch that would reclaim that memory if it had in fact been freed (not sure that this is completely possible?).

kevkrist (changes requested):
Awesome! Just some questions.

src/op/sirius_physical_projection.cpp
wmalpica: select_list should never be empty, that would imply that the projection is removing everything.
So instead I am going to add a check to make sure that its never empty or else throw an error

src/op/sirius_physical_projection.cpp
wmalpica: The actual memory management system will not double count. The alloc_bytes of the data_batch from the view in this case will report the memory consumed by the owner (in the all_passthrough case) or the newly allocated memory (in the all_evaluated case) or the sum of these in the mixed case. 
In the vast majority of scenarios what will be keeping the owner alive will be the data_batch created from the view, so this represents most accurately for most of the scenarios how much memory would be freed  if that batch. But your question has made me re-evaluate, and I realize that this info will be used a lot in algorithms which care how big the data is in the sense of how big therefore will be the output of an operation as opposed to how much memory could be freed, so I will change this. 
NOTE that I actually cannot get a fully accurate size of a subset of columns from a view, since views dont have a size in bytes or alloc_size type of API, so I will have to use an estimate

src/op/sirius_physical_projection.cpp
wmalpica: Yes, in a scenario SELECT A FROM T(A, B, C, D), we would be avoiding a copy of A at the cost of keeping B, C and D around for longer. That being said, if we have high memory pressure and the output of said projection needs to be downgraded, then we only mode A to host, B, C and D could be freed.

In order to avoid this, we would have to change our core data_batch structure to hold a vector of cudf::column as opposed to a cudf::table among other things.

src/op/sirius_physical_projection.cpp
wmalpica: correct. We could implement some sort of cost function to decide if its worth it to do so

test/cpp/operator/test_physical_projection.cpp
wmalpica: It is correct that the memory itself may not be freed because its a memory pool, but the "Data" is really what we care about. It may have gotten confused because the comment in the test refers to "memory" as opposed to data.

src/op/sirius_physical_projection.cpp
kevkrist: Not worth it rn it seems. Can you just document this possibility somewhere so we(I) don't forget? Thanks!

kevkrist (approved):

PR #992, opened 2026-06-23
WIP: fix(agg): ensure first() ungrouped aggregate returns deterministic results

Jedi18:
## Context

`first()` without `ORDER BY` is non-deterministic by SQL definition — no engine
is required to return the same row as another. However, a single engine must
return the same result for the same query on the same data across repeated runs.
Before this fix, Sirius violated that contract: concurrent scan tasks raced to
claim batch IDs, so the value returned by `first()` could change between runs
depending on thread scheduling.

Note that the current implementation is incorrect hence I've marked this as draft and WIP.

### DIFF
--- a/src/include/data/data_batch_utils.hpp
+++ b/src/include/data/data_batch_utils.hpp
@@ -114,6 +114,33 @@ inline std::shared_ptr<cucascade::data_batch> make_data_batch(
   return std::make_shared<cucascade::data_batch>(get_next_batch_id(), std::move(gpu_repr));
 }
 
+/// Overload that uses a caller-supplied @p batch_id instead of allocating a fresh one.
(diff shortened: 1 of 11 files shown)


mbrobbel:
> However, a single engine must return the same result for the same query on the same data across repeated runs.

Who or what requires this?

Jedi18:
> > However, a single engine must return the same result for the same query on the same data across repeated runs.
> 
> Who or what requires this?

Nothing as such requires this, I'm trying to implement this and see what the implementation would be like to see if it is worth providing this guarantee to the user.

A likely outcome of this could be that we decide to just keep the current implementation since that satisfies the sql standard

Jedi18:
Closing for now while I investigate the proper approach for this and if it is worth it

PR #996, opened 2026-06-23
fix(mgpu): enforce partition device pin for cuco-backed operators

felipeblazing:
Partitioned operators (BUILD_PROBE hash_join, grouped_aggregate_merge, ...) build a per-partition cuco hash table that is only valid on the GPU it was built on; touching it from a stream bound to another device trips cudaErrorInvalidValue in cuco counter_storage (a hard SIGABRT). The partition device was therefore a correctness constraint but was only *preferred*, not *enforced*, in two places:

1. task_creator partition affinity computed `partition_idx % num_gpus` over the PHYSICAL hardware topology (hw_topology.gpus). When the configured num_gpus is smaller than the physical GPU count (e.g. num_gpus=2 on a 4-GPU box), the modulo mapped onto a device id with no executor; the scheduler treats a pin to a non-existent executor as "no preference" and round-robins the task, so a partition's build + probe tasks scattered across GPUs and a probe touched the build table cross-device. Fixed by indexing the ACTIVE GPU executor set (memory manager GPU spaces — the same set task_scheduler keys executors on) via the new task_creator::_active_gpu_ids.

2. The OOM-reschedule path rebuilt a task's local_state without copying preferred_device_id, demoting a rescheduled probe task to "no preference" and letting it scatter. Fixed by carrying the pin forward.

Adds test/cpp/operator/test_partition_memspace_mgpu.cpp (multi-GPU repro that SIGABRT'd before the fix + single-GPU control) and a usage_limit_bytes knob in mgpu_test_utils.hpp to impose a fixed per-GPU budget on big-memory parts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -520,6 +520,7 @@ set(TEST_SOURCES
     test/cpp/operator/test_physical_hash_join_mgpu.cpp
     test/cpp/operator/test_physical_order_mgpu.cpp
     test/cpp/operator/test_mgpu_stress.cpp
+    test/cpp/operator/test_partition_memspace_mgpu.cpp
     test/cpp/memory/test_multiple_blocks_allocation_accessor.cpp
     test/cpp/operator/aggregate/test_gpu_merge_impl.cpp
     test/cpp/operator/aggregate/test_physical_grouped_aggregate.cpp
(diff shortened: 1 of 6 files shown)


felipeblazing:
this resolves #994

bwyogatama (approved):

wmalpica (approved):

PR #997, opened 2026-06-24
updated IO framework

### LINKED ISSUE #935: Make duckdb gpu scannner (cold) work with multiple GPUs

aminaramoon:
what does this PR do:

at its core, it redesigns io_cotnext and reactor interfaces to allow for partial reading from cache as well populating cache on read similar to how OS caches pages. 
- smallest unit of cache is a chunk size specified on cucascade fixed size pool.

below is the list of changes made in this PR:

1. update caching strategy to allow for partial reads from cache, when parts of the requests segment are cached.
	- on device_read:
		we copy cahced segment to device buffer.
		- we complete the io for the remaining chunks using the pinned chunks in the cache, and complete the read.
2. allow for caching to be done, on read, instead of just prefetching cache.
	- cache can only populate the buffers for the ranges requested, on read, if the chunks are not cached, we can use them as bounce buffer to read to device.
	- then on the next read, all the chunks should be cached, assuming they are not evicted.
3. introduces a synchornization primitive for io `semi_future`
	- similar to std::future and promise, you can wait on it to get the value or error.
	- but also, you can connect it to an executor and register callback.
4. introduces `balancing_strategy`, an interface used to distribute scan operator data to different gpus. 
	- the only one is `round_robin` but we can extend it.
5. introduces `load_balancing_and_batch_coalecing` in scan_manager.
	- it allows for scan inputs to be chunked to the rqueted size, regardless of the max num of batches we specified in the config.
	- it distributed the load between gpus after coalescing is done.
	X: fixed a flaw we had in `dev` were we got smaller than requested batch size
	X: allows for massively parallel parsing of metada of files.
6. introduces a epoll based implementation of restful reacotr for reading from s3, azure, and ...
7. it changes task creation policy:
	- in a hint, we now ask for the number of tasks we need to create more tasks. 
	- we use that estimate to create tasks from upstream operators
	- for exaple: a pipelinable branch will create only one task at a time, hard barrier makes all of them
8. makes `pop` order of task queue configuratble
	- we cna choose to pop from back or front
9. introduces a new `owning_table_view` that allows for projection, drop and selection of columns on a view constructed from a table or `<owner,view>` pair
	- it is used to remove unnecessary copies in a gpu_scan_operator.
10. fixes a bug in gpu_ingestible where the cahced batches didn;t get filtered in the scan operator.
11. other synschronization primitives like
	- completion_controller, admission_controller, try ( simialr to std::excepted to be used as value type of async producers )


ran-yuan-rui:
The epoll/socket_action + CURLSH/HTTP-2 direction is exactly what I had lined up after #982 — glad you got to it first, and I think `rest_reactor` is the right base.

Rather than trying to finish the whole S3 migration inside this already-large IO-framework PR, I can cherry-pick / transplant the `rest_reactor` pieces into an S3-focused branch and complete the S3 cutover there: routing/config, #982 instrumentation, restored S3 runtime / SQL / perf tests, and AWS connection-reuse measurements.

That should keep this PR reviewable as the core IO-framework rewrite, and give the S3 side a clean branch with complete tests + perf numbers. Sounds like a better plan?

aminaramoon:
@ran-yuan-rui I would greatly appreciate if you could enable s3 on top of this using rest reactor. I used all the logic you'd implemented for parsing the body, and response, credential providers and from your s3 io context. and built the reactor with the pieces of your s3 work. I still haven;t figured out the cleanset way of switching between context between queries, but I am sure that is something you can address.

ran-yuan-rui:
> @ran-yuan-rui I would greatly appreciate if you could enable s3 on top of this using rest reactor. I used all the logic you'd implemented for parsing the body, and response, credential providers and from your s3 io context. and built the reactor with the pieces of your s3 work. I still haven;t figured out the cleanset way of switching between context between queries, but I am sure that is something you can address.

Sure, this will be my plan. Based the discussion on the call, I will wait the PR to be merged then start my part work.

src/data/host_parquet_representation_converters.cpp
```diff
@@ -80,138 +76,20 @@ convert_host_parquet_to_gpu_with_prefetched_data_source(
...
-    *const_cast<cucascade::memory::memory_space*>(target_memory_space),
-    target_stream);
-  return repr;
+  // STRUCTURAL GAP (io rebase): sirius::op::scan::prefetched_data_source — the
+  // in-memory cudf::io::datasource adapter that wrapped a cache_ranges and fed
+  // cudf::io::read_parquet from prefetched host blocks — was removed
+  // (prefetched_data_source.cpp deleted, the type is no longer declared) with no
+  // replacement. The cached-host-parquet -> GPU conversion below cannot be
+  // reconstructed until that adapter is reimplemented, so this path is stubbed
+  // to throw and the dependent body is disabled.
```
bwyogatama, line 85: Are we still/will be using this prefetched data source?

src/include/io/cache/prefetching_cache.hpp
```diff
@@ -0,0 +1,266 @@
...
+    uint64_t hits{0};
+    uint64_t h2d{0};
+    uint64_t misses{0};
+    uint64_t evictions{0};
+  };
+
+  counters _counters;
+  counters_snapshot _last_reported;
+
+  std::jthread _preparation_thread;
```
kevkrist, line 246: Is it possible for the thread, after the destructor body exits, to join *after* the queue is destroyed, while it intermittently wakes up and tries to touch the queue? Seems possible under reverse-order member deconstruction?

src/include/io/cache/types.hpp
```diff
@@ -0,0 +1,523 @@
...
+ * @brief How the prefetching layer should behave on top of a given backend.
+ *
+ * - @c none: no prefetching.  Either the backend does not support vector host
+ *   reads (so the prefetcher cannot batch range requests cheaply) or the
+ *   backend explicitly opted out.
+ * - @c immediate: prefill the cache ahead of consumer demand.
+ * - @c opportunistic: read-ahead on demand — issue extra IO only when triggered by a
+ *   consumer read.
+ * - @c disposable: prefetching is temporary and can be discarded when no longer needed.
+ */
```
bwyogatama, line 60: add just-in-time

src/io/io_context.cpp
```diff
@@ -16,241 +16,55 @@
...
-    }
-    total_copied += sizes[i];
+  // One-shot.  Repeated calls are silent no-ops so callers can be
+  // robust to multiple wiring sites.
+  if (_cache) return;
+  try {
+    _cache = std::make_unique<cache::prefetching_cache>(reservation_manager,
+                                                        this,
+                                                        cache_config.inflight_budget_chunks,
+                                                        cache_config.buffer_pool_bytes,
```
kevkrist: Is `buffer_pool_slabs` the same as `buffer_pool_bytes`?

bwyogatama: this is passing buffer_pool_bytes to an argument which supposed to take buffer_pool_slabs? They are on different units are we sure this is okay?

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -418,41 +533,34 @@ void sirius_scan_manager::insert_pinned_entry(
...
-      // metadata reflects the union of pinned columns.
-      for (auto& cn : column_names) {
-        if (std::find(entry.column_names.begin(), entry.column_names.end(), cn) ==
-            entry.column_names.end()) {
-          entry.column_names.push_back(std::move(cn));
+          entry.data_batches_by_column[std::string{column_names[i]}].emplace_back(
+            std::move(cols[i]));
         }
       }
+      // The union of pinned column names is reflected by entry.cache_info; the
```
kevkrist: It doesn't look like the column_names/column_ids in the cache_info are updated?

src/sirius_extension.cpp
```diff
@@ -1232,12 +1321,22 @@ void SiriusExtension::RegisterGPUFunctions(DatabaseInstance& instance)
...
-  pin_table.named_parameters["cols"]   = LogicalType::LIST(LogicalType::VARCHAR);
-  pin_table.named_parameters["n_rows"] = LogicalType::BIGINT;
-  CreateTableFunctionInfo pin_table_info(pin_table);
+  // pin_table takes either a positional path (parquet) or no positional (duckdb,
+  // where 'name' is the catalog table reference) — register both arities as a set.
+  TableFunctionSet pin_table_set("pin_table");
+  auto add_pin_table_overload = [&](vector<LogicalType> positional_args) {
+    TableFunction pin_table(
+      "pin_table", std::move(positional_args), PinTableFunction, PinTableBind);
+    pin_table.named_parameters["tier"]        = LogicalType::VARCHAR;
```
kevkrist, line 1330: number of rows for partial pinning?

src/op/scan/duckdb_native_gpu_ingestible.cpp
```diff
@@ -41,52 +44,127 @@
...
+    if (!_acc.empty()) { out.push_back(emit_current()); }
+    return out;
+  }
+
+ private:
+  std::unique_ptr<scan_info> emit_current()
+  {
+    auto split           = std::make_unique<duckdb_native_scan_info>();
+    split->row_groups    = std::move(_acc);
+    split->datasource    = _datasource;
```
kevkrist: To mirror parquet, should this be something like:
```
split->datasource = _datasource
  ? std::shared_ptr<io::sirius_datasource>(_datasource->duplicate())
  : nullptr;
```
I'm looking at the docs at sirius_datasource.hpp:39-43 that mention the datasource is not safe to share, though it seems there that it's about sharing cross-scan, not cross-batch, so I may be mistaken here.

src/scan_manager/load_balancing_scan_batch_coalecer.cpp
```diff
@@ -0,0 +1,142 @@
...
+    auto& state = *_slots[pipeline_id];
+    if (state.batch_provider) {
+      process_cached_entries(state, stop);
+    } else {
+      process_provider_inputs(state, stop);
+    }
+  }
+}
+
+void load_balancing_scan_batch_coalecer::process_provider_inputs(metadata_processing_state& state,
```
kevkrist: If an exception is thrown in this loop, should the connector's get closed so that get_next_split() stops blocking on the connector?

src/io/cache/prefetching_cache.cpp
```diff
@@ -0,0 +1,872 @@
...
+  _counters.n_reads.fetch_add(1, std::memory_order_relaxed);
+
+  coverage_policy policy =
+    _io_ctx->supports_host_to_device_read() ? coverage_policy::partial : coverage_policy::full;
+
+  size_t n_chunks = (size + _pool->chunk_bytes() - 1) / _pool->chunk_bytes();
+  std::vector<cached_chunk*> chunks;
+  chunks.reserve(n_chunks);
+  if (out_handle && *out_handle) {
+    auto& requested_chunks = out_handle->get_context()->chunks;
```
bwyogatama: For just-in-time, the prefetch_request_context might have been destroyed already by the time we do this, potentially causing segfault

src/include/op/scan/batch_coalescer.hpp
```diff
@@ -0,0 +1,45 @@
+/*
```
kevkrist, line 1: It should be `coalescer`, not `coalecer`

src/include/io/cache/types.hpp
```diff
@@ -0,0 +1,523 @@
...
+// unique, const or not); the returned pointers preserve the handle's constness.
+template <std::ranges::contiguous_range Chunks>
+  requires cached_chunk_pointer<std::ranges::range_value_t<Chunks>>
+[[nodiscard]] std::vector<
+  decltype(std::to_address(std::declval<const std::ranges::range_value_t<Chunks>&>()))>
+find_entry(const Chunks& chunks, std::size_t offset, std::size_t size, coverage_policy policy)
+{
+  using chunk_ptr_t =
+    decltype(std::to_address(std::declval<const std::ranges::range_value_t<Chunks>&>()));
+  constexpr std::size_t chunk_size = 1 << 20;  // must match buffer_pool's chunk size
```
bwyogatama: This is hardcoded right now. Should follow _pool->chunk_bytes()

src/scan_manager/load_balancing_scan_batch_coalecer.cpp
```diff
@@ -0,0 +1,142 @@
...
+      state.connector->push_split(std::move(op_data));
+    }
+    if (is_closed) {
+      state.connector->close();
+      break;
+    }
+  }
+}
+
+void load_balancing_scan_batch_coalecer::process_cached_entries(
```
kevkrist: stop_token ignored?
Should `!stop.stop_requested()` be added to the loop condition?

src/scan_manager/load_balancing_scan_batch_coalecer.cpp
```diff
@@ -0,0 +1,142 @@
...
+  return [state_ptr = it->second](exec::try_t<std::unique_ptr<op::scan::scan_info>>&& entry) {
+    state_ptr->queue.enqueue(std::move(entry));
+  };
+}
+
+void load_balancing_scan_batch_coalecer::worker_loop([[maybe_unused]] std::stop_token const& stop)
+{
+  for (auto pipeline_id : _pipeline_order) {
+    if (stop.stop_requested()) { break; }
+    auto& state = *_slots[pipeline_id];
```
kevkrist: Use `at()`? Otherwise `nullptr` default inserted to the map.

src/include/scan_manager/load_balancing_scan_batch_coalescer.hpp
```diff
@@ -0,0 +1,154 @@
...
+  /// entries (semaphore-blocked with a poll timeout so stop_token is
+  /// observed promptly) until it hits a closure sentinel (null
+  /// datasource) or stop is requested.
+  void worker_loop(std::stop_token const& stop);
+
+  void process_provider_inputs(metadata_processing_state& state, std::stop_token const& stop);
+
+  void process_cached_entries(metadata_processing_state& state, std::stop_token const& stop);
+
+  /// unique_ptr storage: the slot contains a semaphore and a moodycamel
```
kevkrist, line 144: Stale doc? `shared_ptr` is used.

src/scan_manager/load_balancing_scan_batch_coalecer.cpp
```diff
@@ -0,0 +1,142 @@
...
+    } else {
+      process_provider_inputs(state, stop);
+    }
+  }
+}
+
+void load_balancing_scan_batch_coalecer::process_provider_inputs(metadata_processing_state& state,
+                                                                 std::stop_token const& stop)
+{
+  std::stop_callback stop_cb(stop, [&state] { state.queue.enqueue(nullptr); });
```
kevkrist: I'm not sure that this ever causes `is_closed` to be true, which would leave the connector unclosed. Can you double check?

src/op/scan/parquet_gpu_ingestible.cpp
```diff
@@ -146,390 +279,213 @@ parquet_gpu_ingestible::parquet_gpu_ingestible(std::unique_ptr<io::ingestible_ta
...
+    [[maybe_unused]] auto const stored =
+      sirius_ds->store_metadata(std::make_shared<parquet_metadata>(file_metadata, footer_len));
+  }
+  auto const& metadata = *file_metadata;
+
+  // FLBA-decimal pushdown probe: cudf's row-group stats filter cannot compare a
+  // fixed_point_scalar AST literal against FLBA / BYTE_ARRAY decimal stats, so
+  // pushdown is disabled for such files (the filter still applies post-decode).
+  bool disable_filter_pushdown = false;
+  for (auto const& elem : metadata.schema) {
```
kevkrist, line 402: Maybe this doesn't belong to this PR, but filter_pushdown should only get disabled if an FLBA-decimal column is scanned. It seem to get disabled here by walking the entire schema.

src/op/scan/parquet_gpu_ingestible.cpp
```diff
@@ -49,69 +53,198 @@
...
+
+ private:
+  std::unique_ptr<scan_info> emit_current()
+  {
+    auto split                     = std::make_unique<parquet_split_info>();
+    split->rg_slices               = std::move(_slices);
+    split->reader_options          = _reader_options;
+    split->plan                    = _plan;
+    split->disable_filter_pushdown = _disable_pushdown;
+    split->needs_assembly          = _needs_assembly;
```
kevkrist, line 171: I think this field is dead now.

src/include/io/cache/types.hpp
aminaramoon: good point, I will fix it

src/io/io_context.cpp
aminaramoon: fixed

src/data/host_parquet_representation_converters.cpp
aminaramoon: no

aminaramoon: this is deadcode

kevkrist (changes requested):

src/io/cache/prefetching_cache.cpp
aminaramoon: I don't understand how ? the handle ? or context ? cause they are shared_ptr owned by datasource and cache, so one copy is always active. also, this is in the read-api, which means datasource that owns the handle and context is alive

src/scan_manager/load_balancing_scan_batch_coalecer.cpp
aminaramoon: they are all pre-populated.

src/scan_manager/sirius_scan_manager.cpp
bwyogatama: It does get updated in sirius_scan_manager.cpp:aligned_column_names

src/sirius_extension.cpp
aminaramoon: @bwyogatama ?

src/scan_manager/sirius_scan_manager.cpp
aminaramoon: @bwyogatama ?

src/include/io/cache/prefetching_cache.hpp
aminaramoon: destructor joins the therads, how can they wake up ?

src/io/uring/uring_reactor.cpp
kevkrist, line 526: It doesn't look like the O_DIRECT flag can be bypassed per the uring_config.hpp flag? Fine to defer to later.

src/sirius_extension.cpp
bwyogatama: Yeah right now it's disabled, i can add it on follow-up PR if its needed. The challenge is that previously we use cudf `chunked_parquet_reader` which allows us to do this easily, but now the ingestible path doesn't use this API anymore

src/scan_manager/load_balancing_scan_batch_coalecer.cpp
aminaramoon: this is left over for when I didn;t replace it with `try_t`. good catch

src/io/uring/uring_reactor.cpp
```diff
@@ -217,389 +809,278 @@ size_t uring_reactor::host_read(int fd, size_t offset, size_t size, uint8_t* dst
...
+  static constexpr std::chrono::milliseconds SHUTDOWN_POLL_MS{100};
+
+  std::stop_callback cb(stop_token, [this] {
+    spdlog::trace("uring_reactor worker_loop: stop requested");
+    _requests.enqueue(nullptr);  // unblock the worker if it's waiting on an empty queue
+  });
+
+  using slot_token = slot_pool::token;
+
+  unique_ring ring(2 * NUM_CHUNKS);
```
kevkrist, line 860: `ring_entried` config param? Fine to defer to later.

src/op/scan/parquet_gpu_ingestible.cpp
aminaramoon: it is from your code, no ? can you push a fix if you think it is incorrect!

src/scan_manager/load_balancing_scan_batch_coalecer.cpp
aminaramoon: but using at() is always better, will change it

src/io/uring/uring_reactor.cpp
```diff
@@ -16,31 +16,489 @@
...
+    io_uring_cqe* tmp = nullptr;
+    // Bounded wait so the top-of-loop _stop check is reachable even when
+    // no CQE arrives.  SINGLE_ISSUER means we can't post a NOP SQE from
+    // interrupt() to unblock a plain wait_cqe; the timeout bounds shutdown
+    // latency to SHUTDOWN_POLL_MS.
+    __kernel_timespec ts{};
+    ts.tv_sec  = timeout.count() / 1000;
+    ts.tv_nsec = (timeout.count() % 1000) * 1'000'000L;
+    int rc     = io_uring_wait_cqe_timeout(ring.get(), &tmp, &ts);
+    if (rc < 0 && rc != -EINTR && rc != -ETIME) { return -rc; }
```
kevkrist, line 448: If rc = -ETIME, -ETIME is returned, and the callers below register an error and breaks the worker loop, but shouldn't it just check shutdown state per the doc?

src/include/io/templated_ioctx.hpp
```diff
@@ -119,264 +220,260 @@ class templated_ioctx : public sirius_ioctx {
...
+      try {
+        auto& tobj    = as_typed(obj);
+        int device_id = rmm::get_current_cuda_device().value();
+        auto reactors = next_reactor(tobj, 0, io_op_type::device_async, device_id);
+        if (reactors.empty()) {
+          return exec::make_semi_future<size_t>(std::make_exception_ptr(
+            std::runtime_error("device_read_async_io: no available reactors")));
+        }
+        request_type_ptr req =
+          Reactor::prep_device_rx_request(_config, tobj, dst, offset, size, stream, device_id);
```
kevkrist, line 373: The host path clamps the requested size with the file size above. Any reason not to do that here?

src/io/uring/uring_reactor.cpp
```diff
@@ -217,389 +809,278 @@ size_t uring_reactor::host_read(int fd, size_t offset, size_t size, uint8_t* dst
...
-      // SQE from interrupt() to unblock a plain wait_cqe; the timeout
-      // bounds shutdown latency to SHUTDOWN_POLL_MS instead.
-      static constexpr long SHUTDOWN_POLL_MS = 100;
-      __kernel_timespec ts{};
-      ts.tv_sec  = SHUTDOWN_POLL_MS / 1000;
-      ts.tv_nsec = (SHUTDOWN_POLL_MS % 1000) * 1'000'000L;
-      int rc     = io_uring_wait_cqe_timeout(ring.get(), &tmp, &ts);
-      if (rc < 0 && rc != -EINTR && rc != -ETIME) {
-        spdlog::error("uring_reactor: io_uring_wait_cqe_timeout failed: {}", strerror(-rc));
+  auto clean_up_and_shutdown = [&]() {
```
kevkrist, line 1038: Should anything be done about incomplete requests here? If shutdown happens while a slot index is in incomplete_requests, it looks like that request can be destroyed without on_complete() or on_error() callbacks being invoked. Maybe a non-issue.

src/include/io/cache/prefetching_cache.hpp
kevkrist: Before they are destroyed/joined but while queue is being destroyed?

src/op/scan/parquet_gpu_ingestible.cpp
kevkrist: Lol probably -- yeah I'll push a fix.

src/sirius_extension.cpp
kevkrist: Fine with me to add it as follow-up

src/scan_manager/sirius_scan_manager.cpp
kevkrist: I see aligned_column_names getting called *before* insert_pinned_entry, which does the merging, in SiriusExtension::PinTableFunction to create cache_info for the incoming pin call, not the merged cache_info?

src/op/scan/duckdb_native_gpu_ingestible.cpp
aminaramoon: I think they should be duplicated in the coalescer, because that is where task inputs are actually created

src/io/cache/prefetching_cache.cpp
bwyogatama: Maybe i'm misunderstanding something.
I only see the `shared_ptr` owned by the `prefetching_cache` though? The `prefetching_handle` owns `weak_ptr` via lifecycle manager, and I can't find where the `shared_ptr` is being owned by the datasource.

For `just_in_time` case, during `prepare_loop`, the request is never enqueued and thus get out of scope in the `prefetching_cache`. During `device_read`, the handle is still active but the context might be out of scope already?

src/io/uring/uring_reactor.cpp
aminaramoon: int const fd = (cfg.use_odirect && segment.is_odirect_compatible()) ? file.odirect_handle()
                                                                      : file.buffered_handle();

aminaramoon: they get bypassed in the prep_req functions

src/io/uring/uring_reactor.cpp
aminaramoon: nope, that is dead. it is not a good idea to let people play with this I realized :D

src/op/scan/duckdb_native_gpu_ingestible.cpp
kevkrist: Yeah I think this is in the coalescer so we are aligned here

src/io/uring/uring_reactor.cpp
aminaramoon: removed it

src/include/io/templated_ioctx.hpp
aminaramoon: it is not needed for either case. it short reads are checked against file size to see if it is EOF

src/include/io/cache/config.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+struct config {
+  size_t inflight_budget_chunks   = 2048;
+  size_t initial_pool_reservation = 10;
+  bool dispose_after_use          = false;
+
+  /// Total pinned-memory budget for the prefetch buffer pool (bytes).  Divided
+  /// by the slab size to derive the max-slabs count passed to
+  /// @c sirius_ioctx::initialize_cache.
+  size_t buffer_pool_bytes = 256UL << 20;
```
bwyogatama: Is this currently stale code then? I saw that we are using initial_pool_reservation instead as the constructor for prefetching_cache

src/io/rest/rest_reactor.cpp
```diff
@@ -0,0 +1,1344 @@
...
+      }
+    };
+
+    // Handle one completed transfer.  Returns true iff the slot was parked in
+    // `copying` (held until its H2D copy finishes); otherwise the caller
+    // recycles the slot immediately.
+    auto finish = [&](int i, CURLcode rc, long status) -> bool {
+      io_slot& s          = slots[static_cast<size_t>(i)];
+      auto& req           = *s.req;
+      bool const ok_range = (status == 206) || (status == 200 && req.chunk.offset == 0);
```
bwyogatama, line 1139: I am a noob on this topic, but do we not limit the size of the full read here? If status = 200 and bytes start at 0 seems are we going to download everything no matter what the size

src/scan_manager/sirius_scan_manager.cpp
bwyogatama: oh i see, you are right i will fix this

src/io/cache/prefetching_cache.cpp
aminaramoon: ah, I left that out. very good catch bobbi. initially I made it a weap_ptr because all the request on creations were pushed to the eviction queue, so they were kept alive, weak_ptr was there to protect against evicted requests

src/io/rest/rest_reactor.cpp
bwyogatama: If this is tricky to address we can leave it for follow up PR btw, or Yu can fix this in his PR

src/io/cache/prefetching_cache.cpp
aminaramoon: addressed it

src/io/uring/uring_reactor.cpp
aminaramoon: it is a non-issue, because once the request goes out of scope, it notifies its future if it doesn;t have a state. 
i think it gets an broken_promise message

src/include/io/cache/config.hpp
aminaramoon: yes, it is. it is intended to be use as a threshold for back-pressure on prefetching. but I will have to do it in a follow up PR

src/io/rest/rest_reactor.cpp
aminaramoon: we always request partial chunks, so we get 206. 200 is complete GET request which means offset has to be 0

src/io/uring/uring_reactor.cpp
```diff
@@ -217,389 +809,278 @@ size_t uring_reactor::host_read(int fd, size_t offset, size_t size, uint8_t* dst
...
-    bool work_to_submit = !pending.empty() || !pending_host.empty();
+      cudaError_t err = cudaSuccess;
+      auto hint       = s.copy_h2d_async(err);
+      if (hint == io_slot::h2d_sync_hint::h2d_failed) {
+        s.on_error(err);
+        continue;
+      } else if (hint == io_slot::h2d_sync_hint::h2d_event_based) {
+        copying_slots.push_back(s.release_slot());
+      }
+      s.on_complete(s.bytes_read);
```
bwyogatama, line 1014: Do we need to wait for the copy to finish before incrementing the bytes_read at on_complete?

aminaramoon: no, if the user is giving the boiunce buffer ( cached chunk ), you can mark it as complete, when they call the stream.sync on their side, the copy gets completed. but for internal bounce buffer ( h2d_event_based ), we set complete when the copy is done, so we can reuse thhe buffer again.

when a request is complete ( all teh chunks have called on_complete, or on_error, or the request dies before it gets processed ), we set a value/exception to the promise. in the callback of semi_future, the user calls stream synchronize, and get whatever they want. satifaction of promise means IO is complete, but h2d might still be ongoing. 

this is why semi_future was very helpful in getting rid of all the streamCallback I had to release this buffers,

src/op/scan/parquet_gpu_ingestible.cpp
kevkrist: we'll remove it later

kevkrist (approved):

src/include/scan_manager/sirius_scan_manager.hpp
```diff
@@ -261,136 +221,112 @@ class sirius_scan_manager {
...
-  /// \throws std::runtime_error when no backend supports @p uri, or when the
-  ///         footer fetch / schema inference fails.
-  [[nodiscard]] parquet_bind_result describe_parquet(std::string const& uri);
+  /// \brief Process-wide ioctx used to mint @c sirius_datasource instances.
+  ///        Returns nullptr when the manager was configured with
+  ///        @c use_sirius_datasource=false.
+  [[nodiscard]] sirius::io::sirius_ioctx* io_ctx() const noexcept { return _io_ctx.get(); }
+
+  [[nodiscard]] std::shared_ptr<sirius::io::sirius_datasource> create_datasource(
+    std::string_view path) const noexcept;
```
ran-yuan-rui: Here we declares create_datasource(...) const **noexcept**, while calls _io_ctx->open_datasource(file_path) directly. That can throw through local uring open failures. 
Missing files, permission errors, unsupported local URIs, or O_DIRECT failures should surface as query/open errors, not process termination. Remove noexcept, or catch here and intentionally map unsupported paths to nullptr while rethrowing real I/O errors?

src/scan_manager/split_provider.cpp
```diff
@@ -16,29 +16,56 @@
...
+    auto work = next_split_provider();
+    if (!work) { continue; }
+    scheduler.enqueue([this, on_split, work = std::move(work), token = ctrl.acquire()]() mutable {
+      try {
+        auto split = work();
+        on_split(std::move(split));
+      } catch (...) {
+        on_split(std::current_exception());
+      }
+    });
```
ran-yuan-rui, line 60: Current pattern: split_provider::run() loops through every file/batch and calls scheduler.enqueue(...); scoped_dispatcher::enqueue() is explicitly non-blocking and spills overflow into an unbounded pending_ queue.
Current complexity: O(F) queued closures/memory for F files, with no producer-side backpressure.
Why costly: large lakehouse-style scans can enqueue tens/hundreds of thousands of metadata tasks, holding file paths, callbacks, completion slots, and queue nodes before workers can drain them. This is especially painful because metadata scan opens datasources and parses footers.

should we use the dispatcher’s blocking schedule() path, or add a bounded metadata prefetch/window in split_provider::run() so only ~`num_threadstok*num_threads` tasks are outstanding?
What we will get - O(W) queued work where W is the window size, same total O(F) metadata work but bounded memory and steadier IO pressure.

bwyogatama:
@wmalpica found a segfault when running the benchmark so we can't merge this yet

bwyogatama (approved):

ran-yuan-rui (approved):

ran-yuan-rui:
@aminaramoon 

I found deadlock — the read's future never resolves and the serving worker is stuck on a libcurl share lock (not in
`epoll_wait`):

```
#0  __lll_lock_wait
#3  ___pthread_mutex_lock
#5  Curl_share_lock
#6  Curl_dnscache_prune
#8  Curl_cpool_do_locked
#13 sirius::io::rest::rest_reactor::worker_loop
```

**Root cause.** `curl_share::lock_cb` / `unlock_cb` (`src/io/rest/curl_handle.cpp`) ignore the `curl_lock_data` argument and serialize *every* lock class through a single non-recursive `std::mutex`. libcurl holds one class's lock while acquiring another's —`Curl_cpool_do_locked` holds `CONNECT`, then `Curl_dnscache_prune` takes `DNS` — so the worker re-locks the same mutex on one thread → permanent self-deadlock. Endpoints that
churn connections (`Connection: close`) trip it immediately; keep-alive S3/MinIO rarely prunes the dns cache under the cpool lock, so it hasn't surfaced (the reactor read path also has no fast-endpoint test on the branch yet).

**Fix** — one mutex per `curl_lock_data` class (the standard `CURLSHOPT_LOCKFUNC` contract), selected by the callback's `data` argument:

```cpp
// curl_handle.hpp:  std::mutex _mtx;  ->  std::array<std::mutex, CURL_LOCK_DATA_LAST> _mtx;
void curl_share::lock_cb(CURL*, curl_lock_data data, curl_lock_access, void* userp) {
  static_cast<curl_share*>(userp)->_mtx[static_cast<std::size_t>(data)].lock();
}
void curl_share::unlock_cb(CURL*, curl_lock_data data, void* userp) {
  static_cast<curl_share*>(userp)->_mtx[static_cast<std::size_t>(data)].unlock();
}
```

Verified: the previously-hanging reads pass and the full Catch2 suite is clean.

**Cherry-pickable commit** (on our fork, self-contained — no S3-layer deps):

```
git fetch https://github.com/ran-yuan-rui/sirius.git feature-s3-newioframework
git cherry-pick 226ccb26
```

or browse: https://github.com/ran-yuan-rui/sirius/commit/226ccb26

You could merge this fix - or just leave to me when 997 merge to dev.

ran-yuan-rui:
@aminaramoon 
I may found and fixed a latent bug in the new semi_future primitive. Timed waits (semi_future::get(timeout)) used a raw Linux FUTEX_WAIT_BITSET, but resolution only called std::atomic::notify_all(). Under libstdc++, atomic::wait/notify route through an internal waiter pool (a proxy futex), not a direct futex on the atomic's own address — so notify does not wake a raw futex waiter parked on that address, and a timed waiter could sleep until its deadline even after the value had already been published.

The fix pairs each producer-side notify with a matching raw FUTEX_WAKE_BITSET. Untimed get() was unaffected (it waits on std::atomic::wait, which shares that same pool), which is why this stayed hidden until the restored REST vector-read tests exercised get(timeout).

Fix commit — single self-contained file (src/include/exec/semi_future.hpp, +35), cherry-picks cleanly:

[`69cc018a` Wake raw futex waiters when a semi_future is resolved](https://github.com/ran-yuan-rui/sirius/commit/69cc018a1cdc05734be113e415357f68bb47c3c5)


git cherry-pick 69cc018a1cdc05734be113e415357f68bb47c3c5
(on branch feature-s3-newioframework. ) You can merge or wait for my branch

PR #998, opened 2026-06-24
Add info about Quent to README

dhruv9vats:
Adds:
- A dedicated Quent Telemetry doc file.
- A section in the main README briefing that Quent is used, and points to the more elaborate doc.


docs/README.md
```diff
@@ -94,6 +94,29 @@ SET sirius_log_level = 'trace';
 SET sirius_log_flush_seconds = 1;
 ```
 
+## Tracing
+> **Note:** Tracing is experimental and the telemetry schema may change.
+
+Sirius instruments query execution with [Quent](https://github.com/rapidsai/quent), emitting
+per-query traces of operator and pipeline activity that Quent can render
```
johanpel, line 101: > per-query traces

Q: do we indeed construct one context + engine entity per query?

docs/super-sirius/quent-telemetry.md
johanpel, line 1: Nit: Maybe we can add a screenshot here to get a quick peek at what the outcome is of following these steps.

mbrobbel (approved):

johanpel (approved):
Nice, thanks for making it more obvious from the top readme how to get going with Quent! One nice to have nit and one question:

docs/README.md
dhruv9vats: that's a good question, and the answer is: yes, currently. today, we spin with the telemetry at https://github.com/sirius-db/sirius/blob/321cab86c4e54cc58f10ca8a1f0d3fe55f74e43f/src/include/sirius_context.hpp#L310 as part of the `SiriusContext`, and by default every SQL query is run as a separate session.

docs/super-sirius/quent-telemetry.md
dhruv9vats: [Add screenshots](https://github.com/sirius-db/sirius/pull/998/commits/4eab85ba883633f7134ced9d5ecb2b5ff86aed72)

PR #1001, opened 2026-06-25
Drop the redundant host reservation in the materialized result collector

Jedi18:
### Summary
The materialized result collector no longer creates its own host-tier reservation
before materializing the final GPU result to host memory. It selects a target host
space and hands it to `clone_to<host_data_representation>`; the conversion allocates
directly. This removes a reservation the converter never used and a double-count
against the host memory budget.

### What was the issue

The collector made the reservation, then passed
only the memory space to the converter, the reservation was never used for the actual
allocation. While held, it double-counted against the host budget alongside the
converter's allocation, which can cause spurious "insufficient host memory" failures
under pressure even when the allocation would have fit.

### What changes

We now fail fast instead of blocking on the reservation.

Allocation is still capacity-checked (clean `rmm::out_of_memory`); we just drop the old unbounded `request_reservation` block, which could deadlock holding GPU batches on a pipeline thread.
  
  Addresses "|2|" part of issue https://github.com/sirius-db/sirius/issues/566

### DIFF
--- a/src/op/sirius_physical_result_collector.cpp
+++ b/src/op/sirius_physical_result_collector.cpp
@@ -141,23 +141,27 @@ void sirius_physical_materialized_collector::sink(const operator_data& input_dat
       // Use clone_to to clone directly into HOST representation
       auto sirius_ctx  = _client_ctx.registered_state->Get<duckdb::SiriusContext>("sirius_state");
       auto& memory_mgr = sirius_ctx->get_memory_manager();
-      /// TODO: Find the closest memory space, not just any memory space, in HOST tier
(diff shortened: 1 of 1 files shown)


Jedi18:
@kevkrist I see that the previous attempt at this issue was closed https://github.com/sirius-db/sirius/pull/583, do you recall what the reason was?

kevkrist (approved):
[fistbump]

kevkrist:
> @kevkrist I see that the previous attempt at this issue was closed #583, do you recall what the reason was?

I can't remember, tbh. But this is much better than what I was doing then :).

PR #1002, opened 2026-06-25
chore(telemetry): bump `quent` to latest `main`

mbrobbel:
Update the pinned quent rev across the telemetry crates and bump ts-rs from 11 to 12, migrating to the new APIs:

- `build.rs`: `TS::export_all_to(dir)` -> `Config::with_out_dir` + `export_all`
- server: adopt quent's per-context collector sink factory, `EngineIndex` lister (`index_query_engines`), and context-keyed `Sirius::import_events`, mirroring quent's simulator reference server; drop the obsolete custom lister/importer helpers
- model: add a collector feature so `SiriusContext` implements `CollectorSink`
- FsmCollection: handle breaking change

This adopts quent's new per-context/per-entity capture layout; the old flat-file/events.<ext> readers and format auto-detection are gone.

Closes #975
﻿﻿

### DIFF
--- a/rust/crates/telemetry/server/src/main.rs
+++ b/rust/crates/telemetry/server/src/main.rs
@@ -1,23 +1,14 @@
-use std::{
-    net::ToSocketAddrs,
-    path::{Path, PathBuf},
-};
+use std::{net::ToSocketAddrs, path::PathBuf};
 
 use clap::Parser;
-use instrumentation_model::SiriusEvent;
-use quent_collector::server::CollectorServiceOptions;
-use quent_exporter::{
-    ExporterOptions, FileSystemExporterOptions, FileSystemFormat, FileSystemImporterOptions,
-    ImporterOptions, create_importer,
-};
+use instrumentation_model::{Sirius, SiriusContext};
+use quent_exporter::{ExporterOptions, FileSystemExporterOptions, FileSystemFormat};
 use quent_query_engine_server::{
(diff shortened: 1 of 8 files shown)


rust/crates/telemetry/server/src/main.rs
```diff
@@ -92,25 +75,34 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        })?
+        .serve(collector_addr)
+        .await
+        .map_err(|error| -> Box<dyn std::error::Error> { Box::new(error) })
     };
 
-    let lister = move || list_engine_ids(&lister_output_dir);
+    // Index the exported contexts by engine instance: each engine's telemetry is
+    // the engine's own context plus its workers' contexts.
+    let lister = move || index_query_engines(&lister_output_dir, format);
```
johanpel, line 98: Commenting here to ensure awareness: `index_query_engines` currently rests on the assumption that contexts are created only by things that always emit either worker and/or engine entity events. I don't think this is a problem for Sirius as IIUC it will only spawn one Context for now, but it is likely very relevant for the distributed integrations work.

Made a separate issue just for this: https://github.com/rapidsai/quent/issues/259 which it's actually a stop-gap for a larger issue we need to tackle at some point: https://github.com/rapidsai/quent/issues/40.

johanpel (approved):
Looks good, just one thing to be aware of:

PR #1005, opened 2026-06-26
build(pixi): bump to 0.71 and adopt rich platforms

mbrobbel:
Bumps pixi to **0.71** and migrates to its [rich-platform CUDA requirements](https://github.com/prefix-dev/pixi/releases/tag/v0.71.0).

## Why

pixi 0.71 deprecated the per-feature `[system-requirements]` table and turned the CUDA requirement into a *platform-match gate* — so a CUDA environment no longer matches a driverless host. Left unchanged, `pixi list`/`install` fails on every CPU runner with `no platform supported by environment … matches the current system`.

## What changed

- **Pixi 0.71** across the board: all workflow `pixi-version` inputs (`v0.70.2` → `v0.71.1`), the `nightly-runner` dependency (`0.70.*` → `0.71.*`), and `requires-pixi` (`>=0.68` → `>=0.71`). `setup-pixi` is already on its latest release (`v0.9.6`).
- **Rich/named platforms.** Each conda subdir is declared once per supported CUDA major as a named platform pinning `__cuda`; the `cuda12`/`cuda13` features bind to their matching pair (replacing the deprecated `system-requirements`). `CUDAARCHS` (per feature) still carries the codegen arch list.
- **GPU-less builds.** The CPU build/lint runners set `CONDA_OVERRIDE_CUDA` so the CUDA environment resolves without a driver; GPU test runners keep detecting the real driver (`__cuda` matches minimum-style, so a current driver satisfies either CUDA major).
- `duckdb-python` and `nightly-runner` don't depend on the CUDA major, so they're pinned to the cuda13 pair — otherwise each would carry two `linux-*` platforms differing only by `__cuda`.
- **StarRocks experiment manifest** (`experimental/starrocks/pixi.toml`) gets the same migration (its `engine` feature pinned the deprecated `system-requirements`, which pixi 0.71 synthesized into workspace-wide CUDA platforms and broke the CUDA-agnostic `cn` env).
- **Nightly env generator** now strips the `nightly-runner` feature blocks in full (range delete) so their bodies don't leak into the generated manifest.
- Regenerated both `pixi.lock` files.

> [!NOTE]
> Approach mirrors NVIDIA/cuCascade#153. `__cuda_arch` was explored but dropped: pinning it on a platform breaks pixi 0.71.1's current-system match unconditionally (no host value matches, GPU or not), and `system-requirements` can't carry it either.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -23,6 +23,10 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-24.04
+    env:
+      # CPU runner: no GPU/driver, so pixi can't detect `__cuda` to match the
+      # CUDA-pinned platform. Assert it so the (cuda13) environment resolves.
+      CONDA_OVERRIDE_CUDA: "13"
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
@@ -31,7 +35,7 @@ jobs:
       - uses: prefix-dev/setup-pixi@5185adfbffb4bd703da3010310260805d89ebb11 # v0.9.6
         with:
           cache: false
-          pixi-version: v0.70.2
+          pixi-version: v0.71.1
           global-environments: |
             pre-commit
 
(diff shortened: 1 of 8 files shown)


dhruv9vats (approved):
Thanks!

PR #1021, opened 2026-06-26
feat(starrocks): emit local_files reads for file scans

mbrobbel:
Make the StarRocks plan translator emit Substrait `local_files` parquet reads for `FILE_SCAN_NODE`, so a translated single-fragment `FILES()` plan resolves in DuckDB's Substrait reader (as `parquet_scan(<paths>)`) instead of an unresolvable named table. This is the resolvable-plan source the upcoming one-shot `execute_substrait` FFI runs end-to-end on the GPU.

## What
- A `ScanFilePaths` type (static `from_fragment` constructor + `for_node` accessor) collects per-scan-node parquet paths from the fragment's `per_node_scan_ranges` and — for pipeline fragments — `node_to_per_driver_seq_scan_ranges`, keyed by plan node id, and is threaded through the plan context.
- `scan_rel` emits `ReadType::LocalFiles` (`uri_file` + `parquet`) when paths are present, and keeps the named-table read otherwise (e.g. HDFS scans).

## Validation (fail loud, don't mis-read)
DuckDB's `local_files` reader emits `parquet_scan(<paths>)` and ignores per-file byte offsets, so ranges it can't honor are rejected with `TranslateError::UnsupportedScanRange` rather than silently producing wrong results:
- non-parquet `format_type` (CSV/ORC/JSON/…),
- byte-range splits (non-zero start, or a bounded size we can confirm is shorter than the file),
- path-derived columns (`columns_from_path`), which `parquet_scan` would not produce.

## Known limitation
The read schema is taken from the scan tuple descriptor, assuming the parquet file's physical column order matches the descriptor slot order — true for `FILES()` `SELECT *`. Non-identity column mappings (`columns_from_path`, destination-slot expressions, src≠dest tuples) are a pre-existing translator assumption (the named-table path assumed the same) and are left as a follow-up.

## Tests
`local_files` emission via both `per_node_scan_ranges` and the per-driver map; rejection of non-parquet and split (offset and partial-size) ranges.

### DIFF
--- a/experimental/starrocks/crates/starrocks-plan-translator/src/scan_paths.rs
+++ b/experimental/starrocks/crates/starrocks-plan-translator/src/scan_paths.rs
@@ -1,0 +1,277 @@
+//! Per-scan-node parquet file paths collected from a fragment's broker scan ranges.
+
+use std::collections::HashMap;
+
+use starrocks_thrift::exprs::TExprNodeType;
+use starrocks_thrift::internal_service::{TExecPlanFragmentParams, TScanRangeParams};
+use starrocks_thrift::plan_nodes::{
+    TBrokerRangeDesc, TBrokerScanRangeParams, TFileFormatType, TFileScanType,
+};
+use starrocks_thrift::types::TFileType;
+
+use crate::descriptor_table::DescriptorTable;
+use crate::error::{Result, TranslateError};
+
+/// Parquet file paths for each scan node in a fragment, keyed by plan node id.
+///
+/// Built from the fragment's broker scan ranges so `FILE_SCAN` nodes emit
+/// Substrait `local_files` reads DuckDB resolves as `parquet_scan(<paths>)`.
+/// Nodes without broker ranges have no entry and fall back to a named-table read.
+///
+/// Collection fails closed: only the slice that maps faithfully onto a
+/// `parquet_scan` over local files is accepted — a `FILES()` query (not a load),
+/// reading whole local parquet files whose columns are direct passthroughs to the
+/// scan tuple. Anything else (loads, byte-range splits, remote schemes, casts or
+/// other column transforms, path-derived or flexibly-mapped columns) is rejected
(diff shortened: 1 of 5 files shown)


experimental/starrocks/crates/starrocks-plan-translator/src/scan_paths.rs
```diff
@@ -0,0 +1,266 @@
...
+            paths.add_ranges(*node_id, ranges, desc)?;
+        }
+        if let Some(per_driver) = exec_params.node_to_per_driver_seq_scan_ranges.as_ref() {
+            for (node_id, per_seq) in per_driver {
+                for ranges in per_seq.values() {
+                    paths.add_ranges(*node_id, ranges, desc)?;
+                }
+            }
+        }
+        Ok(paths)
```
aocsa, line 68: The correctness argument today is "StarRocks never puts one node in both maps." That's an external contract enforced in Java (LocalFragmentAssignmentStrategy), invisible from the Rust side. Make it a local, checked invariant:

```
if let Some(per_driver) = exec_params.node_to_per_driver_seq_scan_ranges.as_ref() {
    for (node_id, per_seq) in per_driver {
        debug_assert!(
            !exec_params.per_node_scan_ranges.contains_key(node_id),
            "node {node_id} appears in both per_node_scan_ranges and \
             node_to_per_driver_seq_scan_ranges; whole-file paths would be double-read",
        );
        for ranges in per_seq.values() {
            paths.add_ranges(*node_id, ranges, desc)?;
        }
    }
} ```

aocsa (approved):
LGTM,  The two scan-range maps (per_node_scan_ranges / node_to_per_driver_seq_scan_ranges) looked like they could double-count a node's paths.

experimental/starrocks/crates/starrocks-plan-translator/src/scan_paths.rs
mbrobbel: I'll return an error when this happens, instead of panic in debug.

PR #1022, opened 2026-06-26
feat(sirius): execute Substrait plans on the GPU via the FFI

mbrobbel:
> **Stacked on #1009** (compiles the Substrait reader into `libsirius`). Until that merges, this PR's diff includes its commit; afterwards this rebases onto `dev`.

Adds the one-shot `execute_substrait` FFI: run a serialized Substrait plan on the GPU and get the result back over the Arrow C Data Interface. Stacks on the branch that compiles the Substrait→DuckDB reader into `libsirius`.

## Route
`Context::execute_substrait` (`src/sirius_ffi.cpp`) drives the same lowering the transparent table-function path uses, but standalone:
`SubstraitToDuckDB::TransformPlan` → `Planner::CreatePlan(RelationStatement)` → `Optimizer` → `ResolveOperatorTypes` → `ColumnBindingResolver` → `sirius_physical_plan_generator::create_plan` → `sirius_interface::sirius_execute_query` → `ResultArrowArrayStreamWrapper`. DuckDB is used only to lower the plan; execution runs on the Sirius GPU engine.

## Bindings
- `sirius-sys`: a cxx bridge method `Context::execute_substrait(&CxxString plan, usize out_stream_addr)`.
- `sirius`: `SiriusContext::execute_substrait(&mut self, &[u8]) -> Result<ArrowArrayStreamReader, SiriusError>` — allocates an `FFI_ArrowArrayStream`, hands its address to the engine, then wraps it as an Arrow `RecordBatchReader`.

The public header stays free of Arrow/cxx types: the plan is a `std::string`, and the stream is passed by integer address (`uintptr_t`).

## Bring-up the transparent path gets for free
The embedded `DuckDB` is a fresh instance, so `execute_substrait` does explicitly what the table-function path inherits from its enclosing query:
- load the parquet extension so `parquet_scan` binds (allow-unsigned + a path from `SIRIUS_DUCKDB_PARQUET_EXTENSION` — a dev stopgap until parquet is linked into the engine);
- wrap bind + execution in a transaction (GPU execution is eager and the result materialized, so the transaction closes before the Arrow stream is consumed);
- `converter_registry::initialize()` for the GPU scan/result converters.

## Test
A GPU test writes a parquet fixture, builds a `local_files` plan, runs it through `execute_substrait`, and checks the rows. The two context-constructing tests share a process lock (the engine keeps process-global GPU state). Requires a GPU + `LD_LIBRARY_PATH` to the built engine, like the existing context test.

### DIFF
--- a/src/sirius_ffi.cpp
+++ b/src/sirius_ffi.cpp
@@ -20,29 +20,170 @@
 
 #include "sirius_ffi.hpp"
 
-#include "sirius_config.hpp"   // sirius::sirius_config
-#include "sirius_context.hpp"  // duckdb::SiriusContext
+#include "data/sirius_converter_registry.hpp"              // sirius::converter_registry
+#include "duckdb/common/arrow/result_arrow_wrapper.hpp"    // duckdb::ResultArrowArrayStreamWrapper
+#include "duckdb/common/enums/optimizer_type.hpp"          // duckdb::OptimizerType
+#include "duckdb/execution/column_binding_resolver.hpp"    // duckdb::ColumnBindingResolver
+#include "duckdb/main/client_context.hpp"                  // duckdb::ClientContext
+#include "duckdb/main/config.hpp"                          // duckdb::DBConfig
+#include "duckdb/main/connection.hpp"                      // duckdb::Connection
+#include "duckdb/main/database.hpp"                        // duckdb::DuckDB
+#include "duckdb/main/prepared_statement_data.hpp"         // duckdb::PreparedStatementData
+#include "duckdb/main/query_result.hpp"                    // duckdb::QueryResult
+#include "duckdb/main/relation.hpp"                        // duckdb::Relation
+#include "duckdb/optimizer/optimizer.hpp"                  // duckdb::Optimizer
+#include "duckdb/parser/statement/relation_statement.hpp"  // duckdb::RelationStatement
+#include "duckdb/planner/planner.hpp"                      // duckdb::Planner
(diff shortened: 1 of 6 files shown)


src/sirius_ffi.cpp
```diff
@@ -20,30 +20,171 @@
...
+           pos             = escaped.find('\'', pos + 2)) {
+        escaped.replace(pos, 1, "''");
+      }
+      auto load = conn->Query("LOAD '" + escaped + "'");
+      if (load->HasError()) { load->ThrowError(); }
+    }
+    // Register the engine on the connection and disable the DuckDB-specific optimizer
+    // rewrites the GPU planner does not implement (mirrors SiriusContext's transparent
+    // path: SiriusTableFunctionData::PrepareConnection / OnConnectionOpened).
+    auto& client = *conn->context;
```
aocsa: Nit: this omits STATISTICS_PROPAGATION and disables COLUMN_LIFETIME unconditionally, so it doesn't really "mirror the transparent path" (sirius_extension.cpp:319-322). Not a correctness issue — just match the set or tweak the comment so it doesn't mislead.

aocsa (approved):
LGTM

src/sirius_ffi.cpp
mbrobbel: [fe3557c](https://github.com/sirius-db/sirius/pull/1022/commits/fe3557c6a42026cf739b6ff25d0b83d48c1029a3)

PR #1024, opened 2026-06-29
feat(starrocks): execute fragments on the GPU via SiriusEngine

mbrobbel:
> **Stacked on #1008** (CN execute path), **#1022** (`sirius::SiriusContext::execute_substrait` FFI), and **#1021** (translator `local_files` reads — without it the translator emits a `NamedTable` the embedded DuckDB can't bind, so real `FILE_SCAN` fragments wouldn't resolve). Until they merge, this PR's diff includes their commits.

Replace the compute node's `StubExecutor` with a GPU-backed `SiriusEngine`, so a query runs end-to-end through the StarRocks CN: FE fragment → Substrait → embedded DuckDB lowering → **GPU** → rows back via `fetch_data`.

## `SiriusEngine` actor
`sirius::SiriusContext` is `!Send`/`!Sync` (opaque cxx handle) and the engine serializes queries through one process-global context. So `SiriusEngine` owns the context on a **dedicated thread** — created, used, and dropped there — and the handle talks to it over channels, which are `Send`/`Sync` and carry only owned data (`Vec<u8>` in, `Vec<RecordBatch>` out). No `unsafe`, and it matches the single-thread usage the FFI is already tested under (the #1022 GPU test).

- `start()` brings the context up with a fail-fast ready handshake (a bad config / GPU failure surfaces before any RPC is served).
- `impl FragmentExecutor`: forwards `translated.to_substrait_bytes()` to the thread, waits for the batches.
- `Drop` closes the request channel and joins the thread for an ordered context teardown.

## Wiring (feature-gated)
- `SiriusComputeNodeService::with_executor(executor)` (`new()` is now the test stub constructor).
- `BrpcServer::with_executor(executor)`.
- `main` builds the executor — `SiriusEngine::start` under `sirius-engine`, else `StubExecutor` — and threads it through `BrpcRuntime::start` → `BrpcServer`. Engine bring-up moves off the main thread into the actor.
- `--no-default-features` keeps the pure-Rust `StubExecutor` path (what CI builds).

## Known limitation
`FragmentExecutor::execute` is synchronous, but `exec_plan_fragment` runs it on a `spawn_blocking` worker, so the BRPC runtime stays free to serve `fetch_data`, connection cleanup, and shutdown while a query runs. The single-fragment limitations are that the whole result is materialized before dispatch returns and the single process-global context serializes queries — both lifted by the streaming evolution.

## Test
`engine.rs` carries a GPU integration test (`cfg(sirius-engine)`): write a parquet fixture, build a `local_files` plan, run it through `SiriusEngine`, assert the rows. Requires a GPU + `LD_LIBRARY_PATH` to the built engine (and `SIRIUS_BUILD_DIR`/`SIRIUS_DUCKDB_PARQUET_EXTENSION`), like the `sirius` crate's context test. CI's `--no-default-features` run excludes it.


experimental/starrocks/src/engine.rs
```diff
@@ -0,0 +1,374 @@
...
+//! [`SiriusEngine`] talks to that thread over channels — which are `Send`/`Sync` and carry only
+//! owned data (`Vec<u8>` in, `Vec<RecordBatch>` out) — so it satisfies `dyn FragmentExecutor:
+//! Send + Sync` without ever moving the context across threads.
+//!
+//! The seam is synchronous (see [`FragmentExecutor`]): `execute()` blocks the caller until the
+//! engine thread returns the result. For the single-fragment milestone that caller is the async
+//! `exec_plan_fragment` handler on the BRPC current-thread runtime, so a long GPU query stalls
+//! that runtime — an accepted limitation the trait already documents (the streaming evolution
+//! lifts it).
+
```
aocsa, line 15: NIT: This says a long query "stalls that runtime," but exec_plan_fragment offloads to spawn_blocking(...).await (compute_node_service.rs:73-81), so the reactor stays free. Code's better than the doc — i would suggest to update this and the PR body's "Known limitation" so it doesn't imply blocking that isn't there.

experimental/starrocks/src/engine.rs
```diff
@@ -0,0 +1,374 @@
...
+        // here, on the engine thread, returning owned batches whose buffers are released via their
+        // own Arrow C release callbacks — independent of the context. So the batches are safe to
+        // send to, and drop on, the caller's thread.
+        let result = context
+            .execute_substrait(&request.plan)
+            .map_err(|err| err.to_string());
+        // Ignore a send error: the waiting fragment may have been dropped/cancelled.
+        let _ = request.respond.send(result);
+    }
+    info!("sirius-engine thread shutting down");
```
aocsa, line 107: Query errors return Err and the loop keeps serving. The one gap would be: if execute_substrait ever panics (e.g. an arrow-rs edge case), the thread dies silently, the CN keeps reporting healthy (lib.rs heartbeat is independent of engine liveness), and every subsequent fragment fails. Optional for this milestone, but wrapping the call in catch_unwind (or flipping a readiness flag on loop exit) would turn a silent dead-engine into a visible one.

experimental/starrocks/src/engine.rs
```diff
@@ -0,0 +1,374 @@
...
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner())
+            .as_ref()
+            .ok_or_else(|| "sirius-engine is shutting down".to_string())?
+            .send(request)
+            .map_err(|_| "sirius-engine thread is not running".to_string())?;
+        let batches = respond_rx
+            .recv()
+            .map_err(|_| "sirius-engine thread dropped the response".to_string())??;
+        Ok(FragmentResult::new(batches))
```
aocsa, line 138: NIT: It may be blocking, or at least make it clear in the PR scope: The engine can now return types like Date, Decimal, and Binary, but MysqlResultEncoder doesn’t support them yet (see https://github.com/sirius-db/sirius/blob/207f630cb8adc8838afa9a12f9c143bf9d7356ca/experimental/starrocks/src/result_encoder.rs#L88). This means execution can succeed and then fail while returning results. Could we either add support for these types or reject them before execution?

aocsa (changes requested):
Nice work overall—the engine actor. I left a couple of non-blocking notes about stale docs and engine liveness, plus maybe one blocking issue: the result path needs to handle or reject all the types supported by the translator. Requesting changes for that; once it’s fixed, this looks good for the single-fragment milestone.

experimental/starrocks/src/engine.rs
mbrobbel: I'll handle this in a follow-up: #1090

experimental/starrocks/src/engine.rs
mbrobbel: [8879586](https://github.com/sirius-db/sirius/pull/1024/commits/8879586732fe9f33f31978669f15344547d4748c)

experimental/starrocks/src/engine.rs
mbrobbel: All of this work is intentionally limited in scope and incremental. I'll add support for those types in follow-up PRs.

aocsa (approved):
LGTM

PR #1027, opened 2026-06-29
fix/performance(filter): Don't materialize pure filter columns part 2

kevkrist:
Builds off of #1019 

Uses the projection map from the duckdb filter operator to prevent unnecessary materialization of pure filter columns (i.e., columns not needed downstream) in the sirius filter operator.


9prady9 (approved):

src/planner/sirius_plan_filter.cpp
```diff
@@ -67,13 +77,28 @@ sirius_physical_plan_generator::create_plan(duckdb::LogicalFilter& op)
...
+      duckdb::make_uniq<sirius::op::sirius_physical_filter>(std::move(filter_types),
+                                                            sirius::ast::from_duckdb(*combined),
+                                                            op.estimated_cardinality,
+                                                            std::move(output_indices));
     filter->children.push_back(std::move(plan));
     plan = std::move(filter);
   }
-  if (op.HasProjectionMap()) {
-    // there is a projection map, generate a physical projection
+  if (op.HasProjectionMap() && !projection_folded_into_filter) {
```
dhruv9vats: would `projection_folded_into_filter` not always be `true` if `op.HasProjectionMap()` is also `true`? Since it is set at line 89 under the condition `op.HasProjectionMap() == true`.

9prady9: nice catch, missed that

src/include/op/sirius_physical_filter.hpp
```diff
@@ -34,25 +37,20 @@ class sirius_physical_filter : public sirius_physical_operator {
...
                          std::unique_ptr<sirius::ast::node> expression,
-                         std::size_t estimated_cardinality);
+                         std::size_t estimated_cardinality,
+                         std::vector<cudf::size_type> output_indices = {});
 
   //! The filter expression
   std::unique_ptr<sirius::ast::node> expression;
 
+  //! Output column positions to gather from the filtered input, in output order.
+  //! Empty ⇒ keep every input column. When set, columns referenced only by the
```
dhruv9vats: > //! Empty ⇒ keep every input column.

do we want to consider something that embeds this information into code? maybe like:
```cpp
struct passthrough {};

using output_mask = std::variant<passthrough, std::vector<cudf::size_type>>;
```

kevkrist (commented):
Thanks all!

src/planner/sirius_plan_filter.cpp
kevkrist: Not quite. `projection_folded_into_filter = true` is gated behind the `!op.expressions.empty()`, so this block here fires when the expression list is empty. That being said, the added conditional here is always true, so it is redundant. I've dropped it and reshaped the code a little for clarity.

PR #1036, opened 2026-06-30
Remove legacy IO from sirius

aminaramoon:
This PR removes all the legacy cpu scan tasks, iceberg and hive operators


bwyogatama (approved):

ran-yuan-rui:
@aminaramoon 
One assumption I'd like to confirm : I read the current direction as table-format resolution moving out of Sirius entirely — an external planner (the StarRocks FE via Substrait, or the duckdb-iceberg extension expanding its own plan) resolves iceberg/hive down to concrete parquet file lists, and Sirius's scan layer never sees a table format. Is that the intent?

roaramburu:
this is already resolved by PR #1156

PR #1038, opened 2026-06-30
Partition coalescing + single-materialization for multi-GPU hash joins

felipeblazing:
Partition the input once into a reordered table plus zero-copy per-partition views (hash_partition_sliced), so the existing hash_partition() materializes each partition from a single reordered table instead of re-hashing/copying per partition.

For multi-GPU hash partitioning, bucket the fine partitions into num_gpus GPU slots and concatenate each slot into one coalesced batch (respecting _coalesce_batch_bytes), carried by partition_slotted_operator_data so the slot-aware sink() routes each coalesced batch to its target GPU instead of the running batch index. This cuts cross-GPU shuffle overhead by sending one large batch per GPU rather than many fine partitions.

### DIFF
--- a/src/op/sirius_physical_partition.cpp
+++ b/src/op/sirius_physical_partition.cpp
@@ -27,8 +27,13 @@
 #include "op/sirius_physical_hash_join.hpp"
 #include "pipeline/sirius_pipeline.hpp"
 
+#include <cudf/concatenate.hpp>
+#include <cudf/table/table.hpp>
+#include <cudf/table/table_view.hpp>
+
 #include <nvtx3/nvtx3.hpp>
 
+#include <algorithm>
 #include <mutex>
 
 namespace sirius {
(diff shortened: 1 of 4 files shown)


src/op/sirius_physical_partition.cpp
```diff
@@ -173,6 +178,78 @@ std::unique_ptr<operator_data> sirius_physical_partition::execute(const operator
...
+    for (std::size_t p = 0; p < num_fine; ++p) {
+      slot_members[p % num_gpus].push_back(p);
+    }
+
+    std::vector<std::shared_ptr<cucascade::data_batch>> coalesced_batches;
+    std::vector<std::size_t> coalesced_slots;
+    for (std::size_t g = 0; g < num_gpus; ++g) {
+      std::vector<cudf::table_view> group;
+      std::size_t group_bytes = 0;
+      auto flush              = [&]() {
```
wmalpica, line 226: My understanding of what this is trying to improve, is the fact that when you have a lot of partitions, you end up with multiple small copies and small data movements, and this tries to concatenate them so that there are fewer and larger data movements which is faster.

At a high level, what i think this code actually does is that this is taking data from different partitions and concatenating them into one, for example partition 0 and partition 2 and concatenating them into one batch, and partition 1 and partition 3 and concatenating them into another.
This is a problem, i think on two fronts:
1. If we decided that we needed 4 partitions, why are we reducing it back to 2 by concatenating. If that is really what we want to do, then the logic that controls the number of partitions should be the one to change.
2. This is combining partitions depending on their size, and it does this for each task independently. This means that different tasks could combine things differently depending on each tasks inputs, which can result in different partition combinations, which can result in incorrect results. This is especially true in joins, where each side of the join can have batches of different sizes but their partitions need to align.

wmalpica (changes requested):

PR #1041, opened 2026-07-01
build: drop obsolete CMake/vcpkg workarounds

mbrobbel:
Removes build workarounds that are no longer needed after dependency bumps (RAPIDS 26.06, cuCascade at the current submodule pointer). Each change was verified with a from-source vcpkg build that relinks and runs the extension on GPU.

- **NUMA shim + duplicate yaml-cpp link** (`CMakeLists.txt`): cuCascade now resolves libnuma via its bundled `FindNuma` (`Numa::Numa`), so the bare `-lnuma` imported-target shim is dead; yaml-cpp was already linked to both extension targets in the shared block.
- **cuDF Arrow dependency + DLPack patch**: cuDF 26.06 moved interop to nanoarrow and no longer `find_package(Arrow)` (drops building Arrow + its subtree); the DLPack patch targeted the old single-arg `find_and_configure_dlpack` call and no longer matched, so it was inert.
- **cuDF GTest fetch**: `CUDF_BUILD_TESTUTIL=OFF` gates out the test utilities that were the only thing pulling GTest.
- **rmm cudart rewrite + CUDA flag**: rmm 26.06 exports `CUDA::cudart_static` directly (rewrite matched nothing); rmm is CXX-only so `-DCMAKE_CUDA_FLAGS` was reported unused.
- **nvtx3 hand-written config**: use upstream install rules (`NVTX3_INSTALL`), which provide the exported targets and a correct version file (the hand-written config emitted an empty version).

Kept (verified still required): the `CUDA::nvml_static`→shared-stub redirect, the conda-CCCL strip + include ordering, static-cudart preference, the hand-written cccl port (upstream install rules break vcpkg's `share/` config layout), and the kvikio cuFile force-disable (still needed for toolkits whose cuFile lacks/omits the batch/stream API, e.g. the cuda12 CI image).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,18 +81,7 @@ pkg_check_modules(NUMA REQUIRED IMPORTED_TARGET numa)
 pkg_check_modules(LIBURING REQUIRED IMPORTED_TARGET liburing)
 pkg_check_modules(CURL REQUIRED IMPORTED_TARGET libcurl)
 
-# cuCascade - GPU Memory Reservation Library (submodule) cucascade links `numa`
-# as a bare library name (-lnuma). In the vcpkg build the static libnuma.a lives
-# in the vcpkg installed lib dir, which isn't on the default linker search path.
-# Create an imported target so CMake resolves the bare name to the full path
-# instead of emitting -lnuma.
-if(VCPKG_BUILD)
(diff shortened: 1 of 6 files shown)


9prady9 (approved):
nice cleanup 👍🏾

PR #1042, opened 2026-07-01
enable s3 on new io framework

ran-yuan-rui:
## 1. Correctness fixes (crash / deadlock / shared-state)

- **Host→device copy from a null-derived source — SIGSEGV on multi-GB reads (`c44ed2a6`).** A device read whose prefetch-cache coverage is partial emits "gap" segments with a **null host buffer**. The reactor built the H2D copy source as `buffer + offset`, turning `nullptr` into a non-null near-null pointer that `cudaMemcpyAsync` faulted on while draining a multi-GB read. Fix: gap segments no longer fuse across a null buffer; a null-buffer copy is encoded as `src=nullptr` + offset and resolved to the slot's bounce buffer; bounce-staged reads take the event-synchronized recycle path (closing a pinned-buffer reuse race); `copy_async` validates the buffer/bounds and rejects a null/near-null pointer.

- **CURLSH shared-connection-cache self-deadlock — hangs the read (`ab6a6946`).** The per-worker curl SHARE handle (shares the connection / DNS / TLS-session cache across the reactor's pooled easy handles) **funneled every `curl_lock_data` class through one non-recursive mutex**; libcurl holds one lock class while taking another, so the same thread re-locks that mutex and deadlocks — the read's future never resolves and the query hangs. Fix: one mutex per lock class.

- **`enable_duckdb_fallback` was a shared process global (`571be6ae`).** `SET enable_duckdb_fallback = false` on one connection mutated a process-wide static that every other connection read (and that the extension registration used as its default), so a fresh database inherited the last `SET` regardless of connection, and across the test binary a case that set it false leaked into later cases. Fix: read the policy per context via `ClientContext::TryGetCurrentSetting`, register a literal `true` default, keep no global mirror.

- **`semi_future` timed-wait lost wakeup — hang (`a906fa75`).** A `get(timeout)` parked on a raw `futex_wait_until`, which `std::atomic::notify_all` (libstdc++'s internal waiter pool) can never wake, so the wait hung to its deadline. Fix: pair the atomic notify with a raw `FUTEX_WAKE`.

## 2. All pre-#997 S3 tests ported (except real-AWS)

Every S3 test from the old backend is restored onto `rest_ioctx` and green:

- REST I/O integration — `c7924cb5`
- `describe_parquet` over S3 — `e69a5793`
- SQL-over-S3 surface — `5507749e`
- SF10 large + host→device null-segment edge cases — `a3fa147b`
- s3-bench — `e9f9a122`

All run against testcontainers MinIO (HTTP + TLS, auto-started per phase). **The only tests not included are the real-AWS live cases (`[s3][aws]` / `[aws][live]`)** — they need short-lived assume-role credentials and are operator-only, never run in CI.

## 3. REST reactor perf telemetry

`24acaad9` + `500c348d` add a lock-free per-reactor `rest_perf_snapshot`, aggregated across the pool by `rest_ioctx`:

- chunk-GET latency (total / count / max), queue wait, TTFB, observed host→device bytes, **HTTP payload bytes actually read off the wire**, retries, terminal failures, device-stream syncs.
- Micro-timings gate behind `perf_instrumentation`; the byte / retry / terminal counters are always-on and cheap (per-attempt, relaxed atomics — no per-byte cost). The payload-byte counter is incremented **before** the success/retry/terminal branch, so retried and failed response bodies still count as observed wire traffic.

This is what the s3-bench JSON baseline is built from.

## 4. Benchmark vs the old async S3 backend (rough)

Same SF10 `lineitem` parquet (~2.07 GiB object) on MinIO / HTTP, effective read throughput (bytes ÷ wall-clock). "Then" = the pre-#997 async libcurl backend (`s3_ioctx`, RTX 3060, 2026-05-21); "Now" = this REST backend from the local perf baseline:

| backend | scenario | throughput |
|---|---|---|
| async S3 (pre-#997), mc=8 | full read, cache **off** | ~522 MB/s |
| async S3 (pre-#997), mc=8 | full read, cache **on** (prewarmed) | ~895 MB/s |
| REST (post-#997), mc=32 | full read | **~967 MB/s** |

- The REST backend's **cache-off** throughput (~967 MB/s) already **matches/beats the old backend's best prewarmed-cache case** (~895 MB/s) and is ~**1.85×** its cache-off baseline (~522 MB/s).
- Both read exactly the object bytes (`bytes_read / object_size == 1.00`, no over-read) and keep `device_stream_sync_total == 0` — the async-copy design goal is preserved across the rework.
- HTTP vs HTTPS on the REST path: byte-identical payloads; the only delta is the TLS handshake on the first ranged GET (footer-bind ~2.5–3.5× HTTP), amortized after by the shared TLS-session cache.


src/include/scan_manager/sirius_scan_manager.hpp
```diff
@@ -306,13 +307,27 @@ class sirius_scan_manager {
...
   /// the GPU id set fed to the round-robin scan-balancing strategy.
   std::shared_ptr<const sirius::memory::topology_index> _topology_index;
   exec::static_thread_pool _thread_pool;
   std::unique_ptr<exec::scoped_dispatcher> _dispatcher;
   std::shared_ptr<sirius::io::sirius_ioctx> _io_ctx;
+  /// Lazily-built per-backend ioctxs for path-routed datasources (e.g. an s3://
+  /// rest_ioctx alongside the local uring/kvikio `_io_ctx`).  Built exactly once
+  /// per type under the mutex; drained + torn down in the dtor.
+  mutable std::mutex _routed_io_ctxs_mtx;
+  mutable std::unordered_map<sirius::io::io_context_type, std::shared_ptr<sirius::io::sirius_ioctx>>
```
aminaramoon: why is the map mutable ?

src/io/rest/curl_handle.cpp
```diff
@@ -71,17 +71,19 @@ curl_share::curl_share(bool share_connections)
...
 
 void curl_share::lock_cb(CURL* /*handle*/,
-                         curl_lock_data /*data*/,
+                         curl_lock_data data,
                          curl_lock_access /*access*/,
                          void* userp)
 {
-  // Coarse single-mutex serialization across all shared data classes.
-  static_cast<curl_share*>(userp)->_mtx.lock();
+  // Per-class lock (selected by @p data): libcurl holds one class's lock while
```
aminaramoon, line 78: thanks for the fix,

src/op/scan/duckdb_native_gpu_ingestible.cpp
```diff
@@ -245,14 +245,20 @@ bool duckdb_native_gpu_ingestible::has_processed_all_metadata() const
...
+duckdb_native_gpu_ingestible::next_split_provider(ioctx_resolver resolve)
 {
   auto const idx = _next_range_idx.fetch_add(1, std::memory_order_relaxed);
   if (idx >= _num_ranges) { return nullptr; }  // lost the race for the final range
 
   auto const rg_begin = idx * _chunk_row_groups;
   auto const rg_end   = std::min(rg_begin + _chunk_row_groups, _plan.n_row_groups);
 
+  // All ranges read the one `.duckdb` file, so resolve its backend once by db_path.
+  auto io_ctx = resolve(_info->db_path);
```
aminaramoon, line 258: why not throw in the resolve instead

src/include/op/scan/gpu_ingestible.hpp
```diff
@@ -54,6 +55,12 @@ class gpu_ingestible;
 // include cycle; only used by const-reference below. Full definition pulled in by .cpp.
 class scan_operator_input;
 
+/// Resolves the ioctx that should serve a given file path.  The scan path routes
+/// each metadata task by path (s3:// -> rest, local -> uring/kvikio), so the
+/// scan_manager threads this through split_provider into next_split_provider
+/// rather than binding the whole scan to one ioctx.
+using ioctx_resolver = std::function<std::shared_ptr<io::sirius_ioctx>(std::string_view)>;
```
aminaramoon: shouldn;t this be in io namespace ?

aminaramoon (approved):

src/include/scan_manager/sirius_scan_manager.hpp
ran-yuan-rui: It's mutable because ioctx_for_path() memoizes the off-default (s3://) ioctx into this map from a const method. But that const is misleading — the first touch actually starts the reactor threads and allocates the prefetch cache, so it isn't logically const. Nothing forces the const either (the only caller, describe_parquet, is non-const), so I'll make ioctx_for_path / create_datasource non-const and drop the mutable on the map and its mutex.

src/op/scan/duckdb_native_gpu_ingestible.cpp
ran-yuan-rui: will the throw into the resolver, so ioctx_resolver is now "returns a valid ioctx or throws" and both next_split_provider impls drop the null-check.

src/include/op/scan/gpu_ingestible.hpp
ran-yuan-rui: will move ioctx_resolver to sirius::io, next to sirius_ioctx / io_context_type.

PR #1043, opened 2026-07-01
Use `data_batch` standardized on `std::shared_ptr`

dhruv9vats:
Update Sirius for cuCascade’s shared-only data_batch API:

- Construct batches through data_batch::make() and use the non-templated repository manager.
- Respect const read-only representations in result collection.
- Replace destructive table-scan projection with the existing zero-copy, view-backed ownership
  pattern.
- Add regression coverage verifying projection does not invalidate its input.


src/op/sirius_physical_table_scan.cpp
```diff
@@ -204,40 +205,39 @@ std::unique_ptr<operator_data> sirius_physical_table_scan::execute(const operato
...
 
-    auto projected_table = std::make_unique<cudf::table>(std::move(selected));
-    auto projected_rep   = std::make_unique<cucascade::gpu_table_representation>(
-      std::move(projected_table), *space, stream);
-    output_batch = std::make_shared<cucascade::data_batch>(batch_id, std::move(projected_rep));
+    cudf::table_view projected_view(selected);
+    std::size_t const referenced_bytes = sirius::estimate_referenced_column_bytes(
+      input_view, referenced_indices, output_ro.get_data()->get_size_in_bytes());
+    output_batch = sirius::make_data_batch_from_view(
+      projected_view, std::move(output_ro), referenced_bytes, *space, stream);
```
dhruv9vats, line 240: This view-backed ownership pattern also exists at:

https://github.com/sirius-db/sirius/blob/0dd28871c2d883faee43e8eef4a5e9e93eb2978c/src/op/sirius_physical_projection.cpp#L123-L180

mbrobbel (commented):
This needs a submodule change after https://github.com/NVIDIA/cuCascade/pull/154 is merged to `main`

wmalpica (approved):

mbrobbel (approved):

PR #1045, opened 2026-07-01
Rename gpu_expression_executor to expression_evaluator

9prady9:
## What

Rename `gpu_expression_executor` to `expression_evaluator`. The class runs no threads or event loop, so the "executor" name conflicts with the actual task-executor family (`itask_executor`, `gpu_pipeline_executor`, `downgrade_executor`); "evaluator" reflects what it does.

- Rename the class, its `execute`-flavored members (`execute_result`, `execution_mode`, `execute()`/`execute_ast()` → `evaluate*`), and the `expression_executor/` directories, files, and include paths.
- Rename the `expression_executor_strategy` config type to `expression_evaluator_strategy`.
- Drop the redundant `execute_` prefix on specialization filenames.

## SQL setting

Adds the new `expression_evaluator_strategy` setting. The old `expression_executor_strategy` is kept as a deprecated alias that warns and delegates, so existing `SET` statements keep working; it will be removed in a future release.

## Testing

- `pixi run make` passes.
- Expression evaluator / translator / non-duckdb-proof suites pass (171 cases, 4897 assertions).
- `pre-commit run -a` clean.


9prady9:
resolved the conflicts , gng to push in few mins

kevkrist (approved):

mbrobbel (approved):

PR #1047, opened 2026-07-01
Telemetry: report all queries under one session-scoped query group

felipeblazing:
## What

Previously each `sirius_engine` (i.e. each query) minted its own Quent query group, so a whole run surfaced as N separate `default_group` groups with one query each. This moves query-group ownership to the session-scoped `telemetry_context` so every query in a `SiriusContext` reports under a single `<engine_name>-session-<pid>` group.

## Changes

- `telemetry_context` now creates and declares the query group (engine-parented) and exposes it via `query_group_id()`.
- `sirius_engine` drops its per-query group/observer and passes `telemetry_context_->query_group_id()` as the query's parent group.

## Not affected

No config keys, telemetry-generation scripts, or quent server-invocation change — only the *shape* of the emitted telemetry (one group per session instead of one per query). No CI job invokes quent, and no test asserts on the query-group wiring.

### DIFF
--- a/src/include/sirius_engine.hpp
+++ b/src/include/sirius_engine.hpp
@@ -28,7 +28,6 @@
 #include "pipeline/sirius_meta_pipeline.hpp"
 #include "pipeline/sirius_pipeline.hpp"
 #include "telemetry-bridge/gen/query.rs.h"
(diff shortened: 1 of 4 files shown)


dhruv9vats (approved):
Looks great! Thanks 🎉

PR #1050, opened 2026-07-02
Recover expression result type natively, drop DuckDB round-trip

9prady9:
## What

The expression evaluator recovered each result's expression class and return type by rebuilding a full DuckDB expression tree per output column via `sirius::ast::to_duckdb` — an allocation-heavy round-trip used only to read two fields. This recovers them natively from the Sirius AST.

- Add a uniform `sirius::logical_type return_type()` accessor to every AST alternative, plus a `has_return_type` concept + `static_assert` and a `node::return_type()` `std::visit` dispatcher (mirrors `cudf_ast_op_count`).
- `comparison`/`conjunction`/`between`/`in_list` return `BOOLEAN`; `unary_op` returns `BOOLEAN` and throws `not_implemented` for `op_try`; `cast` returns its target type.
- Convert `constant`/`reference` from aggregate structs to a constructor + private `return_type_` + `return_type()` accessor; update field reads to method calls.
- `expression_evaluator`: `post_process` uses `node.holds<reference>()` + `node.return_type()` + `sirius::get_cudf_type`; the filter path asserts the BOOLEAN return type natively. Swap `duckdb::InternalException` for `sirius::internal_exception` and drop all duckdb includes from the `.cpp`.

## Testing

- `pixi run make` passes.
- Expression evaluator / translator / non-duckdb-proof / AST unit suites pass (346 cases, 5519 assertions).
- `pre-commit run -a` clean.

### DIFF
--- a/src/expression/ast/node.cpp
+++ b/src/expression/ast/node.cpp
@@ -73,6 +73,11 @@ std::size_t node::cudf_ast_op_count() const
   return std::visit([](auto const& alt) { return alt.cudf_ast_op_count(); }, v);
 }
 
+logical_type node::return_type() const
+{
+  return std::visit([](auto const& alt) -> logical_type { return alt.return_type(); }, v);
+}
+
 reference const& require_reference(node const* n, std::string_view context)
 {
   if (n == nullptr) { throw_wrong_node_type(context, "non-null reference node"); }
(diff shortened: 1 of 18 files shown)


Jedi18 (approved):
LGTM

mbrobbel (approved):

PR #1052, opened 2026-07-02
Add StarRocks-compatible CRC32 partition-hash GPU kernel

mbrobbel:
Standalone GPU kernel that reproduces StarRocks BE's `BUCKET_SHUFFLE_HASH_PARTITIONED` CRC32 bucket hash bit-exactly, so GPU-computed partitions land on the same compute node as StarRocks' on-disk bucket layout.

**Why:** see #914 §4 ("The Data Path: In → Compute → Out") — the partitioned sink reuses Sirius's existing `PARTITION` machinery "differing only in the hash function," and must reproduce StarRocks' hash bit-exactly per partition type, or co-located rows land on the wrong node and silently corrupt shuffle joins/aggregations. This PR provides that CRC32 hash. **Not wired into any operator yet** — groundwork for the exchange sink.

**Kernel:** standard zlib CRC-32 (poly `0xEDB88320`), seed 0, key columns folded left→right (null = 4 zero bytes; string = raw chars, empty = no-op; int/bool/float = raw little-endian bytes). Decimals take a caller-supplied `decimal_key`, since cuDF's `data_type` drops precision and the DecimalV2/V3 distinction: DecimalV2 and DecimalV3(precision 27, scale 9) use StarRocks' int64+int32 split fold; other DecimalV3 hash raw unscaled bytes; ambiguous/unspecified decimals are rejected. All key columns fold in one memory-bandwidth-bound pass. Date/datetime, INT64 string offsets, sliced strings, and nested types are rejected (follow-ups).

**Tests:** Catch2 suite anchored to published CRC-32 check values and independent zlib golden vectors (split fold, multi-column, nulls).


src/op/partition/crc32_partition_hash.cu
```diff
@@ -0,0 +1,315 @@
...
+    }
+    hashes[idx] = h;  // one global write per row
+  }
+}
+
+cudf::size_type grid_for(cudf::size_type n)
+{
+  // Widen before the ceiling division: `n` can be near INT32_MAX, which cuDF permits.
+  auto const blocks = (static_cast<int64_t>(n) + kBlock - 1) / kBlock;
+  return static_cast<cudf::size_type>(std::min<int64_t>(blocks, 65535));
```
9prady9: Are you intentionally capping this at lower grid size ? This can be larger for newer cards right.

src/op/partition/crc32_partition_hash.cu
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// DecimalV2 / DecimalV3(27,9) fold: int64 integer part then int32 fractional part of the int128
+// value, matching StarRocks' `DecimalV2Value::int_value()` / `frac_value()` (C++ truncated
+// division).
+__device__ __forceinline__ uint32_t crc32_decimal_split(uint32_t seed,
+                                                        uint8_t const* p,
+                                                        uint32_t const* tab)
+{
+  __int128 v             = *reinterpret_cast<__int128 const*>(p);
```
9prady9, line 103: Since we always use CUDA 12 or higher, we don't need to check for this type's support I guess.

9prady9 (approved):
lg

src/op/partition/crc32_partition_hash.cu
mbrobbel: [0d8844f](https://github.com/sirius-db/sirius/pull/1052/commits/0d8844ffa891f52a5e293ae913960fa0854c4b05)

PR #1055, opened 2026-07-02
Remove the legacy gpu_processing execution path

mbrobbel:
Removes the dead legacy Sirius execution path (the `gpu_processing` engine) entirely, keeping only Super Sirius. Legacy was already gated OFF by default (`ENABLE_LEGACY_SIRIUS` / `SIRIUS_ENABLE_LEGACY`), so this deletes code and support that no longer runs.

## Removed
- **Code:** `src/legacy/` and `src/include/legacy/` (the `gpu_processing` / `gpu_buffer_init` engine, buffer manager, legacy expression executor, plan generator, operators, CUDA kernels), plus every `#ifdef SIRIUS_ENABLE_LEGACY` block in `sirius_extension.{cpp,hpp}`.
- **Build / CI:** the `ENABLE_LEGACY_SIRIUS` option and `SIRIUS_LEGACY_*` CMake wiring, the `legacy-release` Makefile target and CMake preset, and the `variant: legacy` CI build (the now single-valued `variant` matrix dimension is dropped).
- **Config:** the `use_pin_memory` / `use_pin_memory_for_caching` SET options and `USE_PIN_MEM_FOR_*` flags, whose only consumer was the legacy buffer manager.
- **Scripts / tests / docs:** legacy benchmark + SQL runners, the ClickBench harness, the legacy SQLLogic tests, the `cpu_cache` test, orphaned legacy-coupled C++ tests (with `orphan_test_allowlist.txt` reconciled), `docs/gpu_processing.md`, and stale doc/comment references.

## Behavior change
The `gpu_processing` and `gpu_buffer_init` table functions and the `use_pin_memory*` SET options no longer exist — `gpu_execution` (transparent SQL interception) is the sole entry point.

## Validation
- Release build (extension + DuckDB + `sirius_unittest`) clean.
- `sirius_unittest`: all 1469 cases pass on GPU.
- `pre-commit run -a` green, including the `check-orphan-tests` gate.

## Not included (possible follow-ups)
- Stale `parquet_scan_task` references left by an earlier scan-manager cleanup — unrelated to this removal.
- Re-adding a Super-Sirius ClickBench harness / porting the `bugfix.test` regression cases, if that coverage is still wanted.


bwyogatama:
I dont think we can do this yet as our ClickBench #1 result is using this, and if someone wants to reproduce it it will break if we do this

PR #1057, opened 2026-07-02
feat(join): support expressions in hash-join equality conditions (#329)

### LINKED ISSUE #329: Enable support for expressions in join conditions for hash joins
given the following query:
select * from customer c join nation n on c.c_custkey = n.n_nationkey * 10

We would need to have expression evaluation perform the multiplication, but the more challenging part is dealing with partitioning.
For partitioning we have a couple options:
1. we dont do any partitioning, which is simple but wont scale
2. we move the expression evaluation to a projection which happens before projection. This is challenging because then we have to rewrite all the expressions in the joins (and the partition and concat operators which feed the join)
3. we perform the expression evaluation as a way of materializing the columns prior to partitioning (this is redundant work, but may be the easiest and most scalable to do)

9prady9:
Closes #329.

Hash joins rejected equality conditions with an expression key (e.g. `ON c.c_custkey = n.n_nationkey * 10`), throwing `Unsupported join condition expression`.

## Constraint
- DuckDB evaluates join keys inline in the hash join; it never lowers them into a projection.
- Sirius feeds every hash join through `PARTITION -> CONCAT` (all modes: STANDARD / BUILD_PROBE / MIXED_JOIN).
- `PARTITION` — not the join — is the first key consumer, and it hashes columns by index and cannot evaluate expressions.
- So an expression key must be a materialized column *before* partitioning; the inline model can't be reused.

## Change
- In `plan_comparison_join`, for each complex equality-condition side, append it as a column via `push_projection`.
- Rewrite that condition side to a plain `BoundReferenceExpression` at the appended index.
- `PARTITION`/`CONCAT`/hash-join are unchanged — they derive keys from the now-plain-reference conditions.
- The synthetic key column is excluded from the join output (empty projection map made explicit over original columns).
- Co-location is correct because DuckDB binds both comparison sides to a common type, so the materialized column matches the type the other side hashes.

```
scan -> PROJECTION[append key=expr] -> PARTITION[key=col] -> CONCAT -> HASH_JOIN[key ref, not in output]
```

## Scope
- Plain/cast-of-ref key sides keep the existing fast path (no materialization).
- Inequality-condition sides stay inline (evaluated as the mixed-join predicate).
- Untranslatable key expressions keep the existing throw (no new CPU fallback).
- Hash joins only; nested-loop / range joins out of scope.

## Tests
- Planner (`test_join_expression_key.cpp`): materialization into a projection, condition rewritten to a reference, synthetic key excluded from output, plain-column joins untouched.
- Integration (`test_gpu_execution_tpch.cpp`): GPU-vs-CPU correctness for expr on build/probe/both sides + a mixed eq/inequality join, using the `nation`/`customer` example.
- Full suite 1509 passing (exit 0); `pre-commit run` clean.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -639,6 +639,7 @@ set(TEST_SOURCES
     test/cpp/parallel/test_task_executor.cpp
     test/cpp/planner/test_distinct_hash_join_detection.cpp
     test/cpp/planner/test_dynamic_filter_router.cpp
+    test/cpp/planner/test_join_expression_key.cpp
     test/cpp/planner/test_projection_fold.cpp
     test/cpp/pipeline/test_gpu_pipeline_executor.cpp
     test/cpp/pipeline/test_oom_reschedule.cpp
--- a/src/planner/sirius_plan_comparison_join.cpp
+++ b/src/planner/sirius_plan_comparison_join.cpp
@@ -21,6 +21,7 @@
 #include "duckdb/main/client_context.hpp"
 #include "duckdb/main/settings.hpp"
 #include "duckdb/parser/constraints/unique_constraint.hpp"
+#include "duckdb/planner/expression/bound_cast_expression.hpp"
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
 #include "duckdb/planner/operator/logical_aggregate.hpp"
@@ -30,7 +31,9 @@
 #include "duckdb/planner/operator/logical_order.hpp"
 #include "duckdb/planner/operator/logical_projection.hpp"
 #include "duckdb/planner/table_filter.hpp"
+#include "expression/ast/from_duckdb.hpp"
 #include "expression/ast/node.hpp"
+#include "expression/ast/reference.hpp"
 #include "expression/ast/to_duckdb.hpp"
 #include "expression/join_condition.hpp"
 #include "helper/type_conversions.hpp"
@@ -39,6 +42,7 @@
 #include "op/sirius_physical_hash_join.hpp"
 #include "op/sirius_physical_nested_loop_join.hpp"
 #include "planner/sirius_physical_plan_generator.hpp"
+#include "planner/sirius_plan_projection_utils.hpp"
 #include "sirius_context.hpp"
 
 #include <algorithm>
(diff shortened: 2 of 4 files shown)


9prady9:
## TPC-H SF100 benchmark — expression-join-key feature: W/ vs W/O

**Hardware:** NVIDIA RTX PRO 6000 Blackwell Workstation Edition (96 GB)
**Data:** TPC-H SF100 parquet (26 GB, `tpchgen-rs`) · **Config:** default `test/cpp/integration/integration.yaml` (48 GB GPU budget)
**Harness:** `performance_test.py`, `--engine gpu`, each query run in its **own process**, warm-iteration time.
**"without change"** = same tree with the `materialize_expression_join_keys` call disabled and rebuilt. `speedup = without ÷ with`.

| TPC-H q# | without (s) | with (s) | speedup | failure reason (if any) |
|---|---:|---:|---:|---|
| q1–q22 (stock, plain-column joins) | — | — | **~1.0×** (0.95–1.05×) | no-op: byte-identical plan. Pre-existing wrong results on q4, q6, q7, **q9 (catastrophic)**, q12 — reproduce on baseline |
| **q23** — expr key on one side | 1.773 | **0.840** | **2.11×** | |
| **q24** — expr keys on both sides | 2.107 | **0.909** | **2.32×** | |
| **q25** — expr eq keys + plain inequality | 2.281 | **1.039** | **2.19×** | |

**Stock TPC-H (q1–q22):** the feature is a no-op — every stock join key is a plain column, so `materialize_expression_join_keys` early-returns and the plan is byte-identical. Per-query speedups are all ~1.0× (variation is single-iteration measurement noise).

**Expression-key variants (q23–q25):** ~2.1–2.3× faster. Without the change these joins hit `Transparent execution fallback: Unsupported join condition expression` and run on **DuckDB CPU**; with the change they run on **GPU**. All three validated **byte-exact** against DuckDB.

### What q23–q25 are (not standard TPC-H)

Added to `test/tpch_performance/queries.py` to exercise the feature. Each is a `lineitem ⋈ orders` join; the modification introduces an **expression** in the join key — exactly what #329 enables. Diffs show the plain-column join (runs on GPU today) vs the expression-key variant (previously CPU-fallback, now GPU):

**q23 — expression on one side:**
```diff
  select count(*) as c, sum(l_extendedprice) as s
  from lineitem l join orders o
-   on l.l_orderkey = o.o_orderkey
+   on l.l_orderkey = o.o_orderkey * 2
```

**q24 — expression on both sides:**
```diff
  select count(*) as c, sum(l_extendedprice) as s
  from lineitem l join orders o
-   on l.l_orderkey = o.o_orderkey
+   on l.l_orderkey * 2 = o.o_orderkey * 2
```

**q25 — expression equality keys + plain inequality (mixed join):**
```diff
  select count(*) as c, sum(l_extendedprice) as s
  from lineitem l join orders o
-   on l.l_orderkey = o.o_orderkey
+   on l.l_orderkey * 2 = o.o_orderkey * 2
    and l.l_extendedprice < o.o_totalprice
```

> The `*2` scaling is semantics-preserving (both sides scaled identically), so results match the plain join and validate against DuckDB — while forcing the key through the expression-materialization path.

### Note on cumulative (single-session) runs

The table runs **each query in its own process**, where all 25 pass on both builds. When the 22 stock queries are instead run **back-to-back in one connection**, the GPU crashes non-deterministically with `cudaErrorIllegalAddress` / `cudaErrorMisalignedAddress` — the crash lands on a different query each run (observed q1, q7, q19), on **both builds** and at both 48 GB and 81 GB GPU budgets. It is session-cumulative and **not** caused by this change (reproduces on baseline). The per-query correctness mismatches noted above are likewise pre-existing.

wmalpica (approved):

PR #1058, opened 2026-07-02
fix(bench): update parquet_benchmark for buffer_pool API change

roaramburu:
## Summary
- `test/io/parquet_benchmark.cpp` referenced `sirius::io::cache::buffer_pool::CHUNKS_PER_SLAB`, which was removed by #997's IO cache redesign, so the (non-CI) benchmark no longer compiled.
- Replaced it with `cucascade::memory::fixed_size_host_memory_resource::default_pool_size`, which is what the removed constant mirrored — no behavior change to the cache/IO framework itself.

## Test plan
- [x] Built the target: `cmake --build --preset release --target parquet_benchmark`
- [x] Ran both datasource modes against `test_datasets/tpch_parquet_sf30/lineitem/*.parquet`: `cudf` and `uring`, both completed and produced sane throughput numbers
- [x] `pixi run pre-commit run -a` passes

### DIFF
--- a/test/io/parquet_benchmark.cpp
+++ b/test/io/parquet_benchmark.cpp
@@ -288,19 +288,20 @@ int main(int argc, char** argv)
     // Size the buffer pool to fit the working set, plus headroom.
     constexpr uint32_t POOL_MAX_SLABS       = 20;
     constexpr size_t INFLIGHT_BUDGET_CHUNKS = 2048;
+    constexpr size_t CHUNKS_PER_SLAB =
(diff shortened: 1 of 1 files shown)


mbrobbel (approved):

roaramburu:
@aminaramoon i couldn't run the io benchmark test without this change. seems to be something that changed with the big IO PR. LMK if this is okay, please.

aminaramoon (approved):

PR #1065, opened 2026-07-03
Handle zero-input sources end to end: fix S3 pushdown hangs

### LINKED ISSUE #1061: Hang when a scan produces zero input
When a scan produces zero input (no rows/batches), the query hangs instead of completing with an empty result. This is easy to reproduce on the S3 parquet path.

Likely area (hypothesis, needs confirmation): with no batches ever produced, downstream `get_next_task_hint()` / barrier bookkeeping never observes a "source finished with no data" transition, so no task is created and the pipeline never reports finished.

Related: #475 added empty-result support for the general case; the zero-input path through the current IO framework's S3 parquet scan still hangs.

ran-yuan-rui:
## What happened

Run a selective filter over s3:// parquet — say `WHERE n_regionkey = 99` on nation — and filter pushdown prunes **every** row group at runtime. That's the whole point of pushdown on S3: don't download data you don't need. But a fully-pruned scan legally produces **zero tasks**, and the engine quietly assumed every source produces at least one. Result: the query hangs forever. Easy to hit on S3, where selective predicates over remote parquet are the normal case.

Reproduce: before this PR, any all-pruned s3:// scan feeding an aggregate or join hangs. After this PR it's covered by `make s3-test` — the new matrix runs these exact shapes against a CPU oracle with a watchdog, so a regression fails in seconds instead of hanging CI.

```mermaid
flowchart TD
  A[S3 parquet scan with filter pushdown] --> B[Parquet stats prune all row groups]
  B --> C[Scan source legally produces zero splits / zero tasks]
  C --> D{Old engine assumption}
  D -->|Every source creates at least one task| E[Pipeline completion only happens from task-completion callback]
  C --> F[No task exists]
  F --> G[No task-completion callback fires]
  G --> H[Source pipeline never finishes]
  H --> I[Downstream aggregate / join / result collector waits]
  I --> J[Query hangs forever]

  B -. initial workaround .-> W[Emit one synthetic empty parquet split]
  W -. creates a task .-> E
  W -. local workaround only .-> K[Unblocks parquet case but hides broader zero-task protocol gap]
```

## RCA and the layered fix

"Zero input" wasn't one bug — the assumption was baked into several layers, and each needed its own fix (one commit per layer, bottom-up):

1. **Pipeline protocol**: a zero-task source never completed its pipeline. Now the finish guard, a creation-exit re-evaluation, and a completion listener on the result-collector pipeline handle it.

2. **Task scheduling**: only the first scan is scheduled directly; everything else wakes up via task-completion callbacks — which a zero-task finish never fires. Two holes: (a) creation requests now re-evaluate their upstream dependency closure (only while unfinished — re-notifying finished pipelines re-schedules consumers and livelocks the query, we hit that on TPC-H Q4); (b) a probe-side partition waits for the build sibling to negotiate the partition count, but a build side that finished empty never negotiates — the probe now resolves the count locally in that case.

3. **Operator semantics**: zero input is not always zero output. Ungrouped aggregates emit the identity row (`count(*) = 0`, sum/min/max = NULL). Joins with one dead side emit join-type-correct output directly: LEFT/RIGHT/FULL pad from the survivor, ANTI keeps everything, MARK marks all-false. Dead ("never saw input") is distinguished from drained, so regular joins are untouched.

4. **Bugs found on the way**, fixed in the same PR: right/full-outer nested loop joins always failed with "Repository already exists" (DuckDB's build-side scan pipeline was dropped for hash joins but not NLJs — reproduces without any pruning); MARK NLJs (`x < ANY (...)`) were advertised as supported but threw in every path — now implemented; NLJ SEMI/ANTI/MARK outputs ignored the left projection map.

Impact on normal pipelines: none intended, and the full suite agrees. The scheduling changes only fire on the zero-task path (dependency re-eval touches unfinished pipelines only; partition-count fallback only when the build side provably never had
input), and drained sides never take the zero-side join path.

### 2. Final Fix Stack

```mermaid
flowchart TD
  Z["Zero-task source is valid"] --> P["Pipeline protocol fix"]

  P --> P1["Creation exit re-evaluates pipeline status"]
  P --> P2["Result collector pipeline gets a completion listener"]
  P --> P3["Finish guard checks source exhausted and connector drained"]

  P --> S["Scheduling fix"]
  S --> S1["Task creation re-evaluates upstream dependencies"]
  S --> S2["Do not re-notify already finished pipelines"]
  S --> S3["Partition probe resolves count when build side is empty"]

  S --> O["Operator semantics fix"]

  O --> A["Ungrouped aggregate"]
  A --> A1["count emits 0"]
  A --> A2["sum min max avg first emit typed NULL"]

  O --> G["Grouped aggregate"]
  G --> G1["Zero input stays zero rows"]

  O --> J["Join zero-side handling"]
  J --> J1["INNER and SEMI can be empty"]
  J --> J2["LEFT RIGHT FULL preserve survivor rows with NULL padding"]
  J --> J3["ANTI keeps survivor rows"]
  J --> J4["MARK emits false marks"]

  O --> N["Extra bugs found while fixing"]
  N --> N1["RIGHT and FULL nested loop join duplicate wiring"]
  N --> N2["MARK nested loop join was missing"]
  N --> N3["NLJ SEMI ANTI MARK ignored left projection map"]

  J --> R["Disable synthetic empty split fallback by default"]
  R --> Done["All-pruned S3 pushdown completes with correct SQL result"]
```

```mermaid
flowchart LR
  C1[02659288<br/>Parquet-scoped empty split workaround] --> C2[72023c8f<br/>Zero-task pipeline completion]
  C2 --> C3[e0251300<br/>Ungrouped aggregate identity row]
  C3 --> C4[5c9e409<br/>Schedule zero-task dependency chains]
  C4 --> C5[e0b872a<br/>Join zero-side semantics]
  C5 --> C6[4f45ffcf<br/>Disable empty-split fallback by default]
  C5 --> C7[fa2c2bb<br/>Drop NLJ build-side scan pipelines]
  C5 --> C8[7eddc8c<br/>Apply NLJ left projection map]
  C2 --> C9[7232ac8a<br/>Build routed ioctx outside routing-map mutex]
```

## Known limitation (separate issue)

MARK joins return `false` instead of `NULL` for unmatched rows when NULLs are involved (three-valued semantics). This is pre-existing and engine-wide — the hash MARK join on /dev does the same — so it's tracked in its own issue rather than fixed here.

## How it's tested

- **S3 pushdown matrix** (`test_s3_sql_surface.cpp`, runs in `make s3-test`): every query runs twice — GPU over all-pruned s3:// parquet vs DuckDB CPU over the same local parquet — and the results must match. Cells cover hash and nested loop joins × INNER/LEFT/RIGHT/FULL/SEMI/ANTI/MARK/EXISTS/NOT EXISTS/IN × dead-left/dead-right/ both-dead × both fallback modes, plus aggregate identity and count(*) composition. Serial with a per-query watchdog: a hang becomes a fast failure.｜

- **Zero-task protocol unit tests** (`[zero-task-protocol]`).

- **Operator-level projection tests** (`test_physical_mark_join.cpp`): construct the NLJ with a non-identity projection map and assert output schema/values for MARK/SEMI/ANTI on both the regular and empty-side paths.

- **Full suite**: 1506 test cases green with no env vars set (protocol-default path).
  TPC-H Q4 ran 20× to pin the scheduling livelock fix.

### DIFF
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -31,11 +31,9 @@
 #include <duckdb/parallel/thread_context.hpp>
 
 #include <algorithm>
-#include <functional>
 #include <limits>
 #include <optional>
 #include <unordered_map>
-#include <unordered_set>
 
 namespace sirius::creator {
 
@@ -226,60 +224,12 @@ void task_creator::manager_loop()
       continue;
     }
 
-    auto* original_node = request->node;
-    if (original_node == nullptr) { continue; }
+    auto node = request->node;
+    if (node == nullptr) { continue; }
 
-    auto* node = get_operator_for_next_task(original_node);
+    node = get_operator_for_next_task(node);
 
-    if (node == nullptr) {
-      // Zero-task exit before dispatch: the
-      // hint resolved to nothing (e.g. the split connector closed before this
-      // request ran), so no creation lambda will run the paired post-loop
-      // re-evaluation. Re-evaluate the request's pipeline AND its transitive
-      // dependency closure, upstream-first: only the first scan is scheduled
-      // directly, so a zero-task upstream chain (e.g. an all-pruned second
-      // scan feeding a join) may never have had a status evaluation at all —
-      // the finish guard of every pipeline in between blocks on it, and no
-      // task completion will ever come to unblock the cascade. Dependencies
-      // are evaluated only while unfinished: their finish transition performs
-      // its own one-shot downstream notification, whereas re-evaluating an
-      // already-finished dependency re-notifies level-triggered and
-      // re-schedules its consumers, whose zero-task creation exits dispatch
-      // this handler again — a self-sustaining request storm that livelocks
-      // the query. Only the requesting pipeline keeps the unconditional
-      // level-triggered re-evaluation (its re-notify is what re-delivers a
-      // lost query_finished event). Dispatch on the reserved pool slot, never
-      // inline on this manager thread: update_pipeline_status() can block on
-      // _status_mutex held across a scan's blocking get_next_split().
-      if (auto original_pipeline = original_node->get_pipeline()) {
-        duckdb::vector<duckdb::shared_ptr<pipeline::sirius_pipeline>> eval_order;
-        std::unordered_set<pipeline::sirius_pipeline*> visited;
-        std::function<void(const duckdb::shared_ptr<pipeline::sirius_pipeline>&)> visit =
-          [&](const duckdb::shared_ptr<pipeline::sirius_pipeline>& p) {
-            if (!p || !visited.insert(p.get()).second) { return; }
-            for (auto& dep : p->dependencies) {
-              visit(dep);
-            }
-            eval_order.push_back(p);
-          };
-        visit(original_pipeline);
-        _bounded_pool->dispatch(
-          std::move(slot), [this, eval_order = std::move(eval_order), original_pipeline]() {
-            try {
-              for (auto& p : eval_order) {
-                if (p != original_pipeline && p->is_pipeline_finished()) { continue; }
-                p->update_pipeline_status(false);
-              }
-            } catch (const std::exception& e) {
-              SIRIUS_LOG_ERROR("Task Creator: Exception during pipeline status re-evaluation: {}",
-                               e.what());
-              _task_scheduler->terminate_query(std::current_exception());
-              stop();
-            }
-          });
-      }
-      continue;
-    }
+    if (node == nullptr) { continue; }
 
     // Dispatch the task creation work to the pool
     _bounded_pool->dispatch(std::move(slot), [this, node]() mutable {
@@ -458,13 +408,12 @@ void task_creator::manager_loop()
           task_lock.unlock();
           _task_scheduler->schedule(std::move(task));
         }
-        // Unconditional re-evaluation at every creation exit
-        //: with the source-exhaustion finish
-        // guard, "last task completed at T1, connector closed at T2>T1" has no
-        // later mark_task_completed() to re-check the pipeline — this call,
-        // observing the now-exhausted source, is the paired re-evaluation.
-        // Without it, normal fast-GPU queries would hang, so it must not be
-        // gated on the zero-task case.
+        // Unconditional re-evaluation at every creation exit: with the
+        // source-exhaustion finish guard, "last task completed at T1,
+        // connector closed at T2>T1" has no later mark_task_completed() to
+        // re-check the pipeline — this call, observing the now-exhausted
+        // source, is the paired re-evaluation. Without it, normal fast-GPU
+        // queries would hang.
         pipeline->update_pipeline_status(false);
       } catch (const std::exception& e) {
         SIRIUS_LOG_ERROR("Task Creator: Exception during task creation: {}", e.what());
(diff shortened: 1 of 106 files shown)


src/creator/task_creator.cpp
```diff
@@ -224,12 +226,60 @@ void task_creator::manager_loop()
...
-
-    if (node == nullptr) { continue; }
+    auto* original_node = request->node;
+    if (original_node == nullptr) { continue; }
+
+    auto* node = get_operator_for_next_task(original_node);
+
+    if (node == nullptr) {
+      // Zero-task exit before dispatch: the
+      // hint resolved to nothing (e.g. the split connector closed before this
```
wmalpica: This comment is long and confusing and its really only relevant from the perspective of what this PR is fixing.
I would change it to something like:
"Get operator for next task returning nothing can indicate that its waiting on its dependencies. This can happen when there are operators receiving zero tasks such as when all data has been pruned. Lets update the status of all dependencies to see if any pipelines are complete after zero tasks and to re-trigger the appropriate next task creation. "

src/creator/task_creator.cpp
```diff
@@ -408,6 +458,14 @@ void task_creator::manager_loop()
           task_lock.unlock();
           _task_scheduler->schedule(std::move(task));
         }
+        // Unconditional re-evaluation at every creation exit
+        //: with the source-exhaustion finish
```
wmalpica: This comment is confusing, how about we just say:
"For each task creation request, we always want to check the pipeline status at the end in case its a no-task situation to prevent a hang."

wmalpica (changes requested):
I think we should use the empty batch protocol instead of the zero batch protocol, I think that leads to simpler code. I the process of evaluating which protocol would be best I had Claude implement it and I have created a PR against this PR https://github.com/ran-yuan-rui/sirius/pull/1

src/creator/task_creator.cpp
ran-yuan-rui: Sure - as we go with the empty batch protocol - this comment will be removed

src/creator/task_creator.cpp
ran-yuan-rui: OK - as we go with the empty batch protocol - this comment will be removed

ran-yuan-rui:
> I think we should use the empty batch protocol instead of the zero batch protocol, I think that leads to simpler code. I the process of evaluating which protocol would be best I had Claude implement it and I have created a PR against this PR [ran-yuan-rui#1](https://github.com/ran-yuan-rui/sirius/pull/1)

Merged your PR as-is (fast-forward, use the empty batch protocol)

- **One remaining hole, found and fixed (`c4e6b593`):** a truly-empty native table (zero row groups, e.g. freshly created) still produced zero parse ranges, so no scan_info ever reached the coalescer and your empty-split fallback never seeded its template — `count(*)` returned an empty result instead of one row `0`, and a join with such a table on one side hung. Fix is one line (always slice at least one parse range; the `[0,0)` walk is already viable and seeds the template), plus `[empty-table]` tests: count identity, schema-preserving empty scan, left-join padding, and an empty-parquet regression cell.
- Your two inline comments on the old zero-task comments are moot now — that code went away with your rewrite.

wmalpica (approved):

PR #901, opened 2026-06-08
Added ARRAY data type (fixed-sized lists)

yayen-lin:
### Summary
- Maps DuckDB's ARRAY type to cuDF LIST in the type system.
- cuCascade handles the H2D transfer and memory management for the LIST column.
- list_offset_fixup.cpp/.hpp: a temporary workaround (see [#147](https://github.com/NVIDIA/cuCascade/issues/147)).

### Changes
**Type system**
- `src/include/helper/logical_type.hpp`: add the ARRAY logical type.
- `src/helper/type_conversions.cpp`: ARRAY to/from DuckDB conversion.
- `src/include/cudf/cudf_utils.hpp`: map ARRAY to/from cuDF LIST.

**Scan - Metadata walk**
- `src/include/op/scan/duckdb_native_metadata.hpp`: add ARRAY child-segment buckets and is_array flag to column metadata.
- `src/op/scan/duckdb_native_metadata.cpp`: walk an ARRAY column's three segment trees into their buckets.

**Scan - Decode**
- `src/op/scan/duckdb_native_decoder.{cpp,hpp}`: stage/decode ARRAY columns on the GPU; enable ALP/ALPRD codecs.

**Read-back**
- `src/include/op/result/host_table_chunk_reader.hpp`: add child mask accessor / child null count fields.
- `src/op/result/host_table_chunk_reader.cpp`: copy element-level child validity back into the DuckDB ARRAY vector.

**LIST offset fixup**
- `src/helper/list_offset_fixup.cpp`, `src/include/cudf/list_offset_fixup.hpp`: recast LIST offsets back to INT32 after cuCascade's H2D promotes them to INT64.
- `src/include/pipeline/batch_lock_utils.hpp`: apply the fixup after the H2D conversion.
- `CMakeLists.txt`: add list_offset_fixup.cpp and tests to the build.

**Tests**
- `test/cpp/helper/test_logical_type.cpp`: ARRAY logical type tests.
- `test/cpp/helper/test_cudf_utils.cpp`: ARRAY to cuDF LIST mapping tests.
- `test/cpp/scan/test_duckdb_native_walker.cpp`: unit tests for the ARRAY metadata walk.
- `test/cpp/integration/test_gpu_execution_array.cpp`: end-to-end ARRAY scan tests.
- `test/cpp/utils/gpu_execution_fixture.hpp`: shared file-backed native-scan fixture + GPU-vs-CPU comparator

PR #1067, opened 2026-07-03
Re-enable GPU scan and projection of nested Parquet columns (STRUCT/LIST/MAP), based on new IO frame work

### LINKED ISSUE #1075: GPU parquet scan can't read nested columns (struct/list/map) — throws at bind
## Summary

Any parquet file with a top-level STRUCT, LIST, or MAP column is unreadable on the GPU path. The scan throws at bind:

    SELECT * FROM read_parquet('nested.parquet');
    -- Error: nested parquet types not yet supported

On the local path this forces a CPU fallback; on the S3 (GPU-only) path it fails outright, so a file with even one nested column can't be queried at all.

## Expected

A parquet file with nested columns can be **read and projected** on the GPU — selecting a nested column returns its values (matching DuckDB's own `read_parquet`), and selecting the flat columns alongside it works too. Operating *on* a nested column (filter / group by / join) can stay unsupported for now, but must fail with a clear error rather than a crash or hang.

## Scope

**In:**
- Bind: `extract_schema` maps nested parquet schema subtrees to DuckDB STRUCT/LIST/MAP, matching DuckDB's `read_parquet` bind shape (3-level LIST encoding, MAP key_value group).
- Materialize: read cuDF nested columns into DuckDB struct/list/map vectors recursively (list offsets → `list_entry_t`, value/field children recurse, per-level validity distinguishing null vs empty list and null struct vs null field).
- Reject cleanly: a nested column in WHERE / GROUP BY / JOIN ON (including pushed-down table filters) raises a clear "nested column operation … unsupported" error naming the column.
- Coexist with the existing fixed-size ARRAY support (#901) — the ARRAY fast path is unchanged.

**Out:**
- Operating on nested columns (filter / aggregate / join) — deferred; explicit unsupported error.
- MARK three-valued semantics and other flat-path concerns are unrelated.

ran-yuan-rui:
Closes #1075 

## What

The GPU parquet scan threw at bind on any top-level STRUCT / LIST / MAP column —nested-column files were unreadable on the S3 (GPU-only) path and fell back to CPU for local files. This PR reads and projects nested columns on the GPU.

## How

- **Bind**: `extract_schema` recursively maps nested parquet schema subtrees to DuckDB STRUCT/LIST/MAP, matching DuckDB's own `read_parquet` bind shape (3-level LIST encoding, MAP key_value group).
- **Materialize**: `host_table_chunk_reader` recursively fills DuckDB struct/list/map vectors from cuDF nested columns — list offsets → `list_entry_t`, value/field children recurse, validity copied per level (null vs empty list; null struct vs null field; misaligned list-child slices read bit-by-bit). The full DuckDB result types are threaded through the result collector so nested children are built faithfully (the flat GPU type representation can't hold them).
- **Reject**: plan generation accepts a projected nested column but raises a clear unsupported error naming the column when one appears in WHERE / GROUP BY / JOIN ON, including predicates pushed into the scan's table filters.

## Coexistence with #901 (ARRAY)

The fixed-size ARRAY path added by #901 is preserved: a cuDF LIST materializing into a DuckDB ARRAY still takes the contiguous fast copy; only variable-length LIST / MAP /STRUCT go through the new recursive materializer.

## What's out

- Operating on nested columns (filter/aggregate/join) — explicit unsupported error.

## Test plan

- `[parquet_helpers][schema][nested]` — schema mapping (struct/list/map/deep, resume-at-next-scalar): 4/4.
- `[scan_manager][describe_parquet][nested]` — bind shape vs DuckDB CPU, local + S3.
- `[s3][integration][nested]` — GPU scan values vs a local DuckDB CPU oracle across struct/list/map/deep, null + empty edges, a multi-chunk boundary; plus the unsupported-operation boundary (clear error, no crash/hang).
- `make s3-test` 31 → 33 cases; full Catch2 1539/1539; ARRAY (#901) 15/15 — no regression.


src/include/io/parquet_helpers.hpp
```diff
@@ -36,15 +36,20 @@ struct schema_info {
...
+/// and TIMESTAMP.
 ///
-/// @throws std::runtime_error on a nested top-level column, a malformed
-///         schema, or an unsupported physical type.
+/// Nested columns recurse over the preorder schema subtree: a plain group →
+/// @c LogicalType::STRUCT(children); a group annotated LIST (the standard
+/// 3-level @c LIST → repeated group → element encoding) → @c LogicalType::LIST;
+/// a group annotated MAP (@c MAP → repeated key_value → {key, value}) →
+/// @c LogicalType::MAP(key, value). The mapping matches DuckDB's own
+/// @c read_parquet bind shape for the same file.
```
9prady9: nit: the ai tool must have smudged together all the text. it will be better to read for human if these inline code snippets for categories are present as bullet points.

src/include/op/result/host_table_chunk_reader.hpp
```diff
@@ -155,11 +182,22 @@ class host_table_chunk_reader {
...
    * @throw std::runtime_error If there is a mismatch in column metadata and types, if the row count
    * is negative or inconsistent across columns, or if the duckdb output logical type for any column
    * is HUGEINT.
    */
+  host_table_chunk_reader(duckdb::ClientContext& client_ctx,
+                          cucascade::host_data_representation const& host_table,
+                          duckdb::vector<duckdb::LogicalType> const& types);
+
+  /**
+   * @brief Convenience overload taking sirius logical types. NOTE: sirius::logical_type cannot
```
9prady9: @ran-yuan-rui This sounds like a limitation in `sirius::logical_type`. Would it help if we add support for that type to be nested as well ? I am not clear on the requirement yet, please shed some light on that front. My idea was that we can't execute on those nested columns, correct me if that has changed.

I don't know how much effort it is, but since nested columns is being enabled. Was wondering about it.

9prady9 (commented):
thank you. Did a quick pass, couple of comments.

src/include/io/parquet_helpers.hpp
ran-yuan-rui: Good point. I’ll split the leaf and nested mappings into bullets so the schema categories are easier to scan.

src/include/op/result/host_table_chunk_reader.hpp
ran-yuan-rui: You are correct: this PR enables GPU **reading and projection pass-through** of nested columns only. Engine operators still do not execute on nested values — there are no filters, joins, or aggregations over
STRUCT/LIST/MAP — and that has not changed.

That is also why I did not extend `sirius::logical_type` with general recursive nested child metadata in this PR. Nothing inside the engine currently consumes those nested types. The only boundary that needs the complete child types is the result collector, and it already receives them from the DuckDB bind types through the first overload.

Extending `sirius::logical_type` would be an engine-wide change with no consumer yet. I would rather do that together with the follow-up work that makes operators execute on nested data, where it has a concrete payoff. The existing Sirius-type overload remains for flat-column call sites and unit tests, as the note says.

9prady9 (approved):

PR #1068, opened 2026-07-03
Add `data_batch` telemetry

### LINKED ISSUE #1062: Add memory consumption and data movement instrumentation via Quent
Sirius already emits Quent telemetry for operator/pipeline/task activity (see `docs/super-sirius/quent-telemetry.md`). Extend the instrumentation with two new dimensions:

**Memory consumption**

- Per-memory-space usage over time (each GPU + host tiers): allocated/reserved bytes and peaks.
- Attribution to operators/pipelines where possible (reservation vs. actual usage).

**Data movement**

- Transfers between memory spaces — host↔GPU, GPU↔GPU (peer or staged), disk spills — with source/destination space, bytes, duration, and the triggering task/operator.
- Downgrade-executor activity (what was spilled, when, and why).

Goal: analyze runs in the Quent UI alongside the existing operator timeline — e.g. correlate stalls with memory pressure or cross-GPU traffic. This is especially useful for the multi-GPU work, where quantifying cross-device copies is a prerequisite for reducing them (#993, #995).

The ndjson exporter and the `enable_quent` / `output_directory` config already exist; this is about adding new event/metric types to the emitted schema.

dhruv9vats:
Add data_batch telemetry (engine emission → analyzer → quent UI)

Summary

Instruments the lifecycle of every GPU data_batch and surfaces it in the quent UI. Each batch now emits a telemetry FSM — Constructed → Stationary ⇄ InTransit → Destructed — attributed to the pipeline that produced it, so you can see per-batch memory residency and cross-tier transfers, and per-resource (memory/channel) occupancy over a query's execution.

Engine (C++ / Super Sirius)

- quent_data_batch_probe (telemetry/data_batch_probe.hpp): an idata_batch_probe that forwards batch state transitions to quent, resolving memory/channel handles from a memory_context.
- memory_context: registers a quent memory handle per memory space and a channel handle per space-pair, owned by the telemetry_context. The reservation manager is optional, so telemetry-free/no-GPU unit tests build a valid (empty-handle) context.
- batch_telemetry_info POD (telemetry context + producing pipeline_uuid); quent_data_batch_probe::create(...) returns a real probe when a context is present, else a no-op base probe.
- Every production batch-creation site now attaches a probe: the make_data_batch/make_data_batch_from_view factories, direct data_batch::make calls (top-n, limit, table-scan, ungrouped-aggregate, scan path, scan-manager cache provider), and the clone/convert sites (result-collector, partition, grouped-aggregate-merge). In-place convert_to reuses the existing probe.
- Attribution is sourced from the operator's pipeline: both the telemetry context and pipeline UUID ride on pipeline_build_context (constructor-injected once at plan-convert time) and are read via the operator's _pipeline — no per-operator telemetry state, no per-call plumbing.

Analyzer (Rust)

- Ingests the memory and channel resource streams (mirroring the existing task-queue/executor-thread handling).
- Reconstructs DataBatch FSMs and exposes them alongside tasks: registered in fsm_types, fed into single/bulk/bulk-chunked resource timelines and long_fsms, filtered per-query by producing pipeline.
- Per-state breakdowns are type-gated (a memory timeline broken down "by task" yields empty, and vice-versa), so task and data-batch state sets never mix.
- Removed vestigial FsmCollection/Using impls on the model (unused by the timeline path).

Tests

Behavior unchanged. Call sites that build pipelines/batches without an engine pass a null telemetry context (→ no-op probe), so existing C++ unit tests are unaffected apart from the now-explicit argument.

Verification

- C++ extension and Rust analyzer + server both build clean.
- A GPU query run confirmed correct data_batch events with populated, per-pipeline producer_pipeline_uuid and resolved memory handles (no dropped events).
- Pending: end-to-end UI runtime check against a live capture (confirm memory/channel timelines + FSM rendering) — not yet run.

🤖 PR description generated with @claude


experimental/starrocks/starrocks
mbrobbel, line 1: I guess this is unintended?

dhruv9vats: yes! don't know how this got here. maybe a merge got messed up. thanks.

cucascade
mbrobbel, line 1: Should bump to `main` after https://github.com/NVIDIA/cuCascade/pull/156 merge

src/include/telemetry/data_batch_probe.hpp
```diff
@@ -0,0 +1,206 @@
...
+    //   memory_context_->get_memory_handle(new_data.get_memory_space().get_id());
+    // if (not maybe_memory_handle) {
+    //   SIRIUS_LOG_WARN(fmt::format("No quent memory handle found for {}",
+    //                               new_data.get_memory_space().to_string()));
+    //   return;
+    // }
+    // handle_->stationary({
+    //   .memory_resource_id    = (*maybe_memory_handle).get().uuid(),
+    //   .memory_capacity_bytes = new_data.get_size_in_bytes(),
+    // });
```
mbrobbel: Is this intended?

src/op/sirius_physical_result_collector.cpp
```diff
@@ -161,7 +162,11 @@ void sirius_physical_materialized_collector::sink(const operator_data& input_dat
 
       // clone_to: creates new batch with data converted to host_data_representation
       auto result_batch = ro.clone_to<cucascade::host_data_representation>(
```
mbrobbel, line 164: The probe is attached after `registry.convert()` finishes, so the GPU→host transfer never emits `InTransit` or channel usage. This omits result-collection transfers from the advertised telemetry.

src/telemetry/telemetry_context.cpp
```diff
@@ -34,20 +34,23 @@
...
   : engine_uuid_(uuid::now_v7()),
     worker_uuid_(uuid::now_v7()),
     query_group_uuid_(uuid::now_v7()),
     context_(
       quent::create_context(config.enable_quent ? "ndjson" : "noop", config.output_directory)),
     engine_observer_(quent::engine::create_observer(*context_)),
     worker_observer_(quent::worker::create_observer(*context_)),
-    query_group_observer_(quent::query_group::create_observer(*context_))
+    query_group_observer_(quent::query_group::create_observer(*context_)),
+    memory_context_(std::make_shared<memory_context>(engine_uuid_, *context_, manager))
```
mbrobbel: This happens before engine and worker init? I guess we should just move this into the constructor body, and explicitly handle the order in the destructor.

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -271,25 +282,44 @@ impl UiAnalyzer for SiriusUiAnalyzer {
...
+                                &view,
+                                req.entity_filter,
+                                &fsm_filter,
+                                config.span,
+                            )?
+                            .into_iter()
+                            .filter(|db| db.usages().any(|u| u.resource_id() == req.resource_id)),
+                            |id| id == req.resource_id,
+                        )?;
+                    } else {
```
johanpel: Perhaps this should become a pattern `match` instead and error out on unknown entity type names to prevent silent breakages.

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -337,16 +383,39 @@ impl UiAnalyzer for SiriusUiAnalyzer {
...
+                                config.span,
+                            )?
+                            .into_iter()
+                            .filter(|db| {
+                                db.usages()
+                                    .any(|usage| resource_ids.contains(&usage.resource_id()))
+                            }),
+                            |id| resource_ids.contains(&id),
+                        )?;
+                    } else {
```
johanpel: Same here

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -497,7 +588,35 @@ impl UiAnalyzer for SiriusUiAnalyzer {
...
+            }
+
+            for (state_name, usage) in data_batch.usages_with_state_names() {
+                let resource_id = usage.resource_id();
+                if let Some(builder_indices) = per_state_index.get(&resource_id) {
+                    for &builder_idx in builder_indices {
+                        let builder = &mut per_state_builders[builder_idx];
+                        if builder.4 == DATA_BATCH_TYPE_NAME
+                            && data_batch.matches_filter(&builder.3)
+                        {
```
johanpel, line 642: Since this code will get quite repetitive the more FSM types we add, could we perhaps add some kind of umbrella FSM type that we can iterate over that just forwards trait implementations of e.g. usage_with_state_names() to its variants?

Might be possible in multiple places below as well.

Feel free to do in follow-up.

johanpel (commented):
Very nice, a few small comments about de-duplicating code.
Figuring those out for Sirius may provide some nice insights for analysis-related codegen in Quent later.

src/op/sirius_physical_result_collector.cpp
dhruv9vats: There does not seem to be any straight forward way to achieve this, so lets take the "omits result-collection transfers from the advertised telemetry" compromise, _for now_. As this design is refined.

src/include/telemetry/data_batch_probe.hpp
dhruv9vats: I was contemplating removing the `set_data` this corresponds to, but to reduce scope creep, lets just keep that API and this `stationary` -> `stationary` transition.

rust/crates/telemetry/analyzer/src/lib.rs
dhruv9vats: It should be indeed.

dhruv9vats: [use match for entity type matching, streamline](https://github.com/sirius-db/sirius/pull/1068/commits/81611db290300984870cadf2b22b692c9e8bf4a9)

rust/crates/telemetry/analyzer/src/lib.rs
dhruv9vats: [use match for entity type matching, streamline](https://github.com/sirius-db/sirius/pull/1068/commits/81611db290300984870cadf2b22b692c9e8bf4a9)

rust/crates/telemetry/analyzer/src/lib.rs
dhruv9vats: Yes, I started de-duplicating the code, but deferred that for a smaller follow-up.

johanpel (approved):
Approving Rust changes, thanks @dhruv9vats !

substrait
mbrobbel, line 1: this seems unintended?

dhruv9vats: Dont know why these submodule changes keep happening after merges. Thanks.

mbrobbel (approved):

PR #1071, opened 2026-07-03
Update clangd tooling for cuda 13 lsp parsing.

kevkrist:
cuda 13 requires clangd 22+ for correct lsp parsing.


mbrobbel (commented):
We should also bump in pre-commit and reformat with clang format 22

kevkrist:
@mbrobbel It looks like nvcc only supports clang < 22, so cuda developers will need to point their clangd to a local clang 22 for correct parsing. I don't understand the circularity here. Closing for now.

PR #1080, opened 2026-07-06
telemetry: render threads as a collapsible GPU → type tree in Quent

felipeblazing:
## What

Quent currently shows every execution thread as a flat sibling list under the engine, because all thread telemetry declares the engine as its `parent_group_id`. This PR emits the nesting as data instead — no Quent-side changes needed:

```
engine (siriusDB)
├─ gpu-0
│  ├─ executor_thread            → gpu_pipeline-gpu0-exec-0..3
│  ├─ task_manager_loop_thread   → gpu-0-exec-manager
│  └─ gpu_pipeline-task-queue
├─ gpu-1                          (same shape)
└─ shared
   ├─ task-scheduler-thread
   └─ task-scheduler-gpu-queue
```

## How

- **Model** (`rust/crates/telemetry/model`): new `GpuDevice` and `ThreadGroup` resource-group entities (declaration = `instance_name` + `parent_group_id`, plus GPU `ordinal` on `GpuDevice`).
- **Runtime** (`telemetry_context`): `create()` takes the topology's GPU device ids and declares one `gpu-N` group per GPU under the engine, with per-thread-type bucket groups underneath, plus a `shared` group. Unknown device ids fall back to the engine group with a warning.
- **Spawn sites**: the telemetry wrappers take an explicit `parent_group_id`. `gpu_pipeline_executor` parents its executor threads, manager thread, and task queue under its GPU's groups, sourcing the device ordinal from `_memory_space->get_device_id()` (not from name strings). `task_scheduler`'s thread and queue go under `shared`. Executor thread names now include the GPU ordinal (`gpu_pipeline-gpu0-exec-N`), so the two GPUs' threads no longer share identical names.
- **Analyzer**: ingests the new declarations as resource groups; the existing tree machinery does the rest. Also adds a `print_resource_tree` example for inspecting a session's group tree offline.

## Verification

- New `[telemetry_context]` unit test asserts the declared groups, thread parenting, and engine fallback against the emitted ndjson (29 assertions).
- Real 2-GPU TPC-H q1 (SF100) run: `gpu-0`/`gpu-1` declared under the engine, all 8 executor threads point at per-GPU bucket uuids (4+4, none at the engine), and the analyzer builds the full tree with no orphans. Confirmed the collapsible tree renders in the Quent UI.

### DIFF
--- a/src/sirius_context.cpp
+++ b/src/sirius_context.cpp
@@ -58,6 +58,7 @@
 #include <string>
 #include <string_view>
 #include <utility>
+#include <variant>
 
 namespace duckdb {
 
@@ -281,10 +282,15 @@ void SiriusContext::initialize(const sirius::sirius_config& config)
 
   config_ = config;
   // Declare one telemetry device group per GPU so thread/queue telemetry can
-  // nest under its device instead of piling up flat under the engine.
+  // nest under its device instead of piling up flat under the engine. The GPU
+  // memory space configs already reflect the configured topology.num_gpus /
+  // topology.gpu_ids selection, unlike the raw hardware topology.
   std::vector<int> telemetry_gpu_ids;
-  for (auto const& gpu : config_.get_hw_topology().gpus) {
-    telemetry_gpu_ids.push_back(gpu.id);
+  for (auto const& space_config : config_.get_memory_space_configs()) {
+    if (auto const* gpu_config =
+          std::get_if<cucascade::memory::gpu_memory_space_config>(&space_config)) {
+      telemetry_gpu_ids.push_back(gpu_config->device_id);
+    }
   }
   telemetry_context_ =
     sirius::telemetry::telemetry_context::create(config_.get_telemetry_config(), telemetry_gpu_ids);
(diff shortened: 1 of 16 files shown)


src/sirius_context.cpp
```diff
@@ -279,8 +279,15 @@ void SiriusContext::initialize(const sirius::sirius_config& config)
 {
   if (is_initialized_) { throw std::runtime_error("Sirius context is already initialized."); }
 
-  config_            = config;
-  telemetry_context_ = sirius::telemetry::telemetry_context::create(config_.get_telemetry_config());
+  config_ = config;
+  // Declare one telemetry device group per GPU so thread/queue telemetry can
+  // nest under its device instead of piling up flat under the engine.
+  std::vector<int> telemetry_gpu_ids;
+  for (auto const& gpu : config_.get_hw_topology().gpus) {
```
mbrobbel: This should respect `num_gpus` and `gpu_ids`

mbrobbel (approved):

src/sirius_context.cpp
felipeblazing: Done — the telemetry GPU list is now derived from the GPU memory space configs (which honor `topology.num_gpus` / `topology.gpu_ids`) instead of the raw hardware topology.

PR #1084, opened 2026-07-06
taking timer options as chrono string_literals to avoid mistakes like…

aminaramoon:


src/include/yaml_reader.hpp
```diff
@@ -124,6 +124,66 @@ inline std::uint64_t parse_bytes(std::string_view sv)
...
+///   ns / nsec           -> nanoseconds
+///   us / usec           -> microseconds
+///   ms / msec           -> milliseconds
+///   s  / sec / seconds  -> seconds
+///   m  / min / minutes  -> minutes
+///   h  / hr  / hours    -> hours
+///
+/// A unit suffix is required: bare numbers are rejected because the intended
+/// unit would be ambiguous. Fractional values are allowed (e.g. "1.5s").
+inline std::chrono::nanoseconds parse_duration(std::string_view sv)
```
mbrobbel, line 141: Maybe we can use `absl::ParseDuration` from `absl/time/time.h`?

mbrobbel (commented):
Nice improvement. One question.

src/include/yaml_reader.hpp
aminaramoon: I think chronos string literal conversions are more aligned with our needs in sirius.

mbrobbel (approved):
We could consider a temp deprecation warning for `monitor_period_ms` to avoid a hard breaking config change, to give people some time to update their configs?

PR #1087, opened 2026-07-07
perf improvement - Collapse S3 parquet footer bind to one suffix-range GET

ran-yuan-rui:
## Collapse S3 parquet footer bind to one suffix-range GET

Closes #<1086>.

### What changed
- **One suffix-range GET replaces the HEAD.** `open_hint::parquet_footer_probe` makes `rest_ioctx` issue a single `Range: bytes=-N` GET that returns the size (from `Content-Range`) and stashes the object's trailing N bytes on the `rest_io_object`; `rest_reactor::host_read` then serves cuDF's trailer/footer reads from that stash — no extra GET. The stash is per-open, not a cache.
- **Metadata-aware `describe_parquet`.** Cold bind (metadata_store miss) probes; a warm re-bind opens `generic` (one HEAD) and reuses the parsed footer — dev-parity, no footer re-download.
- **Configurable window.** `scan_manager.rest.footer_probe_bytes` (yaml, byte-valued), default **512 KiB** (footer ≈ 0.037 % of file: SF1 207 MB → 78 KiB, SF10 2.2 GB → 771 KiB; N must cover the footer or it falls back to a body re-GET, so err large — over-read is a cheap one-time bind cost).
- **Fallback.** An unusable suffix response (200 full body / 416 / missing-or-`*` `Content-Range`) aborts the body mid-stream and falls back to a plain HEAD.
- **Bench A/B.** `make s3-bench` runs generic vs probe on MinIO + AWS and reports `bind_ms = open_ms + footer_fetch_ms` + `bind_chunk_get_count`.

### Results (footer-probe A/B)
- **Requests 3 → 1**, deterministic: `bind_chunk_get_count 2 → 1` on 24/24 AWS + MinIO samples.
- **MinIO HTTPS (clean link): footer bind 11 ms → 4 ms (2.75×)**, `footer_fetch` 9 ms → 0.
- AWS (via socks5 proxy) confirms 2 → 1; absolute latency is proxy-masked (the 434 KiB over-read ≈ one throttled-proxy RTT) — a clean magnitude needs in-region EC2.

### Test plan
All green on the CI GPU host: `make s3-test` 40/40, `[…rest…footerbind]` 10/10, `make s3-bench` 2/2,
`make s3-test-large` (SF10), full Catch2 1548/1548. Coverage: cold bind = 1 suffix GET / 0 HEAD; size
from `Content-Range`; trailer/footer from stash; footer beyond window → one body re-GET; unusable
suffix → HEAD; transient-5xx retry + exhaustion; small object; concurrent binds; warm re-bind; config.

### DIFF
--- a/src/io/rest/rest_reactor.cpp
+++ b/src/io/rest/rest_reactor.cpp
@@ -140,6 +140,7 @@ size_t capture_header(char* buffer, size_t size, size_t nitems, void* userdata)
 struct suffix_sink {
   std::vector<std::uint8_t> data;
   std::size_t cap{0};
+  std::size_t total_received{0};  // wire bytes, incl. those dropped by cap/abort
   long status{0};
   std::string content_range;
   std::string retry_after;
@@ -176,6 +177,7 @@ size_t suffix_write_cb(char* ptr, size_t size, size_t nmemb, void* userdata)
 {
   auto* s            = static_cast<suffix_sink*>(userdata);
   size_t const bytes = size * nmemb;
+  s->total_received += bytes;
   if (s->status != 206) { return 0; }
   if (s->data.size() < s->cap) {
     size_t const take = std::min(s->cap - s->data.size(), bytes);
@@ -886,6 +888,11 @@ footer_probe rest_reactor::fetch_footer_suffix(std::string_view bucket,
     long status       = 0;
     curl_easy_getinfo(h.get(), CURLINFO_RESPONSE_CODE, &status);
 
+    // payload_bytes_read_total is always-on and per-attempt (see the async
+    // worker's finish()), so credit every attempt's wire bytes outside the
+    // perf_instrumentation gate.
+    _perf.payload_bytes_read_total.fetch_add(sink.total_received, std::memory_order_relaxed);
+
     // suffix_write_cb aborts any non-206 body, so a CURLE_WRITE_ERROR here is our
     // own doing and the HTTP status is still valid; only a different curl error
     // (no HTTP status) is a genuine transport failure.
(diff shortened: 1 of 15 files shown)


aminaramoon (approved):
LGTM!

src/io/rest/rest_reactor.cpp
```diff
@@ -745,17 +843,131 @@ size_t rest_reactor::head_object_size(std::string_view bucket, std::string_view
...
 }
 
+footer_probe rest_reactor::fetch_footer_suffix(std::string_view bucket,
+                                               std::string_view key,
+                                               std::size_t n)
+{
+  footer_probe probe;
+  if (n == 0) { return probe; }
+
+  s3::s3_object_ref const obj{std::string(bucket), std::string(key)};
```
aminaramoon, line 873: why did we not use scatter gater reactor apis to get header/footer ?

ran-yuan-rui: Good question. I did not use the scatter-gather path here because this is still the open/bootstrap step, before we have a sized rest_io_object.

The existing scatter-gather worker path takes closed byte ranges on an already-created object. But the footer probe is an open-ended suffix GET (`Range: bytes=-N`) whose job is to discover the object size from `Content-Range` and stash the footer window. HEAD is similar: it has no body to scatter, it only produces size metadata.

Also cuDF drives footer bind through datasource::host_read, one blocking range at a time. It does not give us a vector of footer ranges to gather there.

After the object is sized, normal host reads do go through the worker path: stash hits are served locally, and stash misses enqueue `prep_host_rx_request`.

So I think using the existing scatter-gather API here is not directly expressible. 
The real follow-up is to teach the worker a metadata/open request type for HEAD + suffix footer probe, so it can reuse the worker connection pool too. I kept this PR scoped to the 3→1 footer-bind RTT collapse.

I plan to do more perf testing when we have multi file support in S3, and use data to avoid wrong design

PR #1088, opened 2026-07-07
Fix hash-join partition sizing

mbrobbel:
## Summary

Fix two independent ways Sirius can underestimate the number of hash partitions:

1. Byte-based partition counts use floor division, so an input just over a partition boundary remains at the lower partition count.
2. Join siblings are always sized from the build input, even when a `RIGHT`, `RIGHT_SEMI`, or `RIGHT_ANTI` join retains and concatenates the probe input.

This change uses ceiling division for byte-based partition counts and lets the retained probe input drive the shared count for right-family joins.

## Why this is needed

Both sides of a hash join must use the same partition count. Previously, the count was effectively based on:

```text
max(1, floor(build_bytes / partition_target))
```

### Q13: the wrong input drives sizing

DuckDB plans Q13's SQL `LEFT OUTER JOIN` as a physical `RIGHT` hash join.

At SF1000:

- Customer build input: approximately 0.62 GB
- Filtered orders probe input: 18,176,930,656 bytes
- Configured partition target: 5 GiB

The old build-driven calculation selects one partition from the small customer input, even though Sirius retains the much larger orders probe input.

The new calculation uses the complete probe input:

```text
ceil(18,176,930,656 / 5,368,709,120) = 4 partitions
```

This gives Q13 four balanced join tasks instead of one large retained probe task. The expected benefit is improved memory and task shape, not lower Q13 latency.

### Q21: floor division independently under-partitions

Q21 contains a build-driven stage with approximately 5.94 GB of input:

```text
old: floor(5,937,804,664 / 5,368,709,120) = 1 partition
new:  ceil(5,937,804,664 / 5,368,709,120) = 2 partitions
```

This decision remains build-driven. Its change from one partition to two comes solely from ceiling division, not the right-family logic.

The sampled logs show the following stage starting roughly 1.2 seconds earlier, consistent with the additional parallelism and observed Q21 runtime improvement.

## Implementation

- Round byte-based partition counts up.
- Size `RIGHT`, `RIGHT_SEMI`, and `RIGHT_ANTI` sibling partitions from the complete probe input.
- Assign the resulting shared count to both join sides.
- Use a `FULL` probe barrier only when the probe drives sizing. Other probe pipelines remain `PARTIAL` and streaming.
- Keep `RIGHT_DELIM_JOIN` build-driven because its probe subtree depends on build-side distinct data.
- Only use build-side byte counts when selecting the `BUILD_PROBE` execution mode.
- Preserve a one-partition fallback when the designated sizing input is empty.
- Centralize right-family classification across concat behavior, execution-mode selection, and pipeline wiring.
- Update the partitioning and task-creation documentation.

`hash_partition_bytes` remains a sizing target rather than a strict per-partition upper bound because hash skew and operator output expansion can still affect individual partition sizes.

## SF1000 evidence

All runs used the same GB300 machine with per-query host pinning after iteration 2.

| Run | Q13 cold / warm-best | Q21 cold / warm-best | Q1-Q22 total cold / warm-best |
|---|---:|---:|---:|
| `dev`, 3 iterations | 3.21 s / 1.35 s | 9.69 s / 4.99 s | 75.10 s / 31.04 s |
| Patch, 3 iterations | 3.18 s / 1.42 s | 8.44 s / 3.80 s | 73.90 s / 29.59 s |
| Patch, 6 iterations | 3.24 s / 1.40 s | 8.37 s / 3.79 s | 73.70 s / 27.84 s |

The three-iteration runs provide the closest direct comparison. The six-iteration run has more warm samples, so its `warm-best` values should not be treated as an identical A/B comparison.

The deterministic evidence is the partition selection:

- Q13: build-driven 1 partition -> probe-driven 4 partitions
- Q21: floor-divided 1 partition -> ceiling-divided 2 partitions

Q13 latency remains effectively unchanged, as expected. Q21 warm-best improves by approximately 24% in the matched three-iteration runs, but the partition-count correction—not a guaranteed timing improvement—is the basis for the patch.

## Validation

- Added classification coverage for right-family and unaffected join types.
- Added a regression test that enters through the non-driving build side and only produces two sibling partitions when both probe-driven sizing and ceiling division work.
- Full pre-commit suite passes.
- Full SF1000 TPC-H Q1-Q22 host-pinned benchmarks completed successfully.

## Review notes

The main behavioral tradeoffs are:

- Ceiling division may add one partition near each configured boundary, increasing task overhead for those inputs.
- Right-family probe sizing waits for the complete probe input before partitioning.
- `OUTER` and all other join families retain their existing build-driven behavior.
- `RIGHT_DELIM_JOIN` is deliberately excluded to avoid introducing a dependency cycle.

### DIFF
--- a/docs/super-sirius/operators.md
+++ b/docs/super-sirius/operators.md
@@ -217,8 +217,8 @@ Repartitions data into N buckets based on partition keys.
 
 - **Modes:** `HASH` (most common), `RANGE`, `EVENLY`, `CUSTOM`, `NONE`
 - **Adaptive count:** `determine_num_partitions()` computes N from actual input data size and `hash_partition_bytes` config
-- **Sibling coordination:** Build-side partition determines count; probe-side waits for the result
-- **Key members:** `_partition_keys`, `_partition_type`, `_num_partitions`, `_is_build`, `_sibling_partition_op`
+- **Sibling coordination:** Build-side partition normally determines the shared count. For RIGHT-family hash joins other than `RIGHT_DELIM_JOIN`, the retained probe side determines it instead.
+- **Key members:** `_partition_keys`, `_partition_type`, `_num_partitions`, `_is_build`, `_drives_partition_count`, `_sibling_partition_op`
 
 ### `sirius_physical_concat` — `CONCAT`
 **File:** `src/include/op/sirius_physical_concat.hpp`
--- a/docs/super-sirius/optimizations.md
+++ b/docs/super-sirius/optimizations.md
@@ -11,7 +11,7 @@ This document catalogs Super Sirius performance optimizations by category. Each 
 **Mechanism:** `determine_num_partitions()` computes partition count from actual input data size:
 ```
 total_bytes = sum of all batch sizes from input repository
-num_partitions = max(1, total_bytes / hash_partition_bytes)
+num_partitions = max(1, ceil(total_bytes / hash_partition_bytes))
 ```
 
 **Code path:** `src/op/sirius_physical_partition.cpp` — `determine_num_partitions()`
--- a/docs/super-sirius/task-creator.md
+++ b/docs/super-sirius/task-creator.md
@@ -143,9 +143,9 @@ Uses base class for both `get_next_task_hint()` and `get_next_task_input_data()`
 
 | Method | Behavior |
 |--------|----------|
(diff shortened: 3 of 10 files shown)


bwyogatama (approved):

bwyogatama:
nice improvement, although this mean that we will have a pipeline breaker before the probe which can be memory consumng

wmalpica (approved):

PR #1089, opened 2026-07-07
fix(scan): account for Parquet filter decode memory

mbrobbel:
## Summary

- track projected-column bytes separately from all decoded Parquet column buffers
- cap coalesced scan batches by the decoded-buffer footprint, including filter-only columns
- use that footprint as a lower bound for fresh scan reservations while preserving the projected-column history basis

## Why

Parquet scans can decode columns that are needed only to evaluate a filter. Those columns are absent from the logical output, so the existing size estimate did not include their GPU buffers.

TPC-H Q13 at SF1000 exposes this with the filter on `o_comment`: a representative scan task reserved about 42.9 GB from the projected-column estimate but reached about 63.9 GB after decoding the filter column. The same underestimate also let the coalescer build batches larger than its intended decoded-memory budget.

## Approach

`estimated_bytes()` remains the normal projected-column basis used by memory history. A separate decoded-buffer estimate includes pure-filter columns. Parquet coalescing uses the latter, and a fresh scan uses:

```
8 * projected_bytes + (decoded_buffer_bytes - projected_bytes)
```

This keeps the existing 8x heuristic for projected data and adds filter-only buffers once instead of applying the expansion factor to them again. History-based estimates are clamped to at least the known decoded-buffer footprint. Resident/pinned inputs are unchanged.

Scans with no projected data column use their decoded read columns as a nonzero history basis; regression coverage includes count-style and Hive partition-only output shapes.

## Validation

- `pixi run make test`: 32,070,873 assertions in 1,544 test cases
- focused coverage for Parquet coalescing, estimate propagation, fresh-scan sizing, and history lower bounds


kevkrist (approved):
Only nit is that we've been trying to enforce a distinction, not very well, between 'split' and 'batch', to reduce the overload on 'batch'. A 'split' is what the split provider emits, and splits (e.g., sets of row groups) are coalesced into 'batches', which are emitted by the scan operator.

PR #1091, opened 2026-07-07
fix(sort): size complete samples from actual input

mbrobbel:
## Summary

- use observed bytes when SORT_SAMPLE contains the complete upstream input
- retain cardinality extrapolation for partial samples
- cap the partition count by sampled rows to avoid duplicate boundary indices

## Why

SORT_SAMPLE currently extrapolates every sample through the logical cardinality estimate, even when upstream is finished and the sample already contains the entire input.

TPC-H Q13 shows the mismatch clearly: the final sort receives 30 rows (480 bytes), while the inherited cardinality estimate is about 1.5 billion rows. That turns a complete 480-byte input into a roughly 24 GB size estimate. A GB300 still selects one partition because its automatic partition budget is larger, but smaller GPUs or an explicit `max_sort_partition_bytes` can create many unnecessary partitions and repeated boundary rows.

## Approach

The boundary task treats its sample as complete only when:

1. the upstream pipeline is finished; and
2. the input repository is empty after the sample was claimed.

Completion is checked first. Upstream output is published before pipeline completion becomes visible, and completion is terminal, so the following locked repository check is stable. A false result is conservative and keeps the existing extrapolation path.

Complete samples use `total_sample_bytes` directly. Partial samples keep the previous average-batch/cardinality calculation. In either case, the partition count cannot exceed the number of sampled rows, because only that many distinct boundary positions are available.

## Validation

- `pixi run make test`: 32,070,846 assertions in 1,540 test cases
- focused SORT_SAMPLE suite: 55 assertions in 8 test cases
- regression cases cover finished+drained, unfinished+drained, and finished+data-remaining states


9prady9 (approved):
nice catch @mbrobbel looks good.

Did you by any chance run any benchmarks to see if this improved any numbers ?  Less likely perhaps but I wonder.

mbrobbel:
> Did you by any chance run any benchmarks to see if this improved any numbers ? Less likely perhaps but I wonder.

Only on gb300 and:

> > A GB300 still selects one partition because its automatic partition budget is larger, but smaller GPUs or an explicit `max_sort_partition_bytes` can create many unnecessary partitions and repeated boundary rows.

PR #1093, opened 2026-07-07
fix(scan): reserve cached HOST materialization

mbrobbel:
## Summary

- Account for cached scan inputs when estimating GPU materialization memory.
- Add the uncompressed upload size for HOST-cached batches while leaving GPU-cached batches unchanged.
- Keep transfer memory separate from the operator peak learned by pipeline history.

## Why

Pinned scans use `scan_operator_input`, which stores the cached batch directly rather than through `pipelineable_operator_data`. The task reservation estimator only inspected pipelineable inputs, so a HOST-cached scan reported zero materialization bytes even though `prepare_for_processing()` uploads the batch to GPU before execution.

This leaves the first tasks under-reserved. In the SF1000 Q13 validation that exposed the issue, a cached scan reserved about 5.23 GiB but reached about 6.54 GiB, a gap of roughly 1.31 GiB. The query completed because reservations are soft by default, but the excess allocation consumes unreserved GPU capacity and weakens admission control under concurrency.

## Approach

`gpu_pipeline_task_local_state::get_estimated_bytes_to_materialize_input()` now recognizes a cached `scan_operator_input`:

- Non-GPU cached batch: add its uncompressed GPU footprint as materialization memory.
- GPU-cached batch: add zero.
- Fresh scan metadata: unchanged; decode memory remains covered by the scan operator estimate.

The existing resident classification is intentionally unchanged. For scans, resident distinguishes cached data from a fresh read; treating a HOST cache hit as fresh would incorrectly apply the 8x decode heuristic. The resulting reservation remains:

```text
operator peak estimate (or history) + input materialization bytes
```

## Tests

- Added regression coverage for HOST-cached input with and without pipeline history.
- Added coverage confirming GPU-cached input does not receive an upload charge.
- `pixi run make test` — 1,539 test cases, 32,070,831 assertions passed.
- Focused regression test — 20 assertions passed.
- `pixi run pre-commit run -a` passed.


aocsa (approved):
LGTM.

While reviewing, I found a related **history-recording issue** that this PR doesn't introduce but does expose: once cached HOST scans carry a nonzero `M`, the blind `peak − M` subtraction in the history-recording paths underreports the operator peak whenever `track_per_stream_reservation: true`. This isn't blocking for the PR since the flag defaults to `false`, I suggest a follow-up PR before we rely on the new `M` with per-stream tracking.

aocsa:
> LGTM.
> 
> While reviewing, I found a related **history-recording issue** that this PR doesn't introduce but does expose: once cached HOST scans carry a nonzero `M`, the blind `peak − M` subtraction in the history-recording paths underreports the operator peak whenever `track_per_stream_reservation: true`. This isn't blocking for the PR since the flag defaults to `false`, I suggest a follow-up PR before we rely on the new `M` with per-stream tracking.

To reproduce:

The PR's new tests only size reservations (they never `execute()`) and the fixture pins the flag to `false`, so they can't see this. A minimal repro on top of the existing fixture in `test/cpp/pipeline/test_gpu_pipeline_task_history.cpp`:

1. Build the fixture with `.track_reservation_per_stream(true)`.
2. Create a HOST-tier input batch of ~20 MiB (`M`) and a stub operator whose `on_execute` allocates ~5 MiB (`P < M`).
3. `execute()` the task, then check `get_memory_history().estimate_peak_memory(input_basis)` and a second task's `get_estimated_reservation_size_info()`.

- Expected: history ≈ `P` (5 MiB), next reservation ≈ `P + M` (25 MiB). 
- Actual: history = 0, next reservation ≈ `M` (20 MiB). 

The same steps with the flag `false` pass, which isolates the tracker as the variable.

Issue #839, opened 2026-05-29
Stream session support


mbrobbel:
Add a session that builds a streaming plan, starts it on the existing task scheduler without blocking, and exposes `push` + `close_input` (#836), `pull` + `wait` (#837). And handles errors and routing by stream id (source operators).

This would benefit from an operator builder API, but initially we can reuse substrait support.

aocsa:
@mbrobbel can you review this design : 

# `stream_session` — API design for review (#839 draft)

This proposal outlines the C++ public entry point for streaming computation, designed to interface with the Rust wrapper over the cxx-FFI boundary. It acts as a plan-builder and channel router leveraging the existing task_scheduler rather than introducing a new thread pool.

## Proposed surface (cxx-FFI boundary)

```cpp
class stream_session {
 public:
  // Builds the streaming plan, wires channels+repos, submits to `sched`, returns
  // immediately (non-blocking). `mem` is borrowed, not owned (#840, option A).
  stream_session(task_scheduler& sched,
                 memory_reservation_manager& mem,
                 streaming_plan plan);              // from translate_fragment (#841)
  ~stream_session();                                // teardown contract → Q3

  stream_session(stream_session&&) = default;       // move-only; not copyable
  stream_session& operator=(stream_session&&) = default;

  // --- producer side: wrapper (nixl recv thread) → engine --------------------
  status push(stream_id_t id, data_batch_ptr batch);   // blocks iff input channel full
  status close_input(stream_id_t id);                  // EOS marker; close-then-drain

  // --- consumer side: engine → wrapper (→ FE / next hop) ---------------------
  result<std::optional<data_batch_ptr>> pull(stream_id_t id);  // non-blocking; nullopt = none now
  status wait(stream_id_t id);                                 // block until pullable or drained
  [[nodiscard]] bool drained(stream_id_t id) const;            // closed && empty

 private:
  std::map<stream_id_t, std::shared_ptr<exchange_channel>>        _input_channels;
  std::map<stream_id_t, std::shared_ptr<shared_data_repository>>  _input_repos;
  std::map<stream_id_t, std::shared_ptr<exchange_channel>>        _output_channels;
  std::map<stream_id_t, std::shared_ptr<shared_data_repository>>  _output_repos;
  std::vector<std::unique_ptr<sirius_physical_operator>>          _operators;
};
```

## Invariants the implementation must hold (from the design)

1. **Register-then-Push Sequence:**
[...]

PR #1094, opened 2026-07-07
feat(streaming): streaming source operator + exchange channel

aocsa:
PR 1 of 2 for #836 (sink: #837). Adds the exchange building blocks for streaming
execution. Deliberately unwired — nothing constructs the operator outside tests;
wiring lands with the stream session (#839).

## What

- `sirius::exec::exchange_channel` — bounded MPMC queue of `{batch_id, size_bytes}`
  handles (item + byte capacity, close-then-drain EOS, push/pop/close hooks fired
  outside the lock).
- `sirius::op::sirius_physical_streaming_source` — multi-shot source: pops one handle
  per task, resolves the batch via `shared_data_repository`, passes it through
  unchanged.

## Design notes

- The channel carries handles, not `shared_ptr`s: the repository stays owner of record,
  so queued batches remain visible to the downgrade/spill sweep and self-heal on resolve.
- Source hint contract: `READY` when non-empty, `WAITING` when open+empty (re-armed by
  the session in #839), done when closed and drained.
- Channel close notifies the pipeline via a weak reference and re-arms downstream
  consumers, so empty or late-closed streams still finish.
- Known gap, deferred to #839: the channel bound does not bound the task-scheduler
  backlog; the session must gate task creation on in-flight work (documented in
  `operators.md`).

## Testing

- `[exchange_channel]` (CPU): capacity/byte admission, overflow guard, FIFO, close
  semantics, MPMC stress, callback contracts.
- `[streaming_source]` (GPU): hint/lifecycle, pass-through identity, spill self-heal,
  input validation, callback lifetime after operator destruction, and pipeline
  completion for empty/late-closed streams including downstream re-arming.

### DIFF
--- a/docs/super-sirius/operators.md
+++ b/docs/super-sirius/operators.md
@@ -71,7 +71,7 @@ each handle via a `cucascade::shared_data_repository`, and publishes the batch i
 as a `pipelineable_operator_data`. Used only when a fragment's input arrives from another node
 over exchange; a leaf fragment keeps its normal `GPU_SCAN` source.
 
-Key design invariants (design §3/§7):
+Key design invariants:
 - The channel carries **handles**, not `shared_ptr`s — the repository owns the batch so queued
   items remain spill-visible to the downgrade executor.
 - Engine workers use `try_pop` only (non-blocking); `push`/`pop` are provided for the wrapper/test side.
@@ -91,9 +91,20 @@ Hint table:
 `all_ports_empty()` is overridden to `_input_channel->drained()`, driving both the task-creation
 loop guard and the port-less source pipeline-finish predicate.
 
+Channel close notifies the pipeline (`update_pipeline_status(false)`, via a weak pipeline
+reference wired in `set_pipeline`), so an empty or late-closed stream still finishes its
+pipeline — and re-arms downstream consumers — even when no task is left in flight.
+
 **Producer contract**: register the incoming batch in the input repository (`add_data_batch`) *first*,
 then push the handle. The session (#839) owns edge-triggered re-scheduling; the plan generator (#838)
 owns channel wiring.
+
+**Backpressure (open integration requirement for #839)**: `try_pop()` frees channel item/byte
+capacity at task-creation time, but the popped batches move into the unbounded task-scheduler
+queue — the channel bound therefore does not bound total outstanding data. When #839 wires the
+session, task creation must be gated on in-flight work (e.g. counting via the channel's `on_pop`
+hook and the task-completion path) so a fast producer cannot accumulate an arbitrarily large GPU
+backlog behind a nominally bounded channel.
 
 ### `sirius_physical_dummy_scan` — `DUMMY_SCAN`
 **File:** `src/include/op/sirius_physical_dummy_scan.hpp`
--- a/docs/super-sirius/operators.md
+++ b/docs/super-sirius/operators.md
@@ -62,6 +62,39 @@ The pipeline converter rewrites a DuckDB parquet or DuckDB-native table scan int
 
 See [Scan](scan.md) for the full scan subsystem (scan manager, `gpu_ingestible`, pinned-table caching, and the IO layer).
 
+### `sirius_physical_streaming_source` — `STREAMING_SOURCE`
+**File:** `src/include/op/sirius_physical_streaming_source.hpp`
+
+Source operator that marks the bottom boundary of an intermediate pipeline fragment. It pulls
+`exchange_batch_handle` records (batch-id + size) from a bounded `exec::exchange_channel`, resolves
+each handle via a `cucascade::shared_data_repository`, and publishes the batch into the pipeline
+as a `pipelineable_operator_data`. Used only when a fragment's input arrives from another node
+over exchange; a leaf fragment keeps its normal `GPU_SCAN` source.
+
+Key design invariants (design §3/§7):
+- The channel carries **handles**, not `shared_ptr`s — the repository owns the batch so queued
+  items remain spill-visible to the downgrade executor.
+- Engine workers use `try_pop` only (non-blocking); `push`/`pop` are provided for the wrapper/test side.
+- EOS is **close-then-drain**: `close()` forbids new pushes; queued handles stay poppable;
+  `drained()` (= `closed() && empty()`) is the terminal predicate.
+- `execute()` is a pure pass-through (COLUMN_DATA_SCAN shape — no GPU work).
+- `no_history_peak_memory_estimate()` returns `stats.bytes` (no extra allocation).
+
+Hint table:
+
+| Channel state | `get_next_task_hint()` |
+|---|---|
+| non-empty (open or closed) | `READY{this}` |
+| open, empty | `WAITING{nullptr}` — re-armable by the session on push (#839) |
+| closed && drained | `std::nullopt` — EOS |
+
+`all_ports_empty()` is overridden to `_input_channel->drained()`, driving both the task-creation
+loop guard and the port-less source pipeline-finish predicate.
+
+**Producer contract**: register the incoming batch in the input repository (`add_data_batch`) *first*,
+then push the handle. The session (#839) owns edge-triggered re-scheduling; the plan generator (#838)
+owns channel wiring.
+
 ### `sirius_physical_dummy_scan` — `DUMMY_SCAN`
 **File:** `src/include/op/sirius_physical_dummy_scan.hpp`
 
@@ -312,6 +345,7 @@ After pipeline finalization, `source` and `sink` are just aliases for the first 
 | DUCKDB_SCAN | Scan | DuckDB table function (CPU / DuckDB-source path) |
 | PARQUET_SCAN | Scan | Parquet reading (CPU / DuckDB-source path) |
 | GPU_SCAN | Scan | Unified GPU scan source served by `sirius_scan_manager` via a per-format `gpu_ingestible` |
+| STREAMING_SOURCE | Scan | Exchange-input source; pulls batch handles from `exchange_channel`, resolves via `shared_data_repository` |
 | DUMMY_SCAN | Scan | Generates 1 row |
 | COLUMN_DATA_SCAN | Scan | Reads ColumnDataCollection |
 | FILTER | Relational | `expression_evaluator::select()` |
--- a/src/include/exec/exchange_channel.hpp
+++ b/src/include/exec/exchange_channel.hpp
@@ -1,0 +1,232 @@
+/*
+ * Copyright 2025, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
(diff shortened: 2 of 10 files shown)


duckdb
mbrobbel, line 1: I think this submodule change was unintended?

substrait
mbrobbel, line 1: Same here

test/cpp/operator/test_physical_partition.cpp
mbrobbel, line 1: Those changes seem unrelated? Can you move them to a separate PR?

test/cpp/operator/test_physical_order.cpp
mbrobbel, line 1: Same here

test/cpp/operator/test_physical_merge_sort.cpp
mbrobbel, line 1: And here

test/cpp/expression_evaluator/test_expression_evaluator_ast_equivalence.cpp
mbrobbel, line 1: +1

test/cpp/memory/test_host_table_utils.cpp
mbrobbel, line 1: +1

test/cpp/expression_evaluator/test_expression_evaluator.cpp
mbrobbel, line 1: +1

test/cpp/operator/operator_test_utils.hpp
mbrobbel, line 1: `shared_data_repository_manager` is just an alias for `data_repository_manager`, so please revert this.

src/scan_manager/sirius_scan_manager.cpp
mbrobbel, line 1: You can use `pre-commit` to revert those formatting changes and pass the `lint` check.

src/include/legacy/expression_executor/gpu_expression_executor.hpp
```diff
@@ -62,7 +62,7 @@ namespace sirius {
  */
 struct GpuExpressionExecutor {
   using data_batch              = cucascade::data_batch;
-  using data_repository_manager = cucascade::data_repository_manager;
+  using data_repository_manager = cucascade::shared_data_repository_manager;
```
mbrobbel: Same here, please revert. Also, note that this is legacy code, we shouldn't be touching it for super sirius changes.

src/include/expression_evaluator/expression_evaluator.hpp
```diff
@@ -73,7 +73,7 @@ class expression_evaluator {
 
  public:
   using data_batch              = cucascade::data_batch;
-  using data_repository_manager = cucascade::data_repository_manager;
+  using data_repository_manager = cucascade::shared_data_repository_manager;
```
mbrobbel: Same here

mbrobbel (commented):
Thanks @aocsa. Can you please fix the merge conflicts and take a look at the failing CI jobs?

src/include/exec/exchange_channel.hpp
```diff
@@ -0,0 +1,271 @@
...
+#include <functional>
+#include <mutex>
+#include <optional>
+#include <stdexcept>
+
+namespace sirius::exec {
+
+/// Lightweight handle pushed through the channel. The shared_data_repository is the owner of
+/// record; queued batches therefore sit idle where the downgrade sweep can see and spill them
+/// (design §3/§7). A channel that owned batches would make them spill-invisible.
```
9prady9:
```suggestion
/// Lightweight handle pushed through the channel. The shared_data_repository is the owner of
/// record; queued batches therefore sit idle where the downgrade sweep can see and spill them.
```

we should remove the running notes of tools

src/include/exec/exchange_channel.hpp
```diff
@@ -0,0 +1,271 @@
...
+///     50-byte bound).
+///   - Oversized-batch rule: a handle whose size_bytes > capacity_bytes is admitted into an
+///     *empty* channel so the stream never wedges. A non-empty channel never admits a handle
+///     that would push the cumulative total past capacity_bytes.
+///   - close()   forbids further pushes; already-queued items remain poppable.
+///   - drained() ≡  closed() && empty()  — terminal EOS predicate.
+///   - Engine workers use try_push / try_pop only (non-blocking). Blocking push / pop are
+///     provided for the wrapper / test side.
+///   - Callbacks (on_push / on_pop / on_close) fire outside the lock; each is single-slot —
+///     last setter wins. on_close fires exactly once, on the first successful close().
```
9prady9: we have to improve this doc string, we should include behavior and avoid in the doc string. I think only extremely important caveats should be mentioned in class docs and the rest will go in as inline code comments.

docs/super-sirius/operators.md
```diff
@@ -62,6 +62,39 @@ The pipeline converter rewrites a DuckDB parquet or DuckDB-native table scan int
...
+### `sirius_physical_streaming_source` — `STREAMING_SOURCE`
+**File:** `src/include/op/sirius_physical_streaming_source.hpp`
+
+Source operator that marks the bottom boundary of an intermediate pipeline fragment. It pulls
+`exchange_batch_handle` records (batch-id + size) from a bounded `exec::exchange_channel`, resolves
+each handle via a `cucascade::shared_data_repository`, and publishes the batch into the pipeline
+as a `pipelineable_operator_data`. Used only when a fragment's input arrives from another node
+over exchange; a leaf fragment keeps its normal `GPU_SCAN` source.
+
+Key design invariants (design §3/§7):
```
9prady9:
```suggestion
Key design invariants:
```

It is hard to come back here when this ref is used inside code as there is no guaranteed way to find out where refs are populated.

src/include/exec/exchange_channel.hpp
```diff
@@ -0,0 +1,271 @@
...
+  /// Non-blocking push. Returns false when the handle can't be admitted (item/byte bound would
+  /// be crossed) or the channel is closed; never blocks.
+  [[nodiscard]] bool try_push(exchange_batch_handle h)
+  {
+    std::function<void()> cb;
+    {
+      std::unique_lock<std::mutex> lock(_mutex);
+      if (_closed || !can_push_unlocked(h)) return false;
+      _queue.push_back(h);
+      _total_bytes += h.size_bytes;
```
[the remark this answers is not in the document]
9prady9: better to have this here +1

src/include/exec/exchange_channel.hpp
9prady9, line 1: `try_push` - `push` and `try_pop` - `pop` share lot of common code, I think we can reduce code duplication there. It also reduces the burden of managing identical logic in two different locations in future.

src/op/sirius_physical_streaming_source.cpp
```diff
@@ -0,0 +1,108 @@
...
+  // docs/super-sirius/streaming-source-p1-fix-plan-v2-no-task-creator.md, Finding 1.
+  //
+  // get_pipeline() is resolved lazily (at callback-fire time, not here at construction time):
+  // set_pipeline() runs once per operator during query::build_indices(), which happens after
+  // every operator is constructed but before any channel activity starts in production, so by
+  // the time close() is actually called the pipeline is always set. The null check makes this
+  // safe even if that ordering assumption is ever violated (e.g. in tests that never wire a
+  // pipeline at all).
+  _input_channel->set_on_close([this] {
+    if (auto pipeline = get_pipeline()) { pipeline->update_pipeline_status(); }
```
9prady9: Blocking: this uses the default `original_pipeline=true`. `notify_downstream_pipelines()` only schedules output consumers when `!original_pipeline`. For an empty stream (zero tasks), or a late close after the last task completed, no executor task remains to schedule consumers, so this pipeline can finish while its downstream pipeline is never re-armed. Please call `update_pipeline_status(false)` here and add multi-pipeline completion regressions; the current BUG-1/BUG-2 single-operator tests only verify this pipeline's finished flag.

src/op/sirius_physical_streaming_source.cpp
```diff
@@ -0,0 +1,108 @@
...
+  });
+}
+
+sirius_physical_streaming_source::~sirius_physical_streaming_source()
+{
+  // The channel may outlive this operator (e.g. held by a producer on another thread); clear
+  // the callback before `this` becomes dangling, per exchange_channel's documented contract.
+  _input_channel->set_on_close(nullptr);
```
9prady9: Blocking lifetime race: clearing the stored callback does not synchronize with a callback already snapshotted by `exchange_channel::close()`. `close()` copies `_on_close` under the mutex and invokes it after unlocking, so a concurrent destructor can return and then the copied `[this]` lambda dereferences a destroyed operator. Since the producer/channel may outlive this operator, use a callback that does not capture `this` (for example, a weak pipeline bound after wiring), or provide synchronized callback unregistration that waits for in-flight invocations.

src/op/sirius_physical_streaming_source.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+  // Open but empty: drop the request. A future push must re-schedule via the session (#839).
+  return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, nullptr};
+}
+
+bool sirius_physical_streaming_source::all_ports_empty() { return _input_channel->drained(); }
+
+std::unique_ptr<operator_data> sirius_physical_streaming_source::get_next_task_input_data()
+{
+  auto maybe_handle = _input_channel->try_pop();
```
9prady9, line 91: Backpressure concern: `try_pop()` frees the channel item/byte capacity at task-creation time. The current `task_creator` then loops over all available input and moves those batches into the unbounded `task_scheduler::_task_queue`. A fast producer can therefore refill the channel while an arbitrarily large GPU backlog accumulates, so the advertised channel bound does not bound total outstanding data. Please account for in-flight tasks (or otherwise gate task creation) when #839 wires `on_pop`; at minimum, document and track this integration requirement explicitly.

9prady9 (changes requested):
Requesting changes for the downstream-completion wakeup and callback lifetime race. I also flagged an end-to-end backpressure concern that should be resolved before the session wiring in #839.

src/op/sirius_physical_streaming_source.cpp
```diff
@@ -0,0 +1,108 @@
...
+  //
+  // get_pipeline() is resolved lazily (at callback-fire time, not here at construction time):
+  // set_pipeline() runs once per operator during query::build_indices(), which happens after
+  // every operator is constructed but before any channel activity starts in production, so by
+  // the time close() is actually called the pipeline is always set. The null check makes this
+  // safe even if that ordering assumption is ever violated (e.g. in tests that never wire a
+  // pipeline at all).
+  _input_channel->set_on_close([this] {
+    if (auto pipeline = get_pipeline()) { pipeline->update_pipeline_status(); }
+  });
```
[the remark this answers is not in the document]
aocsa: Solved — the constructor now throws `invalid_input_exception` on a null channel or
 repository

src/op/sirius_physical_streaming_source.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+sirius_physical_streaming_source::~sirius_physical_streaming_source()
+{
+  // The channel may outlive this operator (e.g. held by a producer on another thread); clear
+  // the callback before `this` becomes dangling, per exchange_channel's documented contract.
+  _input_channel->set_on_close(nullptr);
+}
```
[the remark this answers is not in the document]
aocsa: Fixed — the callback no longer captures `this`; it holds a weak pipeline reference bound in
 `set_pipeline`.

9prady9 (approved):
lgtm

One minor note: `set_on_close` still doesn't synchronize with an in-flight callback snapshot from `close()` — but worth keeping in mind if future callbacks ever need stronger tear down guarantees.

mbrobbel (approved):

PR #1106, opened 2026-07-07
fix(starrocks): serialize result batches into a growable buffer

mbrobbel:
Result batches can be arbitrarily large; the fixed 64 KiB thrift channel truncated them, surfacing as a transport error at `fetch_data`. Serialize into a plain growable buffer instead.

## Stack

1. #1106 ⬅️ this PR
2. #1107
3. #1108
4. #1109
5. #1110
6. #1111

### DIFF
--- a/experimental/starrocks/src/result_encoder.rs
+++ b/experimental/starrocks/src/result_encoder.rs
@@ -10,10 +10,7 @@ use arrow_array::{
 };
 use arrow_schema::DataType;
 use starrocks_thrift::data::TResultBatch;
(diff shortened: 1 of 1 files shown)


aocsa (approved):
One small non-blocking suggestion: this change is easiest to regress accidentally during a future refactor because the buffer size is no longer explicit. Would you mind adding a test that serializes a `TResultBatch` larger than 64 KiB and asserts the output length exceeds that limit? That would lock in the growable-buffer behavior.

PR #1107, opened 2026-07-07
feat(starrocks): encode decimal/date/timestamp result columns

mbrobbel:
Encode decimal128, date32, and microsecond-timestamp result columns (plus large/view string arrays) as MySQL text on the `fetch_data` path.

## Stack

1. #1106
2. #1107 ⬅️ this PR
3. #1108
4. #1109
5. #1110
6. #1111

### DIFF
--- a/experimental/starrocks/src/result_encoder.rs
+++ b/experimental/starrocks/src/result_encoder.rs
@@ -4,11 +4,14 @@
 //! text-protocol resultset rows: each column value is a length-encoded string, NULL is the single
 //! byte `0xFB`. The FE forwards those row bodies straight to the MySQL client.
 
+use arrow_array::temporal_conversions::{as_date, as_datetime};
+use arrow_array::types::{Date32Type, TimestampMicrosecondType};
 use arrow_array::{
-    Array, BooleanArray, Float32Array, Float64Array, Int8Array, Int16Array, Int32Array, Int64Array,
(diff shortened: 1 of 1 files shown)


aocsa (commented):
LGTM 

# Possible improvements

1. Handle timestamp timezones — render_cell currently matches DataType::Timestamp(TimeUnit::Microsecond, _), but ignores the optional timezone field ([result_encoder.rs:103](https://github.com/sirius-db/sirius/blob/e2ae3707f561a9270a3e64346d3fd946da6ff02e/experimental/starrocks/src/result_encoder.rs#L103)). If the CN ever produces timestamps with a non-UTC timezone, they'll be formatted as UTC. It might be worth either documenting that timestamps are expected to be UTC, or applying the timezone offset before formatting.

2. Add a fractional-padding test — the current tests cover timestamps with no fractional part and with a full 6-digit fractional part ([result_encoder.rs:233-236](https://github.com/sirius-db/sirius/blob/e2ae3707f561a9270a3e64346d3fd946da6ff02e/experimental/starrocks/src/result_encoder.rs#L233-L236)). It could be useful to add a case like 1_500 microseconds to verify that %.6f produces "1970-01-01 00:00:00.001500" instead of a variable-width fractional part, which matches what MySQL clients typically expect.

9prady9 (approved):

PR #1108, opened 2026-07-07
feat(starrocks): add fragment and plan dump debug aids to the CN

mbrobbel:
Env-gated debug aids for plan work against a live frontend:

- `SIRIUS_CN_DUMP_FRAGMENTS=<dir>`: dump each received thrift fragment and its translated Substrait plan bytes
- `SIRIUS_CN_TRANSLATE_ONLY=1`: accept untranslatable fragments so the frontend dispatches (and we capture) a query's complete fragment set
- an ignored engine test that replays a dumped Substrait plan in isolation

## Stack

1. #1106
2. #1107
3. #1108 ⬅️ this PR
4. #1109
5. #1110
6. #1111


experimental/starrocks/src/compute_node_service.rs
```diff
@@ -283,9 +307,25 @@ impl SiriusComputeNodeService {
...
+        use std::sync::atomic::{AtomicU64, Ordering};
+        let Ok(dir) = std::env::var("SIRIUS_CN_DUMP_FRAGMENTS") else {
+            return;
+        };
+        static SEQ: AtomicU64 = AtomicU64::new(0);
+        let seq = SEQ.fetch_add(1, Ordering::Relaxed);
+        let path = std::path::Path::new(&dir).join(format!("plan-{seq:04}.substrait"));
+        if let Err(err) = std::fs::write(&path, translated.to_substrait_bytes()) {
+            tracing::warn!(error = %err, path = %path.display(), "failed to dump substrait plan");
+        }
```
aocsa: The plan counter can get out of sync with the fragment counter after a translation failure. Could we reuse the same ID for both files so `fragment-0001.txt` always matches `plan-0001.substrait`?

aocsa (changes requested):
LG, but I added a minor comment that I think requires a small change.

experimental/starrocks/src/compute_node_service.rs
mbrobbel: Ah good catch! Fixed in d62f5cbf.

bwyogatama (approved):

aocsa (approved):

PR #1110, opened 2026-07-07
feat(starrocks): translate one-phase aggregation and sorts

mbrobbel:
Translate `AGGREGATION_NODE` into a Substrait aggregate (sum/count/min/max/avg, `multi_distinct_count` as distinct count), restricted to finalized one-phase aggregation (`new_planner_agg_stage=1`); HAVING conjuncts become a filter over the aggregate output. Translate `SORT_NODE` as sort-tuple projection + sort; any node's limit (plus sort offset) becomes a `FetchRel`.

Rejected: merge aggregates, `avg` with a non-double result, distinct aggregates without grouping keys or over multiple columns, partitioned/rank-based top-N.

## Stack

1. #1106
2. #1107
3. #1108
4. #1109
5. #1110 ⬅️ this PR
6. #1111


aocsa (approved):
PR 1110 — aggregation and sort

Enables: 
```
GROUP BY with sum / count / min / max / avg (one-phase only: new_planner_agg_stage = 1)
COUNT(DISTINCT col) with a GROUP BY, as a Substrait distinct count
HAVING, as a filter on the aggregate output
ORDER BY (ASC/DESC, nulls first/last), via sort-tuple project then SortRel
LIMIT on any node, plus sort OFFSET, as FetchRel
```

After reviewing this PR, I found some issues, so I pushed a few fixes and improvements.

Slots were sorted by id while the FE indexes grouping keys and sort-tuple exprs by wire order, so `GROUP BY` / `ORDER BY` could swap columns silently. [`a24b1337`](https://github.com/sirius-db/sirius/commit/a24b1337be99e9601498ba5b9d9ce56c5d749245) keeps wire order; [`59298518`](https://github.com/sirius-db/sirius/commit/59298518) adds non-monotonic fixtures and a two-measure aggregate.

[`fd361a3c`](https://github.com/sirius-db/sirius/commit/fd361a3c512fc1a87c525ffd7ec7d3f542ee8cc2) pins fetch encoding: offset-only (`count = -1`), limit+offset, `LIMIT 0`, and limit above a non-sort filter.

Sorts that carried a second row tuple, a pre-agg payload, or conjuncts were translated as a different operator. Those shapes are now rejected ([`c66e9585`](https://github.com/sirius-db/sirius/commit/c66e9585), [`008bc078`](https://github.com/sirius-db/sirius/commit/008bc078)).

PR #1111, opened 2026-07-07
feat(starrocks): translate hash and nested-loop joins

mbrobbel:
Translate `HASH_JOIN_NODE` (inner/outer/left-semi; conditions evaluated over the concatenated probe-then-build row) and `NESTLOOP_JOIN_NODE` (cross product + filter, inner/cross only). Node conjuncts become post-join filters; semi joins keep the probe-side layout.

Rejected: anti joins (DuckDB's Substrait consumer has no left-anti conversion), null-safe equality conjuncts, and conjunct-free cross products (no GPU cross-product operator).

With the full stack, single-fragment plans (scan/filter/project/aggregate/sort/join) translate and execute on the GPU. Every TPC-H query still splits at a gather exchange, so TPC-H proper waits for multi-fragment execution.

## Stack

1. #1106
2. #1107
3. #1108
4. #1109
5. #1110
6. #1111 ⬅️ this PR

### DIFF
--- a/experimental/starrocks/crates/starrocks-plan-translator/src/lib.rs
+++ b/experimental/starrocks/crates/starrocks-plan-translator/src/lib.rs
@@ -33,9 +33,11 @@
 //! | `PROJECT_NODE`       | `ProjectRel`       |
 //! | `AGGREGATION_NODE`   | `AggregateRel` (finalized one-phase only, `new_planner_agg_stage=1`) |
 //! | `SORT_NODE`          | `ProjectRel` (sort tuple) + `SortRel` (global row-number top-N only) |
+//! | `HASH_JOIN_NODE`     | `JoinRel` (inner/outer/left-semi; anti joins are rejected) |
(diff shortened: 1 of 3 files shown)


aocsa (approved):
PR 1111 — hash and nested-loop joins

Enables:
```
Inner / le`ft / right / full outer equality joins (HASH_JOIN_NODE → JoinRel)
EXISTS / IN as LEFT_SEMI, probe-side output only
Non-equi joins (NESTLOOP_JOIN_NODE → CrossRel + FilterRel)
Extra ON predicates and post-join WHERE
Left-then-right row layout (right-side slot = left_width + index)
```

After reviewing this PR, I found some minor issues

2ca88c3d: one AND helper; empty list is None, not and(). Removes the dead nestloop “bare cross” branch.

5c3bcd15: check join_op before “missing conjuncts”, so an anti join is named as such.

d80cf1ce: nestloop must have a comparison that reads both sides. A filter that only hits one side still became a cross product and died in Sirius.

9463a5f1: tests for ON extras, multi-AND, post-join filter, and a wide left side so field 2 is observable.

PR #1113, opened 2026-07-08
fix: refuse duckdb-native GPU scan of overflow-length strings

bwyogatama:
## The bug: silent data corruption for strings ≥ 4 KB

DuckDB stores any **single string** whose length reaches `StringUncompressed::GetStringBlockLimit` (4,096 B at the default block size) in a separate **overflow block**, leaving only a negative offset + 12-byte `BIG_STRING_MARKER` (`block_id` + offset) in the segment dictionary — on disk too (`FetchStringFromDict` branches on `dict_offset < 0`).

The GPU duckdb-native scan cannot decode these, and nothing refused them:

1. the walker never rejects marker-bearing segments (viability checks cover codec / CONSTANT / missing-stat / int32-total only);
2. the overflow blocks are never staged — `additional_blocks` is populated via `cf.visit_block_ids`, which **only ZSTD implements** in our pinned DuckDB;
3. `kernel_gather_uncomp` (`src/cuda/scan/strings/uncompressed.cu`) takes `abs()` of the offset (its header comment says the sign bit is "irrelevant on disk", which is not the case) and memcpys the 12 marker bytes as string content.

Net effect on `dev`: a duckdb-native GPU scan of a table containing a ≥4 KB string **silently returns 12 bytes of marker garbage** in place of each big string. No error. TPC-H is unaffected (all strings < 4 KB), which is why CI never caught it.

## The fix: stat-based refusal, three layers, coarsest first

Detection is **exact and metadata-only**: DuckDB maintains `max_string_length` (the longest *individual* string) per segment/row group at write time, so `stat ≥ limit ⟺ the scope contains at least one overflow string`. No string data is read; the checks are O(#varchar columns × #row groups) field reads.

- **Plan time** (`sirius_plan_get.cpp`): `seq_scan` over a table whose varchar column's table-level stat reaches the limit throws `NotImplementedException` inside `OnFinalizePrepare`'s `create_plan()` gate → **transparent execution falls back to DuckDB CPU cleanly**. This layer is load-bearing: an execution-time refusal surfaces as `InternalException` ("Sirius GPU execution failed") with no fallback, which also invalidates the database.
- **Prepare time** (`prepare_duckdb_native_walk`): exact per-surviving-row-group refusal (skips filter-pruned row groups). Covers `pin_table`, which now fails with the reason instead of caching garbage.
- **Range-walk time**: per-segment stat guard as a stats-drift backstop so a marker-bearing segment can never reach the decoder.

Real overflow-string support (marker-resolving kernel + staging of overflow blocks) is a follow-up that would benefit the on-disk path generally; this PR makes the current behavior safe.

## Tests

- `[overflow_string]` walker cases: refuses at 5,000 B and **exactly at the 4,096 B boundary** (DuckDB overflows on `>=`), accepts 4,000 B. Shared walker-test helpers lifted out of the file's disabled `#if 0` block (unblocks future re-enable).
- `[overflow_string]` integration case: intercepted `SELECT` returns the 5,000-char string **intact** via CPU fallback — asserted positively with the transparent-execution counters (2 fallbacks, 0 rebinds, 0 GPU executions); `pin_table` on the table fails cleanly with the refusal reason. Pre-fix, this test's `max(length(s)) == 5000` assertion reads the 12-byte marker garbage.
- Regressions: `[pin_table_duckdb]`, `[pin_table_duckdb_host]`, `[pin_table]`, `[pin_table_host]`, `[pin_table_cols_subset]`, TPC-H `[Q1] [Q6] [Q18]` (both fixtures) all pass — normal varchar tables (< 4 KB strings) still take the GPU path.

### DIFF
--- a/src/op/scan/duckdb_native_metadata.cpp
+++ b/src/op/scan/duckdb_native_metadata.cpp
@@ -28,6 +28,7 @@
 #include <duckdb/main/attached_database.hpp>
 #include <duckdb/planner/table_filter.hpp>
 #include <duckdb/storage/block_manager.hpp>
+#include <duckdb/storage/segment/uncompressed.hpp>
 #include <duckdb/storage/statistics/base_statistics.hpp>
 #include <duckdb/storage/statistics/string_stats.hpp>
 #include <duckdb/storage/storage_manager.hpp>
@@ -296,6 +297,16 @@ std::optional<std::string> walk_standard_column(duckdb::ColumnData& col_data,
                std::to_string(rg_idx) + ": Max String Length stat absent from segment stats";
       }
       desc.max_string_length = duckdb::StringStats::MaxStringLength(segment.stats.statistics);
+      // Defense-in-depth for the prepare-time overflow refusal (stats-drift guard): a
+      // marker-bearing segment must never reach the GPU string decoder, which cannot
+      // resolve BIG_STRING_MARKERs and would emit the marker bytes as string content.
+      if (*desc.max_string_length >=
+          duckdb::StringUncompressed::GetStringBlockLimit(segment.GetBlockSize())) {
+        return "varchar segment on column " + std::to_string(column_id) + " row group " +
+               std::to_string(rg_idx) +
+               ": max string length reaches the overflow-block limit; overflow strings are not "
+               "GPU-decodable";
+      }
     }
     col_md.data_segments.push_back(std::move(desc));
   }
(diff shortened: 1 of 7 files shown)


src/op/scan/duckdb_native_metadata.cpp
```diff
@@ -296,6 +297,16 @@ std::optional<std::string> walk_standard_column(duckdb::ColumnData& col_data,
                std::to_string(rg_idx) + ": Max String Length stat absent from segment stats";
       }
       desc.max_string_length = duckdb::StringStats::MaxStringLength(segment.stats.statistics);
+      // Defense-in-depth for the prepare-time overflow refusal (stats-drift guard): a
```
kevkrist: This is a little more complicated than 4KB, which is correct but conservative.
UNCOMPRESSED: duckdb/src/include/duckdb/storage/string_uncompressed.hpp:139 (4KB)
DICT_FSST: duckdb/src/include/duckdb/storage/compression/dict_fsst/common.hpp:40 (16KB)
I guess DICTIONARY / FSST are rejected codes when any string reaches the uncompressed overflow threshold (4KB).
Conclusion: fine with this, but if there are strings <16KB but >4KB with DICT_FSST codec, these will be deflected to CPU unnecessarily. The codecs aren't available at the higher levels, so this distinction can't be enforced there and the conservative 4KB refusal is necessary; since we are unlikely to reach this refusal point with > 4KB strings, it is essentially moot.

TLDR: can we just document somewhere that DICT_FSST strings < 16KB may get unnecessarily refused here?

src/planner/sirius_plan_get.cpp
```diff
@@ -95,6 +104,46 @@ sirius_physical_plan_generator::create_plan(duckdb::LogicalGet& op)
...
+          continue;
+        }
+        auto const primary = col_idx.GetPrimaryIndex();
+        if (primary >= op.returned_types.size() ||
+            op.returned_types[primary].id() != duckdb::LogicalTypeId::VARCHAR) {
+          continue;
+        }
+        auto stats = table.GetStatistics(context, primary);
+        if (!stats || !duckdb::StringStats::HasMaxStringLength(*stats) ||
+            duckdb::StringStats::MaxStringLength(*stats) >= overflow_limit) {
```
kevkrist, line 133: From Codex:
Preserve CPU fallback when this check runs during prepared execution.
 
create_plan() is not confined to OnFinalizePrepare: PhysicalSiriusExecution::GetDataInternal() calls it again for every execution of a reusable prepared statement. A Connection::Prepare performed while all strings are short passes this check and replaces DuckDB’s CPU plan. If a later data-only insert adds a ≥4 KB string, DuckDB does not rebind the statement because its catalog identity has not changed. The next Execute() reaches this condition, throws after the CPU plan has already been discarded, and returns a query error instead of falling back. 

Could either: 
 -  force Sirius-backed prepared statements to rebind before execution
 - or retain/reconstruct a CPU path when this runtime eligibility check fails
A regression test should prepare and execute against short data, insert/checkpoint a 5 KB string, then execute the same PreparedStatement again and assert that it returns the correct result through CPU fallback.

We fail without fallback in many places as-is, and perhaps this is a niche case, so this is mostly an FYI.

kevkrist (approved):
lg2m. just some fyis.

Issue #819, opened 2026-05-28
Concurrent INSERT/DELETE via DuckDB with Sirius Pin-Table Cache


bwyogatama:
## Goal

Enable concurrent INSERT/DELETE via DuckDB SQL while Sirius serves OLAP queries from a GPU-pinned table cache. The cache is an immutable snapshot of the table's **last-checkpointed disk image** taken at pin time (`V_base`); DuckDB's in-memory MVCC delta (uncheckpointed transient segments + `ChunkInfo` visibility) is merged at query time, honoring each query's transaction snapshot: a query at `T_query` returns exactly what DuckDB would return at `T_query.start_time`.

This is the standard "main + differential" pattern (MonetDB/X100 PDT, HYRISE differential buffer, Apache Doris MOR, Delta Lake deletion vectors). The novel piece is wiring it to DuckDB MVCC and Sirius's GPU scan pipeline.

> **Note:** this body was rewritten 2026-07. The previous version predated the #997 IO-framework refactor and described obsolete components (`cached_split_provider`, `scan_cached_operator_data`, separate per-format scan operators, `DataTable*` cache identity, a `CheckpointLock` design, and a `captured_mask`). It is superseded in full; see edit history for the original. **Revised 2026-07-08:** the scan-manager serving reorg (PR2, #1126) was dropped before merge in favor of prepare-time MVCC jobs on the dispatcher — the serving path stays coalescer-direct, and the ladder slot is now PR2-lite (cached-serving hardening).

## Status

- ✅ **Foundation merged on `dev`**: unified `pin_table` supports DuckDB-native tables — `CALL pin_table(format='duckdb', name='lineitem', tier='gpu'|'host', schema_name=…, cols=[…])` — reading via the `gpu_ingestible` framework at pin time. Cache identity is the qualified name tuple `(catalog, schema, table)` compared in `cache_entry_info::can_serve_with_columns`. The merged foundation has **no MVCC fields and no checkpoint protection** — both are added by this work.
- ✅ `sirius::helper::get_query_start_time(ClientContext&)` merged (#823) — derives `T_query` from the live `ClientContext`.
- ✅ **PR0 — #1113**: overflow-string refusal. Fixes silent corruption on the existing native scan for ≥4KB strings, and introduces the **plan-time viability probe seam** (`create_plan(LogicalGet)`) that this work's guards reuse (see "Fallback constraint" below).
- This issue tracks **PR1–PR6** below.

## Architecture

```
[...]

PR #1115, opened 2026-07-08
feat(pin_table): MVCC snapshot foundation for duckdb-native pins (#819 PR1)

bwyogatama:
Part of #819 (PR1 of the ladder — foundation only, serve path unchanged). Foundation for query-time MVCC delta merge over the duckdb-native pinned cache: everything a pin must remember at pin time so later stages can reconcile the cached snapshot with DuckDB's live MVCC state (DELETE visibility masks, then the INSERT delta).

## Summary

- **New `duckdb_mvcc_metadata`**, composed onto `pinned_entry` (`nullptr` for parquet pins): `v_base` (the pin transaction's `DuckTransaction::start_time` on the pinned table's **own** catalog — each AttachedDatabase has its own MVCC counter domain) and `base_row_count_per_chunk` (parallel to the entry's chunk storage; prefix sums map each chunk to its absolute rowid range; `n_cache()` = the base/delta boundary).
- **`sirius_scan_manager::attach_mvcc_metadata(name, metadata)`** — called by the duckdb pin path right after insert; overwrites on re-pin (fresher fence); throws if the entry is absent.
- **Auto-checkpoint suppressed at duckdb pin time** — a checkpoint would compact tombstones/flush transient appends and silently shift rowids under the positional cache. Both triggers are disarmed: `checkpoint_wal_size` (set to max) **and** the entry-count trigger `wal_autocheckpoint_entries` (zeroed). Deliberately not restored on unpin (needs pin refcounting; PR5 candidate). Manual `CHECKPOINT` — and `DETACH`, whose close runs a shutdown checkpoint — while pinned are documented out of contract.
- **Per-chunk row counts captured on both tiers**: `materialized_pin` gains `base_row_count_per_chunk`; the host path returns a new `materialized_host_pin` mirroring it.
- **Whole-row-group chunk invariant enforced at pin time** (`validate_duckdb_pin_chunk`): every duckdb-native batch must be a contiguous run of whole row groups continuing exactly where the previous batch ended, and its metadata row total must equal the decoded row count. A future coalescer change that splits/reorders row groups fails the pin loudly instead of silently misaligning the downstream visibility masks.
- **Merge-path guard in `insert_pinned_entry`** (pre-existing hazard, surfaced by review): byte-budget coalescing is column-set-dependent, so a re-pin with different columns can produce equal chunk *counts* with different row-group boundaries — passing the existing memory-space checks while corrupting the merged entry positionally. The merge now rejects per-chunk row-count mismatches loudly.

## Notes / observed behavior

- A pin taken while committed-but-uncheckpointed rows exist **fails loudly today** (transient segments stage zero bytes; the decoder's pre-validation throws) — no silent garbage is possible. PR4 will make that case work by excluding transient tails from the pin walk and serving them as the insert delta.
- Same-name same-row-count re-pins across formats remain a pre-existing `insert_pinned_entry` quirk (cache identity is preserved from the first pin); unchanged here.

## Test plan

- [x] New `[pin_table_mvcc]` suite (14 cases, 279 assertions): GPU + host duckdb pins capture `v_base`/counts with counts matching the cached chunks' real row counts elementwise; multi-chunk pin (13M rows > the 100MB batch cap) records the per-chunk rowid partition at row-group multiples; `v_base` equals the pin transaction's `start_time` on the pinned catalog exactly (pinned inside an explicit transaction); parquet pin carries no metadata; re-pin refreshes `v_base` through the merge path; WAL grows past the 16MiB default across 40 commits with both triggers disarmed (entry trigger armed at 5 before the pin); INSERT/DELETE run while pinned; `attach_mvcc_metadata` without an entry throws; `validate_duckdb_pin_chunk` failure paths unit-tested (gap, out-of-order, row-total mismatch, non-duckdb skip).
- [x] Existing pin regression suites pass unchanged: `[pin_table]`, `[pin_table_host]`, `[pin_table_duckdb]`, `[pin_table_duckdb_host]`, `[pin_table_cols_subset]`, `[pin_table_host_streaming]`, `[pin_mgpu]` (17 cases, 887 assertions).
- [x] `pixi run pre-commit run -a` clean.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -596,6 +596,7 @@ set(TEST_SOURCES
     test/cpp/integration/test_pin_table_host_streaming.cpp
     test/cpp/integration/test_pin_table_merge_columns.cpp
     test/cpp/integration/test_pin_table_column_order.cpp
+    test/cpp/integration/test_pin_table_mvcc_foundation.cpp
     test/cpp/integration/test_table_gpu_cache_warm_mgpu.cpp
     test/cpp/io/s3/test_sigv4.cpp
     test/cpp/io/s3/test_s3_default_visibility_guard.cpp
--- a/src/include/pin_table.hpp
+++ b/src/include/pin_table.hpp
@@ -72,6 +72,7 @@ namespace sirius {
 
 namespace op::scan {
 class gpu_ingestible;
+class scan_info;
 }  // namespace op::scan
 namespace io {
 class sirius_ioctx;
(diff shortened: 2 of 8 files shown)


wmalpica (approved):

bwyogatama:
Merge-queue note: the cuda13 Test job in the queue ([run 29034867074](https://github.com/sirius-db/sirius/actions/runs/29034867074)) timed out at \`[964/1666] pin_table mvcc - duckdb pin suppresses WAL auto-checkpoint\`. The uploaded sirius trace log shows the test's pin and all DML completed in ms; the process then sat 36 minutes inside transparent execution of the test's trivial \`SELECT current_setting('wal_autocheckpoint_entries')\` — a `CPU_SOURCE → DUMMY_SCAN → PROJECTION` plan whose engine execution never returned (last log line \`Executing sirius_engine\` at 17:04:28.991, nothing after). The Catch2 progress line was stuck in the stdout buffer, so the GH log misattributes the stall.

That's a latent state-dependent engine hang on the trivial-pipeline path (same family as the known full-suite teardown flakes), not a WAL/pin defect — the full suite passes locally in the same order (1629/1629 on a single-GPU box). `aa19e6eee` removes the trigger from this test: the setting is read back via `Settings::Get` (the same read `AutomaticCheckpoint` uses) and the test's statements run with `gpu_execution` off, since its subject is WAL mechanics. The underlying engine hang is worth its own issue; it needs the 2×L4 / cuda13 runner to reproduce.

Issue #1118, opened 2026-07-08
MARK join: emit NULL mark for the hash-join (equality) path — fixes IN/NOT IN with NULLs


9prady9:
Sub-issue of #1076. Scope: the **hash-join (equality) MARK path** only — the path every `IN` / `= ANY` / `NOT IN` / correlated equality subquery flows through. This resolves all the cases #1076 reproduces.

## Problem

`resolve_mark_join_result` (`src/op/sirius_physical_hash_join.cpp:863-908`) builds the BOOL8 mark column all-`false` and only ever scatters `true` at matched left-row indices — there is no path that writes a `NULL`. By SQL three-valued logic, an unmatched left row is `NULL` (not `false`) when the probe key is NULL **or** the build/right side contains a NULL join key.

## Correct semantics

| case | mark |
|---|---|
| `x` matches an element of `S` | `true` |
| no match, `x` non-NULL, `S` has no NULL key | `false` |
| no match, and (`x` is NULL **or** `S` has a NULL key) | **`NULL`** |

## Approach

The mark **values** are already correct (they equal the matched-indicator); only a **null mask** must be added. Validity decomposes into two cases (a matched row always has a valid probe key under `null_equality::UNEQUAL`):

- `build_has_null == true` → every unmatched row is NULL → `valid = matched` → mask from `cudf::bools_to_mask(mark_view)`.
- `build_has_null == false` → `valid = probe_row_valid` → mask from `cudf::bitmask_and(probe_keys)`.

Then `mark_column->set_null_mask(std::move(mask), null_count)`. No custom kernel.

Threading:
- Extend `resolve_mark_join_result` with `probe_keys` (table_view) and `build_has_null` (bool) params.
- Add a `table_has_any_null(table_view)` helper (any key column with `null_count() > 0`).
- Compute/pass at all four hash-join MARK call sites (BUILD_PROBE, mixed, single-shot build-on-left, single-shot standard). For BUILD_PROBE, compute at build time and store in a new `bool _build_has_null` member (`src/include/op/sirius_physical_hash_join.hpp`, under `op_state_mutex`).
- Includes: `<cudf/transform.hpp>`, `<cudf/null_mask.hpp>`.

## Tests (`test/cpp/operator/test_physical_mark_join.cpp`, tag `[physical_mark_join]`)

The existing cases use non-NULL INT32 keys so the bug hides. Add:
[...]

Issue #1076, opened 2026-07-04
MARK joins never emit a NULL mark — IN/NOT IN wrong when NULLs are involved


ran-yuan-rui:
## Summary

MARK joins never produce a NULL mark — an unmatched left row always comes back `false`. Mechanically, `resolve_mark_join_result` (`src/op/sirius_physical_hash_join.cpp:853-855`) builds the BOOL8 mark column from `numeric_scalar<bool>(false, /*valid=*/true)` and only ever scatters `true` at the matched row indices; there is no path that writes a NULL. The nested-loop MARK path added in #1065 scatters into the same all-`false` column.

That's harmless until NULLs appear. By SQL's three-valued logic an unmatched row is `NULL`, not `false`, when the probe value is NULL **or** the build side contains a NULL in the join key. Because we always emit `false`, everything DuckDB plans as a MARK join — `IN`, `= ANY`, `NOT IN`, correlated subqueries — returns wrong answers once NULLs are involved. The sharpest case is `x NOT IN (subquery containing a NULL)`, which must return no rows but returns them (`false → NOT → true → kept`).

The NLJ MARK path #1065 just added shares the same all-`false` scatter. #1065 flagged it as a known limitation and split it out here.

## Correct semantics

For a left row with key `x` against build set `S`:

| case | mark |
|---|---|
| `x` matches an element of `S` | `true` |
| no match, `x` non-NULL, `S` has no NULL key | `false` |
| no match, and (`x` is NULL **or** `S` has a NULL key) | **`NULL`** ← we return `false` |

## Reproduce (SQL, vs the DuckDB CPU oracle)

```sql
CREATE TABLE l AS SELECT * FROM (VALUES (1),(2),(3)) t(x);
CREATE TABLE r AS SELECT * FROM (VALUES (2),(NULL)) t(y);

-- IN projection:  DuckDB → true, NULL, NULL     Sirius → true, false, false
SELECT x, x IN (SELECT y FROM r) FROM l ORDER BY x;

-- NOT IN filter:  DuckDB → (no rows)            Sirius → 1, 3
SELECT x FROM l WHERE x NOT IN (SELECT y FROM r) ORDER BY x;
```

## Coverage gap / test to add

The existing `[physical_mark_join]` cases all use non-NULL INT32 keys, so the all-`false` output is trivially correct and the bug hides. A regression test needs a NULL in play — e.g. mirror the `create_mark_join()` fixture with a right key column that contains a NULL:
[...]

PR #1121, opened 2026-07-08
fix(nlj): emit NULL mark for conditional MARK joins with NULL keys

### LINKED ISSUE #1119: MARK join: emit NULL mark for the nested-loop (conditional) path
Sub-issue of #1076. Scope: the **nested-loop-join (conditional / non-equality) MARK path** added in #1065 (`src/op/sirius_physical_nested_loop_join.cpp:408-443`). This path is only reached for MARK joins whose conditions are *purely* non-equality (rare); the common `IN` / `NOT IN` cases go through the hash-join path (see sibling sub-issue).

## Problem

Same root cause as #1076: the NLJ `resolve_mark_join_result` builds the mark all-`false` and scatters `true` only at matched rows, never emitting `NULL`. #1065 flagged this as a known limitation.

## Why this is harder than the equality path

`cudf::conditional_left_semi_join` returns only rows where the predicate is **TRUE** for some build row. To emit the correct three-valued mark we must distinguish, for an unmatched row, "predicate FALSE for all build rows" (`false`) from "predicate UNKNOWN for some build row" (`NULL`). That requires a **second pass with a `predicate IS NOT FALSE` predicate**:

- `true_match` = `conditional_left_semi_join(P)` → `true`.
- `maybe_match` = `conditional_left_semi_join(P')` where `P'` is Kleene "P is TRUE or UNKNOWN".
- mark = `true` if in `true_match`; `NULL` if in `maybe_match` but not `true_match`; else `false`.

Kleene "IS NOT FALSE" is **not** a simple wrap of the existing AST, so this needs real work in `gpu_expression_translator` to synthesize `P'`, plus its own three-valued-logic test matrix.

## Tests

Add NLJ MARK regression cases with NULLs (via `create_projected_nlj(JoinType::MARK, …)` / `execute_projected_nlj`) covering: probe-side NULL in a referenced column, build-side NULL in a referenced column, and predicates mixing NULL and definitely-false evaluations. Validate against the DuckDB CPU oracle.

9prady9:
Fixes #1119 (sub-issue of #1076; pairs with the merged #1118 hash-join fix).

## Problem

The NLJ MARK path marked every unmatched left row `false`. `cudf::conditional_left_semi_join` only returns TRUE matches, dropping NULL/UNKNOWN pairs — but by SQL three-valued logic an unmatched row is `NULL` when the predicate was UNKNOWN (never TRUE) for some right row.

## Fix

**Three-valued mark (commit 1):** a second semi-join on `predicate IS NOT FALSE` separates definite-false rows from unknown ones. Since a conjunction is FALSE iff some conjunct is FALSE, the predicate is built per comparison as `ci OR IS_NULL(left) OR IS_NULL(right)` (Kleene `NULL_LOGICAL_OR`). A row is `NULL` iff it's in this "maybe" set but not the matched set; the null mask comes from `bools_to_mask` over `validity == matched OR NOT maybe`.

**Null-safe comparisons (commit 2):** `IS [NOT] DISTINCT FROM` yield a definite TRUE/FALSE for NULL operands, never UNKNOWN, so:
- their conjuncts skip the `IS_NULL` tainting (else `NULL IS NOT DISTINCT FROM 5` would wrongly mark NULL instead of FALSE), and
- `distinct_from` is now lowered as `NOT(NULL_EQUAL(l, r))` instead of null-propagating `NOT_EQUAL` — a pre-existing bug where `NULL IS DISTINCT FROM 5` evaluated UNKNOWN instead of TRUE on every NLJ join type.

Also dedups the two AND-chain builders into one `fold_logical_and` helper.

## Tests

`[physical_nested_loop_join][mark]` regressions: right/probe-side NULLs, definite-false stays false, `NULL AND FALSE` conjunction absorption, mixed null-safe + propagating conjunction, and the full `IS [NOT] DISTINCT FROM` truth table. Full C++ suite: 1649/1649.

### DIFF
--- a/src/op/sirius_physical_nested_loop_join.cpp
+++ b/src/op/sirius_physical_nested_loop_join.cpp
@@ -361,10 +361,7 @@ cudf::ast::ast_operator to_ast_operator(sirius::comparison_type comparison)
     case sirius::comparison_type::equal: return cudf::ast::ast_operator::EQUAL;
     case sirius::comparison_type::not_distinct_from: return cudf::ast::ast_operator::NULL_EQUAL;
     case sirius::comparison_type::not_equal: return cudf::ast::ast_operator::NOT_EQUAL;
-    case sirius::comparison_type::distinct_from:
-      // No single null-safe cuDF operator; the condition loop builds NOT(NULL_EQUAL) instead
-      // (NOT_EQUAL would be null-propagating and wrong for NULL operands).
-      break;
+    case sirius::comparison_type::distinct_from: break;  // built as NOT(NULL_EQUAL) by the caller
     case sirius::comparison_type::lt: return cudf::ast::ast_operator::LESS;
     case sirius::comparison_type::gt: return cudf::ast::ast_operator::GREATER;
     case sirius::comparison_type::le: return cudf::ast::ast_operator::LESS_EQUAL;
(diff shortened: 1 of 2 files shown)


wmalpica (approved):

9prady9:
@wmalpica please don't merge it yet, I found a rare corner case which I am trying to fix.

wmalpica (approved):

Issue #1124, opened 2026-07-08
Dynamic filters: measure, disable, then delete the build-priority pass


kevkrist:
Resolves #1014 by measurement, not assertion. The pass only reorders already-queued feeder tasks; its memory cost and its coverage benefit are both unmeasured today. Sequence: instrument, compare, disable, delete.

- [ ] **A1/A2** — (instrumentation: stable publication/target/channel/filter IDs, publication outcomes, per-query resident high-water, build/hash-table lifecycle counters, channel coverage) and A2 (`dynamic_filter_build_priority={legacy,off}` switch; gates only the dispatch preference, telemetry stays on in both modes) as one PR with stage-per-commit history (A1a, A1b, A2). Zero behavior change at defaults.
- [ ] **A3** — (measurement) three configs (filters disabled / legacy / off) on nested-join TPC-H + a many-join synthetic; gate: no wall-time or resident-peak regression outside run variance, with coverage explaining any movement. Results posted here.
- [ ] **A4** (decision-gated on the measurement) --  delete `collect_filter_build_pipelines`, `_filter_build_pipelines`, and the priority `pop_if` branch. Scheduler becomes filter-agnostic.

kevkrist:
## Track A / #1014 — findings summary

**Verdict: PASS. Ship `dynamic_filter_build_priority=off` as the default.**

[EDIT]: no performance diff, for single gpu SF=300, gb200. 1-2% regression observed for multi-gpu -- not enough to justify this code, which is a hacky bypass of the natural task scheduling and breaks when dynamic filters are pushed higher up in the query plan.

PR #1133, opened 2026-07-08
Issue 1010

### LINKED ISSUE #1014: Dynamic filters: bound build-task prioritization (memory)
The scheduler prioritizes every build-subtree pipeline so the filter publishes before the probe drains. Prioritizing all of them front-loads their hash tables, which can raise the resident-memory peak on many-join or large-build plans.

Options: cap the number of concurrently-prioritized build subtrees, or prioritize each build only relative to its own consumer once that consumer is live (staggering builds instead of clustering them — most relevant once probes are non-scan; see the SIP follow-up).

Code: `src/pipeline/task_scheduler.cpp` (build-subtree priority pass). Follow-up to the Phase-1 dynamic-table-filter work.

kevkrist:
This PR removes the build-task prioritization that was integrated into the scheduler by #794. The code was predicated on an assumption which is false: that probe scan tasks can be scheduled before their corresponding build concat operator completes. Moreover, the code introduces performance regression, as documented in #1124.

Fixes #1014

### DIFF
--- a/docs/super-sirius/issue-1010-dynamic-filter-sip-design.md
+++ b/docs/super-sirius/issue-1010-dynamic-filter-sip-design.md
@@ -1,0 +1,1181 @@
+# General Dynamic Filters: Sideways Information Passing at Hash-Join Probe Inputs
+
+**Status:** finalized proposal — converged after cross-review (Claude ↔ Codex); ready for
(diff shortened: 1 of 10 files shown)


kevkrist:
Accident

PR #1142, opened 2026-07-10
feat: query-time MVCC DELETE masks for the duckdb-native pinned cache (#819 PR3)

bwyogatama:
Third rung of #819 (after #1113, #1115, #1136): a duckdb-pinned table keeps serving GPU queries while DELETEs land. Every query returns exactly what DuckDB returns at its own snapshot — pre-pin tombstones, post-pin committed deletes, and the query's own uncommitted deletes are all filtered out. States the cache can't represent decline to the transparent CPU fallback at plan time.

## How it works

On a duckdb+mvcc cache hit, `prepare_for_query` builds a per-query **keep-mask** over the cached rows before serving starts: one mask job per pinned table, fanned out across the scan-manager dispatcher and joined before serving begins. Serving itself is unchanged — masks ride the existing scan input, and `materialize_table` applies each mask on the GPU (`mask_to_bools` + `apply_boolean_mask`). Clean tables cost almost nothing: a non-loading row-group probe skips the mask entirely when a chunk has no version state.

## Cache-or-CPU guards (plan time)

While a table is MVCC-pinned, a GPU plan either serves from the cache or declines in `create_plan(LogicalGet)` — the only place a refusal still becomes a clean CPU fallback. The plan-time checks are all O(1):

- **snapshot-too-old** — the transaction opened before the pin.
- **insert-present** — committed rows past the pinned prefix, or the transaction's own uncommitted appends (served once PR4's insert-delta reader lands).
- **column-mismatch** — the table is pinned with only a subset of its columns and the query needs one outside that subset (a rowid-only scan like `count(*)` counts too). This isn't about a column missing from the query; the cache simply doesn't hold that column's data, so it can't serve the scan — even when the table is otherwise clean.

If a matched entry can't serve, prepare raises a loud error rather than silently reading the stale disk image.

## Tests

- `[duckdb_mvcc_visibility]`, `[mvcc_mask_job]`, `[host_keep_mask]` — the visibility walk, concurrent mask assembly, and the GPU apply helpers.
- `[cached_serving]` — mask forwarding through the serving path.
- `[pin_table_mvcc_delete]` — end-to-end masked serving and every live guard's fallback, asserting the exact path taken (cache-served vs CPU fallback).
- `[can_serve]`, `[pin_table_mvcc]`, and the existing pin suites are unchanged.

Depends on #1113, #1115, #1136. Follow-ups: #1160 (walk-guard replacement), PR4 (INSERT delta).


src/op/scan/gpu_ingestible.cpp
```diff
@@ -33,6 +38,25 @@ filtered_table gpu_ingestible::materialize_table(const op::scan::scan_operator_i
...
     auto view   = get_cudf_table_view(rbatch);
+    if (split.mvcc_keep_mask) {
+      auto const& mask = *split.mvcc_keep_mask;
+      if (mask.row_count != static_cast<std::size_t>(view.num_rows())) {
+        throw std::runtime_error("[gpu_ingestible::materialize_table] mvcc keep-mask covers " +
+                                 std::to_string(mask.row_count) +
+                                 " rows but the cached chunk has " +
+                                 std::to_string(view.num_rows()));
+      }
+      auto masked = apply_host_keep_bitmask(
```
wmalpica, line 49: We should check the mask.row_count to ensure its > 0.
Otherwise we are making a copy for no reason. Since in apply_host_keep_bitmask we have:
`if (num_rows == 0) { return std::make_unique<cudf::table>(view, stream, mr); }`

wmalpica: Also, another possible optimization is that we return the along the mvcc mask, and we apply it at `post_filter_and_project`. If we do so, and there is a filter that is being applied en the expression_evaluator::select,  we could potentially combine the mask, so that we are basically only calling cudf::apply_boolean_mask once and avoiding an intermediate materialization.

Granted if we did that, then we are applying the expression evaluation on potentially  a larger batch, so there is a bit of a tradeoff.

src/op/scan/sirius_gpu_scan_operator_data.cpp
```diff
@@ -72,7 +72,17 @@ std::size_t scan_operator_input::get_estimated_working_set_size_in_bytes() const
     return std::get<std::unique_ptr<scan_info>>(materialization_info)
       ->estimated_working_set_bytes();
   }
-  return get_estimated_size_in_bytes();
+  auto const batch_bytes = get_estimated_size_in_bytes();
```
wmalpica, line 75: Something this code does not take into account, ether before this PR and in this PR, is that if a filter is applied in post_filter_and_project, then we end up with another intermediate copy (a smaller copy, but its still another copy), whose memory is not being taken into consideration. 
Maybe in this function we check to see if the ingestible has a filter_expression, and if so, we add more to the estimate

src/include/scan_manager/mvcc_chunk_mask.hpp
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * One bit per row in cuDF's bitmask convention (bit `row % 32` of uint32 word
+ * `row / 32`, 1 = keep), so the packed words upload and expand with
+ * @c cudf::mask_to_bools without translation; padding bits past @ref
+ * row_count are don't-care. @ref words points into storage owned via the
+ * type-erased @ref retention (the mask job's pinned {reservation, blocks}
+ * bundle; tests may retain a plain vector). Immutable once published.
+ */
+struct mvcc_chunk_mask {
+  std::span<std::uint32_t> words;   ///< ceil(row_count / 32) packed keep bits
```
wmalpica: src/include/scan_manager/mvcc_chunk_mask.hpp:35 says the mask is immutable once published, and consumers receive shared_ptr<const mvcc_chunk_mask>, but words is
     std::span<std::uint32_t>. Constness of the containing structure does not make the span’s elements const, so a consumer can still modify published mask words.

Store std::span<const std::uint32_t> in the published object while retaining a mutable span only in chunk_work.

wmalpica (changes requested):
the scan documentation needs updating for the new behavior.

docs/super-sirius/scan.md:38 says resident cached batches are forwarded as-is, and line 40 says resident working-set size is just the input size. Line 221 similarly says identity-layout cached batches flow through untouched.

MVCC-pinned DuckDB chunks can now carry a keep-mask, allocate a BOOL8 expansion and filtered table, synchronize the stream, and use a larger memory estimate. The documentation should
     also describe:
      - Per-query DELETE visibility masks and their preparation phase.
      - Snapshot, insert, and column-mismatch fallback guards.
      - Pin-time update-chain rejection.
      - The pinned/pageable host-mask storage behavior.
      - The additional GPU allocation and synchronization for masked chunks.

src/include/scan_manager/mvcc_chunk_mask.hpp
```diff
@@ -0,0 +1,48 @@
...
+  std::size_t row_count{0};         ///< rows covered (== the chunk's row count)
+  std::shared_ptr<void> retention;  ///< keeps the words' storage alive
+};
+
+/// Per-pinned-entry mask set for one query: slot i masks chunk i (the cached
+/// provider's chunk order). A null slot means every row of that chunk is
+/// visible — the chunk serves unmasked (no upload, no kernel).
+using mvcc_chunk_mask_set = std::vector<std::shared_ptr<mvcc_chunk_mask const>>;
```
aminaramoon: this is very fishy! I need to read this carefully to see why we need a `shard_ptr<vector<shared<mask>>>
where it stores a span and shared_ptr

something about the chain of ownership s not right here

src/scan_manager/mvcc_mask_job.cpp
```diff
@@ -0,0 +1,340 @@
...
+  std::size_t row_count{0};
+  std::shared_ptr<void> retention;
+  std::span<op::scan::mvcc_row_group_slice const> slices;
+  duckdb::TransactionData transaction{duckdb::TransactionData::Committed()};
+  std::atomic<bool> any_deleted{false};
+};
+
+}  // namespace
+
+void run_mvcc_mask_jobs(std::span<mvcc_mask_job_request> requests,
```
aminaramoon: this is extremely difficult code to follow and even harder to test. 
all he implemenation are hidden inside this function, why don't have a simple 

`prepare_tasks(requests, ....) -> tasks
and dispathc methods
`

in release mode, this code becomes very diffcult to follow inside gdb, cuase most of it gets inlined and optimized out. that is why you probably want to organize this a bit nicer for review and testing.

aminaramoon (changes requested):

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -807,44 +846,109 @@ void validate_pinned_entry_for_serving(pinned_entry const& entry,
...
+                                     : entry.host_chunks.size();
+        if (stored_chunks != n_chunks) {
+          // Pin-time metadata and entry chunks are kept in lock-step (the
+          // insert merge-path guard); a mismatch means masks would bind to
+          // the wrong rows.
+          throw std::runtime_error("mvcc metadata covers " + std::to_string(n_chunks) +
+                                   " chunk(s) but the entry stores " +
+                                   std::to_string(stored_chunks));
+        }
+        auto masks     = std::make_shared<mvcc_chunk_mask_set>(n_chunks);
```
ran-yuan-rui: Could we deduplicate the MVCC mask job per pinned entry within a query?

`try_assign_cached_entries()` runs once per scan operator, and this block always
creates a new `mvcc_chunk_mask_set` and pushes a new job request. For a self-join
or repeated CTE reference to the same pinned table, `run_mvcc_mask_jobs()` will
therefore repeat the row-group capture, host reservation/allocation, and mask
fill for identical visibility state.

The keep-mask depends on the query transaction and the pinned table/chunk
layout, not on the scan operator's projected columns. All providers for the
same pinned entry in one query should be able to share one immutable mask set;
the existing `shared_ptr<const mvcc_chunk_mask_set>` ownership already supports
that.

A query-local map keyed by the pinned entry (and `DataTable*`, if needed) could
create one request on the first scan and return the same mask set to subsequent
providers. This would also make the implementation match the PR description's
"one mask job per pinned table" behavior.

src/include/scan_manager/mvcc_chunk_mask.hpp
bwyogatama: Thanks for flagging this, i will try to do something about this

src/scan_manager/mvcc_mask_job.cpp
bwyogatama: will do!

src/scan_manager/sirius_scan_manager.cpp
bwyogatama: yes i will incorporate this

src/op/scan/sirius_gpu_scan_operator_data.cpp
bwyogatama: yeah you are right

src/op/scan/gpu_ingestible.cpp
bwyogatama: mask.row_count is the same as the number of rows in our data batch btw
and currently we are already not doing the copy if num_rows = 0

src/op/scan/sirius_gpu_scan_operator_data.cpp
bwyogatama: yeah you are right

wmalpica (approved):

aminaramoon (approved):

PR #1144, opened 2026-07-10
feat: replace CPU_SOURCE with a GPU_VALUES source operator

Jedi18:
## Problem

Queries whose source is plan-time CPU data hang indefinitely on the GPU path:

```sql
SELECT b FROM (VALUES (1), (2), (3)) t(b);   -- COLUMN_DATA_SCAN
SELECT 42;                                    -- DUMMY_SCAN
SELECT n_nationkey FROM nation WHERE 1=0;     -- EMPTY_RESULT
```

PR #913 removed `duckdb_scan_executor`, the only component that dispatched
`cpu_source_task`. Since then nothing runs `CPU_SOURCE` operators: they have no
input ports, so the task creator's normal loop sees `all_ports_empty() == true`
and never creates a task, and downstream pipelines block forever.

PR #980 fixed this by creating **and executing** the task inline in
`task_creator::manager_loop()`. That was reverted (#1114): the task creator
must only create work, and CPU_SOURCE itself is slated for deprecation.

## Fix

Replace `CPU_SOURCE` with a new **`GPU_VALUES`** source operator
(`sirius_physical_gpu_values`) modeled on `GPU_SCAN`, and delete the CPU_SOURCE
path entirely.

`GPU_VALUES` is a proper single-shot pipeline source. It implements the three
source hooks the task creator already drives:

- `get_next_task_hint()` — CAS-latched: `READY` exactly once, `nullopt` after.
- `all_ports_empty()` — false until its one task input has been handed out.
- `get_next_task_input_data()` — returns one `gpu_values_input`, a small
  descriptor (mirroring `scan_operator_input`) whose `prepare_for_processing()`
  captures the task's reserved memory space.

The one task flows through the completely standard machinery —
`gpu_pipeline_task`, GPU reservation, executor, sink publish, pipeline task
accounting.

**Note**: Sirius has no built-in ColumnDataCollection/DataChunk → cudf::table converter. This constructs the cuDF columns directly using the task’s stream and reserved GPU allocator, avoiding the previous intermediate host_data_representation conversion. Pageable staging is intentional because these plan-materialized inputs are expected to be small.

`execute()` converts the DuckDB data to a cudf table on the task's stream:

- **Collection** (VALUES / materialized subqueries / statistics propagation):
  stages each column host-side in cudf layout (fixed-width payload, string
  offsets+chars, validity bitmask; invalid rows zero-filled to stay
  ASAN/compute-sanitizer clean) and uploads with `cudaMemcpyAsync` into
  `rmm::device_buffer`s from the reservation's memory space. Data here is
  KB-scale, so pageable staging is deliberate — no pinned-pool or IO machinery.
- **DUMMY_SCAN**: one all-null row. A 0-output-column dummy scan (`SELECT 42`)
  emits a 1-row all-null TINYINT sentinel column, because cudf derives table
  row count from columns; downstream constant projections read `num_rows()`
  only, and projection output drops the sentinel.
- **EMPTY_RESULT**: a 0-row table with the declared schema, so downstream
  operators (e.g. `count(*)` → 0) see the same shape a filter-everything
  pipeline produces and completion flows through the normal executor path.

A viability gate (`sirius_physical_gpu_values::throw_if_unsupported_types`)
rejects types the conversion can't represent faithfully (HUGEINT/UHUGEINT:
DuckDB stores 16 bytes vs cuDF's 8; DECIMAL with precision <= 4: INT16 storage
with no cuDF equivalent; nested types). It throws during plan construction —
before the source's ColumnDataCollection is moved out — triggering the
existing clean DuckDB CPU fallback.

### Query kickoff

`query::build_indices()` registers `GPU_VALUES` in `_scan_operators` so
`task_scheduler::start_query()` can kick off plans with no GPU_SCAN
(e.g. `SELECT 42`), which would otherwise never schedule anything.

## Removed

- `sirius_physical_cpu_source.{hpp,cpp}`, `cpu_source_task.{hpp,cpp}`
- `split_cpu_source()`, the CPU_SOURCE repository-wiring branch, and the
  `CPU_SOURCE` operator type

## Tests

`[gpu_values]`-tagged integration tests (GPU vs CPU result comparison):

- `WHERE 1=0` empty result, and `count(*)` over an empty result
- Constant-only `SELECT 42` (DUMMY_SCAN sentinel path)
- `VALUES` source, including VARCHAR + NULLs
- 5000-row generated VALUES clause (collection spans multiple DataChunks)
- VALUES-backed CTE consumed twice (multi-repository fan-out)
- VALUES joined against a base table (GPU_VALUES + GPU_SCAN in one plan)

The six existing STATISTICS_PROPAGATION tests were retagged
`[cpu_source]` -> `[gpu_values]`.

## Notes

- `SELECT 42` may log a harmless output-type validation warning (the internal
  sentinel batch has one column while DUMMY_SCAN declares zero); the sentinel
  never reaches results.
- If large collections ever route through this operator, pinned staging and
  chunked upload would be worth revisiting; for VALUES-scale data the simple
  path is preferred.


src/include/op/sirius_physical_gpu_values.hpp
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  /// Feeds the reservation system's input_basis (see
+  /// gpu_pipeline_task_local_state::get_task_consumption_basis).
+  [[nodiscard]] std::size_t get_estimated_size_in_bytes() const override
+  {
+    return _estimated_bytes;
+  }
+
+  ::cucascade::memory::memory_space* gpu_memory_space = nullptr;
```
aminaramoon: make this private with a. getter, so it cannot be changed by users that have acess to a mutable ref

aminaramoon (approved):
LGTM! thanks Akhil, just a small comment

Issue #1123, opened 2026-07-08
Pinned DuckDB pushes join filters below LIMIT/TOP_N (wrong results, latent)


kevkrist:
**Decision (2026-07-08): deferred.** We accept the latent exposure and pick up the fix at the next pin bump to a released DuckDB containing duckdb/duckdb#22963. This issue stays open to track that bump. Until then: LIMIT/TOP-N-shaped tests must use explicit expected rows or a filters-disabled reference run — never an unpatched CPU run (it shares the bug). Nothing else is blocked.

---

The pinned DuckDB (v1.5.4, `08e34c447b`) pushes join filters below `LIMIT` and `TOP_N`: `GetPushdownFilterTargets` recurses through both (`duckdb/src/optimizer/join_filter_pushdown_optimizer.cpp:62-72`). A filter applied below a row selector changes which rows survive it. Upstream fixed this as a wrong-results bug in duckdb/duckdb#22963 (`4c8c90db44`), after our pin.

**Affected:** DuckDB's own CPU min/max pushdown at this pin, and Phase 1 dynamic table filters (#794) — both consume the same `probe_info` metadata.

**Severity: latent, not active at defaults.** `late_materialization` rewrites the common LIMIT/TOP-N-over-scan shapes into a legal rowid semi-join, `compressed_materialization` kills the pushdown on small-integer shapes, and scheduling races mask the rest. Wrong results were reproduced with those two optimizers disabled, on a dependency-ordered TOP-N shape. The exposure is deterministic at plan-metadata level (`probe_info` non-empty below a selector).

kevkrist:
Decision: deferring the fork/backport. Waiting for the fix in a released DuckDB; this issue now tracks the pin bump. Body updated; the B1 plan doc becomes the pin-bump playbook. Nothing blocks Track A/C work.

Issue #1015, opened 2026-06-26
Dynamic table filters: follow-ups & roadmap (Phase 1)


kevkrist:
Tracking issue for follow-ups and known limitations of the Phase-1 dynamic table-filter pushdown. Design & rationale: `docs/super-sirius/dynamic-filters.md`.

### Follow-ups
Tracked as sub-issues of this umbrella (see the **Sub-issues** panel above). #1010 (extend pushdown to non-scan probes / SIP) is the highest priority.

### Known limitations (roadmap; documented in the design doc, not yet separate issues)
- Only `BUILD_PROBE` joins produce filters; extending to `STANDARD` joins needs partition-awareness and multi-batch handling.
- Only parquet scans are served as consumers.
- Only equality join keys produce filters (an inequality ANDed with an equality key still gets the equality filter).
- The zone-map applies only on the parquet `set_filter` path; the pinned/cached path applies membership filters only, so a zone-map there does nothing.

PR #1148, opened 2026-07-10
Enable dynamic table-filter pushdown on duckdb-native scans

bwyogatama:
## Why

Dynamic table-filter pushdown (Phase 1, #794) shipped with a parquet-only scan consumer — #1015 lists "Only parquet scans are served as consumers" as a known limitation. Plan-gen already attaches the `sirius_dynamic_filter_set` channel to **both** parquet and seq_scan scan nodes and the producer side is format-agnostic, so filters published for a duckdb-native probe scan were silently dropped at pipeline conversion. Native tables served from the pin-table GPU cache are the best case for this optimization (scan cost ≈ 0, all savings land on downstream partition/probe work), yet got none of it.

## What changed

- `sirius_physical_dynamic_filter` takes a `dynamic_filter_apply_mode` constructor parameter (defaulted; parquet call sites unchanged).
- `duckdb_native_ingestible_table_info` carries the channel, and the native ingestible ctor installs the channel's column_ids → output-position remap. Decode positions at or past the output arity are pure-filter columns that `post_filter_and_project` drops — they map to the sentinel so `push_filter` rejects filters nothing downstream could apply. The `source_ids` computation moved out of the static-filter guard (a probe target often has no static filters).
- Plan-gen's `wrap_table_scan_source` — the only scan-lowering path since #1147 — wires the channel and places a `DYNAMIC_FILTER` operator directly above the native scan, mirroring parquet. The operator is elided when no producing join wired the channel.
- The native operator runs in `include_ast_row_masks` mode: the native scan has no read-time dynamic path (no reader `set_filter`), so opted-in zone maps are evaluated row-wise alongside the membership masks, all behind the existing selectivity gate. Membership (IN-list / Bloom) works out of the box under the default-on master switch.
- Docs: `dynamic-filters.md` consumer scope + native consumer paragraph; `scan.md` pruning and pipeline notes.

Read-time row-group pruning against dynamic zone-map bounds (the native analog of parquet's reader pushdown) is deliberately out of scope — it needs host-visible bounds and row-group stats retained past prepare, for a filter kind that is off by default.

## Tests

- `test/cpp/scan/test_duckdb_native_dynamic_filter_remap.cpp` — remap install: projection reorder, pure-filter trailing column → sentinel, no-static-filters hoist, null channel.
- `test/cpp/pipeline/test_pipeline_dynamic_filter_native_shape.cpp` — a join over native tables wraps the probe scan in `DYNAMIC_FILTER`; absent for single-table scans and with the master switch off.
- `test/cpp/integration/test_gpu_execution_dynamic_filter_native.cpp` — GPU-vs-CPU over the SF1 native fixture: selective membership join, membership composed with a pure-filter probe column, opt-in zone map through the AST row-mask path.

## Verification

Verified on the rebased branch (2026-07-13):

- Clean build; `pre-commit run -a` passes (21 hooks).
- **Full `make test` (the CI default set): all 1718 test cases pass** (32.1M assertions).
- `[dynamic_filter]` 71/71 (295 assertions) — includes the plan-shape test (native scan wraps in `DYNAMIC_FILTER`; elided when no producer is wired or the master switch is off) and the native GPU-vs-CPU test.
- `[pin_table]` 5/5 (478 assertions) — includes `duckdb-native pin serves the cache under the tree pipeline build`.
- End-to-end SQL over native tables: GPU results bit-identical to CPU, post-decode apply observed in logs (`mode=include_ast_row_masks`, 3M → 296 rows), including the pinned-cache serving path and the no-producer / pushdown-off elision cases.
- Includes a follow-up cleanup commit — `make_gpu_scan_leaf<InfoT>` extraction deduping the `seq_scan`/parquet branches of `wrap_table_scan_source` (behavior-preserving) — rebuilt and re-verified against all of the above.

Running `[gpu_execution]` by tag directly also force-includes 17 Catch2 `[.]`-hidden tests that CI and `make test` exclude (16 `[integration_disabled]` tpch cases whose fixture yields a null registered context, 1 hidden hive `reversed column order` binder error). These are pre-existing — byte-identical with this change git-stashed away — and unrelated to the physical-plan path this PR touches.

## Notes

- Wiring the native consumer extends the accepted-latent #1123 exposure (pinned DuckDB pushes join filters below LIMIT/TOP_N; fixed upstream post-pin) to native-table queries — previously native scans were incidentally shielded by dropping the filters.
- When this lands, #1015's "Only parquet scans are served as consumers" bullet should be updated.
- #1147 (legacy converter deletion) landed first; on rebase the converter-side wiring reduced to the tree path, as anticipated.


src/planner/sirius_physical_plan_generator.cpp
```diff
@@ -233,10 +233,30 @@ void wrap_table_scan_source(
   duckdb::unique_ptr<sirius::op::sirius_physical_operator> leaf;
   bool replace_slot = false;
   if (fn == "seq_scan") {
-    auto info       = build_duckdb_native_table_info(scan, op_params, context);
+    auto info = build_duckdb_native_table_info(scan, op_params, context);
```
kevkrist: nit: This is almost identical to the "parquet" branch -- consider extracting redundant logic into `wrap_scan_with_dynamic_filter(leaf, scan, op_params, apply_mode, info->sirius_dynamic_filters` or something like that.

kevkrist (approved):
Great!

PR #1152, opened 2026-07-11
test: refectory TPC-H-over-S3 correctness test suite

ran-yuan-rui:
Run the canonical TPC-H Q1-Q22 over transparent s3:// parquet views and assert each result equals a separate local DuckDB-CPU oracle, executed GPU-only with zero fallback (verified per query). 

Two scales: a tiny variant over the 8 already-uploaded fixtures runs in `make s3-test`, and an SF1 variant generates and uploads an 8-table SF1 fixture (gated SIRIUS_TEST_S3_TPCH) and runs in a new `make s3-tpch` target and the large gate. A per-query wall-clock benchmark records an S3 arm against a local SF1 arm (gated SIRIUS_BENCH_S3_TPCH), excluded from the routine bench.

The Q1-Q22 texts move into a shared test/cpp/utils/tpch_queries.hpp so the local suite and the S3 tier query from one source; the local suite is unchanged (22 cases, 18491 assertions).

make s3-test 49->50; make s3-tpch runs both scales; routine make s3-bench stays at 2 cases.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -693,7 +693,8 @@ if(SIRIUS_BUILD_S3_TESTS)
     test/cpp/io/s3/test_sirius_httpfs.cpp
     test/cpp/io/rest/test_rest_ioctx_integration.cpp
     test/cpp/scan_manager/test_describe_parquet_s3.cpp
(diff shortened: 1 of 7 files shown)


aminaramoon (approved):
I like the cleanup

PR #1155, opened 2026-07-11
Fail plan-tree-shape tests loudly on planner errors (#1147 follow-up)

bwyogatama:
## Problem

Review feedback on #1147 (P2): the plan-tree-shape suite silently passes on planner failures.

`generate_sirius_plan()` in `test/cpp/planner/test_plan_tree_shape.cpp` catches every `duckdb::InternalException` and returns null, and all nine test paths then `WARN("Plan generation failed; skipping")` and return. A parser, optimizer, resolver, or tree-rewrite regression would therefore turn the entire replacement suite into passing "skips".

## Change (test-only)

- `generate_sirius_plan()` no longer swallows `InternalException`: the remaining `catch (...)` still rolls back the transaction and restores the disabled-optimizer set, then rethrows, so the test fails with the original error.
- All nine `if (!plan) { WARN(...); return; }` skip paths are removed.

These fixtures are supported in CI, so planner failures now fail the tests loudly instead of vanishing.

The remaining #1147 review items (CPU-source plan rejection, raw-schedule determinism comparison, parquet/hive schedule sweeps) come in separate follow-up PRs.

### DIFF
--- a/test/cpp/planner/test_plan_tree_shape.cpp
+++ b/test/cpp/planner/test_plan_tree_shape.cpp
@@ -83,7 +83,10 @@ class scoped_temp_db_path {
   std::string _path;
 };
 
-/// Generate a Sirius physical plan from a SQL query string.
+/// Generate a Sirius physical plan from a SQL query string. Throws on any parse / plan /
+/// optimize / generation failure (after rolling back and restoring the optimizer settings):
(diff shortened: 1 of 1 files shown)


bwyogatama:
Superseded by #1156, which carries this same commit plus the other three #1147 review items as one four-commit PR.

PR #1157, opened 2026-07-12
feat: perf improvement - configurable REST bounce grain

ran-yuan-rui:
On large scans (SF10–SF100 parquet over S3), the REST device-read path splits everyread into windows of `bounce_block_size`, and each window costs one Range GET + one H2D copy + one slot lifecycle (SigV4 signing, curl setup, enqueue/completion/recycle, CUDA event). The grain was hard-wired to the host block size (1 MiB) — the config key
parsed but was silently overwritten — so a ~1 GB scan became **1467 GETs**, and per-request fixed cost dominated. Raising concurrency does not help: fixed cost × request count is unchanged, connections only hide latency (measured flat from 32 to 2048 slots).

## What

**1. `rest.bounce_block_size` is now honored** (whole multiple of the host block size, ≤ 1 GiB per slot). Default stays `0 → host block size` (1 MiB) — behavior and memory byte-identical for every existing config; larger grains are explicit opt-in.

**2. Bounce staging memory: before and after**

**Before**
Each REST reactor allocated one fixed-size host-pool block per curl slot:

```text
staging bytes = reactors × max_connections × host block size
```
With the default 1 MiB host block, every device read was split into 1 MiB windows. Although `rest.bounce_block_size` was parsed, it was overwritten by the host block size. A larger slot could not reuse multiple slab blocks: `allocate_multiple_blocks()` returns independent, potentially non-contiguous blocks, while curl requires each slot to be one contiguous buffer.
**After**
The default path remains unchanged:

```text
grain == host block size
→ one existing FSMR slab block per slot
```
An explicitly larger grain uses a budget-accounted contiguous span:

```text
total = reactors × max_connections × grain
FSMR reserve(total)
upstream.allocate(total)      # one contiguous pinned region
reactor r → fixed slice
slot i    → slice + i × grain
```

The span is allocated once when the REST ioctx starts and reused by all requests; there is no per-GET staging allocation. A slot remains occupied until its asynchronous H2D copy completes, then returns to the slot pool. Retrying requests release their slot during backoff and rebind to an available slot on the next attempt. Memory remains bounded by the selected HOST resource budget, a 1 GiB per-slot limit, and a 2 GiB whole-pool ceiling. The physical span is released before its FSMR reservation during teardown.

The existing default footprint remains 32 MiB:
```text
2 reactors × 16 connections × 1 MiB = 32 MiB
```
Opting into a 4 MiB grain with the same geometry uses 128 MiB:
```text
2 reactors × 16 connections × 4 MiB = 128 MiB
```

## Results

SF10 lineitem, local MinIO, 5 fresh-process samples/point, payload fixed at
1,062,080,163 B. Constant 32 MiB of bounce memory in every row — trading connections
for grain:

| Layout | Grain | GETs | Median | vs baseline |
|---|---:|---:|---:|---:|
| 1r × 32 conn | 1 MiB | 1467 | 0.907 GB/s (CV 3.7%) | — |
| 4r × 2 conn | 4 MiB | 491 | 1.271 GB/s (CV 1.2%) | **+40%** |
| 1r × 4 conn | 8 MiB | 491 | **1.447 GB/s** (CV 0.9%) | **+60%** |

Real EC2-FS3 validation is still pending. Higher per-request latency makes request-count reduction promising, but the magnitude and optimal connection/grain geometry must be measured on EC2.

Not a universal win: small objects or low-parallelism reads may not fill large slots — which is why the default is unchanged and the knob is opt-in.

### DIFF
--- a/src/include/io/rest/config.hpp
+++ b/src/include/io/rest/config.hpp
@@ -53,37 +53,28 @@ struct config {
   std::size_t max_read_split{16};
 
   /// Bounce-slot size (bytes) for the reactor-staged device path — also the
-  /// window grain @c prep_device_rx_request splits device reads by (the static
-  /// prep needs this size without access to the live staging resource, which
-  /// lives on the @c reactor_context).  Nonzero is authoritative and must be a
-  /// whole multiple of (and at least) the host staging block size — validated
-  /// at datasource construction and in the reactor constructor.  Zero means
-  /// auto: the factory substitutes the staging resource's block size, or
-  /// keeps 0 (device reads disabled) when no staging resource exists.  Values
-  /// above @c max_bounce_block_size are rejected outright.  A grain larger
-  /// than one staging block switches the bounce pool to one contiguous pinned
-  /// span BOOKED against the host staging budget (FSMR reservation, stricter
-  /// reservation-limit admission than the block path's capacity admission).
+  /// window grain @c prep_device_rx_request splits device reads by (the
+  /// static prep cannot reach the live staging resource).  Nonzero is
+  /// authoritative (whole multiple of the host block size,
+  /// <= @c max_bounce_block_size); a larger-than-block grain stages through
+  /// one budget-booked contiguous span instead of per-slot staging blocks.
+  /// 0 = auto: the host block size, or disabled when no staging resource
+  /// exists.
   std::size_t bounce_block_size{0};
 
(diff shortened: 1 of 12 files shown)


aminaramoon:
bounce block size is just htere, so when we prepare the requests we dont need to have a cucascade memory resource. otherwise it is set to what cucascade provides. there are other configs that allows merging multiple segments into 1. request without this PR. I am closing the PR

PR #1163, opened 2026-07-13
Use caller reservations in HOST representation conversions

### LINKED ISSUE #566: add missing host reservations

Jedi18:
## Summary
- Adopt cuCascade’s reservation-aware converter API (`convert` / `convert_to` / `clone_to` overloads that take `memory::reservation&`) so HOST allocations draw down a caller-owned reservation instead of committing a second copy of the same capacity.
- Update Sirius call sites that convert/clone into HOST (downgrade, host materialization in `lock_or_prepare_batch`, result collector, pin-table) and Sirius-registered parquet converters to accept and forward the reservation.
- Keep the `memory_space*` overloads as the no-reservation path (GPU/DISK today, probes, and fallback when `make_reservation_or_null` fails).
## Depends on
~cuCascade PR https://github.com/NVIDIA/cuCascade/pull/152. Once that is merged I'll bump up cuCascade in this PR.~
The PR is merged in cuCascade, this PR bumps up cuCascade

Closes https://github.com/sirius-db/sirius/issues/566

### DIFF
--- a/src/include/pipeline/batch_lock_utils.hpp
+++ b/src/include/pipeline/batch_lock_utils.hpp
@@ -158,6 +158,11 @@ inline std::optional<cucascade::read_only_data_batch> lock_or_prepare_batch(
           mut_accessor.convert_to<cucascade::host_data_representation>(
             registry, *host_reservation, stream);
         } else {
+          SIRIUS_LOG_WARN(
+            "lock_or_prepare_batch: host reservation failed for batch {} ({} bytes) — "
+            "proceeding without reservation, converter may OOM",
+            mut_accessor.get_batch_id(),
+            mut_accessor.get_data()->get_size_in_bytes());
           mut_accessor.convert_to<cucascade::host_data_representation>(
             registry, target_space, stream);
         }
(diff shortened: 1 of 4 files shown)


src/include/pipeline/batch_lock_utils.hpp
```diff
@@ -151,8 +151,16 @@ inline std::optional<cucascade::read_only_data_batch> lock_or_prepare_batch(
       // Same non-atomic-upgrade race as the GPU arm: skip if already in the target space.
       if (mut_accessor.get_memory_space() == nullptr ||
           mut_accessor.get_memory_space()->get_id() != target_space->get_id()) {
-        mut_accessor.convert_to<cucascade::host_data_representation>(
-          registry, target_space, stream);
+        auto host_reservation =
```
kevkrist, line 154: If the reservation fails, should we log it? In case the failed-reservation-converter OOMs.

kevkrist (approved):

src/include/pipeline/batch_lock_utils.hpp
9prady9: +1

9prady9 (approved):
lg

src/include/pipeline/batch_lock_utils.hpp
Jedi18: Agreed, I've changed this and other spots to log when reservation fails

PR #1168, opened 2026-07-13
feat(op): streaming sink operator (#837)

aocsa:
PR 2 of 2 for #836 (source: #1094). Adds the streaming sink operator. 

## What

- `sirius::op::sirius_physical_streaming_sink` — boundary operator (source+sink) heading its own pipeline: pops batches from its input port, registers them in the output repository, pushes `{batch_id, size_bytes}` handles to the exchange channel; closes the channel at pipeline finish.

## Design notes

- Zero-copy: the repository stays owner of record; no host clone, no `ColumnDataCollection` — queued batches remain idle and spillable.
- Backpressure is a task-creation condition: a full channel makes the hint report not-ready; `sink()` on a full channel parks the handle in a pending queue and never blocks a worker.
- EOS: finalize flushes what fits and closes immediately, or defers the close until consumer pops drain the pending queue.
- Known gap, deferred to #839: sink re-arm on consumer pop is owned by the session wiring, not the operator.

## Testing

- `[streaming_sink]` (GPU): hint/admission gating, zero-copy identity, full-channel fallback, flush-then-close EOS (immediate/deferred/empty), spill-while-queued, upstream-operator integration, mid-stream backpressure, concurrency conservation, input validation.

### DIFF
--- a/docs/super-sirius/operators.md
+++ b/docs/super-sirius/operators.md
@@ -96,6 +96,47 @@ session, task creation must be gated on in-flight work (e.g. counting via the ch
 hook and the task-completion path) so a fast producer cannot accumulate an arbitrarily large GPU
 backlog behind a nominally bounded channel.
 
+### `sirius_physical_streaming_sink` — `STREAMING_SINK`
+**File:** `src/include/op/sirius_physical_streaming_sink.hpp`
+
+Sink operator that marks the top boundary of an intermediate pipeline fragment. It registers each
+incoming `cucascade::data_batch` in a `cucascade::shared_data_repository` (making it spill-visible
+to the downgrade executor) and then pushes an `exec::exchange_batch_handle` (batch-id + size) to a
+bounded `exec::exchange_channel`. Used only when a fragment's output must be forwarded to another
+node over exchange; a leaf fragment keeps its normal `RESULT_COLLECTOR` sink.
+
+Key design invariants:
+- **Zero-copy emission**: the batch is registered in the output repository first, then only a
+  lightweight handle (batch-id + size) is pushed to the channel — no host clone, no
+  `ColumnDataCollection` materialization.
+- **CONCAT shape**: `is_source() && is_sink()` are both `true`. The operator heads its own
+  single-operator pipeline fed via an `"input"` port; the upstream pipeline pushes batches into
+  that port's repository via the base-class `sink()` → `push_data_batch()` path.
+- **Backpressure as admission control**: if the channel is full or `_pending` is non-empty,
+  `get_next_task_hint()` and `get_next_task_input_data()` return `WAITING{nullptr}` / `nullptr`
+  respectively, preventing new tasks from being created. Worker threads never block.
+- **FIFO `_pending` queue**: when `try_push` fails (channel full), the handle is appended to
+  `_pending`. Subsequent calls to `get_next_task_hint()` or `sink()` call `try_flush_pending()`
+  first, draining `_pending` in order before any new handles are attempted.
+- **Flush-then-close EOS**: `on_finalize_operator()` flushes what fits; if `_pending` is empty
+  it closes the channel immediately; otherwise it sets `_close_when_flushed` and defers the
+  close to the `try_flush_pending()` call triggered by the next consumer pop.
+- **Locking discipline**: whenever `_pending_lock` and the channel's internal mutex are both
+  needed, `_pending_lock` is always acquired first (consistent ordering prevents deadlock);
+  the base-class `lock` guards `ports` and is never held simultaneously with `_pending_lock`.
+- `no_history_peak_memory_estimate()` returns `stats.bytes` (no extra allocation).
+
+Hint table:
+
+| State | `get_next_task_hint()` |
+|---|---|
+| `_pending` non-empty or channel full | `WAITING{nullptr}` — stalled on consumer |
+| input port repo non-empty, channel has room | `READY{this}` |
+| port empty, upstream pipeline finished | `std::nullopt` — EOS |
+| port empty, upstream still running | `WAITING{upstream_op}` — propagate hint up the chain |
+
+**Consumer contract**: pop a handle from the channel, then call `output_repository->pop_data_batch_by_id(handle.batch_id)` to take ownership of the batch. The `try_flush_pending()` flush is triggered by `get_next_task_hint()` (called by the task-creation loop on every iteration), so consumer pops indirectly drive the pending drain.
+
 ### `sirius_physical_dummy_scan` — `DUMMY_SCAN`
 **File:** `src/include/op/sirius_physical_dummy_scan.hpp`
 
@@ -345,6 +386,7 @@ After pipeline finalization, `source` and `sink` are just aliases for the first 
 |----------|----------|-----------|
 | GPU_SCAN | Scan | Unified GPU scan source served by `sirius_scan_manager` via a per-format `gpu_ingestible` |
 | STREAMING_SOURCE | Scan | Exchange-input source; pulls batch handles from `exchange_channel`, resolves via `shared_data_repository` |
+| STREAMING_SINK | Pipeline | Exchange-output sink (CONCAT shape); registers batches in `shared_data_repository`, pushes handles to `exchange_channel` with backpressure |
 | DUMMY_SCAN | Scan | Generates 1 row |
 | COLUMN_DATA_SCAN | Scan | Reads ColumnDataCollection |
 | FILTER | Relational | `expression_evaluator::select()` |
--- a/docs/super-sirius/operators.md
+++ b/docs/super-sirius/operators.md
@@ -115,18 +115,20 @@ Key design invariants:
 - **Backpressure as admission control**: if the channel is full or `_pending` is non-empty,
   `get_next_task_hint()` and `get_next_task_input_data()` return `WAITING{nullptr}` / `nullptr`
   respectively, preventing new tasks from being created. Worker threads never block.
-- **FIFO `_pending` overflow buffer (bounded)**: when `try_push` fails (channel full), the handle
-  is appended to `_pending`. Subsequent calls to `get_next_task_hint()` or `sink()` call
+- **FIFO `_pending` overflow buffer**: when `try_push` fails (channel full), the handle is
+  appended to `_pending`. Subsequent calls to `get_next_task_hint()` or `sink()` call
   `try_flush_pending()` first, draining `_pending` in order before any new handles are attempted.
-  `_pending` is not a second unbounded queue: because admission control refuses new sink tasks
-  while `_pending` is non-empty, its depth is bounded by the operator's in-flight task concurrency,
-  never by the input size.
+  Admission control stops new pulls while `_pending` is non-empty or the channel is full, but the
+  task creator can pull an entire port before any created task executes, so `_pending` depth is
+  bounded only once task creation is paced against completion — the session wiring's job (#839),
+  the same caveat as the streaming source above. Parked batches remain registered, idle, and
+  spill-visible (accounted, spillable state).
 - **Flush-then-close EOS**: `on_finalize_operator()` flushes what fits; if `_pending` is empty
-  it closes the channel immediately; otherwise it sets `_close_when_flushed` and defers the
-  close to the `try_flush_pending()` call triggered by the next consumer pop. Finalize sets a
-  `_closing` flag under `_pending_lock`; a `sink()` that observes it (a batch arriving after
-  finalize — which the pipeline never does) throws rather than stranding the handle behind the
-  closed channel.
+  it closes the channel immediately; otherwise it sets `_close_when_flushed` and the close
+  happens on a later `try_flush_pending()` call. Nothing polls that after finalize — the
+  consumer/session must call `try_flush_pending()` after pops (or wire `on_pop` to it, #839) or
+  deferred EOS stalls. Finalize also sets `_closing` under `_pending_lock`; a `sink()` that
+  observes it throws rather than stranding the handle behind the closed channel.
 - **Locking discipline**: whenever `_pending_lock` and the channel's internal mutex are both
   needed, `_pending_lock` is always acquired first (consistent ordering prevents deadlock);
   the base-class `lock` guards `ports` and is never held simultaneously with `_pending_lock`.
@@ -141,7 +143,7 @@ Hint table:
 | port empty, upstream pipeline finished | `std::nullopt` — EOS |
 | port empty, upstream still running | `WAITING{upstream_op}` — propagate hint up the chain |
 
-**Consumer contract**: pop a handle from the channel, then call `output_repository->pop_data_batch_by_id(handle.batch_id)` to take ownership of the batch. The `try_flush_pending()` flush is triggered by `get_next_task_hint()` (called by the task-creation loop on every iteration), so consumer pops indirectly drive the pending drain.
+**Consumer contract**: pop a handle from the channel, then call `output_repository->pop_data_batch_by_id(handle.batch_id)` to take ownership of the batch. Hint polling drives the pending drain only while the sink's pipeline is still running; after finalize the consumer/session must call `try_flush_pending()` after pops (or wire `exchange_channel::on_pop` to it) so `_pending` drains and the deferred close completes. Channel callbacks run synchronously on the calling thread — they must post/schedule work, never re-enter an operator or take pipeline locks.
 
 ### `sirius_physical_dummy_scan` — `DUMMY_SCAN`
 **File:** `src/include/op/sirius_physical_dummy_scan.hpp`
--- a/docs/super-sirius/operators.md
+++ b/docs/super-sirius/operators.md
@@ -111,7 +111,9 @@ Key design invariants:
   `ColumnDataCollection` materialization.
 - **CONCAT shape**: `is_source() && is_sink()` are both `true`. The operator heads its own
   single-operator pipeline fed via an `"input"` port; the upstream pipeline pushes batches into
-  that port's repository via the base-class `sink()` → `push_data_batch()` path.
+  that port's repository via the base-class `sink()` → `push_data_batch()` path. Being the head,
+  it is simultaneously the *final* operator of that pipeline (`publish_output()` calls its
+  `sink()`) — the same head-and-terminal shape as `RESULT_COLLECTOR` and `MERGE_GROUP_BY`.
 - **Backpressure as admission control**: if the channel is full or `_pending` is non-empty,
   `get_next_task_hint()` and `get_next_task_input_data()` return `WAITING{nullptr}` / `nullptr`
   respectively, preventing new tasks from being created. Worker threads never block.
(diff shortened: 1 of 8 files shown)


src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,187 @@
...
+  // Hold the lock across flush + close decision so they are atomic: close now if the backlog
+  // drained, otherwise defer the close to whichever try_flush_pending() delivers the last handle.
+  std::lock_guard<std::mutex> lk(_pending_lock);
+  flush_pending_locked();
+  if (_pending.empty()) {
+    _output_channel->close();
+  } else {
+    _close_when_flushed.store(true, std::memory_order_release);
+  }
+}
```
mbrobbel, line 209: What happens if a `sink()` is running concurrently?

src/include/op/sirius_physical_streaming_sink.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+ private:
+  /// Push queued handles into the channel in FIFO order until it rejects one (full).
+  /// Caller must hold _pending_lock. Does not close the channel — callers decide that.
+  void flush_pending_locked();
+
+  std::shared_ptr<exec::exchange_channel> _output_channel;
+  std::shared_ptr<cucascade::shared_data_repository> _output_repository;
+  std::mutex _pending_lock;
+  std::deque<exec::exchange_batch_handle> _pending;
```
mbrobbel, line 104: Following https://github.com/mbrobbel/sirius/blob/c8593116000a5fab2228b25d9d937025e526adbe/experimental/starrocks/docs/exchange-design.md?plain=1#L224-L226, this should be bounded.

src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,187 @@
...
+    // The lock is released before add_data_batch so the repo receives the batch in idle state.
+    std::size_t size_bytes = 0;
+    {
+      auto ro = batch->to_read_only();
+      if (const auto* d = ro.get_data()) { size_bytes = d->get_size_in_bytes(); }
+    }
+
+    // Register first: the repo becomes the owner of record and the batch is spill-visible
+    // before any channel push, so _pending handles always point at accountable batches.
+    _output_repository->add_data_batch(batch);
```
mbrobbel: There is a `size` race here.

src/include/op/sirius_physical_streaming_sink.hpp
9prady9: If there is a additional bound-marker/extent-indication, perhaps it would have been fine. Better to have a static sized container ?

src/op/sirius_physical_streaming_sink.cpp
9prady9: I think our current pipeline usage pattern prevents it, but it is always better to avoid relying on that and enforce that check here.

src/op/sirius_physical_streaming_sink.cpp
9prady9, line 1: nit: given both streaming source and sink operators are in the code base, might be worth taking a long if there are common utilities for both and reduce code duplication.

src/op/sirius_physical_streaming_sink.cpp
aocsa: Good catch. They already share `_pending_lock`, but the ordering was unsafe: if finalize closed the channel first, a later `sink()`'s `try_push` would fail and park the handle in `_pending` where nothing ever flushes it — a stranded batch after EOS. The pipeline prevents this today.

Fixed: finalize sets a `_closing` flag under `_pending_lock`; `sink()` checks it under the same lock and throws. `add_data_batch` moved inside that hold so check → register → push is one atomic section. Added a test: sink-after-finalize throws and leaves no orphan.

src/include/op/sirius_physical_streaming_sink.hpp
aocsa: `_pending` isn't a second unbounded queue: admission control refuses new sink tasks while it's non-empty, so its depth is bounded by in-flight task concurrency, never by input size. Documented that bound in the header and operators.md.

Kept `std::deque` over a fixed-size container deliberately: a hard cap would force `sink()` to block or drop on overflow, breaking the never-block / never-drop invariants. Happy to switch to a fixed ring buffer as a follow-up if you'd prefer a structural cap.

src/op/sirius_physical_streaming_sink.cpp
aocsa: I don't think this can corrupt anything, tell me if you saw a specific failure mode. 

The size is read under the batch's shared lock, which excludes a concurrent spill (downgrade needs the exclusive lock), so the read is race-free. The channel then adds and subtracts that *same stored value* — the consumer never re-derives size — so the byte accounting stays balanced no matter what happens to the batch's tier afterwards. A later spill only makes the bound conservative, never wrong.

There's no tier-invariant size to record instead (every size getter is tier-local), so I documented it as a registration-time snapshot.

src/op/sirius_physical_streaming_sink.cpp
aocsa: Fair : they share the constructor null-checks, the pass-through `execute()`, and `no_history_peak_memory_estimate()`. I'd rather factor those out once the sink grows its partitioning/coalescing state (later PR) so we don't dedup code that's about to diverge. Tracking it for that pass.

src/include/op/sirius_physical_streaming_sink.hpp
aocsa: Correction after auditing the task-creation path: admission alone doesn't bound `_pending`. The creation loop can pull an entire input port before any created task executes (task_creator.cpp:252; TODO at :247), and each of those tasks parks a handle once the channel fills — so the depth is bounded only when task creation is paced against completion. That pacing is #839's job, the same caveat the docs already state for the streaming source; I've corrected the header and operators.md accordingly.
  
Fix in this PR: corrected the docs/header to say that, and added a stalled-consumer test that shows the burst behavior (handles stay registered/spill-visible and all deliver on resume).

src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,201 @@
...
+}
+
+void sirius_physical_streaming_sink::try_flush_pending()
+{
+  std::lock_guard<std::mutex> lk(_pending_lock);
+  flush_pending_locked();
+  if (_pending.empty() && _close_when_flushed.load(std::memory_order_acquire)) {
+    _output_channel->close();
+  }
+}
```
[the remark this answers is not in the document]
aocsa: Fixed as suggested — try_flush_pending() now computes the close decision under `_pending_lock` and calls `close()` after releasing it (src/op/sirius_physical_streaming_sink.cpp:57-67). The reorder is safe: `_close_when_flushed` is only set by finalize, which also sets `_closing`, so a late `sink()` throws before parking a handle; `close()` is idempotent under concurrent flushes. Added a regression test that wires `on_close` to re-enter `try_flush_pending()` on the deferred-close path (test_physical_streaming_sink.cpp:640),  it deadlocks without this fix.

src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,201 @@
...
+  // decision or rejects its batch.
+  std::lock_guard<std::mutex> lk(_pending_lock);
+  _closing = true;
+  flush_pending_locked();
+  if (_pending.empty()) {
+    _output_channel->close();
+  } else {
+    _close_when_flushed.store(true, std::memory_order_release);
+  }
+}
```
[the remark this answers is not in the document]
aocsa: Fixed the same way — finalize decides under `_pending_lock` and closes outside it (src/op/sirius_physical_streaming_sink.cpp:189-208), with a re-entrancy regression test for this site (test_physical_streaming_sink.cpp:620). One layer remains that a sink-local fix can't reach: the pipeline calls `finalize_operator()` while holding `_status_mutex` (sirius_pipeline.cpp:388), so the on_close callback still runs under the child pipeline's status mutex. That's a pipeline-level pattern shared by all operators — tracking it as a follow-up for the #839 wiring, where we've also documented that channel callbacks must post/schedule rather than re-enter operators (exchange_channel.hpp:193-195).

docs/super-sirius/operators.md
```diff
@@ -96,6 +96,53 @@ session, task creation must be gated on in-flight work (e.g. counting via the ch
...
+Hint table:
+
+| State | `get_next_task_hint()` |
+|---|---|
+| `_pending` non-empty or channel full | `WAITING{nullptr}` — stalled on consumer |
+| input port repo non-empty, channel has room | `READY{this}` |
+| port empty, upstream pipeline finished | `std::nullopt` — EOS |
+| port empty, upstream still running | `WAITING{upstream_op}` — propagate hint up the chain |
+
+**Consumer contract**: pop a handle from the channel, then call `output_repository->pop_data_batch_by_id(handle.batch_id)` to take ownership of the batch. The `try_flush_pending()` flush is triggered by `get_next_task_hint()` (called by the task-creation loop on every iteration), so consumer pops indirectly drive the pending drain.
```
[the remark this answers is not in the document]
aocsa: fixed

aocsa:
Ready to review again, 

Pushed three follow-up commits for the latest review round:

- f2c46daa — fix(op): close the exchange channel outside `_pending_lock`. `close()` runs the on_close callback synchronously, so both close sites (`try_flush_pending()`, `on_finalize_operator()`) now decide under the lock and close after releasing it, removing the re-entrancy deadlock flagged.

- e0a19fc1 — docs(op): corrected the `_pending` bound claim: admission alone doesn't bound it — the creation loop can pull a whole port before any created task executes, so the bound depends on pacing task creation against completion (#839, same caveat as the streaming source).  

- b913de4e — test(op): re-entrancy regressions for both close sites (they deadlock without the fix) and a stalled-consumer test characterizing the backlog behavior until #839 paces creation.

Follow up draft PR here https://github.com/aocsa/sirius/pull/1/changes

docs/super-sirius/operators.md
```diff
@@ -345,6 +394,7 @@ After pipeline finalization, `source` and `sink` are just aliases for the first
 |----------|----------|-----------|
 | GPU_SCAN | Scan | Unified GPU scan source served by `sirius_scan_manager` via a per-format `gpu_ingestible` |
 | STREAMING_SOURCE | Scan | Exchange-input source; pulls batch handles from `exchange_channel`, resolves via `shared_data_repository` |
+| STREAMING_SINK | Pipeline | Exchange-output sink (CONCAT shape); registers batches in `shared_data_repository`, pushes handles to `exchange_channel` with backpressure |
```
mbrobbel: Maybe:
```suggestion
| STREAMING_SINK | Sink | Exchange-output sink (CONCAT shape); registers batches in `shared_data_repository`, pushes handles to `exchange_channel` with backpressure |
```
or 
```suggestion
| STREAMING_SINK | Result | Exchange-output sink (CONCAT shape); registers batches in `shared_data_repository`, pushes handles to `exchange_channel` with backpressure |
```

src/include/op/sirius_physical_streaming_sink.hpp
```diff
@@ -0,0 +1,105 @@
...
+  static constexpr std::string_view INPUT_PORT     = "input";
+
+  /// Throws invalid_input_exception when output_channel or output_repository is null.
+  sirius_physical_streaming_sink(
+    duckdb::vector<sirius::logical_type> types,
+    std::size_t estimated_cardinality,
+    std::shared_ptr<exec::exchange_channel> output_channel,
+    std::shared_ptr<cucascade::shared_data_repository> output_repository);
+
+  bool is_source() const override { return true; }
```
mbrobbel: I guess:
```suggestion
  bool is_source() const override { return false; }
```

9prady9: Yeah, I thought so too but docs explicitly say it is source and sink, is that intentional @aocsa ?

docs/super-sirius/operators.md
```diff
@@ -96,6 +96,55 @@ session, task creation must be gated on in-flight work (e.g. counting via the ch
...
+Hint table:
+
+| State | `get_next_task_hint()` |
+|---|---|
+| `_pending` non-empty or channel full | `WAITING{nullptr}` — stalled on consumer |
+| input port repo non-empty, channel has room | `READY{this}` |
+| port empty, upstream pipeline finished | `std::nullopt` — EOS |
+| port empty, upstream still running | `WAITING{upstream_op}` — propagate hint up the chain |
+
+**Consumer contract**: pop a handle from the channel, then call `output_repository->pop_data_batch_by_id(handle.batch_id)` to take ownership of the batch. Hint polling drives the pending drain only while the sink's pipeline is still running; after finalize the consumer/session must call `try_flush_pending()` after pops (or wire `exchange_channel::on_pop` to it) so `_pending` drains and the deferred close completes. Channel callbacks run synchronously on the calling thread — they must post/schedule work, never re-enter an operator or take pipeline locks.
```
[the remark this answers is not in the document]
aocsa: Good catch — reworded: `on_pop` should schedule a deferred `try_flush_pending()` call, not call it directly. This matches the callback contract (exchange_channel.hpp:193-195). The same phrasing was also in the EOS bullet (operators.md:129); fixed both places.

docs/super-sirius/operators.md
aocsa: Changed it to `Result` — the sink takes RESULT_COLLECTOR's role when a fragment's output goes over exchange (the #839 wiring mirrors RESULT_COLLECTOR). I applied it locally together with the other doc fixes instead of using the suggestion.

src/include/op/sirius_physical_streaming_sink.hpp
aocsa: Yes, intentional — it is the shape that [exchange-design.md](https://github.com/mbrobbel/sirius/blob/c8593116000a5fab2228b25d9d937025e526adbe/experimental/starrocks/docs/exchange-design.md?plain=1#L218) describes. In §3, the sink is port-fed and is "the final operator of its pipeline ... just like RESULT_COLLECTOR". In §4, backpressure works through the sink's own hint: when the output channel is full, get_next_task_hint() reports not-ready. 

~~The task creator only polls hints on pipeline heads, so the sink is the head of a single-operator pipeline, and it is also its last operator. A pipeline head must return is_source() == true, or sirius_pipeline::reset_source() throws (sirius_pipeline.cpp:99). The #839 wiring depends on this (set_pipeline_source, aocsa/sirius#1).~~


The two operators that the design doc names also return true for both: **RESULT_COLLECTOR** (sirius_physical_result_collector.hpp:58,:65) and **MERGE_GROUP_BY** (sirius_physical_grouped_aggregate_merge.hpp:114,:122). I added a comment at the override and a short note in operators.md, so the header now explains this directly.

mbrobbel: That's confusing to me. Should we revisit the logic on the task creator side?

aocsa:
Pushed two follow-up commits for the latest feedback comments. Quick summary:

Why sink() parks instead of blocking: it runs on a GPU worker thread, and those are a bounded pool. If a push blocks there, we can deadlock the pool whenever the consumer's pop also needs a worker — and with no cancellation on the channel yet (#839), a stalled consumer would strand the thread forever. So a refused handle just goes into _pending (tiny, 16 B each — the actual batches are already in the output repository and stay spillable).

Backpressure is admission control, not blocking: channel full or _pending non-empty → the hint says WAITING → no new sink tasks get created. Upstream output waits in the input port, still spillable. When the consumer pops, _pending drains in order and admission reopens. How deep _pending can get is bounded by tasks already in flight — pacing creation against completion is the session's job (#839).

Scope: this PR is phase 1 only — the plain 1:1 sink, one input, one channel, no partitioning. Phase 2 (GPU hash partitioning, per-destination channels, coalescing) comes as a follow-up PR once #841 provides the partition spec. The interface here was designed so phase 2 just adds on top; nothing in the scheduler or session contract changes.

src/include/op/sirius_physical_streaming_sink.hpp
aocsa: I re-verified instead of leaning on the RESULT_COLLECTOR convention.

There's no task-creator logic to revisit: it never checks is_source(). It polls whatever pipeline head it's handed, and head-ness is structural (operators[0]). The flag has exactly three readers in the codebase, and none runs at query time — the only real enforcement point, sirius_pipeline::reset_source(), has no live caller.

Confirmed empirically by flipping RESULT_COLLECTOR's is_source() to false: clean build, 1,754 unit tests pass, 22/22 TPC-H SF1 validate on GPU.

So I'll apply the suggestion, plus two follow-ons: update the header comment / operators.md that cite the old rationale, and note on #839 that its session wiring (which revives reset_source()) must key the head check on structure, not the flag — otherwise the flip throws there.

mbrobbel (approved):

src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+  }
+  if (_output_channel->full()) {
+    return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, nullptr};
+  }
+
+  std::lock_guard<std::mutex> lg(lock);
+  auto it = ports.find(std::string(INPUT_PORT));
+  if (it == ports.end() || !it->second || !it->second->repo) {
+    return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, nullptr};
```
wmalpica, line 89: This does not seem right. If you cant find the port, I think this would be an error not just waiting. The port should be determined and set during query graph instantiation. It would not be expected to change during runtime, therefore waiting does not seem like the right thing to do here

src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,211 @@
...
+  }
+  auto* port_ptr = it->second;
+
+  if (port_ptr->repo->total_size() > 0) {
+    return task_creation_hint{TaskCreationHint::READY, this};
+  }
+
+  bool upstream_finished = port_ptr->src_pipeline && port_ptr->src_pipeline->is_pipeline_finished();
```
wmalpica, line 97: NIT: there is an improbably and slight race condition here: you should check is_pipeline_finished first before checking total_size(). Technically when you check total_size it could be zero, the imediatelly you could get something before you check is_pipeline_finished. So do it like:

```
  bool upstream_finished = port_ptr->src_pipeline && port_ptr->src_pipeline->is_pipeline_finished();
  if (upstream_finished && port_ptr->repo->total_size() == 0) { return std::nullopt; }

  if (port_ptr->repo->total_size() > 0) {
    return task_creation_hint{TaskCreationHint::READY, this};
  }
```

src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+  std::lock_guard<std::mutex> lg(lock);
+  auto it = ports.find(std::string(INPUT_PORT));
+  if (it == ports.end() || !it->second || !it->second->repo) return nullptr;
+  auto* port_ptr = it->second;
+
+  auto batch_ids = port_ptr->repo->get_batch_ids(0);
+  if (batch_ids.empty()) return nullptr;
+
+  auto batch = port_ptr->repo->pop_data_batch_by_id(batch_ids[0], 0);
```
wmalpica, line 127: Here if you are just going to pop the next data_batch, then use repo->pop_next_data_batch

src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,211 @@
...
+    // the batch is neither orphaned in the repo nor stranded behind the closed channel.
+    if (_closing) {
+      throw sirius::internal_exception(
+        "sirius_physical_streaming_sink::sink() called after finalize — a task was still in "
+        "flight when the pipeline was declared finished");
+    }
+
+    // Register first: the repo becomes the owner of record and the batch is spill-visible
+    // before any channel push, so _pending handles always point at accountable batches.
+    _output_repository->add_data_batch(batch);
```
wmalpica, line 171: when or how or who empties this _output_repository?

wmalpica (changes requested):

PR #1173, opened 2026-07-13
feat(transparent): execution-time fallback to DuckDB CPU

bwyogatama:
## What

When a transparently-intercepted `SELECT` fails **during GPU execution** (not just at plan time), it now completes on DuckDB CPU instead of erroring — under the **same transaction and MVCC snapshot** as the failed GPU attempt.

## How

`OnFinalizePrepare` stashes DuckDB's CPU physical plan (previously discarded) inside `PhysicalSiriusExecution`. On a GPU execution failure, the operator runs that plan itself, from inside `GetData`, on a private `duckdb::Executor` bound to the same `ClientContext`. Reusing the same context (not a fresh connection) keeps the fallback on the same transaction, so it sees the transaction's own uncommitted writes and none of the commits made after it started. The GPU result is fully materialized before any row is emitted, so the fallback can't duplicate rows. No DuckDB submodule changes.

## Behavior

- Both plan-time and runtime fallback are gated by the `enable_duckdb_fallback` session setting (default true, `SET` per connection). When off, failures surface as errors instead of running on CPU.
- Runtime GPU failures no longer throw `InternalException`, which was invalidating the whole database. Errors are sanitized to `ExecutorException`.
- S3-reading queries have no CPU path and surface a clear error. User interrupts propagate.
- A `runtime_fallbacks` counter and a red console banner (shared with the legacy `CALL gpu_execution` path) record each fallback.

## Testing

`test_transparent_runtime_fallback.cpp` drives the runtime path via a test-only injection setting: basic fallback, MVCC same-snapshot (sees own uncommitted writes; ignores a concurrent commit), `enable_duckdb_fallback` gating for both runtime and plan-time, and the setting default/override. The shared `require_transparent_execution_delta` helper now also asserts `runtime_fallbacks`, so existing GPU tests confirm it stays zero.

## Notes

- SQL-level `PREPARE ... AS SELECT` / `EXECUTE` isn't intercepted (the SELECT-only statement-type gate), so runtime fallback doesn't apply there — a separate concern.
- Under query profiling, the nested CPU executor's profile is suppressed so it doesn't corrupt the outer query's profile.

### DIFF
--- a/src/include/sirius_context.hpp
+++ b/src/include/sirius_context.hpp
@@ -67,6 +67,10 @@ class SiriusContext : public ClientContextState {
     uint64_t successful_rebinds = 0;
     uint64_t fallbacks          = 0;
     uint64_t executions         = 0;
+    // GPU execution was attempted and failed at runtime, and the query completed
+    // via DuckDB CPU fallback (same transaction). Distinct from `fallbacks`, which
+    // counts plan-time (create_plan) fallbacks that never reached the GPU.
+    uint64_t runtime_fallbacks = 0;
   };
 
   SiriusContext();
@@ -301,6 +305,10 @@ class SiriusContext : public ClientContextState {
   /// \brief Record that a transparently rebound query actually executed through Sirius.
   void record_transparent_execution() noexcept;
 
+  /// \brief Record that a GPU execution failed at runtime and the query completed
+  /// via DuckDB CPU fallback (same transaction).
+  void record_transparent_runtime_fallback() noexcept;
+
  private:
   void throw_if_not_initialized() const;
   void acquire_query_lifecycle_slot();
@@ -370,6 +378,7 @@ class SiriusContext : public ClientContextState {
   std::atomic<uint64_t> transparent_rebind_success_count_{0};
   std::atomic<uint64_t> transparent_fallback_count_{0};
   std::atomic<uint64_t> transparent_execution_count_{0};
+  std::atomic<uint64_t> transparent_runtime_fallback_count_{0};
 };
 
 /// todo(amin): when duckdb is updated, we need to enable OnExtensionLoaded to support sirius
@@ -402,5 +411,18 @@ class SiriusContextExtensionCallback : public ExtensionCallback {
   sirius::sirius_config config_;
   duckdb::shared_ptr<SiriusContext> context_;
 };
+
+/// \brief Read the per-session `enable_duckdb_fallback` setting (default true).
+///
+/// Gates both plan-time and runtime fallback from GPU to DuckDB CPU. Set per
+/// connection via `SET enable_duckdb_fallback = ...`.
+bool duckdb_fallback_enabled(ClientContext& context);
+
+/// \brief Print the "GPU execution failed, falling back to DuckDB" banner.
+///
+/// Written to stdout in red (ANSI) when stdout is a TTY, plain text otherwise so
+/// piped/redirected output is not corrupted. Shared by the transparent runtime
+/// fallback and the legacy gpu_execution() CALL path so the message stays in sync.
+void print_cpu_fallback_banner();
 
 }  // namespace duckdb
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -592,6 +592,7 @@ set(TEST_SOURCES
     test/cpp/integration/test_pin_table_column_order.cpp
     test/cpp/integration/test_pin_table_mvcc_foundation.cpp
     test/cpp/integration/test_table_gpu_cache_warm_mgpu.cpp
+    test/cpp/integration/test_transparent_runtime_fallback.cpp
     test/cpp/io/s3/test_sigv4.cpp
     test/cpp/io/s3/test_s3_default_visibility_guard.cpp
     test/cpp/io/s3/test_sirius_sigv4_authorizer.cpp
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -429,7 +429,7 @@ SET enable_dynamic_zone_map_filter = false;
 |----------|---------|-------------|
 | `print_gpu_table_max_rows` | - | Max rows to print in debug output |
 | `enable_fallback_check` | - | Enable fallback validation |
-| `enable_duckdb_fallback` | true | Fall back to DuckDB CPU on Sirius errors. Matches the legacy `gpu_processing` path. Set to `false` to surface Sirius errors instead of silently falling back. |
+| `enable_duckdb_fallback` | true | Fall back to DuckDB CPU execution on Sirius errors. Gates both plan-time fallback (unsupported operator/type) and runtime fallback (GPU execution failure) on the transparent path, plus the legacy `CALL gpu_execution(...)` path. Set to `false` to surface Sirius errors instead of falling back. The default is seeded from the `sirius.enable_duckdb_fallback` YAML key; a per-session `SET` overrides it. |
 | `enable_regex_jit_impl` | - | Use JIT regex implementation |
 
 ## Legacy Config Flags
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -429,7 +429,7 @@ SET enable_dynamic_zone_map_filter = false;
 |----------|---------|-------------|
 | `print_gpu_table_max_rows` | - | Max rows to print in debug output |
 | `enable_fallback_check` | - | Enable fallback validation |
-| `enable_duckdb_fallback` | true | Fall back to DuckDB CPU execution on Sirius errors. Gates both plan-time fallback (unsupported operator/type) and runtime fallback (GPU execution failure) on the transparent path, plus the legacy `CALL gpu_execution(...)` path. Set to `false` to surface Sirius errors instead of falling back. The default is seeded from the `sirius.enable_duckdb_fallback` YAML key; a per-session `SET` overrides it. |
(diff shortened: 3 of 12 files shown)


src/include/sirius_context.hpp
```diff
@@ -396,11 +405,30 @@ class SiriusContextExtensionCallback : public ExtensionCallback {
...
   sirius::sirius_config config_;
   duckdb::shared_ptr<SiriusContext> context_;
 };
 
+/// \brief Read the per-session `enable_duckdb_fallback` setting (default true).
+///
+/// Gates both plan-time and runtime fallback from GPU to DuckDB CPU. The default is
+/// seeded from the Sirius YAML config (`sirius.enable_duckdb_fallback`); a
+/// per-connection `SET` overrides it.
+bool duckdb_fallback_enabled(ClientContext& context);
```
mbrobbel, line 419: Could/should this be a member of `ClientContext`?

src/include/sirius_extension.hpp
```diff
@@ -66,7 +66,7 @@ class SiriusExtension : public Extension {
   void Load(ExtensionLoader& loader) override;
   std::string Name() override;
   std::string Version() const override;
-  static void InitialGPUConfigs(DBConfig& db);
+  static void InitialGPUConfigs(DBConfig& db, bool enable_duckdb_fallback_default);
```
mbrobbel: Why is this passed as an explicit additional arg?

mbrobbel (commented):
There are a few more things (identified with agent) we can catch early during planning to avoid runtime fallback. For a follow-up PR:

  1. Non-constant `LIMIT`/`OFFSET`

  The planner only rejects percentage limits, then constructs a streaming-limit operator for everything else. The operator later throws during execute() for expression-valued limits or offsets (`src/op/sirius_physical_limit.cpp:68`).

  This should be rejected directly in `src/planner/sirius_plan_limit.cpp:47`.

  2. `ORDER BY` and `TOP N` expression shape

  These runtime checks use information already present in the logical plan:

  - `ORDER BY` expression must be `BOUND_REF` (`src/op/sirius_physical_order.cpp:58`).
  - `TopN` must have an ordering key.
  - `TopN` expressions must be `BOUND_REF`.
  - `TopN` column indexes must be in range (`src/op/sirius_physical_top_n.cpp:54`).

  The same bound-reference check is repeated at runtime by the inserted sort-sample, sort-partition, and merge-sort operators. It should be checked once in `create_plan(LogicalOrder/LogicalTopN)` before the nodes are moved.

  3. Join capability checks

  Several static join properties are still rejected inside execution:

  - Unsupported hash-join type (`src/op/sirius_physical_hash_join.cpp:1382`).
  - Join types unsupported by a particular `BUILD_PROBE` or `MIXED_JOIN` mode.
  - Unsupported nested-loop comparison/join types (`src/op/sirius_physical_nested_loop_join.cpp:510`).
  - Mixed-join predicate cannot be translated to a cuDF AST (`src/op/sirius_physical_hash_join.cpp:1144`).

  The join type and expression tree are available during planning. AST translation currently needs a stream/device for literals, but a device-free can_translate() walk could validate the expression shape without constructing cuDF objects.

  4. Inserted partition/sort operator modes

  The pipeline rewrite can construct configurations that only fail when executed:

  - `RANGE` or `CUSTOM` partitioning is not implemented.
  - Partition count was not assigned.
  - Sort helper received a non-reference order expression.

  These are properties of the rewritten physical plan and should be covered by `plan->verify()` after `insert_gpu_pipeline_operators()`, rather than discovered by a worker.

  5. Native DuckDB segment viability

  The first native-table viability phase already runs during GPU plan construction and catches unsupported projected types and invalid partition metadata (`src/op/scan/duckdb_native_gpu_ingestible.cpp:195`).

  But the second phase runs asynchronously in `next_split_provider()` and can reject:

  - Unsupported data or validity compression.
  - `VARCHAR` constant compression.
  - Missing max-string-length statistics.
  - Overflow-block strings.
  - Unsupported nested storage layout.
  - `ARRAY` child-element counts over cuDF limits.
  - A split exceeding `cudf::size_type`.

  Those checks are performed by `walk_duckdb_native_row_group_range()` after GPU execution has started (`src/op/scan/duckdb_native_gpu_ingestible.cpp:264`). They could be run eagerly during planning and the resulting range metadata cached for execution.

  This is probably the highest-value change because these are genuine GPU capability failures, not transient runtime failures.

  6. Parquet backend and schema validation

  Currently runtime metadata tasks can discover:

  - No registered backend supports a known URI.
  - A projected column is absent from a parquet file (`src/op/scan/parquet_gpu_ingestible.cpp:438`).
  - Potentially incompatible per-file schema/metadata.

  The paths and requested column names are known during planning, but checking the schema requires reading every footer. I would classify this as optional “metadata preflight,” since putting remote/footer I/O directly in logical planning can make prepare unexpectedly
  expensive.

  7. Required engine configuration

  These currently surface during execution but are knowable before task launch:

  - No HOST memory space configured for native scan staging.
  - No HOST memory space available for result collection (`src/op/sirius_physical_result_collector.cpp:149`).
  - No scan manager or datasource registry for the planned scan.
  - Known URI scheme has no configured backend.

  Check these once when deciding whether the physical plan is GPU-eligible.

bwyogatama:
> There are a few more things (identified with agent) we can catch early during planning to avoid runtime fallback. For a follow-up PR:
> 
> 1. Non-constant `LIMIT`/`OFFSET`
> 
> The planner only rejects percentage limits, then constructs a streaming-limit operator for everything else. The operator later throws during execute() for expression-valued limits or offsets (`src/op/sirius_physical_limit.cpp:68`).
> 
> This should be rejected directly in `src/planner/sirius_plan_limit.cpp:47`.
> 
> 2. `ORDER BY` and `TOP N` expression shape
> 
> These runtime checks use information already present in the logical plan:
> 
> * `ORDER BY` expression must be `BOUND_REF` (`src/op/sirius_physical_order.cpp:58`).
> * `TopN` must have an ordering key.
> * `TopN` expressions must be `BOUND_REF`.
> * `TopN` column indexes must be in range (`src/op/sirius_physical_top_n.cpp:54`).
> 
> The same bound-reference check is repeated at runtime by the inserted sort-sample, sort-partition, and merge-sort operators. It should be checked once in `create_plan(LogicalOrder/LogicalTopN)` before the nodes are moved.
> 
> 3. Join capability checks
> 
> Several static join properties are still rejected inside execution:
> 
> * Unsupported hash-join type (`src/op/sirius_physical_hash_join.cpp:1382`).
> * Join types unsupported by a particular `BUILD_PROBE` or `MIXED_JOIN` mode.
> * Unsupported nested-loop comparison/join types (`src/op/sirius_physical_nested_loop_join.cpp:510`).
> * Mixed-join predicate cannot be translated to a cuDF AST (`src/op/sirius_physical_hash_join.cpp:1144`).
> 
> The join type and expression tree are available during planning. AST translation currently needs a stream/device for literals, but a device-free can_translate() walk could validate the expression shape without constructing cuDF objects.
> 
> 4. Inserted partition/sort operator modes
> 
> The pipeline rewrite can construct configurations that only fail when executed:
> 
> * `RANGE` or `CUSTOM` partitioning is not implemented.
> * Partition count was not assigned.
> * Sort helper received a non-reference order expression.
> 
> These are properties of the rewritten physical plan and should be covered by `plan->verify()` after `insert_gpu_pipeline_operators()`, rather than discovered by a worker.
> 
> 5. Native DuckDB segment viability
> 
> The first native-table viability phase already runs during GPU plan construction and catches unsupported projected types and invalid partition metadata (`src/op/scan/duckdb_native_gpu_ingestible.cpp:195`).
> 
> But the second phase runs asynchronously in `next_split_provider()` and can reject:
> 
> * Unsupported data or validity compression.
> * `VARCHAR` constant compression.
> * Missing max-string-length statistics.
> * Overflow-block strings.
> * Unsupported nested storage layout.
> * `ARRAY` child-element counts over cuDF limits.
> * A split exceeding `cudf::size_type`.
> 
> Those checks are performed by `walk_duckdb_native_row_group_range()` after GPU execution has started (`src/op/scan/duckdb_native_gpu_ingestible.cpp:264`). They could be run eagerly during planning and the resulting range metadata cached for execution.
> 
> This is probably the highest-value change because these are genuine GPU capability failures, not transient runtime failures.
> 
> 6. Parquet backend and schema validation
> 
> Currently runtime metadata tasks can discover:
> 
> * No registered backend supports a known URI.
> * A projected column is absent from a parquet file (`src/op/scan/parquet_gpu_ingestible.cpp:438`).
> * Potentially incompatible per-file schema/metadata.
> 
> The paths and requested column names are known during planning, but checking the schema requires reading every footer. I would classify this as optional “metadata preflight,” since putting remote/footer I/O directly in logical planning can make prepare unexpectedly expensive.
> 
> 7. Required engine configuration
> 
> These currently surface during execution but are knowable before task launch:
> 
> * No HOST memory space configured for native scan staging.
> * No HOST memory space available for result collection (`src/op/sirius_physical_result_collector.cpp:149`).
> * No scan manager or datasource registry for the planned scan.
> * Known URI scheme has no configured backend.
> 
> Check these once when deciding whether the physical plan is GPU-eligible.

Thanks, I will create an issue and a follow up PR on this

src/include/sirius_context.hpp
bwyogatama: ClientContext is a duckdb method that we can't modify

src/include/sirius_extension.hpp
bwyogatama: This already change in the latest commit

mbrobbel (approved):

PR #1174, opened 2026-07-14
Allow task scheduler to look ahead and create more tasks when the task queue is empty

### LINKED ISSUE #1017: Starting scans pre-emptively
In a query with multiple scans we should start the second scan before its depanded by its downstream consumers. We need to have a system which pre-emptively starts scans

aminaramoon:


roaramburu:
this resolves #1017

src/include/creator/config.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <string_view>
+#include <unordered_map>
+
+namespace sirius::creator {
+
+/**
+ * @brief How a scheduling request should be serviced.
+ *
+ * - active:    create tasks in response to input from an active operator
+ * - eager:     create tasks for an operator that received data, but its dependencies are not yet
```
bwyogatama, line 31: this request type is not actually implemented

src/creator/task_creator.cpp
```diff
@@ -210,6 +242,32 @@ void task_creator::schedule(op::sirius_physical_operator* node)
...
+void task_creator::schedule_lookahead(std::optional<int> device_id_hint)
+{
+  if (_config.strategy != request_type::lookahead) { return; }
+  std::lock_guard lock(_lookahead_mutex);
+  for (; _index_of_next_lookahead < _lookahead_queue.size(); ++_index_of_next_lookahead) {
+    auto* node = _lookahead_queue[_index_of_next_lookahead];
+    if (node == nullptr) { continue; }
+    auto hint = node->get_next_task_hint();
+    if (!hint.has_value()) {
+      if (!node->get_pipeline()->is_pipeline_finished()) { return; }
```
bwyogatama, line 254: In what scenario do we hit this branch?
`if (!node->get_pipeline()->is_pipeline_finished()) `

bwyogatama (approved):

bwyogatama (approved):

src/creator/task_creator.cpp
aminaramoon: limjt operator

Issue #1165, opened 2026-07-13
Backend-neutral, query-attributed I/O telemetry (REST/S3 first, Quent exporter)


ran-yuan-rui:
## Background

Sirius reads table data from object stores through its IO framework; the REST backend serves S3 and S3-compatible stores. When an S3-backed query is slow, we can't currently see where the time went in the IO layer for that query. `rest_perf_snapshot` gives process-wide counters, but it can't tell us which query caused a set of reads, how much of a query's data came from the prefetch cache versus a real backend fetch, or the actual timeline of the reads (when they started and finished, rather than when a counter was incremented).

This issue tracks two pieces of work: recording query-attributed I/O inside Sirius, then exporting those records through Quent. It does not describe a defect in Quent. Most of the changes are in Sirius; the Quent-facing question is how asynchronously completed work should carry its original timestamps.

## Attribution

Reads carry query, pipeline, device, and phase identifiers when those values are available. Work performed before query setup — schema discovery at query bind, or an early LIST/HEAD request — carries phase-only attribution.

## Read model

The telemetry records three levels of I/O:

- **Logical read** — one caller-visible datasource read, including cache or stash hits.
- **Backend read** — one backend operation initiated for that logical read; it may fan out into several transport operations.
- **REST transfer** — one chunk transfer, including any retry attempts.

A cache or stash hit has a logical record but no backend record. Joining the two layers distinguishes bytes a query requested from bytes actually fetched remotely.

## Design outline (Sirius side)

- The IO framework exposes an optional backend-neutral sink for logical and backend records. No sink is registered by default, and this layer has no dependency on Quent or any generated bridge header. Quent attaches later as the first concrete sink.
- Attribution is stamped per scan split (each unit of scan work): query and pipeline at scan setup — which covers the Parquet footer metadata reads issued before device assignment — and device at split emission. It then travels by value with each read into the backend layer, so it does not depend on which thread runs the read. This covers foreground column reads, metadata reads issued before device assignment, and background prefetch reads.
- With telemetry disabled, each recording point costs a single null check on the sink pointer: no id generation, no clock read, no allocation, no attribution copy.

## Timestamps — the part we'd like Quent input on

[...]

PR #1175, opened 2026-07-14
feat(io)+ Quent : backend-neutral IO telemetry seam with REST-only activation

ran-yuan-rui:
Part of #1165 (query-attributed IO telemetry). **First PR of a three-PR stack**: it lands the backend-neutral telemetry seam the IO layer will use to feed Quent, and wires it end-to-end for the **REST/S3 backend only** — no uring, no KvikIO (they keep their original read APIs and stay fully inert).
Follow-ups: `T1b` query/pipeline/device attribution, `T1c` cache/prefetch join semantics.

## Design

Three plain-C++ types and one interface, no Quent/Rust dependency (`src/include/io/io_telemetry.hpp`):

- `logical_io_record` — exactly one per caller-visible datasource read, emitted at the three leaf read implementations 

- `backend_io_record` — exactly one per request_manager, aggregating all chunks and retries for one backend read. Per-GET/per-attempt detail belongs to the later Rest Transfer layer.

- `io_attribution` — `{query, pipeline, device, phase}`, carried by value; stamped in the next PR.

- `io_telemetry_sink` — consumer interface: thread-safe, `noexcept`   non-blocking; registered through `scan_manager_config`, propagated at construction to the REST reactor context and ioctx.

## Cost

- **Off (default, no sink)**: per read, 2-3 predictable branches + one pointer load + one extra virtual hop; no atomics, clocks, or allocations; zero per-chunk work in the reactor loop. Verified by an interleaved A/B against the base (5 runs × 6 scenarios, SF10 lineitem on MinIO): deltas −0.58%…+2.05%, all within 2×CV — no measurable drift.
- **On (sink registered)**: the seam adds ~50–130 ns per logical record and ~150–250 ns per backend record (two clock reads, one 64-byte state allocation, POD construction, one virtual call) — ~0.3–0.4 µs per S3 read pair, <0.01% of a wire GET.  On S3 it should be fine.

## Read-path overview

```text
registration:
  scan_manager_config.io_telemetry ─→ rest reactor_context ─→ rest_ioctx (sink)

per read:
  caller ─→ sirius_datasource leaf ──── direct ───→ read API ─→ request_manager
               │                                                     │
               │ logical_io_record                     backend_io_record (dtor)
               │   exactly 1 / read                        exactly 1 / request
               │        └────────── shared read_id ──────────┘
               └─ cache route: silent in this PR (join lands in T1c)
  sink == nullptr  ⇒  none of the above is built (branch-only off path)
```

### DIFF
--- a/src/include/io/io_context.hpp
+++ b/src/include/io/io_context.hpp
@@ -20,6 +20,7 @@
 #include "io/cache/config.hpp"
 #include "io/cache/metadata_store.hpp"
 #include "io/cache/types.hpp"
+#include "io/io_telemetry.hpp"
 #include "io/types.hpp"
 
 #include <rmm/cuda_stream_view.hpp>
@@ -180,6 +181,14 @@ class sirius_ioctx : public std::enable_shared_from_this<sirius_ioctx> {
 
   [[nodiscard]] cache::prefetching_cache* cache() noexcept { return _cache.get(); }
 
+  /// Optional telemetry sink; null (the default) keeps every emission point
+  /// structurally inert.
+  [[nodiscard]] io_telemetry_sink* io_telemetry() const noexcept { return _io_telemetry.get(); }
+  [[nodiscard]] const std::shared_ptr<io_telemetry_sink>& io_telemetry_shared() const noexcept
+  {
+    return _io_telemetry;
+  }
+
   /// True iff @c host_read / @c device_read should consult the cache
   /// before falling through to the backend.  Computed live so it tracks
   /// @ref initialize_cache / @ref shutdown_cache transitions.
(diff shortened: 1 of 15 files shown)


aminaramoon:
I don't think if quent is intended to be integrate to low-level APIs. I think the better thing would be to add CTRACK hooks to the io APIs,

ran-yuan-rui:
> I don't think if quent is intended to be integrate to low-level APIs. I think the better thing would be to add CTRACK hooks to the io APIs,

OK - this might be need more discussion, I am fine with both way (CTRACK/perf_instrumentation in S3 or  quent).
I will move this PR to draft

PR #1176, opened 2026-07-14
fix: two log level debug maybe better then info

yihong0618:
for log level here, maybe SIRIUS_LOG_DEBUG is better than info

as it maybe better for benchmark and the message also shows `coalesce-debug`


johanpel (approved):

mbrobbel (approved):
Thanks @yihong0618

Issue #1145, opened 2026-07-10
Introduce backend-agnostic logging facade


johanpel:
Logging (and some string formatting) is currently hard-wired to spdlog.

I propose to provide a back-end agnostic logging facade, and solve the following problems:

1. Faster alternatives to `spdlog` seem to exist. (Like e.g. [quill](https://github.com/odygrd/quill), that benchmark significantly faster, actively maintained, and ship conda packages)
2. `SIRIUS_LOG_*` macros are not used everywhere. `spdlog` calls exist outside `log/logging.hpp` Every logging call should go through the facade (and the facade API must expose only sirius-owned types).
3. Add a backend-agnostic logging facade, selectable via config at run-time, so users can choose whatever we want to add a back-end for (e.g. OTel logging).
4. Stop using spdlog's bundled fmt transitively to format strings.

In the longer term this can be leveraged to sink logs into Quent as well, so they can be easily rendered alongside a timeline.

PR #1177, opened 2026-07-14
refactor(log): introduce a backend-agnostic logging sink

johanpel:
## Summary

Introduces a backend-agnostic `sink` interface behind the logging API. The public API (`sirius::log`, `SIRIUS_LOG_*`) is spdlog-free, with spdlog confined to a single sink implementation. Next step in #1145.

## What changes

- The API is split by concern: `log/level.hpp` (`enum class level`, `string_to_enum`), `log/sink.hpp` (the `sink` interface plus the `set_sink` / `get_sink` facade), and `log/logging.hpp` (the `SIRIUS_LOG_*` macros).
- The level lives in the sink and is the sole filter; `should_log` gates before formatting, so disabled levels do no work.
- `flush()` returns `bool` to report whether the flush was reliable.
- `noop_sink` is the default sink and discards everything, so logging is safe with no backend installed and after `set_sink(nullptr)`.
- `spdlog_owning_sink` is the spdlog implementation, confined to a single translation unit and exposing no spdlog headers. It installs its daily-file logger via `spdlog::set_default_logger` (a single-lock, atomic swap) and drives periodic flushing with `spdlog::flush_every`.
- The facade is a `std::atomic<std::shared_ptr<sink>>`, swappable at runtime; `get_sink()` never returns null.

## Testing

`pixi run make test`. The `[log]` suite covers the facade contract against an in-memory recording sink: level parsing, macro formatting + call-site attribution, `set_sink(nullptr)` → noop, and concurrent logging racing sink swaps (gated on a `std::latch`). Macro-dependent cases compile out when `SIRIUS_ACTIVE_LOG_LEVEL == OFF`.


src/include/log/level.hpp
```diff
@@ -0,0 +1,67 @@
...
+  critical = SIRIUS_LOG_LEVEL_CRITICAL,
+  /// Emit nothing.
+  off = SIRIUS_LOG_LEVEL_OFF,
+};
+
+/// Parses a level name into a level.
+///
+/// Returns false
+/// (leaving `lvl` unchanged) for an unrecognized name.
+bool string_to_enum(std::string_view name, level& lvl);
```
mbrobbel: Just curious: why not `std::optional`?

src/include/log/sink.hpp
```diff
@@ -0,0 +1,75 @@
...
+  /// `should_log`/`log`, so the new level must become visible atomically.
+  virtual void set_level(level level) = 0;
+
+  /// Whether `level` currently passes the threshold level.
+  [[nodiscard]] virtual bool should_log(level level) const = 0;
+
+  /// Emits `message` attributed to `loc` iff `level` passes the threshold.
+  ///
+  /// Thread-safe.
+  virtual void log(level level, const std::source_location& location, std::string_view message) = 0;
```
mbrobbel, line 44: Should we also support log messages without `std::source_location`?

src/include/log/spdlog_owning_sink.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+// A sink that owns Sirius's spdlog logging: it creates and drives its own spdlog
+// logger — its own daily file, level, and flushing. As the owning variant it
+// controls the spdlog setup and may use global spdlog facilities; the (future)
+// guest sink is the one that leaves a host program's spdlog state untouched.
+
+/// Construction settings of the owning spdlog sink.
+struct spdlog_owning_config {
+  /// Directory the daily log file `sirius.log` is written to.
+  std::string log_dir;
```
mbrobbel: Can't we use `std::filesystem`?

src/log/spdlog_owning_sink.cpp
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  // No destructor deregistration: touching spdlog's registry at process exit
+  // risks a static-destruction-order problem with spdlog's own registry
+  // singleton. On destruction the logger stays registered (harmlessly unused);
+  // spdlog flushes and reaps it at shutdown, and the next constructor's
+  // set_default_logger replaces it.
+
+  // The level lives in the spdlog logger itself — the single source of truth.
+  void set_level(level lvl) override { _logger->set_level(to_spdlog_level(lvl)); }
```
mbrobbel, line 78: Maybe add a note that `_logger->set_level` is thread safe?

src/sirius_context.cpp
```diff
@@ -998,7 +999,15 @@ SiriusContextExtensionCallback::SiriusContextExtensionCallback()
 {
   if (auto* env = std::getenv("SIRIUS_LOG_DIR")) { Config::LOG_DIR = env; }
   if (auto* env = std::getenv("SIRIUS_LOG_LEVEL")) { Config::LOG_LEVEL = env; }
-  sirius::InitGlobalLogger(Config::LOG_LEVEL, Config::LOG_DIR, Config::LOG_FLUSH_SECONDS);
+  sirius::log::level lvl = sirius::log::level::info;
+  sirius::log::string_to_enum(Config::LOG_LEVEL, lvl);
```
mbrobbel: Should we check the return value? Or we just default to info if the value is bad without a warning?

src/include/log/sink.hpp
johanpel: Probably useful, but there are no such instance in the code at this time, so let's add when we need it.

src/include/log/level.hpp
johanpel: It's following the convention of some other overloads of `string_to_enum` but I agree std::optional is nicer, so I'll change it

johanpel: https://github.com/sirius-db/sirius/pull/1177/commits/35132386c75cbe3e314c21216e45ad8b0519dd02

src/include/log/spdlog_owning_sink.hpp
johanpel: https://github.com/sirius-db/sirius/pull/1177/commits/35132386c75cbe3e314c21216e45ad8b0519dd02

src/log/spdlog_owning_sink.cpp
johanpel: https://github.com/sirius-db/sirius/pull/1177/commits/35132386c75cbe3e314c21216e45ad8b0519dd02

src/sirius_context.cpp
johanpel: https://github.com/sirius-db/sirius/pull/1177/commits/209323e56ee9a3f25fcb15b3357af95d7a344024

mbrobbel (approved):

PR #1180, opened 2026-07-14
feat(telemetry): per-operator activity in the quent operator timeline

felipeblazing:
## Summary

Quent's "Operator timeline" previously showed one activity bar per **pipeline** (`Pipeline Id N`). This PR splits that into one bar per **physical operator** (`GPU_SCAN-34`, `HASH_JOIN-17`, …), each with its own active span and metadata. No changes to the quent repo are needed — the UI already renders any operator entity that has an active span.

## Emission (C++)

- Every `sirius_physical_operator` gets a lifetime-stable `operator_uuid` (uuidv7) used as its quent entity id. The existing `operator_id` counter resets per query, so it stays as the human-readable label only.
- `emit_plan_telemetry` declares one quent `Operator` per physical operator (deduped — a join can be the sink of one pipeline and the source of the next) and emits a `Statistics` event per operator: `operator_id`, `pipeline_id`, `estimated_cardinality`, plus `params` when an operator implements `params_to_string()` (none do yet — follow-up).
- Port declarations reference their owning source/sink operator instead of the pipeline; `Computing` task events carry the `operator_uuid`; data batches carry `producer_operator_uuid`.
- The `gpu-N`/`shared` thread groups and memory spaces/channels now parent to the **worker** instead of the engine. The query page prunes empty resource groups, and the quent UI injects the operator-timeline lane under the worker row — so the worker must own its execution resources for the lane to render at all.
- Queries get a real name: with no explicit `sirius_set_query_label`, the label defaults to the (whitespace-collapsed, 120-char-truncated) SQL text instead of `unnamed_query`.

## Model + analyzer (Rust)

- `task::Computing` gains `operator_uuid`; `data_batch::Constructed` gains `producer_operator_uuid`.
- New `TaskExt::operator_active_spans()` splits a task's timeline into one span per `Computing` run; `try_build()` aggregates these into each operator's `active_span`. The pipeline-keyed data_batch attribution is dropped (batch lifetime ≠ operator activity).
- Per-query views and `OperatorFilter` matching now key off computed operator UUIDs / `producer_operator_uuid`. The old convention ("pipeline uuid *is* the operator entity id") matched nothing once entities became per-operator, which left query pages empty.

## Compatibility

⚠️ Telemetry captures recorded **before** this change can't be loaded by the new analyzer (`Computing`/`Constructed` gained required fields). Re-capture rather than replaying old event dirs.

## Verification

- `cargo test -p sirius-telemetry-analyzer` — new unit tests for `operator_active_spans` (per-operator split, trailing-Computing edge case) and filter matching.
- Full C++ Catch2 suite (`pixi run make test`) — all 1760 test cases pass (the telemetry hierarchy test was updated for the new worker-parented layout).
- End-to-end: TPC-H SF100, all 22 queries on GPU, telemetry served via `sirius-telemetry-server` — verified per-operator declarations (e.g. 23 operators on Q3, all with stats), per-operator gantt bars under the worker's "Operator timeline", operator-filtered lanes, and named queries.

## Follow-ups (not in this PR)

- Implement `params_to_string()` per operator so the info panel shows join conditions / projections.
- Hover tooltips on the gantt bars (quent-repo change — the tooltip is explicitly disabled in `OperatorGanttChart`).


dhruv9vats:
Trying this out locally, it seems the physical operator based query plan is missing some egdes / may be incomplete. for example filters and projections appear as source operators, while hash joins appear as sink operators: (this is tpch 21)

Screenshot 2026-07-16 at 16 20 00
Screenshot 2026-07-16 at 16 18 38

src/telemetry/telemetry_context.cpp
```diff
@@ -175,54 +193,64 @@ void emit_plan_telemetry(
...
+    const auto operators = pipeline->get_operators();
+
+    // One quent Operator per physical operator (one timeline bar each); the operator DAG is
+    // conveyed via the plan edges below, so parent_operator_ids stays empty.
+    for (const auto& op_ref : operators) {
+      const auto& op = op_ref.get();
+      if (!declared_operators.insert(op.get_operator_uuid()).second) { continue; }
+      operator_obs->declaration(
+        op.get_operator_uuid(),
+        quent::operator_::Declaration{
```
dhruv9vats:
```suggestion
        {
```
nit: can improve readability

src/telemetry/telemetry_context.cpp
```diff
@@ -175,54 +193,64 @@ void emit_plan_telemetry(
...
-      quent::operator_::Declaration{
-        .plan_id             = plan_id,
-        .parent_operator_ids = {},
-        .instance_name       = operator_chain,
-        .type_name           = std::format("Pipeline Id {}", pipeline->get_pipeline_id()),
-        .custom_attributes   = {},
-      });
+      operator_obs->statistics(op.get_operator_uuid(),
+                               quent::operator_::Statistics{.custom_attributes = std::move(stats)});
+    }
```
dhruv9vats, line 281: nit: readability improvement.
```suggestion
      quent::CustomAttributes stats{};
      stats.i64_attrs = {
        {.key = "operator_id", .value = static_cast<int64_t>(op.get_operator_id())},
        {.key = "pipeline_id", .value = static_cast<int64_t>(pipeline->get_pipeline_id())},
        {.key = "estimated_cardinality", .value = static_cast<int64_t>(op.estimated_cardinality)},
      };
      if (auto params = op.params_to_string(); !params.empty()) {
        stats.string_attrs.push_back({.key = "params", .value = std::move(params)});
      }
      operator_obs->statistics(op.get_operator_uuid(), {.custom_attributes = std::move(stats)});
    }
```

src/telemetry/telemetry_context.cpp
```diff
@@ -175,54 +193,64 @@ void emit_plan_telemetry(
...
+                                  .operator_id   = source->get_operator_uuid(),
                                   .instance_name = std::format("{}_receiver", port_id),
                                 });
         }
       }
     }
 
-    // Sender ports on pipeline sink(last) operators.
+    // Sender pseudo-ports on the pipeline's sink operator.
+    const auto sink = pipeline->get_sink();
```
dhruv9vats: So what we seem to be missing here are edges between physical operators that belong to the same pipeline.

src/sirius_extension.cpp
```diff
@@ -555,6 +557,25 @@ unique_ptr<FunctionData> SiriusExtension::GPUExecutionBind(ClientContext& contex
...
+      if (is_space && last_was_space) { continue; }
+      label.push_back(is_space ? ' ' : c);
+      last_was_space = is_space;
+      if (label.size() >= MAX_QUERY_LABEL_CHARS) {
+        label += "...";
+        break;
+      }
+    }
+    if (not label.empty()) { query_label = std::move(label); }
+  }
```
dhruv9vats: This is great, thanks!

src/include/op/sirius_physical_operator.hpp
```diff
@@ -357,9 +357,16 @@ class sirius_physical_operator {
...
   std::size_t estimated_cardinality;
-  //! The unique ID of this operator (auto-incremented at creation)
+  //! The unique ID of this operator (auto-incremented at creation). NOTE: this counter is reset
+  //! to 0 at the start of each query (see sirius_context), so it is a per-query human-readable
+  //! label, NOT a globally-unique id. Use `operator_uuid` when a stable entity id is required.
   size_t operator_id;
 
+  //! A globally-unique, lifetime-stable id for this operator, used as its quent entity id so
+  //! declaration and per-operator `Computing` telemetry events agree. Mirrors the port
+  //! `source_port_uuid` pattern; unlike `operator_id` it does not reset per query.
```
dhruv9vats: Comments too specific to this PR?
```suggestion
  //! A globally-unique, lifetime-stable id for this operator, used as an operator's quent 
  // entity id; unlike `operator_id` it does not reset per query.
```

src/include/op/sirius_physical_operator.hpp
```diff
@@ -380,6 +387,9 @@ class sirius_physical_operator {
   //! Get the unique operator ID
   size_t get_operator_id() const { return operator_id; }
 
+  //! Get this operator's stable, globally-unique UUID (its quent entity id).
```
dhruv9vats:
```suggestion
  //! Get this operator's stable, globally-unique UUID.
```

src/include/expression/ast/to_string.hpp
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+// standard library
```
dhruv9vats:
```suggestion
```

src/expression/ast/to_string.cpp
```diff
@@ -0,0 +1,167 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "expression/ast/to_string.hpp"
+
+// sirius
```
dhruv9vats:
```suggestion
```

src/expression/ast/to_string.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "expression/aggregate_id.hpp"
+#include "expression/ast/node.hpp"
+#include "expression/function_id.hpp"
+#include "expression/join_condition.hpp"
+#include "expression/value.hpp"
+
+// duckdb
```
dhruv9vats:
```suggestion
```

src/expression/ast/to_string.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "duckdb/common/types/value.hpp"
+
+// standard library
```
dhruv9vats:
```suggestion
```

src/include/telemetry/data_batch_probe.hpp
```diff
@@ -195,15 +201,18 @@ class quent_data_batch_probe : public cucascade::idata_batch_probe {
    *
    * @param ctx The telemetry context providing instrumentation and memory UUIDs.
    * @param producer_pipeline_uuid UUID of the pipeline that produced this batch.
+   * @param producer_operator_uuid Stable UUID of the physical operator that produced this batch.
```
dhruv9vats: It might become confusing if it says here (and at other places) that both the pipeline and the operators produce a data_batch. It might be useful to clarify what the relation between the two is, and what does produce actually mean?

src/sirius_extension.cpp
dhruv9vats: It seems like this is not being used somehow, I am trying to debug this locally.

dhruv9vats:
A more high level question:

If we simple want to replace the current pipeline based query DAG view with a physical operator based view, this is the correct approach.

But, if we instead want to have both the views live along side each other, quent also has provision for that
https://github.com/rapidsai/quent/tree/main/docs/domains/query_engine#plan:
> Mutually exclusive:
>    `query_id: uuid`: the ID of the [Query](https://github.com/rapidsai/quent/tree/main/docs/domains/query_engine#query), if this is a root Plan
    `parent_plan_id: uuid`: the ID of the parent Plan, if this is a derived (or "lowered") Plan. The Query is reachable by traversing up the parent chain.

And depending on what is preferable, we can choose to implement this while keeping the pipeline based view still available.

Subjectively, I prefer to keep both.

src/sirius_extension.cpp
felipeblazing: Good catch — you found a real gap. This fallback only ran in the explicit `gpu_execution()` bind path; the transparent interception path (`PhysicalSiriusExecution::GetGlobalSourceState`) only consulted `take_pending_query_label()`, so any intercepted query without a `sirius_set_query_label()` call showed `unnamed_query`.

Fixed in 20bcc981: the fallback is factored into `sirius_interface::query_label_from_sql()` and applied in both paths (a pending label still takes precedence). Verified locally — an unlabeled transparent parquet query now shows its whitespace-collapsed SQL text as the telemetry instance name.

felipeblazing:
Latest pushes address all the open feedback:

**Both plan views now live side by side** (181cd3ef), per the plan-hierarchy suggestion: the root plan keeps the original pipeline-level DAG (one operator per pipeline, real boundary ports/edges), and the physical-operator DAG is emitted as a derived ("lowered") plan parented to it. The per-operator activity timeline attaches to the derived plan's operators, and the analyzer's `query_plans` traversal picks up operators from both plans.

**Missing edges** (55300ecd, carried into the derived plan in 181cd3ef): consecutive operators of each pipeline are now chained through synthesized port pairs, pairs shared by child pipelines are deduped, and the pseudo-sink self-edge is skipped. Verified on TPC-H SF100, all 22 queries: both plan graphs resolve every edge with zero unconnected operators (q21 previously had 14 floating — filters/projections as sources, joins as sinks).

**Doc/readability nits + the `unnamed_query` label gap** (20bcc981): all suggestions applied, and the SQL-text label fallback now also covers transparently intercepted queries — see the inline reply for details.

**On the CI failure**: the Jul 17 run passed everything substantive (1940/1940 unit tests, TPC-H snapshot 22/22, 0 validation errors) and failed only on the final benchmark-artifact upload with a transient `FinalizeArtifact: 403 Forbidden`. Rerunning just that job can't recover because the run's `build-cuda-13` artifact has expired; the new push triggered a fresh full run.

src/telemetry/telemetry_context.cpp
```diff
@@ -27,13 +27,26 @@
...
 #include <memory>
-#include <ranges>
 #include <string>
+#include <unordered_set>
 
 namespace sirius::telemetry {
 
+namespace {
+//! Hasher for `uuid::UUID` (the cxx-generated type provides `operator==` but no `std::hash`),
+//! so operator UUIDs can be deduplicated in an `unordered_set` while declaring the plan.
```
dhruv9vats, line 40:
```suggestion
//! so operator UUIDs can be deduplicated in an `unordered_set`
```

rust/crates/telemetry/model/src/data_batch.rs
```diff
@@ -6,6 +6,10 @@ state! {
         attributes: {
             data_batch_id: u64,
             producer_pipeline_uuid: Uuid,
+            // Stable id of the physical operator that produced this batch. Operator entities are
+            // declared per-operator (not per-pipeline), so per-operator attribution and filtering
+            // key off this rather than `producer_pipeline_uuid`.
```
dhruv9vats, line 11: now stale
```suggestion
producer_pipeline_uuid: Uuid,
            // Stable id of the physical operator that produced this batch. Per-operator attribution and filtering
            // key off this rather than `producer_pipeline_uuid`.
```

src/include/op/sirius_physical_operator.hpp
```diff
@@ -398,9 +398,15 @@ class sirius_physical_operator {
...
-  //! The unique ID of this operator (auto-incremented at creation)
+  //! The unique ID of this operator (auto-incremented at creation). NOTE: this counter is reset
+  //! to 0 at the start of each query (see sirius_context), so it is a per-query human-readable
+  //! label, NOT a globally-unique id. Use `operator_uuid` when a stable entity id is required.
   size_t operator_id;
 
+  //! A globally-unique, lifetime-stable id for this operator, used as the operator's quent
+  //! entity id; unlike `operator_id` it does not reset per query.
+  uuid::UUID operator_uuid{uuid::now_v7()};
+
```
dhruv9vats, line 409: these should ideally be, either:
1. private and accessed via const accessors
2. or const public members (less preferrable, since that completely disables the class's implicitly generated copy assignment and move assignment operators)

src/include/telemetry/data_batch_probe.hpp
```diff
@@ -194,16 +200,22 @@ class quent_data_batch_probe : public cucascade::idata_batch_probe {
    * data's current memory tier.
    *
    * @param ctx The telemetry context providing instrumentation and memory UUIDs.
-   * @param producer_pipeline_uuid UUID of the pipeline that produced this batch.
+   * @param producer_pipeline_uuid UUID of the pipeline whose execution emitted this batch.
+   * @param producer_operator_uuid Stable UUID of the physical operator within that pipeline
```
dhruv9vats, line 204: Words like `stable` and `globally-unique` may be omitted from docs, etc., as UUIDs are understood to be a _universally unique ID_.
```suggestion
   * @param producer_operator_uuid UUID of the physical operator within that pipeline
```

src/include/telemetry/data_batch_probe.hpp
```diff
@@ -194,16 +200,22 @@ class quent_data_batch_probe : public cucascade::idata_batch_probe {
...
+   *        whose execute() actually produced the batch. Both attribute the same event at
+   *        different granularities: the pipeline is the schedulable unit (a node in the root
+   *        plan's DAG), the operator is the step inside it (a node in the derived plan's DAG).
```
dhruv9vats, line 207: `derived` might not be the best choice here -- it suggests that the physical operator based plan was derived from the pipeline based plan, which is reversed in reality. Maybe 
```suggestion
   *        different granularities: the pipeline is the schedulable unit (a node in the pipeline based
   *        plan's DAG), the operator is the step inside it (a node in the physical operator based plan's DAG).
```

src/include/sirius_interface.hpp
```diff
@@ -65,6 +65,13 @@ class sirius_interface {
  public:
   sirius_interface(duckdb::ClientContext& client_context,
                    std::optional<std::string> query_label = std::nullopt);
+
+  //! Build a telemetry query label from raw SQL text (whitespace-collapsed, truncated) so
+  //! viewers show a recognizable name instead of a placeholder. Shared fallback for every
```
dhruv9vats: `Shared fallback`: sounds like this is something that will happen passively, but nothing _in this API_ enforces that this _will_ be used as a fallback if the query label is not present. Its just the current limited implementation happens to do that. Maybe instead suggest action? (BTW, love that we are using this though, much better for the common case than "unnamed_query")


```suggestion
  //! viewers show a recognizable name instead of a placeholder. Should be used as the fallback for every
```

src/telemetry/telemetry_context.cpp
```diff
@@ -84,15 +99,19 @@ telemetry_context::telemetry_context(const sirius::telemetry_config& config,
     });
 
   // Per-GPU device groups plus per-thread-type buckets underneath, so the
-  // viewer renders threads as an engine -> gpu-N -> thread-type tree instead
+  // viewer renders threads as a worker -> gpu-N -> thread-type tree instead
   // of a flat sibling list. Threads with no single GPU go under `shared`.
+  // These groups parent to the WORKER (not the engine): the query page's
+  // resource tree prunes empty groups, so the worker row — which the quent UI
+  // needs in order to inject the per-operator "Operator timeline" lane — only
+  // renders if the execution groups hang beneath it.
```
dhruv9vats, line 107: too specific a comment? we make no guarantee that the UI will keep that behavior + not relevant for this part of the codebase ?

```suggestion
  // These groups parent to the WORKER (not the engine)
```

src/telemetry/telemetry_context.cpp
```diff
@@ -172,38 +191,98 @@ void emit_plan_telemetry(
   auto port_obs     = quent::port::create_observer(context);
   auto plan_obs     = quent::plan::create_observer(context);
 
-  // Collect edges while iterating
-  rust::Vec<quent::plan::Edges> edges;
+  // Two plan views (see rapidsai/quent docs/domains/query_engine#plan): `plan_id` stays the
+  // root plan and keeps the original pipeline-level DAG (one quent Operator per pipeline,
+  // real ports, edges only at pipeline boundaries). A derived ("lowered") plan carries the
+  // physical-operator DAG that the per-operator activity timeline attaches to, so both views
+  // live side by side.
```
dhruv9vats, line 198: Again, seems like a comment to specific to this PR, maybe instead:

```suggestion
  // Two plan views (see rapidsai/quent docs/domains/query_engine#plan): one quent-operator per pipeline, real ports, edges only at pipeline boundaries); another nested plan that carries the
  // physical-operator DAG that the per-operator activity timeline attaches to, so both views
  // live side by side.
```

>that carries the physical-operator DAG that the per-operator activity timeline attaches to

this is another thing I wanted to bring up, the physical operator plan in the UI is currently nested under the pipeline based plan, but the timeline filtering is keyed by the physical operator plan. (we have an ongoing quent effort that should allow the timeline to be keyed by multiple operators among the plans). So the question is, what is more important for execution semantic's analysis ? Pipeline based plan or the  physical operator based plan? If the former, we keep the nesting as is, if the latter we reverse the nesting.

The quent effort should soon allow timeline filtering from different plan levels, so that's alright for now.

src/telemetry/telemetry_context.cpp
```diff
@@ -172,38 +191,98 @@ void emit_plan_telemetry(
...
+    // each); the operator DAG is conveyed via the derived plan's edges, so
+    // parent_operator_ids stays empty.
+    for (const auto& op_ref : operators) {
+      const auto& op = op_ref.get();
+      if (!declared_operators.insert(op.get_operator_uuid()).second) { continue; }
+      operator_obs->declaration(
+        op.get_operator_uuid(),
+        {
+          .plan_id             = operator_plan_id,
+          .parent_operator_ids = {},
```
dhruv9vats, line 245: we can consider populating this, to point to pipeline_uuids.

src/telemetry/telemetry_context.cpp
```diff
@@ -172,38 +191,98 @@ void emit_plan_telemetry(
...
+      quent::CustomAttributes stats{};
+      stats.i64_attrs = {
+        {.key = "operator_id", .value = static_cast<int64_t>(op.get_operator_id())},
+        {.key = "pipeline_id", .value = static_cast<int64_t>(pipeline->get_pipeline_id())},
+        {.key = "estimated_cardinality", .value = static_cast<int64_t>(op.estimated_cardinality)},
+      };
+      if (auto params = op.params_to_string(); !params.empty()) {
+        stats.string_attrs.push_back({.key = "params", .value = std::move(params)});
+      }
+      operator_obs->statistics(op.get_operator_uuid(), {.custom_attributes = std::move(stats)});
```
dhruv9vats, line 260: nice!

rust/crates/telemetry/analyzer/src/model.rs
```diff
@@ -607,16 +608,8 @@ impl SiriusModelBuilder {
...
-                    {
-                        operator.active_span = Some(match operator.active_span() {
-                            None => data_batch_span,
-                            Some(existing) => existing.extend(&data_batch_span),
-                        });
-                    }
-
+                    // `producer_pipeline_uuid` is a pipeline id, not an operator entity id, so
+                    // batches are intentionally not folded into operator activity here.
                     data_batches.insert(data_batch_id, data_batch);
```
dhruv9vats, line 613: Do we not want to use: `producer_operator_uuid` here too as with tasks ? Or this this because we want the active_spans of an operator to only be based on task activity ?

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -1079,11 +1077,21 @@ impl SiriusUiAnalyzer {
...
+                            self.model.query_engine.operators.get(&operator_uuid)
+                        {
+                            let name = operator.instance_name();
+                            if names.last() != Some(&name) {
+                                names.push(name);
+                            }
+                        }
+                    }
+                    let pipeline_name = (!names.is_empty()).then(|| names.join(" -> "));
+                    Some(task.try_to_ui_fsm(epoch, pipeline_name.as_deref()))
```
dhruv9vats, line 1094: With us keeping both plans now, this is not longer needed? And we can continue to rely on the pipeline node's name ?

rust/crates/telemetry/analyzer/src/task.rs
```diff
@@ -70,6 +76,33 @@ impl TaskExt for Task {
...
+                    return None;
+                };
+                // Runs until the next transition; a trailing Computing has no span.
+                let next = transitions.get(i + 1)?;
+                let span =
+                    SpanUnixNanoSec::try_new(transition.timestamp(), next.timestamp()).ok()?;
+                Some((data.operator_uuid, span))
+            })
+            .collect()
+    }
```
dhruv9vats, line 95: this means that the operator would be shown as active only when a task first enters a computing state for that operator, and would not include times for "preparing" + upgrading data. Just making sure that we are aware and that's what we want. Since this will likely be aggregated as a metric at some point and compared, and for high SF runs this will affect the calculation / comparison.

dhruv9vats (approved):
Thanks @felipeblazing! The analyzer streamlining can be done as part of the de-duplication that is coming down the pipes.

mbrobbel:
Marking as draft, there are some conflicts.

PR #1181, opened 2026-07-14
feat(scan): background host→GPU prefetcher for pinned-cache scan splits

felipeblazing:
## Problem

On hot host-pinned runs, per-iteration nsys traces show the slow TPC-H queries (q21, q18, q12, q9) alternating between **compute-only phases** (100% SM busy, zero H2D traffic) and **copy-only phases** (~350 GB/s H2D, ~20% SM busy). Example from q21's hot iteration (100 ms bins):

| window | H2D | SM busy | running |
|---|---|---|---|
| 900–1600 ms | zero | 100% | Pipeline 28 (lineitem self-join probe) |
| 1600–2000 ms | 123 GB burst | ~20% | Pipeline 38 (`GPU_SCAN → FILTER`) |

The cause: scan splits for pinned tables sit ready in their `split_connector`s from query start, but the H2D upload of each resident batch happens only inside `scan_operator_input::prepare_for_processing` when the scan's pipeline task executes — and scan tasks are created strictly on topology demand. Data movement is glued to task execution, task execution to topology order, so copy engines idle during compute and vice versa.

## Change

An opt-in **`scan_prefetcher`** (`SIRIUS_SCAN_PREFETCH=1`) owned by the scan manager: worker threads walk pending resident splits in scan (execution) order and convert them to GPU tier **ahead of task creation**, on dedicated exclusive streams, gated on GPU-memory headroom. When the scan task later runs, `prepare_for_processing` finds the batch already GPU-resident and skips the upload; the task creator's cached-scan locality derivation reads the batch's post-conversion space, so device dispatch stays consistent. Races with consumers are arbitrated by the `data_batch` state machine (`try_to_mutable`, skip on contention) plus the existing tier re-check in `prepare_for_processing`.

Two contention guards proved essential during benchmarking:

- **Never prefetch an actively-draining connector** — the scan's own tasks convert concurrently (one stream per pipeline thread); fewer prefetch threads holding exclusive locks would serialize them. A connector counts as draining until `SIRIUS_SCAN_PREFETCH_QUIET_MS` (default 100) pass without a pop, re-checked between individual conversions.
- **Dedicated `exclusive_stream_pool`** — the memory_space's shared round-robin pool would interleave multi-GB prefetch copies with task kernels on the same stream.

## Results

TPC-H SF1000, hot host-pinned, GB300, 3 iterations, all query results byte-identical to baseline:

- **Full 22-query suite (parquet): -5.9% total hot runtime** — q12 -18.2%, q18 -16.4%, q6 -11.5%, q21 -11.1%, q9 -9.8%, q2 -9.0%; worst regression +2.1% (run noise).
- **duckdb-native source (q1/9/13/18/21 subset): -8.5%** — q18 -15.7%, q9 -9.9%, q21 -9.9%.
- Config sweep (threads 1/2/4, min-free 0.25–0.55, quiet 50–200 ms): defaults (2 / 0.4 / 100 ms) sit on a flat optimum; the mechanism is robust to ±2× knob changes.

## Scope / caveats

- **Disabled by default**, env-gated (`SIRIUS_SCAN_PREFETCH`, `_THREADS`, `_MIN_FREE_FRACTION`, `_QUIET_MS`). Happy to move these into `sirius.yaml` config if preferred.
- **Single-GPU only**: multi-GPU needs the task creator's per-batch NUMA-locality device derivation; the prefetcher logs a warning and disables itself when more than one GPU space is configured.
- Prefetched-but-unconsumed batches live in connector queues, which the downgrade executor does not scan — hence the conservative headroom floor (default keeps 40% of the GPU space free).


src/scan_manager/scan_prefetcher.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <cucascade/cudf/gpu_data_representation.hpp>
+#include <cucascade/data/data_batch.hpp>
+
+#include <chrono>
+#include <cstdint>
+#include <exception>
+
+namespace sirius::scan_manager {
+
+scan_prefetcher::scan_prefetcher(config cfg,
```
aminaramoon: can we call this something rather than scan_prefetching ? cause it is not scan_prefetching ? it is memory_prefetching or something like that ?

src/scan_manager/memory_prefetcher.cpp
```diff
@@ -0,0 +1,189 @@
...
+    _last_pop_counts(_connectors.size()),
+    _last_pop_time_ms(_connectors.size())
+{
+  if (_gpu_space == nullptr || _connectors.empty()) {
+    _running.store(false);
+    return;
+  }
+  _stream_pool = std::make_unique<cucascade::memory::exclusive_stream_pool>(
+    rmm::cuda_device_id{_gpu_space->get_device_id()}, _config.num_threads);
+  _workers.reserve(_config.num_threads);
```
aminaramoon, line 43: why do we need threads, do they do anything beyond h2d copy ? if that is the case, then the concurrency is drived by hardware not threads in user space

src/scan_manager/memory_prefetcher.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+      // Exclusive lock, skip on contention (a task is consuming this batch —
+      // its own prepare_for_processing does the upload).
+      auto mut = batch->try_to_mutable();
+      if (!mut) { continue; }
+      if (mut->get_data() == nullptr || mut->get_current_tier() == cucascade::memory::Tier::GPU) {
+        continue;
+      }
+
+      auto reservation = _gpu_space->make_reservation_or_null(batch_bytes);
```
aminaramoon, line 153: I think we should probably reserve at a lower threshold than the processing threadhols, otherwise we will be fighting for memory with the executors

src/include/op/scan/sirius_gpu_scan_operator.hpp
```diff
@@ -125,6 +125,13 @@ class sirius_gpu_scan_operator : public sirius_physical_operator {
 
   scan_manager::split_connector& get_split_connector();
 
+  /// Shared handle to the connector, for components (e.g. the scan prefetcher)
+  /// that must outlive-safely reference it from a background thread.
+  [[nodiscard]] std::shared_ptr<scan_manager::split_connector> get_split_connector_ptr() const
```
aminaramoon: do we need `get_split_conenctor_`ptr`` in the name of the method ?

src/include/scan_manager/scan_prefetcher.hpp
```diff
@@ -0,0 +1,128 @@
...
+ * Memory safety: converted batches live in connector queues, which the
+ * downgrade executor does NOT scan (it walks data repositories), so
+ * prefetched-but-unconsumed bytes cannot be reclaimed under pressure. The
+ * headroom gate must therefore stay conservative: prefetch only while
+ * (available - batch_size) >= min_free_fraction * max_memory.
+ */
+class scan_prefetcher {
+ public:
+  struct config {
+    /// Number of prefetch worker threads (each with its own stream).
```
aminaramoon: we should hook this to sirius_config instead of using `env` variable to configure this ? no ?

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -438,6 +439,45 @@ void sirius_scan_manager::start_metadata_processing()
     if (it == _providers_by_op.end()) { continue; }
     it->second->run(*_dispatcher, _metadata_processor->get_split_provider_bridge(op));
   }
+  maybe_start_scan_prefetcher();
+}
+
+void sirius_scan_manager::maybe_start_scan_prefetcher()
+{
+  const char* enabled = std::getenv("SIRIUS_SCAN_PREFETCH");
+  if (enabled == nullptr || std::string_view{enabled} != "1") { return; }
```
aminaramoon: we need to use sirius_config for this, this goes against how we do similar thing for all the other components ?

src/scan_manager/scan_prefetcher.cpp
```diff
@@ -0,0 +1,189 @@
...
+{
+  std::size_t converted = 0;
+  const auto min_free_bytes =
+    static_cast<std::size_t>(_config.min_free_fraction * _gpu_space->get_max_memory());
+  auto& registry = sirius::converter_registry::get();
+
+  // Walk connectors in scan (execution) order so the head-of-line pipeline's
+  // data lands on the GPU first.
+  for (std::size_t ci = 0; ci < _connectors.size(); ++ci) {
+    const auto& connector = _connectors[ci];
```
aminaramoon: so we are completly relying on order of connectors to preload input to gpu memory.

since we reserve at the same threshold that executors reserve, this can cretate memory pressure.
if the order of connectors is different than the order at which the engine processes splits, this prefetching only increases memory pressure which is very problmatic because the cahced scan_data are not downgradable. so we create pressure both for host and gpu by caching and prefetching.

aminaramoon (changes requested):

src/scan_manager/scan_prefetcher.cpp
```diff
@@ -0,0 +1,189 @@
...
+    if (!_running.load(std::memory_order_relaxed)) { return converted; }
+
+    // Actively-draining connector: its scan tasks convert their own batches
+    // on 1 stream per pipeline thread. Grabbing exclusive locks here would
+    // serialize those conversions behind the (fewer) prefetch threads and
+    // slow scan-bound queries down. A pure between-sweeps delta is not enough
+    // — sweeps run every ~2ms while an active scan pops only every ~10-40ms —
+    // so a connector stays "draining" until drain_quiet_ms pass with no pops.
+    auto is_draining = [&]() {
```
wmalpica: Why are we doing this weird check to see if the connector did work, as a means of indicating if its draining? why not just add a flag in the connector to indicate that it is draining. Maybe there is something i am not understanding here, but this just seems weird.

src/scan_manager/sirius_scan_manager.cpp
wmalpica: I agree. And when you add it to the sirius_config, please also update the configuration.md file.

src/scan_manager/memory_prefetcher.cpp
wmalpica: I think that is why Felipe added the check against min_free_bytes which ensures that there is at least 40% of GPU memory available

src/scan_manager/scan_prefetcher.cpp
```diff
@@ -0,0 +1,189 @@
...
+      if (!mut) { continue; }
+      if (mut->get_data() == nullptr || mut->get_current_tier() == cucascade::memory::Tier::GPU) {
+        continue;
+      }
+
+      auto reservation = _gpu_space->make_reservation_or_null(batch_bytes);
+      if (!reservation) { return converted; }
+
+      try {
+        mut->convert_to<cucascade::gpu_table_representation>(registry, _gpu_space, stream);
```
aminaramoon: here is a bug. you made te reservation, but then ddnt attach it, so the reserved arena remian unsed while you are converting

aminaramoon (changes requested):

felipeblazing:

PR #1182, opened 2026-07-14
feat(telemetry): report input size and rate on the task preparing transition

felipeblazing:
## Summary

In the quent UI, task computing sections show `input_bytes` and a derived `bytes_per_sec` rate, but the preparing section that precedes them showed neither. This PR adds both — plus an `origin_tier` attribute — so the preparation phase (staging/upgrading input onto the GPU) is visible in the UI:

- **Model** (`rust/crates/telemetry/model/src/task.rs`): add `origin_tier: String` and `input_bytes: u64` attributes to the task FSM's `Preparing` state. The C++ bridge API is code-generated from the model, so the fields flow through automatically.
- **Emitter** (`src/pipeline/gpu_pipeline_task.cpp`): the preparing transition now reports the input's `get_estimated_size_in_bytes()` — the same size the first computing transition reports — and its origin tiers.
- **Origin reporting** (`src/include/op/sirius_physical_operator.hpp`, `src/include/op/scan/sirius_gpu_scan_operator_data.hpp`): a new `operator_data::get_origin_tiers()` virtual. `pipelineable_operator_data` returns the '+'-joined set of its batches' tiers in tier order (e.g. `GPU+HOST`); `scan_operator_input` returns `SOURCE` for a fresh datasource read or the cached batch's tier when resident (via the non-blocking lock accessor, since `data_batch` keeps its tier private); the base default is `UNKNOWN`. `tier_display_name()` is a shared tier-to-string helper.
- **Analyzer** (`rust/crates/telemetry/analyzer/src/task.rs`): the `bytes_per_sec` derivation that previously only ran for `Computing` transitions now also covers `Preparing`, dividing `input_bytes` by the span to the next transition.

No UI changes needed — the quent UI infers formatting from attribute names (`*_bytes` → IEC size, `bytes_per_sec` → SI B/s rate), so the preparing section picks up the same display treatment the computing sections already have.

## Compatibility note

This changes the `Preparing` state schema: telemetry captured before this change cannot be loaded by the newer analyzer (and vice versa) — recapture instead.

## Testing

- `cargo check`/`cargo test` on `instrumentation-model`, `sirius-telemetry-analyzer`, and `telemetry-bridge` (bridge codegen verified: generated C++ struct field order matches the emitter's designated initializer).
- Full `pixi run make test` — all 1802 test cases pass.
- Verified live on an SF100 TPC-H capture (Q1/Q5/Q9): all 557 tasks report a real origin (`GPU` for pipeline tasks, `SOURCE` for fresh scans, zero `UNKNOWN`), and the quent UI renders size/rate/origin on preparing sections.

### DIFF
--- a/rust/crates/telemetry/analyzer/src/task.rs
+++ b/rust/crates/telemetry/analyzer/src/task.rs
@@ -82,14 +82,19 @@ impl TaskExt for Task {
             .map(|(i, transition)| {
                 // Derive the processing rate from input_bytes over the span.
                 let mut derived_attributes = vec![];
-                if let ModelTaskTransition::Computing(data) = &transition.data
+                let input_bytes = match &transition.data {
+                    ModelTaskTransition::Preparing(data) => Some(data.input_bytes),
+                    ModelTaskTransition::Computing(data) => Some(data.input_bytes),
+                    _ => None,
+                };
+                if let Some(input_bytes) = input_bytes
                     && let Some(next) = raw.get(i + 1)
                 {
                     let span_secs = (next.timestamp() - transition.timestamp()) as f64 / 1e9;
                     if span_secs > 0.0 {
                         derived_attributes.push(Attribute::f64(
                             "bytes_per_sec",
-                            data.input_bytes as f64 / span_secs,
+                            input_bytes as f64 / span_secs,
                         ));
                     }
                 }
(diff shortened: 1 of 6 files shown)


johanpel (approved):
Rust changes LGTM

PR #1187, opened 2026-07-15
feat(telemetry): batch placement lifecycle + memory-tier data-flow view

felipeblazing:
## What

Instruments **batch placements** — the lifecycle of each data batch published to a consuming pipeline's input port — and implements quent's new per-operator data-flow distribution endpoint, so quent's DAG view can scrub through time and show, per pipeline: how many batches (and bytes) are queued waiting for the scheduler, packaged into tasks, or actively processing, and **which memory tier / GPU device the data resides on**.

Companion PR (merged): rapidsai/quent#393 — the generic categorical-timeline protocol + playhead UI.

## Model (`rust/crates/telemetry/model`)

New `Batch` FSM: `batch_registered{batch_id, pipeline_uuid, port_uuid, origin} → batch_queued → batch_packaged{task_uuid} → batch_processing{task_uuid} → batch_consumed{reason}`, where every non-terminal state carries a tier usage on a new `MemoryTier` resource — **one per GPU device** (`GPU-0`, `GPU-1`, …) plus engine-wide `HOST`/`DISK` — weighted by the batch's bytes. Tier changes (downgrade/spill/prepare-time upgrade) are self-transitions with the new tier usage. Placements share the engine's `batch_id` across fan-out and OOM re-packaging, so batch identity survives task reschedules. Complementary to `DataBatch` (#1068), which tracks physical residency; `Batch` tracks scheduling lifecycle per consumer.

## Engine instrumentation (`src/`)

A process-global `batch_telemetry_registry` (sharded by batch id) maps port repositories to consumer pipelines (registered during plan telemetry), so publish sites only pass `(batch, repo)`:

- Publishes (operator sink, partition consumer) emit `registered → queued` before the batch becomes poppable.
- `gpu_pipeline_task`: ctor claims inputs (`queued → packaged`, with lazy registration for OOM-reschedule intermediates), execute emits `packaged → processing` post-prepare (tier re-read to capture upgrades; an id-based path covers merge/concat inputs consumed during materialization), dtor releases claims **by recorded batch id** (the weak batch refs are dead by then). OOM re-claims transfer ownership to the rescheduled task.
- Tier moves reported from `convertible_data_batch::convert` and `lock_or_prepare_batch` while the exclusive lock is held (values passed, never re-locked).
- Leftovers drained as `consumed{query_end}` in `QueryEnd`; gated by new `telemetry.enable_batch_events` (default on).

Every placement in TPC-H SF1/SF300 (2-GPU) test runs completes `consumed{processed}` inside the query window.

## Analyzer (`rust/crates/telemetry/analyzer`)

- Ingests Batch/MemoryTier events; implements `UiAnalyzer::data_flow_timeline` (states × tiers × count/bytes per pipeline; `Unsupported` for datasets without batch events, so old recordings keep working and the UI hides the view).
- `batch` is a third entity kind in the per-state resource timelines and entity lists, beside `task`/`data_batch`; `batch_registered` (instantaneous bookkeeping entry state) is omitted from aggregated series.
- Byte quantities declared with SI prefixes (GB rather than GiB).
- 7 analyzer unit tests incl. hand-computed bin math with a mid-queue tier-change split.

## ⚠️ Do not merge yet (draft)

- The quent crates are pinned to upstream `rapidsai/quent` main (the rapidsai/quent#393 merge commit); the temporary `[patch]` block used during development has been removed.
- Includes a merge of latest `dev` (resolved against #1068/#1080/#1112; the #1112 per-transition tooltip attributes are wired for batches too).

## Testing

- `pixi run make test` (1538 C++ cases) and `cargo test` green; full E2E: TPC-H SF1 + SF300 parquet on 2 GPUs (incl. a constrained-GPU-memory run), NDJSON lifecycle invariants verified, served through the embedded quent UI and scrubbed.


wmalpica (approved):

rust/crates/telemetry/analyzer/src/batch_placement.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    fn pipeline_uuid(&self) -> Option<Uuid> {
+        self.first_data().and_then(|t| match t {
+            ModelBatchPlacementTransition::BatchRegistered(data) => Some(data.pipeline_uuid),
+            _ => None,
+        })
+    }
+
+    fn task_uuid(&self) -> Option<Uuid> {
```
johanpel: Nit: for clarity
```suggestion
    fn last_task_uuid(&self) -> Option<Uuid> {
```

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -958,6 +1147,226 @@ impl UiAnalyzer for SiriusUiAnalyzer {
 
         Ok(BulkChunkedTimelinesResponse { entries })
     }
+
+    fn data_flow_timeline(
+        &self,
+        request: CategoricalTimelineRequest<QueryFilter>,
+    ) -> AnalyzerResult<DataFlowTimelineBinned> {
+        // Datasets recorded before BatchPlacement instrumentation existed: HTTP 501.
```
johanpel: This shouldn't be a concern since we are blessed with `quent-open`, so this build shouldn't really open old datasets.

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -958,6 +1147,226 @@ impl UiAnalyzer for SiriusUiAnalyzer {
...
+        if self.model.batch_placements.is_empty() {
+            return Err(AnalyzerError::Unsupported);
```
johanpel: In which case unsupported might be misleading.

Is it possible no batch placements instances exist when you run corner case queries (e.g. `select 1;`)?

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -958,6 +1147,226 @@ impl UiAnalyzer for SiriusUiAnalyzer {
...
+                    }
+                }
+            }
+        }
+
+        // Task processing space: tasks report their memory reservation as a
+        // tier usage while preparing/computing. Surface it as one synthetic
+        // series per operator so the DAG shows working space beside the batch
+        // lifecycle states (count = tasks holding space, bytes = reservation).
+        for task in view.tasks() {
```
johanpel, line 1266: Would it be possible to dedupe some of the logic here with the previous loop
`for batch in view.batch_placements() {`

johanpel (commented):
Partial review comments so far:

rust/crates/telemetry/analyzer/src/lib.rs
```diff
@@ -958,6 +1147,226 @@ impl UiAnalyzer for SiriusUiAnalyzer {
...
+                display_name: "Batches".to_owned(),
+                quantity: "unit".to_owned(),
+                kind: CapacityKind::Occupancy,
+            });
+        }
+        if want_bytes {
+            measures.push(MeasureDecl {
+                name: MEASURE_BYTES.to_owned(),
+                display_name: "BatchPlacement bytes".to_owned(),
+                quantity: "capacity_bytes".to_owned(),
```
johanpel: Should this use MEMORY_TIER_BYTES_CAPACITY_NAME ?

rust/crates/telemetry/model/src/batch.rs
```diff
@@ -0,0 +1,137 @@
...
+            task_uuid: Uuid,
+        },
+        usages: {
+            tier: MemoryTier,
+        },
+    }
+}
+
+state! {
+    // reason: "processed" | "task_failed" | "query_end".
```
johanpel: Same here as below 🔽

rust/crates/telemetry/model/src/batch.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+state! {
+    // batch_id: the engine's process-unique batch id, shared by all
+    //   placements/re-packagings of one physical batch.
+    // pipeline_uuid: the consumer pipeline (== quent Operator id).
+    // port_uuid: the consumer pipeline's source port receiving the batch.
+    // origin: "operator_output" | "partition_output" | "cpu_source" |
+    //   "reschedule_intermediate" | "pinned_cache".
```
johanpel: If this is a closed set of options, perhaps it would be best to provide an enum class on the C++ side with a constexpr std::string_view to_string_view(...)

I don't see "cpu_source" or "pinned_cache" being used, perhaps this is intentional?

johanpel (approved):
Rust changes look good overall, nice addition to the model @felipeblazing.

Most of my comments should be trivial to solve, so approving ahead.

rust/crates/telemetry/analyzer/src/batch_placement.rs
felipeblazing: Renamed in 4836d469.

rust/crates/telemetry/analyzer/src/lib.rs
felipeblazing: Dropped the empty check in 4836d469; the guard now keys off the `memory_tier` resources (next thread).

rust/crates/telemetry/analyzer/src/lib.rs
felipeblazing: Yes, zero placements is possible. Now `Unsupported` is returned only when the recording has no `memory_tier` resources (installed iff batch telemetry was enabled, e.g. not with `enable_batch_events: false`); a query with zero placements returns a valid empty response. Both cases have tests.

rust/crates/telemetry/analyzer/src/lib.rs
felipeblazing: Deduped: both loops call `push_tier_state_spans`, with a `state_for` closure as the only difference (placements use the state name and skip `batch_registered`; tasks label spans `task_working_space`).

rust/crates/telemetry/analyzer/src/lib.rs
felipeblazing: Fixed, here and at the `quantity_specs` key in `query_bundle`.

rust/crates/telemetry/model/src/batch.rs
felipeblazing: Same treatment: `batch_consumed_reason` enum with `constexpr to_string_view`.

rust/crates/telemetry/model/src/batch.rs
felipeblazing: Added `enum class batch_origin` with `constexpr to_string_view`, used at all emit sites. `"cpu_source"`/`"pinned_cache"` were never emitted (those paths aren't instrumented) — dropped from the docs; the enum is where they'd be added later.

johanpel (approved):
Thanks for addressing the comments 🚀

PR #898, opened 2026-06-08
Fold pipelineable downstream ops into merge pipelines after GROUP BY and TOP_N

Jedi18:
## Summary

After `HASH_GROUP_BY` / `TOP_N` splitting, Sirius previously created a small merge pipeline (`PARTITION → MERGE_*`) and then a separate downstream pipeline for streaming ops (typically `RESULT_COLLECTOR`). That extra pipeline boundary is unnecessary when downstream is a dead end.

This PR folds pipelineable downstream operators into the merge pipeline when safe, so merge and collector run in a single `gpu_pipeline_task`:

Before: `PARTITION → MERGE_GROUP_BY | RESULT_COLLECTOR `
After: `PARTITION → MERGE_GROUP_BY → RESULT_COLLECTOR`

Fusion is applied from `split_group_aggregate_sink` and `split_top_n_sink` via `attach_merge_and_fuse_downstream`, with three guards in `try_fuse_downstream_pipelineable`:

- **Guard 1:** block when a direct downstream pipeline ends in a structural sink (`ORDER BY`, `HASH_JOIN`, nested `GROUP BY`, etc.) that needs its own `split_*` path
- **Guard 2:** block when a later structural pipeline reads from the candidate downstream’s output (e.g. `GROUP BY → PROJECTION → HASH_JOIN`) — fusion would remove a pipeline boundary that join wiring still depends on
- **Guard 3:** only absorb pipelines whose operators are plain intermediates (`!is_source() && !is_sink()`)

**UNGROUPED_AGGREGATE is explicitly excluded:** the same physical op is both the partial-aggregate sink and the merge pipeline source. Folding downstream broke partial→merge repository wiring and produced wrong results (e.g. `avg` off by orders of magnitude).

Adds `absorbed_pipelines_` so absorbed downstream pipelines are skipped in `split_pipelines()` and not double-scheduled.

Also includes regression tests in `test_pipeline_merge_fusion.cpp`

PR #1190, opened 2026-07-16
Fuse merge operators into downstream pipelines

Jedi18:
## Summary

Fuse eligible `GROUP BY` and `TOP_N` merge operators into their downstream pipeline instead of cutting a pipeline boundary. When a merge's downstream chain is a unary streaming path ending in a fusable sink, the merge joins that sink's pipeline as an intermediate operator, removing a task launch and a repository round-trip (typically `merge → RESULT_COLLECTOR` at the query tail).

Supersedes #898.

## How it works

- Eligibility is decided at plan time by `mark_fusable_merge_pipelines` (`sirius_physical_plan_generator.cpp`), which walks parent pointers from each merge to the first downstream sink. It recurses through `children`, delim-join subtrees, and the result collector's plan.
- Marked merges override `build_pipelines` to add themselves to the current pipeline and recurse into their child, which still cuts the upstream boundary.
- Controlled by the new `fuse_merge_pipelines` setting (default **on**).

## Fusion policy

**Fused** — total-input structural sinks already buffer their full input, so folding the merge in does not change when they observe complete data:
- Terminal merges into `RESULT_COLLECTOR`
- `MERGE_GROUP_BY` into a downstream `ORDER_BY`, `TOP_N`, or an outer `GROUP BY`

**Not fused** (standalone merge boundary preserved):
- Join / CTE / delim terminals (`HASH_JOIN`, `NESTED_LOOP_JOIN`, `CTE`, `LEFT`/`RIGHT_DELIM_JOIN`) — multiple inputs or bespoke sink wiring
- Partition sinks (`PARTITION`, `SORT_PARTITION`) — require complete upstream input in one task
- Delim-owned distinct merges — need dedicated sink wiring
- `MERGE_AGGREGATE` (ungrouped aggregate) — uses a different partial-result handoff

## Tests

`test/cpp/pipeline/test_merge_pipeline_fusion.cpp`:
- **Pipeline shape** — terminal `GROUP BY`/`TOP_N` fold into the collector; the flag off restores the merge boundary; ungrouped-aggregate and join terminals stay unfused; structural-sink fusion (`GROUP BY→ORDER BY`, `GROUP BY→TOP_N`, nested `GROUP BY`) lands in the sink's pipeline.
- **Correctness** — each case runs on the GPU with fusion on vs off, asserts both executed transparently, and verifies identical results. Toggling via `SET fuse_merge_pipelines` also exercises the extension-option wiring end to end.

## Docs

- `configuration.md` — new `fuse_merge_pipelines` setting
- `physical-plan-generation.md` — new "Merge fusion" section + notes on the `HASH_GROUP_BY`/`TOP_N` shapes
- `optimizations.md` — "Merge Pipeline Fusion" entry

### DIFF
--- a/src/include/op/sirius_physical_top_n_merge.hpp
+++ b/src/include/op/sirius_physical_top_n_merge.hpp
@@ -62,10 +62,20 @@ class sirius_physical_top_n_merge : public sirius_physical_operator {
  public:
   bool is_sink() const override { return true; }
 
+  //! Allow this merge to join its downstream pipeline.
+  void set_fuse_into_parent(bool fuse) noexcept { _fuse_into_parent = fuse; }
+  [[nodiscard]] bool fuse_into_parent() const noexcept { return _fuse_into_parent; }
+
+  void build_pipelines(pipeline::sirius_pipeline& current,
+                       pipeline::sirius_meta_pipeline& meta_pipeline) override;
+
   std::unique_ptr<operator_data> execute(const operator_data& input_data,
                                          rmm::cuda_stream_view stream) override;
 
   std::unique_ptr<operator_data> get_next_task_input_data() override;
+
+ private:
+  bool _fuse_into_parent = false;
 };
 
 }  // namespace op
--- a/src/include/op/sirius_physical_top_n_merge.hpp
+++ b/src/include/op/sirius_physical_top_n_merge.hpp
@@ -25,6 +25,9 @@ struct DynamicFilterData;
 }  // namespace duckdb
 
 namespace sirius {
+namespace planner {
+class sirius_physical_plan_generator;
+}  // namespace planner
 namespace op {
 
 //! Represents a physical ordering of the data. Note that this will not change
@@ -62,8 +65,8 @@ class sirius_physical_top_n_merge : public sirius_physical_operator {
  public:
   bool is_sink() const override { return true; }
 
-  //! Allow this merge to join its downstream pipeline.
-  void set_fuse_into_parent(bool fuse) noexcept { _fuse_into_parent = fuse; }
+  //! Whether this merge joins its downstream pipeline. Set only during planning (see the friend
+  //! below); read-only thereafter to honor the no-mutation-after-planning contract.
   [[nodiscard]] bool fuse_into_parent() const noexcept { return _fuse_into_parent; }
 
   void build_pipelines(pipeline::sirius_pipeline& current,
@@ -75,6 +78,10 @@ class sirius_physical_top_n_merge : public sirius_physical_operator {
   std::unique_ptr<operator_data> get_next_task_input_data() override;
 
  private:
+  //! The plan-generator is the sole writer of the fusion flag, and only at plan time.
+  friend class sirius::planner::sirius_physical_plan_generator;
+  void set_fuse_into_parent(bool fuse) noexcept { _fuse_into_parent = fuse; }
+
   bool _fuse_into_parent = false;
 };
 
--- a/src/include/op/sirius_physical_top_n_merge.hpp
+++ b/src/include/op/sirius_physical_top_n_merge.hpp
@@ -65,8 +65,7 @@ class sirius_physical_top_n_merge : public sirius_physical_operator {
  public:
   bool is_sink() const override { return true; }
 
-  //! Whether this merge joins its downstream pipeline. Set only during planning (see the friend
-  //! below); read-only thereafter to honor the no-mutation-after-planning contract.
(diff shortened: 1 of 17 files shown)


src/planner/sirius_physical_plan_generator.cpp
```diff
@@ -756,6 +756,74 @@ void sirius_physical_plan_generator::set_parent_ops(sirius::op::sirius_physical_
...
+  for (auto* cur = merge.get_parent_op(); cur != nullptr; cur = cur->get_parent_op()) {
+    if (cur->is_sink()) { return terminal_sink_supports_fusion(*cur); }
+    if (cur->is_source() || cur->children.size() != 1) { return false; }
+  }
+  return false;
+}
+
+}  // namespace
+
+void sirius_physical_plan_generator::mark_fusable_merge_pipelines(
```
kevkrist, line 803: I think it would be better to capture the global once and pass it through recursive invocations of `mark_fusable_merge_pipelines()`. Grab it at engine initialization from the context. I think it's possible for this flag to get mutated by a concurrent connection invoking SET on this configuration variable, in which case the recursion could be inconsistent.

src/include/op/sirius_physical_top_n_merge.hpp
```diff
@@ -62,10 +62,20 @@ class sirius_physical_top_n_merge : public sirius_physical_operator {
  public:
   bool is_sink() const override { return true; }
 
+  //! Allow this merge to join its downstream pipeline.
+  void set_fuse_into_parent(bool fuse) noexcept { _fuse_into_parent = fuse; }
```
kevkrist: [nit] Consider making this private and friending it with the plan-generator, since this flag shouldn't be mutated post-planning. Then consider calling fuse_into_parent() method to honor the no-mutation-after-planning contract.

test/cpp/pipeline/test_merge_pipeline_fusion.cpp
```diff
@@ -0,0 +1,507 @@
...
+  std::vector<std::vector<std::string>> rows;
+  for (duckdb::idx_t r = 0; r < result.RowCount(); r++) {
+    std::vector<std::string> row;
+    row.reserve(result.ColumnCount());
+    for (duckdb::idx_t c = 0; c < result.ColumnCount(); c++) {
+      row.push_back(result.GetValue(c, r).ToString());
+    }
+    rows.push_back(std::move(row));
+  }
+  std::sort(rows.begin(), rows.end());
```
kevkrist: Don't you want to *not* sort when the rows are supposed to be emitted in order?

test/cpp/utils/pipeline_conversion_test_utils.cpp
```diff
@@ -121,7 +136,7 @@ void with_conversion_result(
     duckdb::unique_ptr<duckdb::LogicalOperator> logical_plan;
     {
       optimizer_disable_guard guard(context);
-      logical_plan = extract_logical_plan_sirius_order(context, query);
+      logical_plan = extract_logical_plan_sirius_order(context, query).logical_plan;
```
kevkrist, line 139: This stops at create_plan, and mark_fusable_merge_pipelines never gets called here, as far as I can tell, and the fuse_merge_pipelines = false path is effectively always taken. If fusion is default on, we should capture that in the plans generated here.

test/cpp/pipeline/test_merge_pipeline_fusion.cpp
```diff
@@ -0,0 +1,507 @@
...
+    sirius::test::with_initialized_engine(
+      *con,
+      "SELECT n_nationkey FROM nation ORDER BY n_nationkey DESC LIMIT 3",
+      [&](sirius::sirius_engine& engine) {
+        require_unfused_terminal_shape(engine, SiriusPhysicalOperatorType::MERGE_TOP_N);
+      });
+  }
+}
+
+TEST_CASE_METHOD(merge_fusion_fixture,
```
kevkrist, line 340: [Claude] Both sections pass regardless of the policy: the walk never reaches the T::HASH_JOIN branch (it stops at the PROJECTION under PARTITION_join, and merge != join holds either way), and MERGE_AGGREGATE is never marked at all, so deleting either switch case leaves this green. The two guards that actually fire — T::CTE and the owning_delim_join() early-return — have no coverage; could we add a CTE-body GROUP BY and a correlated subquery that each fail if the guard is removed? 
(I dunno lol)

kevkrist (approved):
nits

Jedi18:
Thanks @kevkrist! I've addressed all the review comments

PR #1194, opened 2026-07-16
SNMG: Introducing Broadcast join and BUILD_PROBE mode for multiple partitions with multiple GPUs

### LINKED ISSUE #1141: Add ability for HASH_JOIN_MODE::BUILD_PROBE to support more than one partition in multi-gpu mode
HASH_JOIN_MODE::BUILD_PROBE currently only supports one partition. This is because you have to build the hash table and then other batches can join against the same hash table. With only one GPU, the logic of allowing partitioning means you have the challenge of having to create a separate hash table for each partition, and what this implies for memory management and lifetime management is tricky, so we only support one partition.
In the multi-gpu case, we can do partitioning of one partition per GPU, and in this case, the memory resource and lifetime management should be manageable because you have one hash table per GPU.

NOTE: at some point we will likely want to ability to make a number of partitions that is not a multiple of the number of GPUs. For the purposes of this issue, that is a separate item of work

wmalpica:
Resolves #1141 

## Multi-GPU multi-partition BUILD_PROBE joins + broadcast small-build joins

Extends the GPU hash join to make real use of multiple GPUs by allowing the BUILD_PROBE mode to be used with multiple partitions (one per GPU). And in the case of small build tables, it implements a BROADCAST mode where it broadcasts the small table to all GPUs. Before this, for small tables it would just move everything (all build and all probe) into a single GPU.

Benchmarks show that on 4 GB200s at SF1K (pinned on host) total time went from 25.1 sec to 20.7 sec with this PR.

### What changed

**1. Multi-partition BUILD_PROBE (one hash table per GPU).**
BUILD_PROBE previously required exactly one partition (a single hash table on one GPU).
It now admits up to one partition per GPU (`num_partitions <= num_gpus`), building one hash
table per partition/GPU and streaming probe batches against each. On a single GPU
to the historical single-partition behavior.

**2. Broadcast small build tables.**
When a build side is small (`< small_table_bytes`, i.e. `num_gpus * 16 MiB`) on multi-GPU,
instead of funneling the whole build to one GPU the PARTITION operator proposes `num_gpus`
partitions and *replicates* the build to every GPU (the `_broadcast` flag). The build sink
deposits the build batch into every slot; the probe sink routes each batch to the slot for its
current GPU. Once the probe side finishes, any slot that received replicated build data but no
probe rows is discarded, freeing its hash table on the owning GPU.

**3. Dynamic filters under broadcast.**
Dynamic-filter publication was gated to single-partition joins. Because a broadcast join gives
every partition the *full* replicated build, publication is now allowed for broadcast too: the
first GPU whose build batch arrives wins the existing `OPEN -> PUBLISHING` compare-exchange and
publishes once; the others fall out at the CAS before doing any work. Genuinely hash-partitioned
(non-broadcast) multi-partition builds still skip pushdown (each partition holds only a slice).
The existing multi-GPU filter-replication/consumer path is reused unchanged.

**4. Correctness / limit guards.**
- Full `OUTER` joins are excluded from BUILD_PROBE: it streams `full_join` per probe batch and
  would re-emit unmatched build rows on every batch/GPU with no global accumulation -- full outer
  joins run on the STANDARD path.
- Broadcast eligibility charges the **full** replicated build size against
  `max_build_hash_table_bytes` (each GPU builds the entire table), not a per-partition slice.

### Notable APIs / functions

- `build_probe_mode_eligible(...)` - pure eligibility gate; now `num_partitions <= num_gpus`,
  per-GPU size check, and excludes right-family / mixed / full-outer joins. 
- `make_broadcast_partition_decision(...)` -- pure two-phase broadcast decision (candidate +
  proposed partition count, finalized once BUILD_PROBE eligibility is known).
- `select_build_probe_action(...)` -- pure per-partition build/probe scheduling decision.
- `broadcast_slots_to_discard(...)` / `discard_build_only_slots_if_probe_complete()` -- identify and
  free build-only slots after the probe side completes.
- `sirius_physical_partition::slot_for_device(...)` -- map a batch's residence GPU to its slot.
- `update_join_exec_mode(..., is_broadcast_candidate)` and new setters `set_broadcast`,
  `set_num_gpus`, `set_active_gpu_ids`; the sorted active-GPU-id list is plumbed through
  `pipeline_build_context` and the engine.

### Tests

Unit tests for the pure decision helpers (`build_probe_mode_eligible`,
`make_broadcast_partition_decision`, `select_build_probe_action`, `broadcast_slots_to_discard`),
covering multi-partition eligibility, the broadcast candidate/finalize logic, full-outer
exclusion, and the full-build-size limit. Added 2-GPU-gated integration tests asserting broadcast
replication across GPUs and dynamic-filter publication under broadcast (both verify GPU == CPU
results). Docs under `docs/super-sirius/` updated (operators, optimizations, dynamic filters,
physical plan generation).


src/op/sirius_physical_partition.cpp
```diff
@@ -275,6 +305,24 @@ void sirius_physical_partition::set_num_partitions(int num_partitions)
...
+    join_port->repo->set_num_partitions(static_cast<std::size_t>(num_partitions));
+  }
+}
+
+std::size_t sirius_physical_partition::slot_for_device(int device_id) const
+{
+  for (std::size_t i = 0; i < _active_gpu_ids.size(); ++i) {
+    if (_active_gpu_ids[i] == device_id) { return i; }
+  }
+  return 0;  // fallback: keep the batch on a valid slot if the device is unexpectedly absent
```
Jedi18: nit: should we also log a warning here?

src/op/sirius_physical_partition.cpp
```diff
@@ -345,11 +404,27 @@ std::unique_ptr<operator_data> sirius_physical_partition::get_next_task_input_da
...
+      // (non-right, non-mixed, folds to one build batch); right/mixed joins reject it and fall back
+      // to the normal count. Only the build side can drive broadcast (sizing_partition._is_build).
+      // See make_broadcast_partition_decision for the pure, unit-tested decision.
+      auto const num_gpus = _active_gpu_ids.size();
+      auto const decision = make_broadcast_partition_decision(
+        sizing_partition._is_build, num_gpus, total_bytes, _small_table_bytes, num_parts);
+
       if (sizing_partition._is_build) {
-        hash_join.update_join_exec_mode(num_parts, total_bytes, build_foldable);
+        hash_join.update_join_exec_mode(
```
Jedi18: There might be a bug here where we need to resize join repositories before enabling BUILD_PROBE

`update_join_exec_mode()` currently enables `BUILD_PROBE` and creates N partition states before the build and probe repositories are resized to N partitions.

During that short window `get_next_task_hint()` can call `snapshot_build_probe_slots()`. It loops over all N states and calls `repo->size(p)`, but a new cuCascade repository contains only partition 0. Accessing partition 1 or higher throws `std::out_of_range` and does not return zero.

We should resize both join input repositories before making BUILD_PROBE observable - ideally inside `update_join_exec_mode()` while holding `op_state_mutex`, with `_join_mode` set last. Also need to update the comment claiming size(p) safely returns zero, which is incorrect

src/include/op/sirius_physical_hash_join.hpp
```diff
@@ -192,21 +279,40 @@ class sirius_physical_hash_join : public sirius_physical_partition_consumer_oper
...
+  // that only read the table. `build_state` is atomic so get_next_task_hint can observe a slot's
+  // progress without holding op_state_mutex while execute() flips it.
+  struct per_partition_build_state {
+    std::atomic<BUILD_HASH_TABLE_STATE> build_state{BUILD_HASH_TABLE_STATE::NOT_BUILT};
+    std::unique_ptr<cudf::hash_join> hash_table;  // general path (INNER/LEFT/OUTER)
+    std::unique_ptr<cudf::distinct_hash_join>
+      distinct_hash_table;  // used instead of hash_table when build keys are proven unique
+    std::unique_ptr<cudf::filtered_join>
+      filtered_table;             // reusable build-on-right object for MARK/SEMI/ANTI joins
+    bool build_has_null = false;  // whether the build/right side has a NULL in any join key column;
```
Jedi18: The new multi-partition BUILD_PROBE path records `build_has_null` separately for each partition. For MARK/IN semantics, NULL presence must cover the complete build side: a NULL in partition A must make an unmatched probe row in partition B evaluate to NULL, not FALSE.

We should either aggregate a join-wide NULL flag before probing or exclude non-broadcast multi-partition MARK joins from BUILD_PROBE (if the same issue exists on the standard path then can be fixed in a single PR targeting both). 

Note that the existing standard path may have a similar issue, but this new execution mode should not extend it.

src/op/sirius_physical_partition.cpp
Jedi18: Thanks, I've verified that this has been fixed now

src/op/sirius_physical_hash_join.cpp
```diff
@@ -439,33 +447,186 @@ void sirius_physical_hash_join::build_pipelines(pipeline::sirius_pipeline& curre
...
+    uint64_t const per_gpu_build_bytes =
+      broadcast_candidate ? total_bytes : total_bytes / static_cast<uint64_t>(proposed);
+    build_probe = per_gpu_build_bytes < max_build_hash_table_bytes && build_foldable &&
+                  !is_right_family && !is_mixed && !is_full_outer;
+  }
+
+  bool const broadcast = broadcast_candidate && build_probe;
+  // MARK single-GPU is clamped to one partition regardless of the natural count; every other case
+  // takes num_gpus when broadcasting and the natural count otherwise.
+  int const num_partitions = broadcast ? num_gpus : ((is_mark && num_gpus <= 1) ? 1 : natural);
```
Jedi18, line 572: There is a bug here where large MARK joins can fall into an invalid partitioned path.

When forced broadcast is rejected—for example, because the complete build exceeds `max_build_hash_table_bytes`—broadcast becomes `false` and this returns the natural `partition` count. On multi-GPU, that can be greater than one.

Different build partitions can legitimately have different NULL states, but `set_build_has_null()` requires every call to agree and throws otherwise.

This is not limited to large multi-GPU builds. Any MARK join that remains in STANDARD/MIXED mode can process multiple build batches—including non-foldable and single-GPU cases—and those batches may also disagree about NULL presence.

Clamping to one partition alone is therefore insufficient because it does not guarantee one complete build batch. We should either:

* guarantee one complete build before probing, when memory permits; or
* reject/downgrade unsupported MARK executions.

A future multi-GPU implementation can globally reduce NULL presence before emitting marks. The “forcing broadcast” warning should also only fire when broadcast is actually selected.

src/op/sirius_physical_hash_join.cpp
```diff
@@ -989,38 +1269,44 @@ std::unique_ptr<operator_data> sirius_physical_hash_join::execute(const operator
...
             join_type == duckdb::JoinType::ANTI) {
           // MARK/SEMI/ANTI: build a reusable filtered_join on the right (filter) keys; each probe
           // batch's semi_join/anti_join returns left-row match indices (scattered into a BOOL8 mark
           // for MARK, gathered as the output rows for SEMI/ANTI).
-          _filtered_table = make_right_filtered_join_ptr(build_keys, stream);
+          slot.filtered_table = make_right_filtered_join_ptr(build_keys, stream);
           // Record whether the build keys contain a NULL; MARK three-valued logic needs it at probe
           // time, but the build keys are not retained beyond this scope.
-          _build_has_null = table_has_any_null(build_keys);
+          set_build_has_null(_build_has_null, table_has_any_null(build_keys));
```
Jedi18: This if branch handles MARK, SEMI, and ANTI joins, but calls `set_build_has_null()` unconditionally. In multi-partition `BUILD_PROBE`, partitions can legitimately differ in `NULL` presence, causing the consistency check to throw.

SEMI and ANTI joins never use `_build_has_null`, so this should be restricted to MARK:
```
if (join_type == duckdb::JoinType::MARK) {
  set_build_has_null(_build_has_null, table_has_any_null(build_keys));
}
```

src/op/sirius_physical_hash_join.cpp
```diff
@@ -474,52 +635,116 @@ bool sirius_physical_hash_join::is_build_probe_mode()
...
+  if (!build_port || !probe_port) {
+    throw std::runtime_error(
+      "In sirius_physical_hash_join:snapshot_build_probe_slots: missing expected ports in "
+      "operator " +
+      std::to_string(this->get_operator_id()));
+  }
+  std::vector<build_probe_slot_view> slots(_partition_build_states.size());
+  for (std::size_t p = 0; p < _partition_build_states.size(); ++p) {
+    slots[p].state = _partition_build_states[p].build_state.load(std::memory_order_acquire);
+    // repo->size(p) safely returns 0 for partitions whose data has not been produced yet.
```
Jedi18: Nit: this comment is a bit misleading. `repo->size(p)` returns 0 for an in-range partition that is still empty, but throws `std::out_of_range` if `p >= num_partitions()`. After the repo pre-size fix that is fine here, but the comment should not read as if out-of-range is safe

src/include/op/sirius_physical_hash_join.hpp
Jedi18: The single gpu and multi-gpu mark join broadcast are correct but there is a fix needed for when large build that is too large to broadcast. I've posted a review comment that goes into the details.

src/op/sirius_physical_hash_join.cpp
wmalpica: Good catch, that comment was added a while back, when Claude had modified the size function to be "safe" when calling out of range. But i changed that so that it would throw instead, but the comment stayed.

Jedi18:
Thanks looks good now, I think the tests need to be updated though. Also just two minor nits:

1)`if (broadcast && !build_probe) throw` is unreachable after `broadcast = broadcast_candidate && build_probe`.

2) MARK now forces BUILD_PROBE even when the build exceeds `max_build_hash_table_bytes`. That’s fine for correctness but can OOM instead of downgrading. Maybe we should create an issue to investigate this in the future for other designs?

Jedi18 (approved):
LGTM once the tests are updated and CI is green

PR #1203, opened 2026-07-16
Removed cuCascade related offset fixup and added array type to estimate_decoded_bytes_budget

yayen-lin:
### Summary
- Removed cuCascade related offset fixup
- Added ARRAY type to `estimate_decoded_bytes_budget`

### Files Changed
- **`src/helper/list_offset_fixup.{hpp, cpp}`, `src/include/pipeline/batch_lock_utils.hpp`**: Removed from the codebase. This workaround for the cuCascade fix ([#147](https://github.com/NVIDIA/cuCascade/issues/147)) is pulled into the submodule.
- **`src/op/scan/duckdb_native_metadata.cpp`**: Add the missing decode budget estimation calculation for ARRAY.
- **`CMakeLists.txt`**: Removed list_offset_fixup.cpp.
- **`test/cpp/integration/test_gpu_execution_array.cpp`**: Added the unit test for decode budget estimation calculation.


ran-yuan-rui:
It looks like this PR is missing commit `645aa7edc90d1282167522faa829ac44570bc426` (`Added array type to estimate_decoded_bytes_budget`).
That commit contains the ARRAY branch described in the PR title and summary, but it is an ancestor of neither the current base nor head. The current PR diff only removes the LIST offset workaround (`0 additions / 189 deletions`).

Without that commit, `estimate_decoded_bytes_budget()` still calls `fixed_width_byte_size()` for ARRAY, which throws when row-group statistics prune a row group while an ARRAY column is projected.

Could you add the missing commit to this PR? A regression test covering ARRAY projection with scalar-filter row-group pruning would also be helpful.

yayen-lin:
Thank you Yu-Teng for the catch I really appreciate it 🙏 Will do the regression test in a bit!

Update: regression test added

mbrobbel (approved):

yayen-lin:
Thank you @mbrobbel @ran-yuan-rui for your review! Happy Friday and have a good weekend all.

PR #1204, opened 2026-07-16
feat(vss): add exact brute-force vector search (ENN)

yayen-lin:
## Description

Adds `sirius_knn_search`, a GPU table function for exact brute-force (ENN) vector
search over a GPU-pinned table.

### Heads-up

0. Most of the diff is not core logic: **~2810** lines are `pixi.lock` / `vcpkg` / `CMake` / etc., and **~1217** are tests, so ~4000 of the changed lines are that. The actual code logic is only **~1225** lines (~500 of which are header declarations).
1. Only the DuckDB-native scan path is supported (parquet has no vectors, so it's unsupported for now).
2. Supported distance metrics: euclidean `l2` and `cosine`.
3. A sliced vector column and row-level nulls are compacted with a boolean mask. The implication is that if `result size < k`, the result has fewer rows than the requested `k`. Element-level nulls are rejected for now as it's rarely a case and would degrade the performance.
4. Avoided using concat.

### Summary

- **New table function**: `sirius_knn_search`, **requires** the table to be pinned on GPU.
- **Memory**: allocate-on-demand, but still routed through the pinned table's GPU `memory_space` allocator, so every transient is tracked by the same memory manager and freed when the query ends.
- **Impl**: local search on each pinned table batch (zero-copy reinterpreted to RAFT's device matrix view, which cuVS uses under the hood), then a k-way merge into the global top-k result.

### Sample queries

<details>
    <summary><strong>Expand to see sample queries</strong></summary>

```bash
# start Sirius with a persistent file
SIRIUS_LOG_LEVEL=debug ./build/release/duckdb test.duckdb
```

```sql
-- creating dataset
CREATE TABLE test_arrays (id INTEGER, vec FLOAT[3]);
INSERT INTO test_arrays SELECT i, [i::float, (i+1)::float, (i+2)::float] FROM range(5000) t(i);
CHECKPOINT; -- checkpoint is needed

-- cache the table
SELECT * FROM pin_table(name => 'test_arrays', tier => 'gpu', format => 'duckdb');

-- sample queries
SELECT * FROM sirius_knn_search(
    'test_arrays',                    -- table   (positional; must be GPU-pinned)
    'vec',                            -- column  (positional)
    [4980,4981,4982]::FLOAT[3],       -- query   (positional)
    k              => 20,             -- default 10
    output_columns => ['id','vec'],   -- default to pinned columns
    metric         => 'l2',           -- default 'l2' ('l2', 'cosine')
);

SELECT * FROM sirius_knn_search(
  'test_arrays',
  'vec',
  [4980,4981,4982]::FLOAT[3],
  k => 20
);
```

</details>

### Changes

**Table Function**

- **`src/sirius_extension.cpp`**: Register the table function with bind & execute logic.

**cuVS / cuDF interop**

- **`src/include/vss/cudf_raft_interop.hpp`, `src/cuda/vss/cudf_raft_interop.cu`**: zero-copy reinterpretation of a fixed-size FLOAT LIST column into a `[n_rows, dim]` raft device matrix view; rejects nulls, slices, and non-FLOAT children.

**ENN search**

- **`src/include/vss/brute_force_search.hpp`, `src/cuda/vss/brute_force_search.cu`**: Exact brute-force search kernel.
- **`src/include/vss/distance_metric.hpp`, `src/vss/distance_metric.cpp`**: Helper functions routing the metric to the corresponding distance function.
- **`src/include/vss/enn_top_k.hpp`, `src/vss/enn_top_k.cpp`**: Per-chunk brute-force top-k (`compute_enn_top_k`, with null/sliced-row compaction) and the k-way merge of per-chunk candidates into the global top-k (`merge_enn_top_k`).
- **`src/include/vss/pinned_column.hpp`, `src/vss/pinned_column.cpp`**: Borrow a pinned table column's GPU chunks as zero-copy `cudf::column_view`s in pin order; rejects missing columns and multi-GPU spread.
- **`src/include/vss/vector_search.hpp`, `src/vss/vector_search.cpp`**: Top-level entry (`run_vector_search`), resolves GPU/host space and pinned table, uploads the query vector, dispatches ENN, and converts the result to host.
- **`src/vss/vector_search_enn.cpp`**: ENN implementation; per-chunk brute-force search, then merge and slice into the global top-k.
- **`src/include/vss/vector_search_internal.hpp`**: Shared `vector_search_context` (resolved handles + request) and the ENN impl declarations.

**Build**

- **`pixi.toml` / `pixi.lock`**: add the `libcuvs` (26.06) dependency; `libraft-headers` comes in transitively (used by the pixi / merge-queue build path).
- **`CMakeLists.txt`**: add the ENN sources and test files to the build; `find_package(raft REQUIRED CONFIG)` before cuvs and link `raft::raft`, since cuvs's config does `find_dependency(raft)` and the ENN sources include raft headers directly; alias `cuvs::cuvs` to `cuvs::cuvs_static` so the static vcpkg build and the shared conda build link the same target name.
- **`scripts/generate_nightly_env.sh`**: arm64 nightly env fix.

**Tests**

Unit tests

- **`test/cpp/vss/test_brute_force_search.cpp`**: exact nearest ids + distances (L2 & cosine), and rejection of out-of-range k / mismatched dimensions.
- **`test/cpp/vss/test_cudf_raft_interop.cpp`**: zero-copy view + malformed-input rejection (non-LIST, dim mismatch, parent-row nulls, element-level nulls, slices).
- **`test/cpp/vss/test_distance_metric.cpp`**: metric-routing helper validation.
- **`test/cpp/vss/test_enn_top_k.cpp`**: per-chunk top-k (including null/sliced compaction) and the global k-way merge.
- **`test/cpp/vss/test_pinned_column.cpp`**: zero-copy chunk views in pin order; missing-column and different-GPU rejection.
- **`test/cpp/vss/test_vector_search.cpp`**: empty-output schema helper and rejections.

Integration tests

- **`test/cpp/integration/test_gpu_execution_vector_search.cpp`**: end-to-end ENN through the SQL table function, compared against a CPU reference; plus `output_columns` defaulting/validation and element-level null coverage.

**vcpkg**

The ci-release (distribution) build compiles cuVS and its RAPIDS deps statically from source via overlay ports, so the single-file extension carries no runtime `.so` dependency on them.

- **`vcpkg_ports/cuvs/portfile.cmake`, `vcpkg_ports/cuvs/vcpkg.json`, `vcpkg.json`**: cuvs overlay port (builds cuvs statically from source), plus the x64 build fix and removal of an unused runtime var. `cuvs/vcpkg.json` and the top-level `vcpkg.json` now also declare `raft` as a dependency.
- **`vcpkg_ports/raft/portfile.cmake`, `vcpkg_ports/raft/vcpkg.json`**: raft overlay port. cuvs's installed config does `find_dependency(raft)` and Sirius includes raft headers directly, so raft must be a findable package in the vcpkg tree. Built header-only (`RAFT_COMPILE_LIBRARY=OFF`); drops its duplicate `rapids_logger` files so the `rmm` port stays the sole owner (raft still resolves rapids_logger via `find_dependency` against rmm's install).

## Checklist
- [x] Read CONTRIBUTING.md

## References
Refs #1446 (ARRAY pin support this builds on)
Refs sirius-db/extension-ci-tools#2 (CI Docker image: add cuBLAS/cuSOLVER/cuSPARSE for the cuVS distribution build)


CMakeLists.txt
```diff
@@ -88,6 +88,7 @@ endforeach()
 
 # Dependencies
 find_package(cudf REQUIRED CONFIG)
+find_package(cuvs REQUIRED CONFIG)
```
ran-yuan-rui, line 91: This will fail the merge-queue (Distribution/vcpkg) build. This `find_package(cuvs REQUIRED CONFIG)` is unconditional and `cuvs::cuvs` is linked below, but the vcpkg side is missing both halves: there is no `cuvs` entry in `vcpkg.json`, and no `vcpkg_ports/cuvs` overlay port (current overlay ports: cccl, cuco, cudf, kvikio, nanoarrow, numactl, nvcomp, nvtx3, rmm). Only `pixi.toml` gained `libcuvs`.

On this PR the `distribution` cuda-12 / cuda-13 jobs show **"skipping"** — the 5-second `distribution` check that goes green is just the gate job. The real vcpkg `ci-release` builds run in `merge_group`, where CMake configure will fail with *"Could not find a package configuration file provided by cuvs"*. 

**Fix:** add a `vcpkg_ports/cuvs` overlay port (pattern it after `vcpkg_ports/cudf`, same RAPIDS 26.06 pin) and add `"cuvs"` to `vcpkg.json`. Prefer that over compiling the feature out of the `ci-release` build, which would make the shipped artifact diverge from the pixi dev build.

src/vss/distance_metric.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include "vss/distance_metric.hpp"
+
+#include <stdexcept>
+#include <string>
+
+namespace sirius::vss {
+
+cuvs::distance::DistanceType enn_distance_type_from_metric(std::string_view metric)
+{
+  if (metric == "l2sq") { return cuvs::distance::DistanceType::L2SqrtUnexpanded; }
```
ran-yuan-rui: `l2sq` maps to `L2SqrtUnexpanded` — the square-rooted L2, not the squared one. The integration test confirms it by matching `array_distance` (also unsquared). DuckDB-VSS uses `l2sq` for the squared distance, so this ships the same name with different values.

Pick one: rename the metric to `l2` and keep this mapping, or keep `l2sq` and map it to `L2Unexpanded`. Update the unit test name ("maps l2sq to the unexpanded form") to match.

src/sirius_extension.cpp
```diff
@@ -1352,6 +1366,163 @@ static void SiriusSetQueryLabelFunction(ClientContext& context,
...
+
+  // Resolve the vector column's dimensionality and each output column's type
+  // from the catalog so the return schema and the host reader agree.
+  auto const qname          = QualifiedName::Parse(req.table_name);
+  std::string const catalog = qname.catalog;
+  std::string const schema  = !qname.schema.empty() ? qname.schema : schema_name;
+  auto& entry_base =
+    Catalog::GetEntry(context, CatalogType::TABLE_ENTRY, catalog, schema, qname.name);
+  auto& entry             = entry_base.Cast<DuckTableEntry>();
+  req.table_name          = entry.name;  // catalog-resolved name (matches query-side derivation)
```
ran-yuan-rui, line 1686: This resolves catalog + schema + table, then overwrites `req.table_name` with the bare name. Runtime then looks the pin up by that bare name (`find_pinned_entry(req.table_name)` in `vector_search.cpp`).

Two same-named tables in different schemas, one pinned: bind validates dim/types against one, the search runs over the other's pinned data — silent wrong result when the dims match, opaque error when they don't. `find_pinned_entry_for_duckdb_table(catalog, schema, table)` already exists on dev; carry catalog + schema in the request and use it instead of the bare-name lookup.

yayen-lin:
Hi @ran-yuan-rui, thank you for a good round of reviews! All of them were fixed! 🙏

ran-yuan-rui:
Comments as below:

### Snapshot and identity

Please confirm that `sirius_knn_search` searches the GPU-pinned snapshot, so INSERT, UPDATE, or DELETE operations after pinning require the table to be pinned again. Also confirm that `catalog.schema.table` is the identity boundary for this version.

### Global top-k and later payload materialization

This could be a perf issue. Use late materialization: compute the global top-k using only row IDs and distances, then fetch the full payload for the final results.

```text
ENN/IVF backend
    -> {global_row_id, distance}
    -> bounded global top-k
    -> group final results by owner chunk
    -> gather payload for only the final k rows
    -> host materialization
```

This reduces wide-payload processing from:

```text
O(C * k * row_width)
```

to:

```text
O(k * row_width)
```

### GPU memory plan and control

Do we need a short design covering the memory estimate, when the reservation is acquired, behavior when reservation fails, and release order on success and error paths.

### Multi-GPU support (in the future)

There is one issue to clarify: `pin_table` distributes chunks across GPUs, while vector search currently selects the first GPU. A multi-chunk pin may therefore be unusable in a multi-GPU configuration.

Restrict this function to a single-GPU topology, allow pinning to a specified GPU, or compute local top-k on each GPU and merge the results?

### Empty tables

A valid zero-chunk pin cannot currently return an empty result because output types are inferred from the first data chunk. The empty result schema should come from bind or pin metadata instead.

The execution path should be:

```text
ResolvedVectorSearchRequest
    -> PinSnapshotLease
    -> backend returns narrow neighbor references
    -> canonical, memory-bounded global top-k
    -> final payload gather
    -> D2H and stream synchronization
```

src/sirius_extension.cpp
```diff
@@ -1352,6 +1366,165 @@ static void SiriusSetQueryLabelFunction(ClientContext& context,
...
+  req.schema              = entry.ParentSchema().name;
+  req.table_name          = entry.name;  // catalog-resolved name (matches query-side derivation)
+  auto const& columns     = entry.GetColumns();
+  auto const schema_names = columns.GetColumnNames();
+  auto const schema_types = columns.GetColumnTypes();
+
+  // Default output_columns to every base-table column (in schema order) when omitted.
+  if (req.output_columns.empty()) {
+    req.output_columns.assign(schema_names.begin(), schema_names.end());
+  }
```
aocsa, line 1694: I wrote a quick regression test for this — [gist](https://gist.github.com/aocsa/20a347b21ec8a14cfe00f0541a26eae9):

Drop it into test/cpp/integration/test_gpu_execution_vector_search.cpp and run:

pixi run build/release/extension/sirius/test/cpp/sirius_unittest

"sirius_knn_search - default output_columns on a subset-pinned table"

It pins t(id, vec FLOAT[3], payload) with cols => ['id','vec'], then does a plain
sirius_knn_search with no output_columns:

CHECK_FALSE( r->HasError() ) ... FAILED
default output_columns error: Invalid Error: VSS: pinned table missing column 'payload'

The control in the same test passes (output_columns => ['id'] → 5 rows), so the pin is
fine — it's specifically the catalog-derived default. A second case shows
output_columns => [] returning all four columns instead of being rejected.

I think this path needs improvement: defaulting to every catalog column is unsatisfiable whenever pin_table(cols => [...]) pinned a subset, and it surfaces as Invalid Error
(untyped, since sirius::internal_exception isn't a duckdb::Exception) after the GPU
space and pin are already set up. Validating in bind against the pin would fix  it.

aocsa (changes requested):
Overall LGTM, and core implementation look solid.


Some feedback comments. 

**1. `output_columns` defaults off the catalog, not the pin.** `pin_table(cols => [...])`
pins a subset, so the default at `sirius_extension.cpp:1465-1467` is unsatisfiable on any subset pin — the no-args call dies at execution with `Invalid Error: VSS: pinned table
missing column 'x'`. Detailed comment inline, plus a regression test in the gist below.

**2. Element-level null rejection is untested.** `enn_top_k.cpp:73-74` relies on it, but the only null test builds a null *parent row* — `cudf_raft_interop.cu:45` never runs.

yayen-lin:
Hey @aocsa, thanks for the review! I've addressed both points and the PR is ready for another look.

1. `output_columns` now defaults to the pinned columns with validation.
2. Element-level null rejection is now tested.

Thank you again I appreciate it!

yayen-lin:
Hey @ran-yuan-rui, I really appreciate your follow-up comment. You're sketching the right long-term shape for this path. One heads-up is that I was targeting this PR as a working prototype of the vector search path, so I've kept a few of the deeper items as follow-up rather than folding them in here.

Thanks again, this is a very useful map of where this goes next. And my apologies on the late response 🙏.

aocsa (approved):
LGTM

aocsa:
@9prady9 is helping us land this PR to fix the CI issue related to cuBLAS, one of cuVS’s required libraries. cc @yayen-lin

9prady9:
> @9prady9 is helping us land this PR to fix the CI issue related to cuBLAS, one of cuVS’s required libraries. cc @yayen-lin

Basically, the required dependency is missing the concerned images that are used for merge queue builds. Hence you see that only there. I don't think there is any CMake fix required here. Just having the images updated with required dependencies should do.

yayen-lin:
From the research I did yesterday it looked like we might need to make some changes to [extension-ci-tools](https://github.com/sirius-db/extension-ci-tools) to add the 3 match libraries that cuVS needs (libcublas, libcusolver, and libcusparse). @aocsa @9prady9 Would it be okay if I go and make a branch and a PR there?

9prady9:
> From the research I did yesterday it looked like we might need to make some changes to [extension-ci-tools](https://github.com/sirius-db/extension-ci-tools) to add the 3 match libraries that cuVS needs (libcublas, libcusolver, and libcusparse). @aocsa @9prady9 Would it be okay if I go and make a branch and a PR there?

I missed something earlier, we need to add `libcublas-dev` to our pixi.toml. It might take care of the merge queue CI jobs too depending on how the dependencies are brought in for that run.

About the `extension-ci-tools` sure we can upstream it there as long as it is sirius specific branch/fork. Otherwise, it would be incorrect place. But I honestly don't get one thing here, why would we need over there ? because duckdb extensions are single files and that forces us to build all of these dependencies as static versions. Isn't it ? 

The issue is only build time related. @yayen-lin try adding the `libcublas-dev` in pixi.toml - it might just work. I wonder if the cmake target will pick up static library or not.

If you are okay with it, I will test this out tomorrow first thing and let you know. In the meantime, you can try the pixi toml change.

yayen-lin:
@9prady9 Will do, thank you Pradeep! Will follow up on this in a bit.

yayen-lin:
Hi @9prady9 , quick follow-up. I dug into the pixi.toml idea and I don't think it'll fix the failing job.

The build that's red is the vcpkg distribution build in the docker image. That path compiles cuVS from source as a static lib and gets its CUDA from the image's yum install. So a dependency added to pixi.toml will not reach that build.

And the pixi env already has the three math libs, where libcublas-dev, libcusolver-dev, and libcusparse-dev are already resolved in pixi.lock, so there's nothing to add to the pixi side.

I put up a PR that adds those three to the cuda toolchain case in both Dockerfiles: https://github.com/sirius-db/extension-ci-tools/pull/2

Whenever you get a chance to test the image build against it, Thanks!

mike-wendt:
@yayen-lin I've merged the PR https://github.com/sirius-db/extension-ci-tools/pull/2 to help with the build error. Your `pixi.lock` is now in conflict, can you ensure that you have run `pixi lock` locally on this branch to ensure it's updated with any changes and commit it? Thanks

yayen-lin:
Working on it now, thank you Mike for your help 🙏

9prady9:
> @yayen-lin I've merged the PR [sirius-db/extension-ci-tools#2](https://github.com/sirius-db/extension-ci-tools/pull/2) to help with the build error. Your `pixi.lock` is now in conflict, can you ensure that you have run `pixi lock` locally on this branch to ensure it's updated with any changes and commit it? Thanks

Ah so those dockerfiles are in this repo, I was thinking they are in CI related files somewhere else. Good to know.

Glad this got in, great work @yayen-lin keep it coming :)

yayen-lin:
Will do, thank you @9prady9 !

PR #1229, opened 2026-07-21
feat(s3): read S3 object keys literally to match AWS CLI semantics

ran-yuan-rui:
S3 keys were being decoded in the wrong layer.

`uri_parser` percent-decoded the key before signing, so a literal key such as `report%2024.parquet` could resolve to `report 24.parquet` instead. Hive paths had the same problem: Sirius decoded `col=a%20b/`, then DuckDB decoded it again when reading the partition value. The glob path rejected `%HH` keys to avoid silently opening the wrong object.

This change keeps the S3 key literal through parsing and glob expansion:

- `uri_parser` no longer percent-decodes S3 keys or treats `?` and `#` as URL delimiters.
- Glob expansion passes the raw key returned by S3 LIST directly to the file list.
- The existing SigV4 signer encodes the raw key once when building the wire path.
- DuckDB remains responsible for decoding Hive partition values.

A object key such as `report%2024.parquet` now addresses the literal object, and a Hive path `col=a%20b/` produces the partition value `a b` after DuckDB's single decode, matching the same tree on local storage.

This changes the meaning of directly typed S3 URLs. To read an object whose key contains a space, use a literal space:

- `s3://bucket/my file.parquet` addresses `my file.parquet`
- `s3://bucket/my%20file.parquet` addresses the literal key `my%20file.parquet`

The same rule applies to `%`, `?`, and `#`.

A literal `?` inside a Hive partition directory, such as `col=a?b/`, is rejected because DuckDB would otherwise omit that partition value. The encoded form, `col=a%3Fb/`, remains supported and produces the value `a?b`. A `?` in the final filename is allowed.

The change also removes the old escape/decode round trip and the temporary guard that rejected `%HH` keys. The behavior is documented in `docs/super-sirius/scan.md`.

Test coverage includes:

- Literal S3 parsing and unchanged behavior for other URI schemes
- SigV4 single encoding of raw keys
- Literal `%2F` keys versus slash-separated decoys
- Literal `%20` filenames
- Hive `%20` and `%3F` partition values
- Direct and glob routes resolving to the same cache identity
- Literal `?` guard boundaries
- Raw-key sort order

### DIFF
--- a/docs/super-sirius/scan.md
+++ b/docs/super-sirius/scan.md
@@ -401,6 +401,10 @@ The scan manager builds one ioctx for the run: `uring_ioctx` when `use_sirius_da
 
 `rest_ioctx = templated_ioctx<rest_reactor>` handles `s3://` paths for AWS S3 and compatible stores such as MinIO. The `gs://` and `azure://` schemes parsed by `uri_parser` have no registered backend. DuckDB uses the read-only `sirius_httpfs` to bind transparent `read_parquet('s3://...')` queries, while scan-manager callers can open the same path directly through the datasource registry. S3 scans require GPU execution and have no DuckDB CPU fallback.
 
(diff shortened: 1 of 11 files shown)


aminaramoon (approved):
LGTM!

Issue #1218, opened 2026-07-20
GPU NULL-handling correctness bugs surfaced by null-data tests


Jedi18:
Sub-issue of #1095. Bugs found while adding null-data test coverage (#<A>).

Each is a confirmed GPU-vs-DuckDB divergence, reproduced by a quarantined
`[!shouldfail]` integration test under `test/cpp/integration/`

## Bugs (most impactful first)

- [x] **1. Three-valued `OR` drops rows.** GPU evaluates `TRUE OR NULL` as
  non-TRUE (naive NULL propagation) instead of `TRUE`, so rows where one OR
  branch is TRUE and the other is NULL are wrongly filtered out. **Severity: high**
  (silent wrong results on a very common query shape).
  Repro: `SELECT id FROM nt WHERE i = 10 OR b = 200` → 3 rows on GPU vs 4 on
  DuckDB (drops the `i=10, b=NULL` row). `AND` is unaffected.
  Suspected: GPU predicate/conjunction evaluator.
  Test: "three-valued OR with NULL operand [known divergence]".
https://github.com/sirius-db/sirius/pull/1271

- [x] **2. Ungrouped `AVG` divides by row count, not non-null count.**
  `AVG(v)` returns `335/8 = 41.875` instead of `335/5 = 67`. `SUM`/`COUNT` are
  individually correct, and **grouped** `AVG` is correct — isolated to the
  ungrouped aggregate. **Severity: high** (silent wrong results; `AVG` is common).
  Repro: `SELECT AVG(v) FROM agg_n`.
  Suspected: `sirius_physical_ungrouped_aggregate`.
  Test: "ungrouped AVG denominator counts NULL rows [known divergence]".
https://github.com/sirius-db/sirius/pull/1231

- [x] **3. Wholly-NULL column read as sentinel values.** A column that is
  entirely NULL loses its validity mask in the GPU native scan and is read as
  `INT_MAX`. `SUM(allnull)` returns `8*INT_MAX`; `COUNT(allnull)` returns the row
  count (8) instead of NULL / 0. All-NULL *groups* of a nullable column are fine —
  only wholly-NULL *columns* are affected. **Severity: med-high** (scan-level;
  likely corrupts any read of such a column, not just aggregates — needs
  blast-radius confirmation).
  Repro: `SELECT SUM(allnull), COUNT(allnull) FROM agg_n` (`allnull` is all NULL).
  Suspected: GPU DuckDB-native scan — validity dropped for all-null segments.
[...]

PR #1231, opened 2026-07-21
fix(aggregate): use non-null count as ungrouped AVG denominator

Jedi18:
Fixes one of the bugs uncovered in null test coverage addition, bug number 2) tracked here https://github.com/sirius-db/sirius/issues/1218

Ungrouped AVG divided SUM by the row count instead of the count of non-null values, so AVG over a column containing NULLs was wrong (e.g. 335/8 instead of 335/5). Use a COUNT reduction (null_policy EXCLUDE) for the denominator, matching SQL semantics and grouped AVG. Skip the reduction when the column has no NULLs to preserve the previous fast path.


mbrobbel (approved):
Maybe add a test?

Jedi18:
> Maybe add a test?

So there is a PR that adds null test coverage. That includes a test for this, but marked in that PR as !shouldfail
So perhaps I'll wait for that PR to go in first?
https://github.com/sirius-db/sirius/pull/1228/changes

mbrobbel:
> > Maybe add a test?
> 
> So there is a PR that adds null test coverage. That includes a test for this, but marked in that PR as !shouldfail So perhaps I'll wait for that PR to go in first? https://github.com/sirius-db/sirius/pull/1228/changes

We can also do it in a follow-up.

PR #1233, opened 2026-07-21
fix(starrocks): materialize common project slots

mbrobbel:
## Description
**Motivation.** A `PROJECT_NODE` whose `common_slot_map` is non-empty failed to translate: its `slot_map` expressions reference common slot ids that no tuple materializes, so `translate_slot_ref` hit a descriptor error. The FE emits this shape whenever a sub-expression is shared by several output columns.

**What changed.** Common expressions are materialized first as hidden columns (one `ProjectRel` per common slot, in ascending slot-id order, chaining allowed), visible expressions read them through `ExprContext::slot_overrides`, and the emit mapping hides the scratch columns from parents. `SELECT_NODE`/`HASH_JOIN_NODE`/`NESTLOOP_JOIN_NODE` with a `common_slot_map` are refused ("common slots are only materialized on PROJECT_NODE") so the unsupported shape is reported up front instead of failing later with an opaque descriptor error (`slot N (tuple T) is not part of row_tuples`) when a conjunct references one of them.

**Verified.** fmt/clippy/`cargo test --workspace --no-default-features`: `project_common_slots_are_materialized_before_visible_expressions`, `nested_common_slots_are_appended_in_slot_id_order`, `common_slots_outside_a_project_are_rejected`. Local run on a33e1ffb (rebased onto dev 909d6cb1 after #1235 and #1236 merged): fmt and clippy clean; starrocks-plan-translator 6 unit + 104 `translate.rs` tests, sirius-starrocks-cn 49 tests, all passing. Merge check: merges cleanly with #1232 in either order.

**Intentionally not handled.** Common slots consumed *above* the project (TPC-H q14 shape) — lands with the carried-slots PR carved out of feat/pin-table-cn (no issue yet); the over-broad refusal on non-PROJECT nodes is tracked in #1692.

## Checklist
- [x] Read CONTRIBUTING.md and ensure PR meets "reviewability" checklist
- [x] Cover changes with new or existing tests
- [ ] Document configuration changes in code and summarize in the description above
- [x] Update human and agent documentation (README.md, docs/, skills, CLAUDE.md)

## References
Refs #1111 (merged). Follow-up issue: #1692.

PR #1234, opened 2026-07-21
fix(starrocks): lower cross joins to equality joins

mbrobbel:
Lower Cartesian products to constant-key equality joins supported by the GPU planner.

Depends on #1111.


aocsa (approved):
LGTM. Land after #1232. This PR should keep append_project, emit_columns, and field_selection. #1235 and #1233 drop their copies.

PR #1242, opened 2026-07-21
feat(starrocks): translate materialized exchanges

mbrobbel:
Translate exchange nodes from explicit materialized local Parquet inputs.

Depends on #1111.

### DIFF
--- a/experimental/starrocks/crates/starrocks-plan-translator/src/descriptor_table.rs
+++ b/experimental/starrocks/crates/starrocks-plan-translator/src/descriptor_table.rs
@@ -238,23 +238,43 @@ impl DescriptorTable {
 
     /// Builds the Substrait schema for a StarRocks tuple.
     pub fn named_struct(&self, tuple_id: i32) -> Result<NamedStruct> {
-        let (names, types): (Vec<_>, Vec<_>) = self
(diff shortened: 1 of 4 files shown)


aocsa:
Closing as superseded on design grounds, not staleness (the diff still applies cleanly). The exchange landed differently on feat/pin-table-cn: `translate_exchange` lowers `EXCHANGE_NODE` to a `ReadRel` over the engine stream view `sirius_stream_<node_id>` — no parquet materialization or GPU→file→GPU round trip, a remote sender is expressible, and a merging exchange becomes a `SortRel` over the stream read (this PR refuses it with "merging exchanges are not supported by sequential execution", which blocks TPC-H Q1/Q3's top fragment). Your API survives verbatim in the replacement PR (`translate_fragment_with_exchange_inputs`, `ExchangeInput`), and its three guard tests `exchange_without_input_row_tuples_is_rejected`, `exchange_names_must_match_the_row_layout_arity`, `exchange_offset_becomes_a_fetch` will be ported, rewritten for streams (the `paths.is_empty()` guard becomes an empty-`stream_view` guard). Thanks — the guard/offset coverage is exactly what the replacement needed.

PR #1193, opened 2026-07-16
dynamic filters: planning sidecar (c1a2a)

kevkrist:
This PR creates a read-only planning sidecar. It does not replace the legacy dynamic filter publication plan, and no runtime code consumes the sidecar in this PR. Thus, it is scaffolding-only, and no functional or performance change is introduced.

Implements stage 2 of the rollout plan in #1179 
Builds off of #1185 

A significant amount of this code's footprint is from documentation and building off of #1185
There is a markdown document that serves as a guide for how these new components fit together with the planning lifecycle: docs/suprer-sirius/issue-1010-c1a2a-component-flow.md. Please use that if you need additional assistance. A summary is in the comment below. It will be deleted before merging.

PR #1179, opened 2026-07-14
docs(dynamic-filter): add SIP design and #1010 delivery plan

kevkrist:
Concise, self-contained design for join-probe sideways information passing and the sub-issue/PR delivery plan that splits the architecture contract into independently reviewed units. Detailed working plans and audits stay with the implementers; each sub-issue carries its own acceptance criteria. Index both docs in the Super Sirius README.

Refs #1010

Issue #1010, opened 2026-06-26
Dynamic filters: extend pushdown to non-scan probes (SIP)


kevkrist:
Phase-1 dynamic table filters only serve probe inputs that are leaf `SCAN` operators. Joins higher up in the join graph — whose probe source is another operator's output rather than a scan — get no dynamic filtering.

The highest-value follow-up is sideways information passing: let a build-side filter prune a *downstream* join's probe input, so filters from dimension joins reach the fact table through intermediate joins. The producer side, the filter zoo, and the channel are unchanged; this adds a new consumer (hash-join probe) and Sirius-owned producer/consumer pairing.

Design & rationale: `docs/super-sirius/dynamic-filters.md` (Phase 2). Follow-up to the Phase-1 dynamic-table-filter work.

PR #1244, opened 2026-07-21
dynamic filters: Sirius-owned producer-key seam (R1a)

kevkrist:
Runtime dynamic-filter publication no longer reads DuckDB's `JoinFilterPushdownInfo`. Admitted-key metadata is decided at plan time and carried on `dynamic_filter_publish_plan`; the publisher consumes only that. See below for a diagrammatic overview.

Behavior-preserving for every accepted configuration. One deliberate exception: the YAML surface now rejects a `dynamic_filter_domain_coverage_threshold` that the SQL `SET` surface already rejected. On `dev` that value was accepted from YAML and reached a constructor built for every GPU hash join, so a non-positive or NaN setting failed planning for every such query.

Also here: the dynamic-filter substrate moves under `op/dynamic_filter/` (this accounts for a large amount of the code footprint), the dead nested-loop-join pushdown constructor is deleted, and the previously inert coverage gate is left inert — its fix is a separate unit.

Second commit resolves a code review of the first; the two are worth reading separately.

Refs #1010
Closes #1193 as obsolete given design changes to #1179


kevkrist:
# Dynamic-filter producer-key data flow

## Summary

This PR moves dynamic-filter key eligibility, scan routing, and coordinate translation to plan time. The result is frozen in a Sirius-owned immutable publish plan, so runtime publication no longer reads DuckDB planner metadata.

The change preserves the existing opportunistic contract: dynamic filters may remove impossible probe rows early, but the hash join remains authoritative for correctness.

## 1. Plan-time ownership transfer

```mermaid
flowchart TD
    DJ["DuckDB logical comparison join"]
    SS["Snapshot filter-pushdown candidate<br/>and collect domain evidence"]
    CP["Create Sirius child plans"]
    KS["Capture original key shapes<br/>before materialization"]
    MK["Materialize computed keys<br/>as runtime columns"]
    WR["Wrap physical join conditions"]
    RR["Resolve DuckDB-named scan routes<br/>and register producers"]
    AK["Admit legal keys and build<br/>sparse target bindings"]
    FP["Freeze immutable<br/>dynamic_filter_publish_plan"]
    HJ["Construct Sirius hash join"]

    DJ --> SS --> CP --> KS --> MK --> WR --> RR --> AK --> FP --> HJ
    SS -. "DuckDB hint and domains" .-> RR
    KS -. "pre-materialization shapes" .-> AK
```

The important boundary is before runtime:

- Sirius snapshots DuckDB’s dynamic-filter candidate while the logical join is still intact.
- It records whether each equality side was a reference, cast, or computed expression before computed keys are rewritten into columns.
- Routing names the eligible scan targets. A route requires valid DuckDB metadata, a filtered build hint, GPU and host placement, and a live filter channel.
- Admission accepts equality keys without casts whose materialized build side is a cuDF-representable reference. Accepted keys are compacted into a dense vector; each target receives only the bindings it needs.
- The immutable plan owns the admitted keys, sparse bindings, replica placement, and source-selection policy. Runtime code consumes only this plan.

If routing produces no target, the plan is disabled even if one or more keys were otherwise admissible.

## 2. Coordinate translation and dense compaction

Several index spaces coexist. They are translated at plan time and must not be treated as interchangeable.

| Coordinate | Meaning | Lifetime |
|---|---|---|
| Planner condition index | Position in the original comparison-join conditions; used for provenance, captured key shape, and domain evidence | Stored per admitted key |
| Admitted-key index | Dense position in the immutable plan’s admitted-key vector | Stored in every target binding |
| Build-key ordinal | Column in the materialized runtime build table from which the filter source is constructed | Stored per admitted key |
| Channel push ordinal | Target-local coordinate supplied to the channel; for scan routes, this is a position in that scan target’s `column_ids` vector | Stored per target binding |
| DuckDB filter ordinal | Alignment position that zips DuckDB’s filtered conditions to each target’s column entries | Planning only |
| Consumer output position | Position expected by the scan-side dynamic-filter operator after its channel remaps the push ordinal once | Derived by the channel |

For example, suppose DuckDB’s hint identifies original conditions `[1, 3]` and a scan target maps them to `column_ids` positions `[8, 2]`. If condition `1` contains a cast and Sirius rejects it, condition `3` becomes dense admitted key `0` while retaining its other coordinates:

```mermaid
flowchart LR
    C1["condition 1<br/>target ordinal 8"] -->|"cast: reject"| X["no admitted key"]
    C3["condition 3<br/>target ordinal 2"] -->|"legal: admit"| K0["admitted key 0<br/>planner condition 3<br/>independent build ordinal"]
    K0 --> B["binding<br/>admitted_key_index = 0<br/>channel_push_ordinal = 2"]
    B --> R["channel remaps 2 once<br/>to consumer output position"]
```

Dense compaction changes only the admitted-key coordinate. It does not renumber the original condition, the materialized build column, or the target-local channel coordinate.

## 3. Runtime construction, replication, and publication

Publication occurs once the concat-folded build batch reaches an enabled `BUILD_PROBE` hash join on its `build` port.

```mermaid
sequenceDiagram
    participant BW as Build writer stream
    participant HJ as Hash join
    participant PS as Publication stream
    participant RG as Replica GPUs
    participant CH as Filter channels
    participant PR as Probe consumers

    BW->>BW: Finish build-batch writes
    BW->>HJ: Record writer event with batch representation
    HJ->>HJ: Acquire read-only batch view before routing
    HJ->>PS: Wait on build writer event
    PS->>PS: Apply gates and construct filter sources
    PS->>PS: Synchronize completed source construction
    PS->>RG: Copy filters to requested device spaces
    RG-->>PS: Complete each successful replica copy
    PS->>CH: Push through sparse target bindings
    HJ->>PR: Schedule immediate probe after publication
    CH-->>PR: Transitive consumers observe available snapshots
```

The writer event orders the build producer against the publication stream. Probe batches do not wait on that event. Instead, the hash join completes publication synchronously before scheduling its immediate probe producer.

For each admitted key, the publisher:

1. Applies the coverage and source-selection policy.
2. Optionally constructs a zone map and chooses a membership representation: tiny in-list, hash in-list, or Bloom filter.
3. Builds the source once, then creates replicas in the plan’s requested device spaces.
4. Publishes only completed, usable replicas through each target’s sparse bindings.

Replica placement and routing are orthogonal. A filter may exist on several GPUs while being routed to only one channel, or several targets may consume the same replica. Failed optional replicas are skipped for that device; they do not affect query correctness.

## 4. Consumer paths

```mermaid
flowchart TD
    PUB["Publisher pushes completed filter<br/>using target-local ordinal"]
    REMAP["Scan channel remaps ordinal once<br/>to consumer output position"]
    SNAP["Consumer snapshots currently<br/>available append-only filters"]
    PZ["Parquet reader zone map"]
    PM["Parquet membership operator"]
    DN["DuckDB-native dynamic-filter operator"]
    PRUNE["Row-group pruning<br/>before decode"]
    POST["Row filtering<br/>after decode"]
    JOIN["Authoritative hash join"]

    PUB --> REMAP --> SNAP
    SNAP -->|"zone map, when enabled"| PZ --> PRUNE --> JOIN
    SNAP -->|"membership"| PM --> POST --> JOIN
    SNAP -->|"zone map + membership"| DN --> POST
```

- Parquet can pass zone maps into `read_parquet` for row-group pruning. Its membership filters run after decode.
- DuckDB-native scans apply both zone-map and membership filters after decode.
- Only Parquet reader-side zone maps can avoid reading and decoding data; zone-map publication is currently disabled by default.
- Channels are append-only opportunistic snapshots, not readiness barriers. Immediate probes run after synchronous publication, while transitive scans may observe none, some, or all filters that have completed so far.
- A missing filter, missing replica, closed channel, or unsupported key passes rows through. Final join semantics do not depend on dynamic-filter availability.

## Current scope

- Production planning creates DuckDB-derived scan targets. Direct join-edge routing exists as groundwork but is not wired into production planning yet.
- Coverage-policy plumbing exists, but current domain tracing yields unknown cardinality after DuckDB rewrites condition sides to bound references, so coverage gates are presently inert.
- This is an ownership and data-flow boundary change. It does not claim a demonstrated wrong-column correctness bug in the previous implementation.

## Review map

- Plan construction: `src/planner/sirius_plan_comparison_join.cpp`
- Route resolution: `src/planner/dynamic_filter_scan_routes.cpp`
- Key admission and binding construction: `src/planner/dynamic_filter_key_admission.cpp`
- Immutable plan schema: `src/include/op/dynamic_filter/dynamic_filter_publish_plan.hpp`
- Runtime trigger: `src/op/sirius_physical_hash_join.cpp`
- Filter construction and fan-out: `src/op/dynamic_filter/dynamic_filter_publisher.cpp`
- Channel remapping and scan consumption: `src/op/dynamic_filter/sirius_dynamic_filter.cpp`

kevkrist:
Superseded by #1277, which consolidates this work into a single PR.

PR #1253, opened 2026-07-21
fix: retry transient CUDA kernel-launch failures as task reschedules

felipeblazing:
## Problem

`cudf::hash_partition()` launches a murmur3 hash kernel via the CUDA Runtime API. When the pipeline executor runs 4 concurrent streams per GPU (`num_threads: 4`), all 4 streams can reach `cudf::hash_partition()` at the same time. The simultaneous PDL (Programmatic Dependent Launch) kernel launches exhaust the GPU's PDL scheduling credit pool, causing some launches to fail synchronously with `cudaErrorLaunchOutOfResources` or `cudaErrorInvalidValue`.

The failed launch sets the per-thread CUDA sticky last error. The immediately-following `thrust::exclusive_scan` (inside `cudf::hash_partition`) calls CUB's `cudaPeekAtLastError()`, which catches the error and throws — triggering the GPU→CPU fallback (~7 s vs ~1.1 s on SF1000 4-GPU). This caused q13 and q18 to hit ~50% slow iterations on the GB200×4 bench machine.

**Diagnosis trail:**
- Added `cudaGetLastError()` in `run_one_operator()`'s catch block → discovered `thrust::throw_on_error()` clears the sticky error before throwing (intentional Thrust design), so the catch always sees success
- Added operator-name logging to the catch → confirmed every failure is `PARTITION (id=32/52)`, never JOIN or GROUPBY
- The error is set inside `cudf::hash_partition()` (pre-check before the call is always clean)
- Three tasks fail simultaneously at the same millisecond with *different* error codes (`cudaErrorInvalidValue` vs `cudaErrorLaunchOutOfResources`) → PDL credit pool contention across 4 concurrent streams

## Fix

Catch `thrust::system_error` for those two retryable error codes in `compute_task()`'s operator loop and convert to a new `cuda_launch_reschedule_exception`, handled by the existing executor back-off + reschedule path (50 ms, up to 100 retries). All failures recovered at retry 1/100 in testing.

**Exception hierarchy** (`oom_reschedule_exception.hpp`):
```
task_reschedule_exception          ← new base (intermediate_data, resume_index, message)
├── oom_reschedule_exception       ← unchanged API, inherits ctor via using
└── cuda_launch_reschedule_exception  ← new, adds cuda_error_code (int)
```

`gpu_pipeline_executor` now catches `task_reschedule_exception&` for both cases; log messages use `ex.what()` to distinguish OOM vs CUDA launch errors.

Also adds `(void)cudaGetLastError()` before `cudf::hash_partition()` to defensively clear any stale per-thread sticky error that could be misattributed to the scan inside `hash_partition`.

## Test plan

- [ ] Ran 8-iteration SF1000 4-GPU investigation for q13 and q18 before fix: ~50% iterations at ~7 s (GPU→CPU fallback)
- [ ] Ran same investigation after fix: all 16 iterations at 1.1–1.4 s, zero fallbacks
- [ ] Log shows `reschedule (retry 1/100) ... : CUDA launch error [701] ... at operator PARTITION` — correct exception type and message
- [ ] All `[pipeline]` unit tests pass; `oom_reschedule_exception` API is unchanged (existing callers unaffected)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/pipeline/gpu_pipeline_executor.cpp
+++ b/src/pipeline/gpu_pipeline_executor.cpp
@@ -308,7 +308,7 @@ void gpu_pipeline_executor::manager_loop()
         try {
           task->execute(exc_stream);
           _tasks_executed.fetch_add(1, std::memory_order_relaxed);
-        } catch (oom_reschedule_exception& oom) {
+        } catch (task_reschedule_exception& ex) {
           if (_completion_handler && _completion_handler->has_error()) {
             // If the completion handler is already in an error state, then we can just return and
             // not try to reschedule
@@ -316,10 +316,10 @@ void gpu_pipeline_executor::manager_loop()
           }
           auto* gpu_task = cast_to_gpu_pipeline_task(task.get());
           if (!gpu_task) {
-            SIRIUS_LOG_ERROR("GPU Pipeline Executor: Failed to cast task for OOM reschedule");
+            SIRIUS_LOG_ERROR("GPU Pipeline Executor: Failed to cast task for reschedule");
             if (_completion_handler) {
               _completion_handler->report_error(
-                "GPU Pipeline Executor: Failed to cast task for OOM reschedule");
+                "GPU Pipeline Executor: Failed to cast task for reschedule");
             }
             return;
           }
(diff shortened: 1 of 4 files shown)


src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -359,6 +390,31 @@ std::unique_ptr<op::operator_data> gpu_pipeline_task::compute_task(rmm::cuda_str
         std::move(operator_input_output_data),
         i,
         "OOM at operator " + op.get_name() + " (index " + std::to_string(i) + ")");
+    } catch (const thrust::system_error& cuda_err) {
+      auto err = static_cast<cudaError_t>(cuda_err.code().value());
+      if (err == cudaErrorLaunchOutOfResources || err == cudaErrorInvalidValue) {
```
bwyogatama, line 395: are we sure that `cudaErrorInvalidValue` should always be a retryable exception? Wouldn't it also be possible for this error to pop up for invalid argument?

bwyogatama (approved):

src/pipeline/gpu_pipeline_task.cpp
felipeblazing: This can be thrown by any kernel when oversubscribed. when we are launching too many kernels basically. If it gets retried and will always fail because of an invalid pointer for example it makes failure a bit slower but the benefit is that we recover from some failures that are recoverable that we currently fail on.

src/pipeline/gpu_pipeline_executor.cpp
```diff
@@ -345,34 +345,35 @@ void gpu_pipeline_executor::manager_loop()
           // too short. With 100 retries × 50 ms backoff (~5 s) the probe
           // tasks get enough patience to clear the contention window while
           // still bailing out on truly wedged queries.
-          static constexpr uint32_t MAX_OOM_RETRIES = 100;
-          if (next_retry_count > MAX_OOM_RETRIES) {
+          static constexpr uint32_t MAX_RETRIES = 100;
```
bwyogatama, line 348: 100 seems a bit excessive, should we do 10?

PR #1254, opened 2026-07-21
adding primary key to duckdb data generation

wmalpica:


wmalpica:
I attemped this to see if telling DuckDB which were the primary keys would result in better query plans, but instead it increases the size of the data dramatically because it creates indexes and it does not actually change the query plans

PR #1260, opened 2026-07-22
perf: narrow column widths

kevkrist:
Serve pinned numeric columns in the narrowest physical type their values fit, and keep them narrow through the plan, until an upcast is required (by, e.g, a PROJECTION, PARTITION, JOIN, etc.). Values remain narrow through BOUND_REF and direct COMPARISON and BETWEEN operators. Relies on pinned min/max stats for the narrowing at the batch-level. Currently off by default.

There is a tier-aware cost model: GPU resident pins have additional gates since they benefit less from saved H2D upload -- there must be transport benefit (join payload / exchange / group key) or a free narrow-domain comparison. Host pins always narrow.

NOTE: while chunks of a pinned column may be stored at different widths in the cache, a plan fixes one carrier width per column -- the widest across the entry's chunks -- and the scan normalizes every batch to that carrier, as needed. This normalization is necessary in order to guarantee that CONCAT, PARTITION, and JOIN always see one type per column, without which they would otherwise break or require complex operator-level normalization.

- **Results Summary** (TPC-H SF300, 4×GB200, interleaved warm medians): host-pinned **−18.6%** (1 GPU) / **−8.5%** (4 GPU); GPU-pinned **−7.0%** / **−7.6%**; SF50 host-pinned −15.4%. Per-query breakdown in the comment below. Also, Simpatico + narrowing results in the comment below that.

LIMITATIONS
 - ~~Simpatico currently omits compression of 16b types: see #1310. This is currently addressed by simply widening those columns when Simpatico and narrowing are both on. If #1325 lands before this PR, those changes will be incorporated and the automatic widening for these types will be eliminated. Otherwise, this will be addressed in a follow-up.~~
 - MVCC: see #1311. The existence of INSERT deltas on a scan will decline narrowing and expand any narrowed data for the scan in the cache.


kevkrist:
# Compressed materialization — performance summary

**Setup.** TPC-H SF300 on 4×GB200 (per-query pinning; 1-GPU and 4-GPU topologies; host-pinned and
GPU-pinned tiers) plus SF50 on GB10. Every number is a same-binary flag A/B
(`enable_compressed_materialization` off vs on) using interleaved OFF/ON rounds (2 rounds × 12
iterations, warm per-query medians within round, median across rounds). **Every leg validates
22/22 against DuckDB; zero GPU→CPU fallbacks.**

## Aggregate results (SF300)

| Cell | OFF | ON | ON vs OFF |
|---|---:|---:|---:|
| 1-GPU host-pinned | 9.925 s | 8.079 s | **−18.60%** |
| 4-GPU host-pinned | 10.367 s | 9.486 s | **−8.49%** |
| 1-GPU GPU-pinned | 5.849 s | 5.442 s | **−6.95%** |
| 4-GPU GPU-pinned | 8.616 s | 7.962 s | **−7.59%** |

SF50 (GB10): host-pinned **−15.42%**, unpinned **+0.31%** (parity — the unpinned state plans
identically to flag-off).

## Per-query ON-vs-OFF (SF300)

| Q | 1-GPU host | 4-GPU host | 1-GPU GPU | 4-GPU GPU |
|---:|---:|---:|---:|---:|
| 1 | −5.7% | +2.8% | +3.9% | +2.7% |
| 2 | −7.2% | −2.6% | +5.3% | −3.0% |
| 3 | −31.6% | −19.5% | −27.7% | −17.3% |
| 4 | −16.4% | +1.5% | +0.0% | +1.8% |
| 5 | −30.7% | −22.3% | −6.6% | −19.6% |
| 6 | −32.3% | −5.3% | +0.0% | +3.9% |
| 7 | −27.4% | −8.8% | −5.6% | −7.7% |
| 8 | −25.8% | −9.2% | −5.8% | −3.6% |
| 9 | −19.6% | −14.2% | −4.2% | −11.7% |
| 10 | −36.9% | −8.5% | −39.0% | −6.6% |
| 11 | −6.9% | −32.4% | −8.3% | −32.3% |
| 12 | −11.8% | −2.0% | −5.2% | +3.7% |
| 13 | +0.0% | −7.0% | −1.2% | −5.6% |
| 14 | −22.6% | −6.0% | +4.0% | +5.6% |
| 15 | −24.1% | +0.0% | −0.0% | +5.0% |
| 16 | −2.2% | +4.9% | +4.8% | +1.7% |
| 17 | −22.1% | −16.8% | −4.7% | −15.8% |
| 18 | −21.8% | −14.8% | −5.3% | −5.6% |
| 19 | −18.0% | −4.7% | −5.6% | +4.1% |
| 20 | −13.3% | −4.3% | +0.7% | −8.1% |
| 21 | −9.9% | −2.7% | +0.7% | −0.6% |
| 22 | +0.0% | +4.7% | −0.0% | −2.4% |

Q1's small GPU-tier residual is the per-batch widening cast of columns the tier policy plans
native over a narrow pin; Q16's is an IN-list evaluator restore on a transport-kept column
(narrow-domain IN evaluation is a known follow-up).

## Mechanism confirmation

- Probe **PARTITION bytes/row** (4-GPU, TRACE): Q5 27.57 → 19.75 (**−28.4%**), Q9 28.66 → 19.58
  (**−31.7%**) — identical on host and GPU tiers; narrow exchange transport is tier-symmetric.
- `partition_narrow_columns` (4-GPU ON): Q5 = 214, Q9 = 65–68 — narrow bytes crossing the
  inter-GPU exchange, derived from actual batch types.
- `scan_narrow_targets_retracted`: **0 on every host leg** (host planning is provably unaffected
  by the tier policy); nonzero on GPU legs only where restoration-only columns exist.

kevkrist:
## PR #1260 stacking matrix — Simpatico × narrowing @ `1333d403`

TPC-H SF300, 2 rounds × 12 iters (drop iter 0), interleaved legs, fresh process per (leg, round), 22/22 validation every leg. Same binary; YAML-only differences:

| leg | narrowing | pin-table compression |
|---|---|---|
| off | off | off |
| narrow | on | off |
| comp | off | on |
| both | on | on |

Machine: `presto-gb200-gcn-17`, 4× GB200, driver 580.105.08, CUDA 13.0.

### Aggregates vs baseline (`off`)

| tier | off | narrow | comp | both | narrow/off | comp/off | both/off | both/comp |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| **host-pinned 1-GPU** | 9.939 s | 8.052 s | 7.989 s | 7.354 s | **−19.0%** | −19.6% | **−26.0%** | **−8.0%** |
| **GPU-pinned 1-GPU** | 5.782 s | 5.391 s | 7.272 s | 6.510 s | −6.8% | **+25.8%** | +12.6% | −10.5% |

- **Host merge gate (narrow/off):** −19.0% (expected ~−15%) — pass.
- **Stacking (both/comp):** −8.0% host / −10.5% GPU — narrowing still helps on top of compression. No query with both/comp > +5%.
- **GPU tier:** compression alone is a large regression (+25.8%); stacking recovers some of it but both remains +12.6% vs off. Worth gating by pin tier separately from this PR.

### Host footprint (max `[host_pool] HOST:0 QueryEnd allocated`, sum over queries)

| leg | footprint | vs off | vs comp |
|---|---:|---:|---:|
| off | 1332 GB | — | |
| narrow | 918 GB | −31% | |
| comp | 425 GB | −68% | — |
| both | 399 GB | −70% | **−6%** |

### Multi-GPU (host-pinned 4-GPU)

Incomplete: `comp` / `both` hit a deterministic Simpatico decode `cudaErrorIllegalAddress` on Q1 (reproduced at merge-base `50352d1b` — **pre-existing, not this PR**). See issue #1330 

Provisional single-round only: `narrow/off` ≈ **−10.8%** (off 8.91 s → narrow 7.95 s), in line with prior ~−9–11% expectation.

src/op/scan/sirius_gpu_scan_operator.cpp
```diff
@@ -38,23 +43,131 @@
...
+  // freshly decoded carrier, but only after exact materialized bounds confirm that its values fit
+  // the planned target.
+  auto const table_view = table->view();
+  for (std::size_t column_idx = 0; column_idx < targets.size(); column_idx++) {
+    auto const& column = table_view.column(column_idx);
+    auto const actual  = column.type();
+    auto const target  = targets[column_idx];
+    if (actual == target || can_restore_to(actual, target)) { continue; }
+
+    if (!has_explicit_physical_schema) { continue; }
```
Jedi18: Without an explicit sidecar, should an incompatible resident carrier throw here instead of continuing?

The comment above says resident batches may only widen a narrow stored carrier to native. That implies `actual == target || can_restore_to(actual, target)`. Today a non-restorable actual passes through unchanged, so the batch can disagree with the scan’s declared native schema.

This could occur if Simpatico’s `apply_stored_dtype` width-mismatch path leaves a decoder type outside the restore lattice. The documented fail-closed check only applies when a sidecar is installed.

Could we require `actual == target || can_restore_to(actual, target)` on the no-sidecar path and throw otherwise?

~Codex

src/pipeline/gpu_pipeline_task.cpp
Jedi18, line 704: Should the historical estimate account for carrier conversion?

Once `peak_opt` exists, this branch ignores `needs_carrier_conversion` and `conversion_destination_bytes`. An earlier pinned chunk may require no cast, while a later chunk must widen its carrier and allocate an additional destination column. Reusing the earlier estimate can under-reserve memory and cause an avoidable OOM retry.

Could we apply the conversion-aware minimum even when history is available?

test/cpp/integration/integration-gb10-compressed-materialization.yaml
```diff
@@ -0,0 +1,42 @@
...
+    downgrade:
+      num_threads: 2
+      monitor_period: 10ms
+  operator_params:
+    scan_task_batch_size: 536870912
+    max_sort_partition_bytes: 0
+    hash_partition_bytes: 536870912
+    concat_batch_bytes: 536870912
+    max_build_hash_table_bytes: 90000000
+    enable_compressed_materialization: true
```
Jedi18: Since compressed materialization now defaults to true, both this config and `integration-gb10.yaml` enable it. Could we set the baseline config (`integration-gb10.yaml`) explicitly to false so the documented benchmark remains an OFF versus ON comparison?

src/sirius_extension.cpp
```diff
@@ -2052,6 +2095,18 @@ static void SetEnablePinnedZoneMapPruning(ClientContext& context, SetScope scope
                    params->enable_pinned_zone_map_pruning);
 }
 
+static void SetEnableCompressedMaterialization(ClientContext& context,
```
Jedi18: This setting is stored per connection by DuckDB, but this setter also updates the shared `SiriusContext`. One connection can therefore change another connection’s effective behavior while its `current_setting` still reports the previous value. YAML overrides can similarly disagree with `current_setting`.

I ran into the same issue in [PR #1190](https://github.com/sirius-db/sirius/pull/1190). The approach there was to read the setting from the active ClientContext using TryGetCurrentSetting. Could we do the same here for both planning and pinning?

(seeing how PRs adding configs seem to run into this issue, I'll look into creating a minor PR that should make this hopefully less likely in the future)

src/op/scan/sirius_gpu_scan_operator.cpp
```diff
@@ -38,23 +43,131 @@
...
-    _split_connector(std::make_shared<scan_manager::split_connector>())
+    _split_connector(std::make_shared<scan_manager::split_connector>()),
+    _compressed_materialization_observer(compressed_materialization_observer)
 {
+  // `this->types` reads the base-class member; the constructor argument was consumed above.
+  _native_physical_types.reserve(this->types.size());
+  for (auto const& type : this->types) {
+    auto const native = sirius::try_get_cudf_type(type);
+    if (!native) {
+      _native_physical_types.clear();
```
Jedi18: Clearing the whole target vector here also disables normalization for columns that were mapped successfully. For example, if a resident batch contains a narrowed `BIGINT` plus one unmappable type, the empty target vector causes `normalize_physical_schema` to return without restoring the `BIGINT`.

Could we preserve targets per column or reject the scan instead of silently skipping normalization for the entire batch?

test/cpp/integration/integration-gb10-compressed-materialization.yaml
kevkrist: This just shouldn't be here. Deleting.

src/sirius_extension.cpp
kevkrist: Your PR idea is great!

src/op/scan/sirius_gpu_scan_operator.cpp
kevkrist: For now I'm just going to reject the scan for simplicity. If this issue surfaces later, we can handle it then with some preservation of targets per column.

kevkrist (commented):
Thanks @Jedi18 !

test/cpp/integration/integration-gb10.yaml
Jedi18, line 1: You also deleted the pre-existing `integration-gb10.yaml` along with `integration-gb10-compressed-materialization.yaml`, we should restore the `integration-gb10.yaml`

src/helper/numeric_narrowing.cpp
```diff
@@ -0,0 +1,277 @@
...
+  if (cudf::is_fixed_point(column.type())) {
+    auto const cudf_scale = column.type().scale();
+    if (cudf_scale > 0 || cudf_scale < -static_cast<int32_t>(std::numeric_limits<uint8_t>::max())) {
+      return std::nullopt;
+    }
+    decimal_scale = static_cast<uint8_t>(-cudf_scale);
+  }
+
+  auto [minimum, maximum] = cudf::minmax(column, stream, mr);
+  if (!minimum || !maximum || minimum->type() != maximum->type()) { return std::nullopt; }
```
Jedi18, line 273: Could we check `minimum->is_valid(stream)` and `maximum->is_valid(stream)` before reading their values? For an all-null column with an unknown null count, minmax can return invalid scalars. Reading those as bounds could select an incorrect narrow carrier.

Jedi18 (approved):
LGTM, thanks for the PR!

PR #1261, opened 2026-07-22
feat(pipeline): priority-based task scheduling via inspectable priority queue

aminaramoon:
## Summary

Adds priority-based pipeline task scheduling so pipelines are dispatched in plan order instead of arbitrary ready-order.

- **`inspectable_priority_queue`** — a priority-ordered replacement for `inspectable_mpsc` as the pipeline task queue (used by the task scheduler, downgrade executor, and convertible task path).
- **`query_index`** — partitions the pipeline DAG into branches (linear chains between branch points) and exposes them for priority assignment.
- **Pipeline priority computation** (`task_creator::compute_pipeline_priorities`): branches are ordered by plan order (an earlier branch is always strictly lower / runs first); within a branch the source ranks the head lowest (sink reverses); a pipeline shared by several branches takes the MIN priority of the branches that reach it. Priorities are compacted to a dense `0..N-1` range. Each pipeline's priority is carried on its shared global state and reported by every task via `gpu_pipeline_task::get_priority()`.
- Adds a `priority_order` config knob (source/sink) in `creator/config`.

## Test plan

- [x] `pixi run make` — clean build
- [x] `pixi run pre-commit run -a` — all hooks pass
- [x] Unit tests: `test_inspectable_priority_queue`, `test_query_index`, `test_convertible_gpu_pipeline_task`
- [ ] TPC-H timing sweep to confirm scheduling order improvements

## Note

The hash join task-creation fixes and `notify_downstream_pipelines` change that were originally on this branch have been split out to #1265.

### DIFF
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -22,6 +22,7 @@
 #include "pipeline/gpu_pipeline_task.hpp"
 #include "pipeline/task_scheduler.hpp"
 #include "planner/query.hpp"
+#include "planner/query_index.hpp"
 #include "sirius_context.hpp"
 
 #include <cucascade/cudf/gpu_data_representation.hpp>
@@ -107,6 +108,8 @@ void task_creator::prepare_for_query(const sirius::planner::query& query)
   std::shared_ptr<const telemetry::telemetry_context> telemetry_context =
     sirius_ctx->get_telemetry_context();
 
+  auto pipeline_priorities = compute_pipeline_priorities(query);
+
   for (const auto& pipeline : pipelines) {
     pipeline->set_task_creator(this);
     auto source_operator = pipeline->get_source();
@@ -117,6 +120,9 @@ void task_creator::prepare_for_query(const sirius::planner::query& query)
     size_t operator_id = source_operator->get_operator_id();
     auto gs =
       std::make_shared<pipeline::gpu_pipeline_task_global_state>(pipeline, telemetry_context);
+    if (auto it = pipeline_priorities.find(pipeline.get()); it != pipeline_priorities.end()) {
+      gs->set_priority(it->second);
+    }
     _gpu_operator_global_state_map.emplace(operator_id, std::move(gs));
   }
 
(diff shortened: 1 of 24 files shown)


src/sirius_config.cpp
```diff
@@ -88,6 +88,7 @@ static void from_yaml(const YAML::Node& node, creator::task_creator_config& opt)
   r.optional("thread_name_prefix", opt.thread_pool.thread_name_prefix);
   r.optional("cpu_affinity", opt.thread_pool.cpu_affinity_list);
   r.optional("strategy", opt.strategy);
+  r.optional("priority_order", opt.priority);
```
wmalpica, line 91: Please update the configuration.md file if you are adding a new config option

src/include/exec/inspectable_priority_queue.hpp
```diff
@@ -0,0 +1,305 @@
...
+    std::unique_lock<std::mutex> lock(_mutex);
+    if (_queue.empty()) { return nullptr; }
+    return pop_front_unlocked();
+  }
+
+  /**
+   * \brief Removes and returns the lowest-priority (first-to-run) item, or nullptr if empty.
+   *        Non-blocking.
+   */
+  std::unique_ptr<T> pop_front() { return try_pop(); }
```
wmalpica, line 154: rename this to `try_pop_front()`  and `pop_back()` to `try_pop_back()` to be more explicit that its non-blocking

src/creator/task_creator.cpp
```diff
@@ -195,7 +273,6 @@ op::sirius_physical_operator* task_creator::get_operator_for_next_task(
 
 void task_creator::stop()
 {
-  std::lock_guard<std::mutex> lock(_global_state_mutex);
```
wmalpica, line 198: We did we remove this lock?

wmalpica (changes requested):

aminaramoon:
we should use this PR instead
https://github.com/sirius-db/sirius/pull/1269

PR #1265, opened 2026-07-22
fix(hash-join): create all probe×build tasks in standard mode

aminaramoon:
## Summary

Fixes standard-mode hash join dropping probe×build task combinations, which produced validation mismatches (TPC-H Q9 @ SF100 summed to ~1/3 of the DuckDB result). Four coupled changes:

- **Flatten the batch cross-product across all partitions** (`get_next_task_input_data`). The old per-partition product (`Σ left[i]×right[i]`) only created tasks for the one matching partition and silently skipped the rest when the probe side is unpartitioned (all data in partition 0) but the build side is partitioned. The correct product is `total_left × total_right`; `execute()` builds a fresh hash table per task and does not rely on partition alignment. The build batch's partition still drives GPU routing.
- **Override `all_ports_empty()`** to use the cross-product cursor under `op_state_mutex`. The base-class check reads raw repo `total_size()` without the lock, racing with `get_next_task_input_data()`'s pops so two concurrent creation lambdas could both see non-empty repos while only one produced a task.
- **Force-wait for the probe pipeline to finish** before the one-time snapshot (`get_next_task_hint`). A PARTIAL probe-side CONCAT — enabled by the converter barrier change here — returns READY as soon as any probe data arrives, but the snapshot is taken once, so late-arriving probe batches were lost.
- **Notify finished-pipeline consumers unconditionally** (`sirius_pipeline::notify_downstream_pipelines`): schedule downstream task creation even for the original pipeline (`if (_task_creator)` instead of `if (_task_creator && !original_pipeline)`).

Also drops a dead `std::unordered_set<void*> seen_ops;` declaration in the converter.

## Test plan

- [x] `pixi run make` — clean build on `dev`
- [x] `pixi run pre-commit run -a` — all hooks pass
- [x] TPC-H Q9 @ SF100 validation: `✓ q9: byte-exact match` (was a ~3× sum mismatch before the fix)
- [ ] Broader TPC-H validation sweep across all queries

### DIFF
--- a/src/include/op/sirius_physical_hash_join.hpp
+++ b/src/include/op/sirius_physical_hash_join.hpp
@@ -254,6 +254,7 @@ class sirius_physical_hash_join : public sirius_physical_partition_consumer_oper
   void discard_build_only_slots_if_probe_complete();
 
   std::optional<task_creation_hint> get_next_task_hint() override;
+  [[nodiscard]] bool all_ports_empty() override;
 
   std::unique_ptr<operator_data> execute(const operator_data& input_data,
                                          rmm::cuda_stream_view stream) override;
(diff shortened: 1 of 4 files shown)


wmalpica:
This PR is superceeded by https://github.com/sirius-db/sirius/pull/1280. Closing

PR #1267, opened 2026-07-22
dynamic filters: live domain-coverage gate (R1b)

kevkrist:
Stacked on #1244. Until that merges, this diff also carries its commits; R1b is `e018979e..HEAD`.

Makes the domain-coverage gate live. It was inert: the build-key domain was always 0, because the lineage walk looked for a node kind that cannot survive column-binding resolution. A positional lineage walk now resolves a build key down to an allowlisted DuckDB-native `seq_scan` and reads `NodeStatistics::max_cardinality` as a row upper bound, and the gate fires only for a key proven unique in its base relation -- for a key with duplicates the ratio measures row retention, not domain coverage. Setting `dynamic_filter_domain_coverage_threshold` above 1.0 disables the gate.

Also lands the pre-registered SIP benchmark-gates document and a publication counter surface on `SiriusContext`, which the new end-to-end tests assert against.

Query results and plan shapes are unchanged; only which advisory filters get published changes.

Reviewer guide: `docs/super-sirius/issue-1010-r1b-code-flow.md`. (will be deleted before merge)


kevkrist:
Superseded by #1277, which consolidates this work into a single PR.

PR #1269, opened 2026-07-23
feat(exec): add multi_index_priority_queue

aminaramoon:
## What

Adds `sirius::exec::multi_index_priority_queue` — a thread-safe priority queue
whose elements are additionally reachable through secondary indexes on
**operator type**, **pipeline id**, and **query id**.

Tasks are stored once in a pointer-stable backing container (`plf::colony` by
default, `std::list` optional) and referenced from four ordered indexes: one
global priority order plus one bucket per operator/pipeline/query key. Every
index sorts by `(priority, insertion-seq)` — ascending priority (smaller popped
first, matching `queue_priority`), FIFO among equal priorities. Each node caches
its iterator in all four indexes for O(log n) erase; empty buckets are pruned.

### API
- **Blocking:** `pop()` / `pop_back()` wait for a task, return `nullptr` on interrupt.
- **Non-blocking:** `try_pop()` / `try_pop_back()`, and the per-index
  `try_pop_from()` / `try_pop_back_from()` for `operator_index` / `pipeline_index`
  / `query_index` (secondary-index pops never block — "wait for a task of this
  operator/pipeline/query" has no well-defined completion).
- **Lifecycle:** `interrupt()` / `reactivate()` / `is_open()`; `drain()` drops all
  tasks; `drain(query_index)` drops one query's tasks.
- **Introspection:** `size()` / `empty()`, per-key `size(index)`, per-dimension
  `*_bucket_count()` and `*_bucket_sizes()` maps.

## Testing
- 18 Catch2 cases / 116 assertions (`[multi_index_priority_queue]`), covering
  ordering/FIFO, all secondary-index pops, cross-index consistency, bucket
  prune/refill, the `std::list` backend, blocking-pop wakeups, interrupt, drain,
  and a 4-producer/3-consumer stress test.
- **ThreadSanitizer clean** (0 warnings) across repeated runs.


src/include/exec/multi_index_priority_queue.hpp
```diff
@@ -0,0 +1,734 @@
...
+    }
+    return total;
+  }
+
+  mutable std::mutex _mutex;
+  std::condition_variable _cv;
+  bool _active{true};  ///< false after interrupt(): blocked pops stop waiting.
+
+  level_map _levels;                                                 ///< Spine: priority -> level.
+  std::unordered_map<pipeline_key, queue_priority> _pipeline_level;  ///< pipeline -> its level.
```
wmalpica: _pipeline_level needs to be  `  std::unordered_map<pipeline_key, std::set<queue_priority>>
otherwise when we have multiple queries, you can have two pipeline_keys that are the same but have different queue_priority. Then when you remove an entry from  _pipeline_level you cant distinguish between which one to remove

wmalpica (changes requested):

src/include/exec/multi_index_priority_queue.hpp
aminaramoon: it is assumed that is is fifo, but if someone keep poping from the same pipeline id, that is a possiblity.  I am removing pipline index because it not needed for now

wmalpica (approved):

PR #1270, opened 2026-07-23
refactor(sirius): consolidate NUMA↔GPU maps onto topology_index

aminaramoon:
## Summary

`SiriusContext` and its owned components each built their own ad-hoc device↔NUMA maps for NUMA-aware GPU routing. This consolidates them onto the existing single source of truth, `sirius::memory::topology_index`: one shared `shared_ptr<const topology_index>` is built once in `SiriusContext::initialize()` and injected into every NUMA-aware component instead of each rebuilding its own map.

## What changed

- **`sirius_context`**: build the shared `topology_index_` once (after the memory manager is populated); reuse it for `task_creator` and `scan_manager` (previously a *second* index was built inline for the scan manager); reset it in `terminate()`.
- **`numa_small_pinned_mr`**: drop the `device_to_numa_` map; resolve the current device's NUMA node via the shared index. Pools stay keyed by the host-space device id verbatim.
- **`task_creator`**: drop `_sys_topology` / `_numa_to_gpu`; host-locality routing uses `gpus_of()`, and the sorted active-executor set derives from `gpu_ids()` (kept sorted+deduped to stay inverse to `sirius_physical_partition`'s device→slot mapping).
- **downgrade executors**: per-GPU `preferred_numa_node` now uses `numa_node_of(device_id)`.

## Design decision

Adopts `topology_index`'s `-1`-as-"unknown" sentinel semantics and **drops the prior `-1`→`0` NUMA normalization**. The index's manager-based constructor cross-checks each GPU's NUMA node against the HOST memory-space device ids, so `numa_node_of()` / `gpus_of()` resolve verbatim (single-NUMA hosts report host-space device id `-1`, consistent with the pre-existing defensive `(raw_numa < 0)` checks).

Scoped to `SiriusContext` and its owned components. `pin_table` / `sirius_extension`'s `host_space_by_gpu` (a gpu→`memory_space*` map — a different kind) is intentionally left untouched.

## Incidental fix

The old downgrade code indexed `topo.gpus[dev_id].numa_node` by device id, whereas `numa_node_of()` matches on the topology entry's `.id` field — so this also fixes a latent bug on hosts where device id ≠ topology-vector index.

## Testing

Reviewed by an internal C++ reviewer (correctness of lifetime/teardown, pool key-matching, partition-affinity ordering, null-safety — all clear). **Draft** because the full `pixi run make test` could not be run in my local environment (uninitialized `duckdb` submodule); relying on CI to validate the build. Existing `test/cpp/memory/test_topology_index.cpp` covers the index itself.

### DIFF
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -17,6 +17,7 @@
 #include "creator/task_creator.hpp"
 
 #include "log/logging.hpp"
+#include "memory/topology_index.hpp"
 #include "op/scan/sirius_gpu_scan_operator_data.hpp"
 #include "op/sirius_physical_delim_join.hpp"
 #include "pipeline/gpu_pipeline_task.hpp"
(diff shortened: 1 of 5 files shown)


mbrobbel (approved):
Nice cleanup

PR #1274, opened 2026-07-23
CI: Switch to T4 GPU Testing

### LINKED ISSUE #1258: [CI-BUG] Increasing Wait Times with Self-Hosted `gpu-2xl4` Runners
### Failing job URL

https://github.com/sirius-db/sirius/actions/runs/29926727052

### Expected behavior

Wait times should be 2 mins or less. The linked run shows an additional 44 wait for the GPU runner to be available before running the `test-run`

### Actual behavior

We've seen wait times of 30-45 mins on several days over the last 30 days as AWS does not have capacity for these L4 nodes

### Error output

```shell

```

### Bug frequency

Intermittent (some runs pass)

### Bug affects

GPU runners

### Additional details

_No response_

mike-wendt:
Closes #1258

This changes our GPU testing from current L4s to T4s. This is to address the availability concerns raised in #1258. While we originally were going to test T4s in parallel to gather data, the observed 2+ hour wait of one of the L4 runners forced us to switch fully in order to unblock the development team. The same work done in #1258 will be done in a new issue to track the stability of this change after merge

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -78,7 +78,7 @@ jobs:
 
       - name: Build
         env:
-          CUDAARCHS: "89" # 75 for T4 capability; 89 for L4 capability
+          CUDAARCHS: "75" # 75 for T4 capability; 89 for L4 capability
         run: make -j$(nproc)
 
       - name: Package build artifacts
@@ -110,7 +110,7 @@ jobs:
           - cuda:
               run-on-pr: ${{ fromJSON(github.event_name == 'pull_request' && 'false' || 'null') }}
       fail-fast: false
-    runs-on: ${{ fromJSON(format('["self-hosted", "{0}", "gpu-2xl4"]', matrix.cuda.runner)) }}
+    runs-on: ${{ fromJSON(format('["self-hosted", "{0}", "gpu-2xt4"]', matrix.cuda.runner)) }}
     timeout-minutes: 60
     env:
       CUDA_CACHE_PATH: /home/runner/actions-runner/_work/.cuda-cache


mbrobbel (approved):
We could also update our existing matrix with a new gpu dimension (with the different gpu labels) and update the build to build for both archs, then the two test jobs can share the one build job.

But, if we just want to keep this around for testing briefly it's also fine to duplicate and change those things in the copy I guess.

mike-wendt:
> We could also update our existing matrix with a new gpu dimension (with the different gpu labels) and update the build to build for both archs, then the two test jobs can share the one build job.
> 
> But, if we just want to keep this around for testing briefly it's also fine to duplicate and change those things in the copy I guess.

Good point, it is actually the way I implemented it at first, but I wasn't sure the overhead of building two archs at once just for `build`. To get the test data I duplicated the full pipeline to mitigate that extra overhead. I will do a quick test in this PR before merging to see what the dual arch penalty is

mike-wendt:
Looks like ~30s extra, so good call to combine the `build`

mike-wendt:
@mbrobbel from Karpenter alerts:
Screenshot 2026-07-23 at 16 19 03
Do we even want to test T4s or just make the cutover in this PR? These are the worst I've seen and there are 3 runners including for this PR stuck waiting.

I also know it's late for you so I will leave this PR up to merge to test T4s in parallel OR I will open another PR that just switches to T4 and let you decide when you're back. Appreciate your perspective here

mbrobbel:
> @mbrobbel from Karpenter alerts: Screenshot 2026-07-23 at 16 19 03 Do we even want to test T4s or just make the cutover in this PR? These are the worst I've seen and there are 3 runners including for this PR stuck waiting.
> 
> I also know it's late for you so I will leave this PR up to merge to test T4s in parallel OR I will open another PR that just switches to T4 and let you decide when you're back. Appreciate your perspective here

Let's just switch to T4s then!

Issue #1125, opened 2026-07-08
Dynamic filters: SIP consumers at hash-join probe inputs


kevkrist:
Implements #1010 (part of #1015). A producing join's membership filters are applied at intermediate hash joins' probe inputs: after the probe repository pop, before key prep and hashing. Scan consumers stay; each eligible intermediate join adds a catch-up checkpoint. Consumption is opportunistic in v1; ordered activation is built only if telemetry shows a route class that needs it. Candidates come from DuckDB's join-filter metadata — Sirius narrows, never broadens.

- [ ] **C1a** — version-pinned DuckDB metadata adapter; publisher and claim decoupling. No behavior change under fixed config.
- [ ] **C1b** — strong target-key types, materialization telemetry, shadow selectivity signal.
- [ ] **C1c** — per-target-key publish restructure (fail-closed arity check kept).
- [ ] **C1d** — enforce the repaired selectivity signal (A/B; blocked on C1b).
- [ ] **C1e** — remove the blanket `build_side_has_filter` gate (A/B, default off).
- [ ] **C2** — `probe_batch_handle` refactor, hash-join probe consumer component, mode-aware memory estimator. No routes yet.
- [ ] **C3** — route discovery/lineage/topology freeze; opportunistic routes behind `enable_dynamic_filter_sip` (default off). Coverage/value experiment.
- [ ] **C4** — default on if the coverage and value gates pass.

Ordering note: A1 (#1124), C1a, and C2 all touch `sirius_physical_hash_join.cpp` — land in that order.

Design: `docs/super-sirius/issue-1010-dynamic-filter-sip-design.md`. Plans: `docs/super-sirius/issue-1010-plans/` (docs land with the first PRs).

Note: the LIMIT/TOP-N pin bug (#1123) is deferred by decision and blocks nothing here; while on the unpatched pin, LIMIT/TOP-N-shaped tests use explicit expected rows or a filters-disabled reference — never an unpatched CPU run.

PR #1277, opened 2026-07-23
feat(dynamic filters): SIP above non-probe scan operators & decoupling from DuckDB

### LINKED ISSUE #1010: Dynamic filters: extend pushdown to non-scan probes (SIP)

kevkrist:
Supersedes #1179, #1193, #1244, #1267.  
Closes #1010, #1125.

## Summary

Decouples Sirius dynamic-filter routing from DuckDB planner metadata. Sirius now admits equality keys and traces each key independently through safe projections, filters, grouping keys, join sides, and scan or join-edge endpoints. This enables SIP across nested joins, grouped inputs, correlated subqueries, and materialized CTEs while preserving type, null, and outer-join boundaries.

Discovery requires either visible build-side filtering or an opaque `CTE_REF`/`DELIM_GET` root (optionally projection-wrapped). Unfiltered visible joins and aggregates therefore cannot publish large no-op filters, while opaque Q11/Q17 routes remain eligible.

The PR also adds the domain-coverage publication gate, marginal keep-ratio suppression, publication counters, consolidated `enable_dynamic_filter` configuration, and end-to-end/plan-shape coverage. DuckDB target discovery remains only as a test parity oracle.

A significant part of the diff is file relocation into the dynamic-filter module.

Giving about ~1.8% perf gain on SF1000.


kevkrist:
## Performance — TPC-H SF1000

Using the optimized PR #1371 Parquet setup, the final all-Q1–Q22 activation audit against current `dev` found one query with additional filters:

| Query | Publications / filters, `dev` → PR | 15-iteration warm median |
|---|---:|---:|
| Q11 | 1/1 → 3/3 | **0.139533s → 0.096099s (-31.13%)** |

Q11’s two new opaque-CTE in-lists filter both `partsupp` scans to about 4% of their input. No other query has a positive activation delta.

Containment removes measured no-op publishers instead: Q8 changes 6/6 → 5/5 and Q9 changes 2/2 → 1/1. Q9 is **0.808990s** vs. **0.828742s** on `dev` (-2.38%; prior regression eliminated).

Suite totals are unchanged at 47 publications / 48 filters per iteration, and all 22 saved results are byte-identical to `dev`. The original counterbalanced pre-containment suite was neutral (+0.19%); post-containment timings are focused confirmations.

Jedi18:
Looks good so far, can we please have this rebased with dev given the related PRs that have been merged with dev

docs/super-sirius/dynamic-filters-multi-gpu.md
```diff
@@ -1,17 +1,11 @@
 # Dynamic Filters — Multi-GPU Publication
 
-> **Status: implemented.** The hash-IN-list, Bloom, and zone-map replica paths
```
joosthooz, line 3: Do we want to keep this file? It looks like a crash/bug investigation, is that problem still present?

docs/super-sirius/dynamic-filters.md
```diff
@@ -4,13 +4,13 @@ A **dynamic filter** is a predicate that is computed at query runtime by one ope
...
+- **Sideways information passing (SIP)** — a hash-join build applies its membership filter inside its own probe subtree, as deep as the key stays a faithful pass-through, so intervening operators also skip rows the join would discard. Controlled by `enable_dynamic_filter`.
 - **Aggregation-driven pushdown** — a `GROUP BY` or `DISTINCT` exposes its distinct-value set to downstream consumers.
 - **Sort- or top-N-driven pruning** — a post-sort min/max is exact and free; a top-N's current threshold tightens upstream filters.
 - **Adaptive runtime predicates** — operators that observe data and refine filters over the lifetime of a pipeline.
 
-This document describes the implemented Phase 1 framework and the design-only directions that could generalize it. Phase 1 has a `BUILD_PROBE` hash-join-build producer, a GPU-scan consumer (parquet and duckdb-native), and three filter kinds (zone map, IN-list, and Bloom). Membership pushdown is enabled by default; the workload-specific zone-map path remains opt-in. Phases 2–4 below are not implemented.
+Phases 1 and 2 are implemented under `enable_dynamic_filter`; zone maps remain opt-in. Phases 3 and 4 are design only.
```
joosthooz: Do we want to talk about phases here, are they explained in the doc?

docs/super-sirius/dynamic-filters.md
```diff
@@ -282,23 +368,22 @@ flowchart TB
...
+**Goal:** establish the framework end-to-end against a single (producer, consumer) pair — hash-join build → GPU scan (parquet and duckdb-native) — and exercise filter-kind polymorphism via progressively richer filters.
 
-**Producer:** `BUILD_PROBE` hash-join build side.
+**Producer:** hash-join build side, in any join mode, whose build port delivers the whole build side as one batch.
 **Consumer:** GPU scan — parquet (`parquet_gpu_ingestible`: reader zone-map + post-decode operator) and duckdb-native (post-decode operator only).
-**Routing:** DuckDB-paired (`DynamicTableFilterSet*` route key).
+**Routing:** Sirius-owned discovery (a walk over the built probe subtree binds each admitted key into the scan its trace bottoms out at).
 **Coordination:** synchronous build-side CONCAT publication strictly precedes the producing join's immediate probe data scan; transitive scan targets remain nonblocking and race publication under normal scheduler order.
 
 ### 1.1 Foundational wiring
```
joosthooz: It looks like this section only has phase 1 and 1.1

docs/super-sirius/dynamic-filters.md
```diff
@@ -450,26 +568,43 @@ For both representations, the constructor enqueues creation of an owned source r
 
 ---
 
-## Phase 2 — Sideways information passing
+## Phase 2 — Sideways information passing (join-edge endpoints)
```
joosthooz: Ah here is phase 2. This doc is seriously long, can we do with a (much) smaller version?

src/include/op/dynamic_filter/dynamic_filter_device.hpp
```diff
@@ -20,9 +20,12 @@
...
-/// @brief Resolve a dynamic-filter device argument, treating a negative ID as the current device.
-///
-/// @return The explicit device ID, the current CUDA device, or -1 if CUDA cannot report one.
+/**
+ * @brief Resolve a dynamic-filter device argument, treating a negative ID as the current device
+ *
+ * @param[in] device_id The requested device ID, or a negative value for the current CUDA device
+ * @return The explicit device ID, the current CUDA device, or -1 if CUDA cannot report one
+ */
 [[nodiscard]] inline int resolve_dynamic_filter_device_id(int device_id) noexcept
```
joosthooz, line 24: This looks more general than for dynamic filters, does there exist other functions that do the same?

joosthooz (commented):
Here's some early questions, and some brief findings from Claude, I will need to continue tomorrow

Review of PR #1277 (dynamic-filter SIP routing decoupling) complete — 8 findings, most-severe first:

  Correctness
  1. dynamic_filter_publish_plan.cpp:126 — restrict_replicas_to() can zero out all replica spaces while enabled() stays true, so
     the next publish call throws std::logic_error instead of degrading gracefully — potential crash on multi-GPU queries after
     replica restriction.
  2. sirius_plan_comparison_join.cpp:525 — ordinal bounds only guarded by assert(), which is compiled out in Release builds (the
     project's default pixi run make), right before unconditional array indexing.
  3. build_filter_evidence.cpp:38 — "opaque build side" detection is a narrow two-node-kind allowlist, so many genuinely
     selective build sides (plain GROUP BY, SEMI/ANTI join, differently-wrapped CTE_REF) silently get no dynamic-filter evidence.
  4. dynamic_filter_target_discovery.cpp:37 — join value-preservation classification is now duplicated in two unsynchronized
     switch statements (physical-tree discovery vs. logical-tree key-domain estimation) — a future join-type extension could
     update one and miss the other.
  5. sirius_dynamic_bloom_filter.cu:186 — Bloom filter build ingests nulls unlike its sibling IN-list filters, silently breaking
     the "preserves null boundaries" invariant this PR otherwise claims.

  Efficiency
  6. dynamic_filter_publisher.cpp:275 — per-key replicate() calls sync per device inside their own loop instead of batching,
  causing up to 2×K×D stream syncs instead of D.
  7. sirius_plan_comparison_join.cpp:510 — probe subtree is re-walked once per equality key instead of once total.

  Docs
  8. docs/super-sirius/configuration.md:391 — says enable_dynamic_filter only applies to BUILD_PROBE joins, contradicting both
  the code comment and this PR's own dynamic-filters.md.

kevkrist:
Thanks @joosthooz!

1 (restrict_replicas_to zeroing replicas): fixed in 73fd45d1.

2 (assert-only bounds guards): fixed in 66045b70.

3 (opaque-build allowlist too narrow): This widening was explored in 5d098297 and deliberately walked back in e774f012 after TPC-H measurement -- unselective filters were being published.

4 (duplicated join-type switches): Disagree with the premise. The two switches answer different questions and correctly disagree: discovery asks "does the trace exit this block un-null-padded" (INNER/LEFT/MARK yes), the domain walk asks "is the output an injective image of the traced child's rows" (INNER/LEFT no — the other side multiplies rows). Unifying them would conflate distinct semantics. New-type drift is already guarded: discovery's switch is exhaustive (-Wswitch fires on a new JoinType), the domain walk defaults to refuse.

5 (Bloom ingests nulls): fixed in 66045b70.

6 (per-key replication syncs): agreed but deferred -- will address in #1451.

7 (probe subtree re-walked per key): not fixed. Each trace is ordinal-specific (every accepted step remaps the ordinal), so a single walk means threading K ordinals through one traversal -- complexity not worth the reward.

8 (configuration.md BUILD_PROBE claim): fixed in 66045b70

docs/super-sirius/dynamic-filters-multi-gpu.md
kevkrist: Thank you! deleted

docs/super-sirius/dynamic-filters.md
kevkrist: Agreed. Simplified.

docs/super-sirius/dynamic-filters.md
kevkrist: See below

docs/super-sirius/dynamic-filters.md
kevkrist: See below

bwyogatama:
One that codex flagged for me:

1. NaN join rows can be incorrectly removed
This happens when `enable_dynamic_zone_map_filter` is enabled.
Simple example:
- The build-side join key contains NaN.
- The probe side also contains NaN.
- DuckDB and cuDF’s hash join treat those keys as equal.
- This PR creates a range filter: probe >= min AND probe <= max.
- Normal floating-point range comparisons return false for NaN.
- The filter deletes the probe row before the join, producing the wrong result.

bwyogatama:
@kevkrist Also I noticed that the code comments are long and complex, can we have an agent cleaning up the comment, simplify the language to be more digestible for human reader, and remove the comments unless it's absolutely necessary to be added

bwyogatama (approved):

kevkrist:
@bwyogatama I've reduced the doc blast radius by ~1,200 lines, and the range filter now simply refuses floating point types.

Jedi18 (approved):

bwyogatama:
@kevkrist several questions that im curious about as i'm learning about the PR
- Why don't we push the filter below ORDER BY?
- Why don't we push the filter below CONCAT and PARTITION? I know for PARTITION we can't do it for bloom filter, but I thought we are supposed to be able to do it for min-max and IN list?

bwyogatama:
@kevkrist i also probably need to sync with you tomorrow cause the condition of when we can pushdown filter below join is something that i don't follow in this PR

kevkrist:
@bwyogatama We do push the filter below CONCAT and PARTITION, for all dynamic filters. Or do you mean filter construction? With the exception of a few *additional* dynamic filters that fire in this PR, the occurrence and placement of filters doesn't change in this PR relative to current dev.

This is the normal plan:
HASH_JOIN
└─ CONCAT
   └─ PARTITION
      └─ DYNAMIC_FILTER (membership)
         └─ GPU_SCAN

You are right that we do not currently push below ORDER BY -- we can do that relatively easily.

PR #1279, opened 2026-07-24
Test Code only - Test deep S3 glob matcher patterns

ran-yuan-rui:
Guard the glob matcher memo against backtracking blowup

The S3 glob matcher, match_glob_segments in src/io/s3/sirius_httpfs.cpp, memoizes on (key_index, pattern_index). Drop that memo and matching a pattern with several ** segments against a deep key goes exponential — the function comment says as much. Nothing tested it, so a refactor that broke the memo would pass review and CI and only show up in production as a hang on one particular pattern shape.

This pins the behavior with a regression test.

The load-bearing case globs stress/**/**/.../**/x.parquet (ten **) against a 30-segment key that ends in y.parquet. The tail mismatch is deliberate: to conclude "no match" the matcher has to exhaust every way of splitting the 28 middle segments across the ten ** groups — about 124 million arrangements — which is exactly what the memo folds down to a few hundred states. With the memo it returns in well under a millisecond; without it the test hangs. The elapsed time is logged and a soft warning fires past one second, but the assertion is only that the call returns and returns empty, so there is no wall-clock flake.

The same case also checks the ** semantics against fixed keys: zero- and multi-segment **, leading and trailing **, and * / ? / [...] mixed into a segment next to a spanning **, asserting both the keys that should match and the near-misses that shouldn't (wrong prefix, wrong character class, extra trailing character, wrong extension).

Test only, no production change. It uses the in-process fake LIST server, so it needs no bucket and no network. 
Runs in the normal suite under [s3][filesystem][glob] (make s3-test) and adds one Catch2 case.

### DIFF
--- a/test/cpp/io/s3/test_sirius_httpfs.cpp
+++ b/test/cpp/io/s3/test_sirius_httpfs.cpp
@@ -18,7 +18,6 @@
 #include "utils/s3_container.hpp"
 
 #include <arpa/inet.h>
-#include <cucascade/memory/topology_discovery.hpp>
 #include <duckdb.hpp>
 #include <duckdb/common/file_system.hpp>
 #include <duckdb/common/open_file_info.hpp>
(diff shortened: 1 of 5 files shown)


test/cpp/io/s3/test_sirius_httpfs.cpp
```diff
@@ -297,6 +310,264 @@ class exposed_sirius_httpfs final : public sirius::io::s3::sirius_httpfs {
   using sirius::io::s3::sirius_httpfs::SupportsOpenFileExtended;
 };
 
+cucascade::memory::system_topology_info single_gpu_topology()
+{
+  cucascade::memory::system_topology_info topology;
+  topology.num_gpus = 1;
+  cucascade::memory::gpu_topology_info gpu;
+  gpu.id        = 0;
```
aminaramoon: use configrutor to have a portable memory_manager, if you don't have numa_node 0 this code wouldn't be portable

aminaramoon (changes requested):

test/cpp/io/s3/test_sirius_httpfs.cpp
ran-yuan-rui: Sure, did the cleanup for all places , I should not use the fake multi GPU setting here

mbrobbel:
Marking as draft. There is a conflict.

PR #1280, opened 2026-07-24
Partial-barrier build & probe for the hash join operator (STANDARD / MIXED_JOIN)

wmalpica:
Summary

Makes sirius_physical_hash_join support partial barriers on both its build and probe input ports when running in HASH_JOIN_MODE::STANDARD or MIXED_JOIN. Previously these modes required a full barrier: the join could not schedule any work until both upstream pipelines had completely finished, and it snapshotted all batch IDs from both sides exactly once. Now batches arrive progressively on each side and the join schedules per-partition buildÃprobe cross-product pairs as they become available, freeing each batch as soon as it can no longer be needed.

BUILD_PROBE mode already streamed its probe side and is unchanged (it served as the reference pattern).

Why this is correct for every join type

The upstream sirius_physical_concat operator already folds whichever side must be seen "whole" for a given join type down to a single batch (_concat_all):

- LEFT / SEMI / ANTI -> build folded to 1 batch; probe streams
- RIGHT / RIGHT_SEMI / RIGHT_ANTI -> probe folded to 1 batch; build streams
- OUTER (FULL) -> both folded (stays 1x1)
- INNER -> neither folded; both may stream (inherently safe for batch-by-batch union)
- MARK -> forced into BUILD_PROBE mode (out of scope here)

Because the must-be-whole side always presents a single folded batch, streaming the other side batch-by-batch against it is correct for all join types. No cross-batch match-accumulation logic is added -- this is purely a scheduling/lifetime change. A defensive invariant in refresh_cross_schedule throws loudly if a concat regression ever violates the "whole side stays one batch" assumption.

What changed

- src/include/op/sirius_physical_hash_join.hpp -- replaced the one-shot snapshot members with a growable, mutex-guarded partition_cross_schedule per partition; added three pure helper functions and a refresh_cross_schedule() declaration.
- src/op/sirius_physical_hash_join.cpp:
  - next_cross_schedule_pair, peek_cross_schedule_kind, collect_cross_schedule_discards -- pure, unit-tested scheduling logic.
  - refresh_cross_schedule() -- re-polls both ports for new batch IDs, enforces the whole-side invariant, and frees fully-consumed batches.
  - Rewrote the STANDARD/MIXED path of get_next_task_input_data and added get_next_task_hint.
- src/pipeline/sirius_pipeline_converter.cpp -- resolve_barrier now stamps the CONCAT -- HASH_JOIN edge as PARTIAL instead of FULL.
- docs/super-sirius/operators.md -- new "Partial-barrier scheduling" subsection with an expanded per-join-type capability table (concat fold / streams-multi-batch / correctness
/ multi-partition / broadcast / supported modes).
- test/cpp/operator/test_cross_schedule.cpp (+ CMakeLists.txt) -- new GPU-free unit tests for the scheduling helpers.

Design notes

- Popping is decoupled from pair-emission. A batch's final pairing may be scheduled before the opposite side finishes, so freeing is handled by an idempotent discard sweep in
refresh_cross_schedule: a batch is freed only once the opposite producer aired with every opposite batch. Borrowers hold live shared_ptr copies across the pop, so no use-after-free.
- Barrier lever: we flip the ports to PARTIAL (rather than only overriding the hint). Accepted trade-off: query_index.cpp reads port->type for lookahead branch-cut ordering, so this alters scheduling order (not result correctness) for joins routed through these port. Easily reverted by dropping the resolve_barrier case if it regresses  the hint/input-data rewrite stands on its own.
- Multi-GPU partition routing is preserved (partitioned_operator_data still tagged with the real partition index).
- We also need to check to if any of the join sides emits no data, if do we need to populate an empty batch to then do a join against the empty data. This is the easiest way to capture all the scenarios that arise (like left outer joins still may need to produce nulls)

Testing

- New [cross_schedule] unit tests: 8 cases / 34 assertions 
- All other tests pass
- Tested 4 GPU SF1K benchmarks. Validated and appear faster.


docs/super-sirius/operators.md
```diff
@@ -158,6 +158,27 @@ Three execution modes:
...
+
+Joining each build batch against each probe batch and unioning the results is only *inherently* correct for INNER. For every other join type, `sirius_physical_concat` folds the side that must be seen whole into a **single batch** (an implicit full barrier), so the streamed side is joined against that one folded batch — correct for all types. `refresh_cross_schedule` asserts this "whole side stays one batch" invariant and throws if a concat regression ever violates it. The following table summarizes, per join type, what concat folds, which side streams under this scheme, and the partition / broadcast / mode support:
+
+| JoinType | concat fold (whole side) | Streams multi-batch | batch×batch + union | Multi-partition | Broadcast | Supported modes |
```
aminaramoon, line 171: this is so useful.

aminaramoon (approved):

PR #1281, opened 2026-07-24
test only (s3): cover common SQL operator shapes over the transparent S3 scan

ran-yuan-rui:
Adds six op-shape coverage tests that run a query over an S3 table and check the result against a
plain-DuckDB-CPU oracle reading the same bytes locally, so a mismatch points straight at an S3-path
bug (DECIMAL scale, NULL bitmask, VARCHAR offsets, mixed-scan planning). The features already work;
these pin them.

The shapes:

- T1 — LEFT/RIGHT outer join (null-producing side) plus a populated LEFT join with aggregation.
- T2 — multi-key GROUP BY with HAVING, COUNT(DISTINCT), and a DECIMAL SUM.
- T3 — VARCHAR LIKE (`%green%`, `forest%`) and IN-set predicates.
- T4 — one local scan joined to one S3 scan in the same gpu_execution plan.
- T5 — NULL / DECIMAL-edge / TIMESTAMP-edge round-trip (new `edge_types.parquet` fixture, ~20 rows).
- T6 — CTE + join + scalar subquery with a DESC ORDER BY and LIMIT.

Every query ends in a total ORDER BY so the order-sensitive row compare is deterministic, and any
DECIMAL/float column is compared with a tolerance. All queries go through the transparent
`read_parquet('s3://')` surface, not the retiring `sirius_read_parquet` rewrite.

Test only, no production change. The comparison helper for watchdog-wrapped queries gained an
optional tolerant-column list (empty keeps its old exact-compare behavior, so existing cases are
unchanged). Runs under `make s3-test`; +6 Catch2 cases (83 → 89).


test/cpp/integration/data/parquet/edge_types.parquet
mbrobbel, line 1: Maybe we can use `CREATE TABLE` instead of adding binary files to the repo? If we really need a parquet file it can be written as a temp file during test setup.

ran-yuan-rui: Good call — dropped the binary. The fixture is now written during test setup into the same temp fixture directory generate_fixtures.py populates and upload_fixtures uploads from, so the S3 object and the local oracle still read the identical file. The edge values are now explicit SQL

mbrobbel (approved):

Issue #990, opened 2026-06-22
Investigate removing operators if estimated data size is small


wmalpica:
We should see if DuckDB plans can provide us with an estimate of the size of data arriving at a LOGICAL operator.
If the data is small, then we might be able to make some operator combinations reduce down to 1.
For example, if data is small, instead of having grouped_aggregate -> partition -> merge_aggregate
We could just have a single grouped_aggregate operator.
Similarly with ungrouped aggregate, joins and order by.

First we need to see what information is available at query graph construction time.
Then we would need to implement this on an operator by operator basis (we probably dont want to do this in one single PR).
For each operator we might need to make some changes.
For example, in the grouped and ungrouped aggregators case, we would need to ensure all the data is in just 1 task, which changes the pipelineability requirements. And the prepare the data for the task would need to collect all batches coming in and concatenate them during the execute portion of the task.

Jedi18:
## Small-Data Operator Simplification: Investigation and Proposal

### Background

For supported queries, Sirius builds a GPU pipeline tree where aggregates, joins, and sorts are expanded into multi-pipeline chains. A grouped aggregate, for example, becomes three pipelines:

```text
P1: input         -> HASH_GROUP_BY (sink)       // per-batch local aggregates
P2: HASH_GROUP_BY -> PARTITION (sink)           // bucket by group keys
P3: PARTITION     -> MERGE_GROUP_BY (sink)      // combine partials into final result
```

The proposal is to reduce this structure for data that can fit comfortably on one GPU. For grouped aggregate, this likely means reducing `3 -> 2` pipelines, not truly collapsing everything into one pipeline, because the final collect-all merge still needs to wait for local aggregation output. For ungrouped aggregate or some sort paths, a `2 -> 1` reduction may be possible. Similar ideas may apply to joins and order by, but those operators need separate measurement because their current fast paths and constraints differ.

### 1. Option: CPU fallback for genuinely small whole queries

Sirius already has whole-query CPU fallback through `run_internal_cpu_fallback_query` in `src/sirius_extension.cpp`. Today this is used for unsupported plans or GPU execution errors, not as a size-based optimization.

We could add a plan-time, size-gated trigger: if the whole query's source cardinality is tiny, skip GPU execution entirely and let DuckDB run it on CPU.

[...]

PR #1282, opened 2026-07-24
feat(bypass): skip partition stages for small whole-query scans (#990)

Jedi18:
## What & why

When a whole query's data is small, the partition / sort-sampling machinery is pure
overhead — it exists to chunk large data across tasks and GPUs. This adds a **whole-query
small-data gate**: when the summed estimated bytes across all base scans fall below
`small_query_bytes_threshold` (default **256 MiB**, clamped to `scan_task_batch_size` so
every scan is single-batch), the plan generator skips `PARTITION` / `SORT_SAMPLE` /
`SORT_PARTITION` and wires producers directly into the retained terminal `CONCAT` / `MERGE`
operators.

## How it works

- **Gate** (`estimate_base_scan_bytes`): sums `TABLE_SCAN` (metadata cardinality × row width,
  with VARCHAR width taken from the column's max-length statistic and **failing closed** when
  that stat is missing) and `COLUMN_DATA_SCAN` (exact materialized size). Decided once,
  whole-query, before wrapping.
- **Retained operators carry correctness**: `CONCAT` now accepts non-partitioned input as
  partition 0, and MARK folds the complete build side via `_concat_all` so the unpartitioned
  path stays correct. `ORDER_BY → MERGE_SORT` uses a `FULL` barrier; join sibling-linking
  guards against the absent `PARTITION`.
- New `small_query_bytes_threshold` config / `SET` option (`0` disables).

## Eligibility & limits

Ineligible (fall through to the normal partitioned path): CTE / delim / recursive plans,
VARCHAR scans with no max-length statistic, and — **on multi-GPU** — any plan containing a
join. That last guard is deliberate: the bypass runs everything on one partition (GPU 0), so
a join whose output amplifies past cuDF's per-column row limit would crash where the
partitioned path spreads it across GPUs. Bypassed joins stay STANDARD (no BUILD_PROBE /
dynamic filters). The gate models leaf-scan **input** only, not join / aggregate **output**
fan-out.

## Correctness

Skipping partitioning never changes results — it's a scaling device, not a semantic one. The
`CONCAT` `_concat_all` fold, the STANDARD join batch grid, and the k-way `MERGE_SORT` are all
partition-count-agnostic, and the gate only ever sees supported operators (anything else has
already fallen back to CPU at plan time).

## Testing

- `test_small_query_bypass_converter` — asserts the bypassed plan omits the partition stages
  and keeps the terminal `CONCAT` / `MERGE` operators; covers gate eligibility (threshold,
  materialized CTE, `COLUMN_DATA_SCAN`).
- `test_gpu_execution_small_query_bypass` — GPU-vs-CPU correctness across aggregates,
  `ORDER BY` / TOP-N + OFFSET, all join types, and the correlated-subquery exclusion.
- `test_physical_concat` — MARK build-side `_concat_all` fold.
- Full C++ suite green on-device.

## Follow-ups (deferred)

- Runtime coverage for the multi-GPU join guard (needs a `num_gpus` test seam) and the
  multi-partial merge path under bypass (needs integration-scale amplifying data).
- Optional: a coarse output-cardinality estimate to allow selective joins to bypass on
  multi-GPU instead of the blunt "any join → partitioned path".

### DIFF
--- a/src/pipeline/sirius_pipeline_converter.cpp
+++ b/src/pipeline/sirius_pipeline_converter.cpp
@@ -199,7 +199,14 @@ op::MemoryBarrierType sirius_pipeline_converter::resolve_barrier(
   // Sort sinks process batches as they arrive. GPU_SCAN is intentionally excluded
   // (legacy gives it FULL/PARTIAL, not PIPELINE); SORT_SAMPLE is never a pipeline
   // sink (it runs as an intermediate in the SORT_PARTITION pipeline).
-  if (sink.type == T::ORDER_BY) { return op::MemoryBarrierType::PIPELINE; }
+  // Small-query bypass (issue #990): ORDER_BY's tree parent is MERGE_SORT directly
+  // (no SORT_SAMPLE/SORT_PARTITION) — merge must wait for ALL local sorts (FULL).
+  if (sink.type == T::ORDER_BY) {
+    if (dest.get_sink() && dest.get_sink()->type == T::MERGE_SORT) {
+      return op::MemoryBarrierType::FULL;
+    }
+    return op::MemoryBarrierType::PIPELINE;
+  }
   // Producers that feed CONCAT can drain incrementally (PARTIAL); otherwise wait
   // for the upstream pipeline to finish (FULL).
   if (sink.type == T::PARTITION || sink.type == T::UNGROUPED_AGGREGATE || sink.type == T::TOP_N ||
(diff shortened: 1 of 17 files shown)


src/pipeline/sirius_pipeline_converter.cpp
```diff
@@ -422,10 +429,27 @@ void sirius_pipeline_converter::link_join_partition_siblings()
     // both can stream the probe, so the upstream→probe-partition edge is PARTIAL for both.
     if (pipeline->source->type == op::SiriusPhysicalOperatorType::HASH_JOIN ||
         pipeline->source->type == op::SiriusPhysicalOperatorType::NESTED_LOOP_JOIN) {
-      auto build_concat_pipeline    = pipeline->dependencies[0];
+      D_ASSERT(pipeline->dependencies.size() >= 2);
+      if (pipeline->dependencies.size() < 2) { continue; }
```
wmalpica, line 433: i think we might want to just throw an error if the number of dependencies is not explicitly 2

src/pipeline/sirius_pipeline_converter.cpp
```diff
@@ -422,10 +429,27 @@ void sirius_pipeline_converter::link_join_partition_siblings()
...
+        continue;
+      }
+
       auto build_partition_pipeline = build_concat_pipeline->dependencies[0];
-      auto probe_concat_pipeline    = pipeline->dependencies[1];
       auto probe_partition_pipeline = probe_concat_pipeline->dependencies[0];
+      using T                       = op::SiriusPhysicalOperatorType;
+      const bool build_has_partition =
+        build_partition_pipeline->get_sink()->type == T::PARTITION ||
+        build_partition_pipeline->get_sink()->type == T::RIGHT_DELIM_JOIN;
```
wmalpica, line 450: do the RIGHT_DELIM_JOIN operator perform partitioning?  The delim joins changed a lot recently. Can you double check this

wmalpica (commented):
In general I am concerned that this PR introduces more complexity than its worth. Its not too complex, but i also does not do much. My understanding is that if all the scans together are less than a certain size, then we can skip the partition operators.  Which means it does not apply often and when it does, it does not do much.

If we wanted this to be more effective, we would need to do more work. We would need to have this apply to parts of a plan (to plan branches that are guaranteed to be small), not just all or nothing. And it would also have to reduce grouped aggregations and sorts to just one operator. 

I am inclined to suggest we either postpone it until it becomes more necessary to have this, or we expand the scope so that its more useful.
Also there are other similar initiatives which are similar in spirit which may result in more meaningful benefits today, like: https://github.com/sirius-db/sirius/issues/1283

Jedi18:
> In general I am concerned that this PR introduces more complexity than its worth. Its not too complex, but i also does not do much. My understanding is that if all the scans together are less than a certain size, then we can skip the partition operators. Which means it does not apply often and when it does, it does not do much.
> 
> If we wanted this to be more effective, we would need to do more work. We would need to have this apply to parts of a plan (to plan branches that are guaranteed to be small), not just all or nothing. And it would also have to reduce grouped aggregations and sorts to just one operator.
> 
> I am inclined to suggest we either postpone it until it becomes more necessary to have this, or we expand the scope so that its more useful. Also there are other similar initiatives which are similar in spirit which may result in more meaningful benefits today, like: #1283

My tests showed that this doesn't really give any performance benefit even at small SFs so I agree that as it stands, this may not be worth the added complexity as you said.

I think I'll put this on hold for the moment for tasks like #1283 which I've started looking into it and come back to this later.

bwyogatama:
Im putting this on draft as this seems to be on hold

PR #1286, opened 2026-07-24
feat(benchmark): TPC-H power & throughput runs (RF1/RF2 refresh)

bwyogatama:
## What

Adds TPC-H power and throughput runs that interleave the RF1 (insert) and RF2 (delete) refresh
functions with the query streams, exercising the pinned duckdb-native insert-delta / delete-mask
path and reporting Power@Size, Throughput@Size, and QphH@Size. The methodology follows
`duckdb/duckdb-tpch-power-test`, with the spec followed where that harness diverges (table below).

```
power:      [clean] ─ RF1 ─ stream 0 (timed → Power@Size) ─ RF2 ─ post-RF2 (mask, timed)
throughput: N query streams (spec perms) + 1 refresh stream, concurrent submission
```

## Files

- `test/tpch_performance/tpch_power_throughput.py` — the runner. Power run: optional clean pass,
  RF1, timed stream 0, RF2, timed post-RF2 pass (delete mask active), with GPU-vs-CPU validation
  after RF1 and RF2. Throughput run: N concurrent query streams plus a refresh stream.
- `test/tpch_performance/generate_tpch_refresh.sh` — refresh sets (`orders.tbl.u*`,
  `lineitem.tbl.u*`, `delete.*`) via classic dbgen `-U`.
- `test/tpch_performance/generate_tpch_queries.sh` — per-stream query sets (`stream<N>.sql`) via
  `qgen`, the reference substitution-parameter generator.
- `test/tpch_performance/tpch_query_streams.py` — load those files: split on qgen's `(Q<n>)`
  tags and fold the `:n` row limit into a `LIMIT`.
- `test_datasets/tpch-dbgen.zip` — upgraded from dbgen 2.14.0 to TPC tools 3.0.1, with the EULA;
  `test/tpch_performance/dbgen_bootstrap.sh` builds `dbgen`/`qgen` on demand.
- `test/tpch_performance/tpch_stream_permutations.py` — spec Appendix A stream orderings
  (verified identical to `qgen -p` for all 41 streams) + spec-minimum stream counts.
- `test/tpch_performance/dbgen_bootstrap.sh` — shared unzip/build of `dbgen` / `qgen`.
- `.claude/skills/benchmark/SKILL.md` — prompts regular vs official run, always asks for existing
  data paths before generating, gpu/host pin tier, fixed vs varied predicates.
- `test/tpch_performance/CLAUDE.md` — usage, metrics, spec-fidelity table, serialization caveat.

## Notes

- Requires a file-backed `.duckdb` with native tables; parquet views can't be mutated. The run is
  always pinned, tier `gpu` or `host`. RF1/RF2 run on DuckDB CPU; the GPU then serves the queries
  from `base + delta − mask` with no checkpoint between.
- `--mode` is `power`, `throughput`, or `both`. In `both`, throughput continues on the same
  pinned database right after power (no unpin/repin), picking up the update-set-1 state and
  running sets 2..N+1 to match the spec's continuous refresh sequence.
- Predicates are fixed by default — all streams share one literal set, so passes stay comparable
  and validation is supported. `--vary-predicates` runs each stream's own qgen parameters, which
  is what an official run requires; validation is rejected there.
- Each query runs in its own transaction, `READ ONLY` except q15, which creates and drops a view.
  qgen names it `revenue<stream>`, so concurrent streams don't collide.
- The engine serializes queries on one lifecycle lock, so the throughput run measures throughput
  of concurrent submission on one GPU, not overlapped execution. This is expected, not a bug.

## Spec fidelity vs duckdb-tpch-power-test

| Area | That repo | Here |
|------|-----------|------|
Checked against the TPC-H v3.0.1 specification.

| Area | That repo | Here |
|------|-----------|------|
| Power run stream | stream 1, reused as a throughput stream | stream 0, ordering number O(00) (5.3.5.2) |
| Throughput refresh pairs | `max(SF/10, 1)`, unrelated to stream count | one RF1/RF2 pair per stream (5.3.7.7) |
| RF1/RF2 in throughput | merged into one transaction | separate transactions, RF1 before RF2 (5.3.7.8) |
| Row limits (`:n`) | dropped, so q2/q3/q10/q18/q21 run unlimited | folded into a `LIMIT` |

Inherited as-is: query text from `qgen`, one transaction per query, and Power@Size as the geomean
over 22 query times plus T_RF1 and T_RF2 (5.4.1).

`test_datasets/tpch-dbgen.zip` now bundles **TPC tools 3.0.1**, the version that repo vendors,
replacing the previous 2.14.0. 3.0.1 draws q4 and q22 within the spec's ranges natively and drops
q1's ANSI `day (3)` qualifier, so no source patching or SQL rewriting is needed. Generated data is
unchanged — 2.14.0 and 3.0.1 produce byte-identical `.tbl` and refresh files at SF1, so existing
datasets stay valid. The bundle carries the TPC EULA and the legend it requires and ships no
compiled binaries, so `setup_test_datasets.sh` builds `dbgen` first and the bootstrap refreshes an
older extraction in place, leaving generated `.tbl` data beside it untouched.

## Verified

- `qgen` output parses to the 22 queries in the right order for streams 0–2, with the spec's row
  limits (q2 100, q3 10, q10 20, q18 100, q21 100); all of it executes on CPU DuckDB at SF1.
- The 3.0.1 upgrade is data-safe: SF1 base tables and refresh sets are byte-identical between
  2.14.0 and 3.0.1 (14/14 files), and the bundle builds `dbgen` and `qgen` from a clean unzip.
- Our permutation table matches `qgen -p` for all 41 streams; the stream-count table matches spec
  Table 11 through SF100000, and beyond the largest valid scale factor the runner refuses to
  default rather than silently under-provisioning.
- After the substitution fixes, q22 codes stay within 10..34 and q4 dates within
  1993-01…1997-10 across nine generated streams.
- Prior SF1/SF10 runs validated end-to-end (all 22 power queries GPU == CPU after RF1 and RF2,
  row counts exact, zero GPU→CPU fallbacks) — but that predates the both-mode, transaction, and
  qgen changes, so QphH needs re-measuring and a GPU smoke run is still outstanding.


aocsa (commented):
LGTM, I ran this at SF1 on an RTX 4070 Laptop with `--pin gpu`, both predicate modes. Main RF1/RF2 path works: Power@Size 38250, Throughput@Size 34366, QphH 36256, validation passed, 800 `GPU_SCAN` markers, no fallbacks. Serialization matches the docs (0.10s alone → 0.20s with 2 streams). SF1 mostly shows overhead, not real perf; `clamped_query_times` was 0.

**May be a blocking: `--vary-predicates` gives every stream the same predicates.** `generate_tpch_queries.sh` only passes qgen `-r` with `--seed`. Without it, qgen uses `time(NULL)`, which is second-granularity, so the fast loop stamps all streams with the same seed. `-p` only reorders queries. Fix: `SEED="${SEED:-$(date +%s)}"` and always `-r "$((SEED + n))"`, plus a duplicate-seed check. This matters because that mode drops validation and is what the PR calls the official path.

aocsa (approved):

PR #1290, opened 2026-07-26
test(tpcds): GPU-vs-CPU NULL differential suite over TPC-DS (#1095)

### LINKED ISSUE #1219: Add a TPC-DS (with NULLs) differential correctness run
Sub-issue of #1095. Broad, realistic regression guard for NULL handling, to
complement the targeted unit-style tests (#<A>).

## Motivation
The targeted null tests pinpoint specific operators. A full query-suite run over
data that naturally contains NULLs catches *interaction* bugs across
scan→join→aggregate→filter that isolated tests miss.

## Why TPC-DS (not TPC-H)
TPC-DS's schema naturally carries NULLs (nullable measures, dates, dimension
keys), so it exercises NULL paths in realistic multi-operator queries without
synthetic injection. TPC-H is almost entirely `NOT NULL`, so it would require
injecting NULLs and regenerating expected answers.

## Approach
GPU-vs-DuckDB-CPU **differential** comparison (not fixed answers), reusing the
`GpuExecutionFixture` pattern over a generated TPC-DS dataset at a small scale
factor. Run a curated subset of queries that touch NULL-bearing columns; assert
GPU output matches CPU (and, where expected, that the query runs on GPU without
fallback).

## Sequencing (important)
Do this **after** the NULL-handling bugs (#<B>) are fixed. Running it now would go
red across many queries that all trace back to the same known root bugs (`OR`,
ungrouped `AVG`, etc.) — high noise, low signal. Once the engine handles NULLs
correctly, this run lands green and becomes a durable regression guard.

## Open questions
- Scale factor for CI (correctness ≠ perf; keep it small/fast).
- Which query subset — all TPC-DS, or a NULL-focused curated set?
- Where unsupported operators force CPU fallback, decide per-query whether to
  assert GPU execution or just result-correctness.

Jedi18:
## What

A NULL-focused GPU-vs-CPU differential test suite over a real TPC-DS dataset,
Closes #1219. Complements the existing per-operator
synthetic NULL tests with an integration layer: NULL handling over scattered,
real NULLs at scale, through operator combinations and across tables.

Tests-only — no production code changes.

## Coverage (52 queries / 11 cases)

- **Filters / 3VL** — `IS [NOT] NULL`, three-valued `OR`/`AND`, `IS NOT DISTINCT FROM`, `BETWEEN`, `IN`, `NOT IN (…, NULL)`
- **Aggregates** — `COUNT(*)` vs `COUNT(col)`, integer `SUM`/`AVG`, `MIN`/`MAX`, `GROUP BY` nullable keys (single + multi), and decimal `SUM`/`AVG` via a tolerance comparator
- **Joins** — `LEFT JOIN` NULL-padding (incl. probe-side nullable key), `INNER JOIN` NULL-key drop, star joins
- **Strings** — `concat` (ignores NULL) vs `||` (propagates), `length`, `LIKE`/`NOT LIKE`, `COALESCE`
- **Expressions** — arithmetic / `CAST` / `COALESCE` / `NULLIF` / `CASE` propagation
- **Ordering** — `ORDER BY … NULLS FIRST/LAST`, `DESC`, and top-N `LIMIT`
- **Date functions** — `year()` over NULL, returns-table measures

## Notes

- **Targeted queries, not canonical q1–q99** — the goal is NULL-interaction coverage; canonical queries lean on operators that mostly fall back today.
- **Deterministic by construction** — exact comparison uses only order-independent aggregates (`COUNT`, integer `SUM`/`AVG`, `MIN`/`MAX`). Decimal `SUM`/`AVG` go through a new `compare_gpu_vs_cpu_approx` (relative tolerance on numeric cells, exact otherwise), since GPU parallel reduction and CPU serial summation differ in the low bits. Signed cancellation-prone measures (`ss_net_profit`) are covered exactly via `COUNT`/`MIN`/`MAX` only.
- **Data** — a pre-generated TPC-DS (sf=0.01) DuckDB file committed under `test/cpp/integration/data/duckdb/tpcds.duckdb` (produced offline by `generate_tpcds_duckdb.sh`), attached read-only at test time — so the suite needs no network or tpcds extension and runs in CI. A precondition case asserts it contains real NULLs (~4–5%) in every column under test. Override the path with `SIRIUS_TPCDS_TEST_DB_PATH`.


bwyogatama (approved):

bwyogatama:
@Jedi18 Quick comment, is it necessary to check the parquet path or do you think duckdb path should be enough?

Jedi18:
> @Jedi18 Quick comment, is it necessary to check the parquet path or do you think duckdb path should be enough?

I think duckdb path should be sufficient for these tests (adding parquet path would be redundant imo), though we should add a specific targetted test for how the parquet reader materializes/interacts with nulls, which I can do in a follow-up.

wmalpica (approved):

PR #1292, opened 2026-07-27
fix(join): make IS NOT DISTINCT FROM joins null-safe on GPU

### LINKED ISSUE #1291: Bug: GPU silently drops `NULL=NULL` matches in `IS NOT DISTINCT FROM` joins
## Summary
GPU execution lowers `IS NOT DISTINCT FROM` join predicates to plain equi-joins (`=`). This silently drops `NULL=NULL` matches instead of falling back to CPU, resulting in undercounted results.

## Reproduction (TPC-DS sf=0.01)
```sql
SELECT count(*)
FROM store_sales ss
JOIN store_returns sr ON ss.ss_addr_sk IS NOT DISTINCT FROM sr.sr_addr_sk;
```
- **CPU (expected):** `280,213`
- **GPU (actual):** `148,018` *(missing all `NULL=NULL` matches)*

## Minimal Repro
```sql
CREATE TABLE a(k INTEGER); CREATE TABLE b(k INTEGER);
INSERT INTO a VALUES (1), (NULL), (NULL);
INSERT INTO b VALUES (1), (NULL);
SELECT count(*) FROM a JOIN b ON a.k IS NOT DISTINCT FROM b.k;
```
- **Expected:** `3` *(1↔1 + two NULL↔NULL pairs)*
- **Actual:** `1` *(NULLs dropped)*

## Root Cause & Proposed Fix
The planner translates `IS NOT DISTINCT FROM` into a standard equi-join key without null-safe matching. 
- **Fix:** Implement null-safe key matching on the GPU, or fall back to CPU.

## Tracking
Quarantined in `test/cpp/integration/test_gpu_execution_tpcds_nulls.cpp` (`feat/tpcds-null-tests`).
https://github.com/sirius-db/sirius/pull/1290

Jedi18:
## Problem

Fixes https://github.com/sirius-db/sirius/issues/1291

A join keyed on `IS NOT DISTINCT FROM` must be **null-safe** — `NULL` matches `NULL` — unlike a plain `=` join. On the GPU it wasn't: `sirius_physical_hash_join` hardcoded `cudf::null_equality::UNEQUAL` at every cuDF join call site. The condition was correctly recognized as an equi-key (so the query ran on the GPU with **no fallback**), but cuDF was told NULL never equals NULL, so NULL-to-NULL pairs were silently dropped — a **wrong result, not a fallback**.

Observed on TPC-DS (sf=0.01):

```sql
SELECT count(*) FROM store_sales ss JOIN store_returns sr
  ON ss.ss_addr_sk IS NOT DISTINCT FROM sr.sr_addr_sk;
-- GPU: 148018   CPU: 280213
```

Found by the TPC-DS NULL differential suite (#1219), where it is currently quarantined.

## Fix

- Add `sirius_physical_hash_join::compare_nulls()`, which returns `null_equality::EQUAL` only when **every** equi-key is `not_distinct_from` (null-safe) and `UNEQUAL` otherwise, and thread it through all join calls (inner/left/full/hash/distinct_hash/mixed/filtered/mark) and the three build helpers — replacing the hardcoded constant.
- **IN / EXISTS (MARK joins)** compare with `=`, so `compare_nulls()` yields `UNEQUAL` for them; their behavior and three-valued NULL logic are unchanged. The fix only ever selects `EQUAL` for an explicit `IS NOT DISTINCT FROM`.
- **Mixed keys** (`a = b AND c IS NOT DISTINCT FROM d`): cuDF applies one `null_equality` flag to all key columns, so per-column semantics can't be expressed. Such joins keep `UNEQUAL` (the prior behavior) — `EQUAL` is selected only for a *pure* `IS NOT DISTINCT FROM` join, and any `=` key forces `UNEQUAL`. A mixed join whose null-safe column has NULLs that should match remains a pre-existing limitation, not introduced here. (Delim joins for correlated subqueries are legitimately mixed, so they must keep running on the GPU rather than being rejected.)

## Test

Adds `test_gpu_execution_null_safe_join.cpp`: GPU-vs-CPU null-safe inner and left joins over tables with NULL keys on both sides, a plain-`=` regression guard (ordinary equality joins must still drop NULL keys), and a mixed `=` + `IS NOT DISTINCT FROM` join that must stay on the GPU.

### DIFF
--- a/src/op/sirius_physical_hash_join.cpp
+++ b/src/op/sirius_physical_hash_join.cpp
@@ -87,26 +87,28 @@ static bool is_equality(sirius::comparison_type c)
 }
 
 static cudf::filtered_join make_right_filtered_join(cudf::table_view const& right_keys,
+                                                    cudf::null_equality compare_nulls,
                                                     rmm::cuda_stream_view stream)
 {
 #if CUDF_VERSION_MAJOR > 26 || (CUDF_VERSION_MAJOR == 26 && CUDF_VERSION_MINOR >= 6)
-  return cudf::filtered_join(right_keys, cudf::null_equality::UNEQUAL, stream);
+  return cudf::filtered_join(right_keys, compare_nulls, stream);
 #else
-  return cudf::filtered_join(
-    right_keys, cudf::null_equality::UNEQUAL, cudf::set_as_build_table::RIGHT, stream);
+  return cudf::filtered_join(right_keys, compare_nulls, cudf::set_as_build_table::RIGHT, stream);
 #endif
 }
 
 // Heap-allocated variant for BUILD_PROBE mode, where one filtered_join is built once on the right
 // (filter) keys and reused across many streamed left probe batches via semi_join.
 static std::unique_ptr<cudf::filtered_join> make_right_filtered_join_ptr(
-  cudf::table_view const& right_keys, rmm::cuda_stream_view stream)
+  cudf::table_view const& right_keys,
+  cudf::null_equality compare_nulls,
+  rmm::cuda_stream_view stream)
 {
 #if CUDF_VERSION_MAJOR > 26 || (CUDF_VERSION_MAJOR == 26 && CUDF_VERSION_MINOR >= 6)
-  return std::make_unique<cudf::filtered_join>(right_keys, cudf::null_equality::UNEQUAL, stream);
+  return std::make_unique<cudf::filtered_join>(right_keys, compare_nulls, stream);
 #else
   return std::make_unique<cudf::filtered_join>(
-    right_keys, cudf::null_equality::UNEQUAL, cudf::set_as_build_table::RIGHT, stream);
+    right_keys, compare_nulls, cudf::set_as_build_table::RIGHT, stream);
 #endif
 }
 
@@ -114,9 +116,10 @@ static std::unique_ptr<cudf::filtered_join> make_right_filtered_join_ptr(
 // Wins over make_right_filtered_join only when the left side is substantially smaller than the
 // right; gated by mark_join_build_switch_ratio at the call site.
 static cudf::mark_join make_left_mark_join(cudf::table_view const& left_keys,
+                                           cudf::null_equality compare_nulls,
                                            rmm::cuda_stream_view stream)
 {
-  return cudf::mark_join(left_keys, cudf::null_equality::UNEQUAL, cudf::join_prefilter::NO, stream);
+  return cudf::mark_join(left_keys, compare_nulls, cudf::join_prefilter::NO, stream);
 }
 
 bool sirius_physical_hash_join::are_conditions_supported(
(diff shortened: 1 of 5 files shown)


src/op/sirius_physical_hash_join.cpp
```diff
@@ -234,6 +237,27 @@ sirius_physical_hash_join::sirius_physical_hash_join(
...
 
+  // Cache the null-matching flag for cuDF joins (conditions and join_type are fixed
+  // hereafter). cuDF applies one flag to all key columns, so EQUAL (null-safe) is
+  // used only when EVERY equi-key is IS NOT DISTINCT FROM; any plain `=` key
+  // (including mixed / delim joins) forces UNEQUAL. MARK joins (IN/EXISTS) drive
+  // three-valued logic that is only correct under UNEQUAL and always compare with
+  // `=`, so force UNEQUAL there to keep that invariant even if a null-safe key ever
+  // reached the mark path.
+  compare_nulls_ = cudf::null_equality::UNEQUAL;
+  if (join_type != duckdb::JoinType::MARK) {
```
kevkrist, line 253: I suppose this is not quite right about MARK joins always comparing with equality:

-- l.k contains: NULL, 1, 2
-- r.k contains: NULL, 1
SELECT l.k,
               EXISTS( SELECT 1
                              FROM r
                              WHERE r.k IS NOT DISTINCT FROM l.k ) as e
FROM l;

Here the MARK_JOIN uses IS NOT DISTINCT FROM condition. The result set should be {(NULL, true), (1, true), (2, false)}, but Sirius is emitting {(NULL, NULL), (1, true), (2, NULL)} because of this gate at this line.

[Codex]

test/cpp/integration/test_gpu_execution_null_safe_join.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+TEST_CASE_METHOD(NullSafeJoinFixture,
+                 "gpu_execution null-safe RIGHT and FULL OUTER joins match NULL to NULL",
+                 "[integration][gpu_execution][join][nulls]")
+{
+  // Cover the right-swap and full-outer null-safe paths (in addition to
+  // inner/left): NULL keys still match, and unmatched rows are NULL-padded on the
+  // appropriate side.
+  compare_gpu_vs_cpu("SELECT l.id, r.id FROM l RIGHT JOIN r ON l.k IS NOT DISTINCT FROM r.k");
```
kevkrist: Can you ensure that DuckDB doesn't rewrite this into a LEFT JOIN with the tables swapped? If so, you may need to disable the `build_side_probe_side` optimizer, or use a direct operator-level test.

kevkrist (changes requested):

src/op/sirius_physical_hash_join.cpp
```diff
@@ -234,6 +237,27 @@ sirius_physical_hash_join::sirius_physical_hash_join(
...
+    bool saw_null_safe   = false;
+    bool saw_plain_equal = false;
+    for (auto const& cond : conditions) {
```
wmalpica, line 256: So what happens if you have 
`FROM A LEFT OUTER  JOIN B on  A.a = B.a AND A.b is not distinct from B.b`
in other words when you have both?
in this case `compare_nulls_` would be cudf::null_equality::UNEQUAL

So if we had 
A.a  = 3
B.a  = 3
A.b = null
B.b = null
then:  3 = 3 AND null is not distinct from null
should be true, but with  compare_nulls_ = `cudf::null_equality::UNEQUAL` it becomes false.

So, do we still have a problem? we should double check this.
I am not sure if this is something that gets solved by the MIXED_JOIN join_mode or if it gets handled by the nested loop join operator. We should double check.

There is a possibility we may need to add a not supported error for this scenario

kevkrist: It seems that this might be a limitation in cuDF applying the same null_equality flag across all join keys.

Jedi18: Yes this is a known issue for which I've created https://github.com/sirius-db/sirius/issues/1293. This PR fixes the pure null-safe case and the mixed case is known to be wrong and will be fixed as part of the issue I just mentioned.

I had earlier attempted to reject this mixed case just as you suggested but that broke tests in the CI, so I deferred it to a follow-up task instead.

src/op/sirius_physical_hash_join.cpp
Jedi18: You're right, and this seems to belong to the same limitation as #1293 , so I'll keep the if check since that matches what is on dev right now, and fix it in the followup PR for #1293.

I'll update the comment though since it's wrong

test/cpp/integration/test_gpu_execution_null_safe_join.cpp
Jedi18: Thanks, added an operator level right join test

src/op/sirius_physical_hash_join.cpp
wmalpica: @Jedi18 Can you file an issue with cudf about this missing feature. (first look if there is already an issue about this, and if so make a comment on it)

kevkrist (approved):

src/op/sirius_physical_hash_join.cpp
Jedi18: Yes will do, I'm investigating a work-around for sirius, once I have more information on that I'll file an issue with cudf

PR #1302, opened 2026-07-27
feat(scan): enable cuDF hardware decompression when supported on all GPUs

aminaramoon:
## Summary

Adds an operator config `use_hw_decompression` (default **true**) that, when the installed CUDA driver supports hardware (on-GPU) decompression on **every** GPU, exports `LIBCUDF_HW_DECOMPRESSION=ON` for the lifetime of the `SiriusContext` so cuDF's parquet reader routes supported codecs through the hardware decompression engine.

## Changes

- **New RAII helper `sirius::util::env_guard`** (`src/util/env_guard.{hpp,cpp}`): sets an environment variable on construction and restores its *prior* state (previous value, or unset) on destruction. Move-only so it composes with `std::optional`.
- **New config `use_hw_decompression`** in `operator_params` (`sirius_config.hpp`): defaults to `true`, settable via the `sirius.operator_params` YAML section and at runtime via `SET use_hw_decompression = ...`. Wired into the YAML loader, a `SetUseHwDecompression` handler, and `AddExtensionOption` in `sirius_extension.cpp`.
- **`SiriusContext` wiring** (`sirius_context.cpp`): in `initialize()`, when the flag is set, probe `rmm::detail::hwdecompress::is_supported()` under each GPU's device context (enumerated via the authorized `get_hw_topology()` source, not raw CUDA APIs). If all GPUs support it, `emplace` an `env_guard` exporting `LIBCUDF_HW_DECOMPRESSION=ON`; otherwise skip and log. The guard is `reset()` in `terminate()` (and, being RAII, also restored on context destruction).

## Behavior notes

- The env var is decided once at context-initialize (extension-load) time from the config value; the effective control point is the YAML/default value read before `initialize()`. A runtime `SET` updates the param but does not retroactively toggle the export for the live context. Documented in the option description.
- If any GPU lacks driver support, the export is skipped and cuDF falls back to software decompression.

## Testing

- `sirius::util::env_guard` was compiled standalone and exercised with a runtime test covering: var absent → restored to unset; var present → restored to previous value; move-then-destroy → moved-from restores nothing.
- Verified the `rmm::detail::hwdecompress::is_supported()` signature against the installed rmm header and confirmed all added includes/symbols resolve in the build's include configuration.
- A full `pixi run make` was **not** run in the authoring environment (the `duckdb`/vcpkg submodules were network-blocked). CI build/tests should confirm the remaining translation units.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -367,7 +367,8 @@ set(EXTENSION_SOURCES
     src/transparent/physical_sirius_execution.cpp
     src/transparent/sirius_optimizer_extension.cpp
     src/util/stream_check_wrapper.cpp
-    src/util/segfault_backtrace_handler.cpp)
+    src/util/segfault_backtrace_handler.cpp
+    src/util/env_guard.cpp)
 # cmake-format: on
 
 # Legacy Sirius — all legacy-specific config lives in src/legacy/CMakeLists.txt.
--- a/src/include/sirius_config.hpp
+++ b/src/include/sirius_config.hpp
@@ -130,6 +130,13 @@ struct operator_params {
   /// pin-time statistics capture and the serve-side survivor plan: a table pinned while the flag is
   /// off carries no zone maps and cannot prune until re-pinned with the flag on.
   bool enable_pinned_zone_map_pruning = true;
+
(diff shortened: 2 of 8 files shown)


bwyogatama (approved):

mbrobbel:
Marking as draft. There are some conflicts.

Issue #1276, opened 2026-07-23
Replace channel backpressure with resource-aware flow control for streaming


aocsa:
## Problem

Sirius needs a principled way to complete workloads that are much larger than available memory and to regulate streaming sessions when downstream consumers cannot keep pace. The mechanism must reduce pressure on constrained resources at a point where upstream work can actually pause; a local bounded channel between a repository and a consumer does not provide that guarantee.

## Decision (2026-07-23 streaming design review)

- Remove bounded-channel backpressure from the streaming source/sink design ([#914](https://github.com/sirius-db/sirius/pull/914)).
- Expose the batch repository directly — producers push, consumers pull.
- Rely on the existing downgrade executor to relieve memory pressure (GPU → host → disk).
- Revisit backpressure only when a real workload shows the need, and co-design it with scheduling, prioritization, memory management, and query concurrency.

## Impact on in-flight streaming work

The compute-node exchange design ([#914](https://github.com/sirius-db/sirius/pull/914), tracked by [#826](https://github.com/sirius-db/sirius/issues/826)) specified the streaming operators around an external bounded channel with channel-level backpressure. This decision changes that contract, so the following are affected:

- **[#836](https://github.com/sirius-db/sirius/issues/836) Streaming source operator** — closed, but implemented against an "external bounded channel ... with backpressure." Needs rework to push directly into the repository, dropping channel-level backpressure.
- **[#837](https://github.com/sirius-db/sirius/issues/837) Streaming sink operator** — in progress; pull directly from the repository, no `exchange_channel` backpressure.
- **[#838](https://github.com/sirius-db/sirius/issues/838) Streaming sink partitioning support** — partition fan-out over direct repository access rather than per-partition channels.
- **[#839](https://github.com/sirius-db/sirius/issues/839) Stream session support** — owns lifecycle/end-of-stream over the repository, without treating channel depth as pressure.

These issues should be implemented (or reimplemented) against the direct push/pull model before any backpressure mechanism from this issue is introduced.

## Why a local bounded channel is insufficient

1. It doesn't stop upstream work — scans keep producing while data piles up before the sink.
2. Channel depth ≠ receiver memory state (drained handles can still queue tasks).
3. No global context — one sink can't reason across concurrent queries, fragments, and workers.
4. It can make things worse — local throttling can drive extra spilling/placement churn instead of prioritizing work that frees resources.

Treat backpressure as resource-pressure control and scheduling, not queue occupancy.

## Design goals

[...]

PR #1320, opened 2026-07-28
feat(exec): repository-backed streaming source with sender-aware EOS and a producer-error plane (#836)

aocsa:
Implements #836. Part 1 of a five-PR stack.

**What.** The plan leaf for a *stream*. A scan stands where the plan reads a table; this operator
stands where there is no table at all — a fragment boundary, fed by whatever another fragment or
external producer pushes in at runtime.

**Design: one difference drives everything.** A scan owns its input, so "empty" means exhausted. This
source owns nothing and its senders are remote: "empty" means wait, waiting must not block an engine
thread, and the stream ends only once every expected sender has closed. The queue is a
`shared_data_repository`, not a bounded channel (#1276 removed it; `exchange_channel.hpp` is deleted
rather than ported), and `exec::batch_stream` binds that repository to the stream state.

- **End-of-stream counts senders, not closes.** With M senders, the stream ends when all M have
  closed; two closes from one sender are not two senders, and an unknown id throws. A plain counter
  would end the stream early and silently return partial results.
- **`on_data` fires on every `push()`.** The scheduler asks a source for work once; if it answers
  "waiting" it is never asked again. The hook re-schedules it. Firing on every push leaves no window
  for a batch to slip in unannounced.
- **`fail_input(error)` ends the stream as a failure.** The consumer wakes as if data arrived and its
  next pull rethrows the first error. Without this verb a dead producer looks like a finished one.
- **`query.cpp` makes it reachable.** STREAMING_SOURCE joins GPU_SCAN and GPU_VALUES as a schedulable
  kickoff; a receiver fragment has no scan, so without it `start_query()` throws "query has no
  schedulable scan sources" before a single task runs.

The rest is task-protocol glue: `READY{this}` when a batch or error is pullable, `WAITING{nullptr}`
when open and empty (no upstream operator to redirect to), `nullopt` at a clean end;
`all_ports_empty()` is `stream.drained()`; `execute()` is a pass-through.

**Reading order.**
1. `src/include/op/sirius_physical_streaming_source.hpp` — class comment is the design note.
2. `src/include/exec/batch_stream.hpp` — the primitive's contract, S1–S5 and the error plane P1–P4.
3. `src/exec/batch_stream.cpp` — push/close/fail vs pull/classify under one lock.
4. `src/op/sirius_physical_streaming_source.cpp` — glue; `set_pipeline` wires the two hooks.
5. `src/planner/query.cpp` — 5 lines; `sirius_pipeline.hpp` gains `get_task_creator()`, which is how
   a producer thread re-arms a starved head.
6. `docs/super-sirius/operators.md` — the STREAMING_SOURCE entry and hint table.
7. `test/cpp/exec/test_batch_stream.cpp`, `test/cpp/operator/test_physical_streaming_source.cpp` —
   BSTR-1…17, SRC-1…27, BUG-1…4, REARM-1…4 (need a GPU). `test_exchange_channel.cpp` goes with the
   header it covered.

### Stack: part 1 of 5
- **Base:** `dev`
- **Depends on:** nothing


aminaramoon:
[@aocsa](https://github.com/aocsa) we already have all the capabilities on stream source in sirius, no additional operator is needed.

you can send end_stream, wait_for_data and data batches to scan operators throught split_connector ?

do you think we need to add this or just a rewiring of split_connector would be adequate ?

aocsa:
> we already have all the capabilities on stream source in sirius, no additional operator is needed.

Agreed — no new capability is invented: the operator is thin glue over the existing `shared_data_repository` ([#1276](https://github.com/sirius-db/sirius/issues/1276): the repository *is* the queue, spillable, no bounded channel). The only thing it adds is what a repository can't say — fan-in end-of-stream across an expected **set** of senders ([#836](https://github.com/sirius-db/sirius/issues/836)).

> you can send end_stream, wait_for_data and data batches to scan operators through split_connector?

For the special case of exactly one sender — probably, yes. But an exchange input is a **fan-in** in general: N remote senders into one stream, where end-of-stream means *every expected sender* has closed. `split_connector` models the opposite shape: one local producer, one `close()` bool, no notion of who is sending.

Also worth clarifying the topology — `STREAMING_SOURCE` does **not** replace scans. In this query plan, for example:

- **Fragment 1 (sender):** `GPU_SCAN(lineitem) → filter → project → STREAMING_SINK` —
  starts with the normal GPU scan; its `split_connector` is untouched by this stack.
- **Fragment 2 (receiver):** `STREAMING_SOURCE → aggregate → RESULT_COLLECTOR` — the source
  appears only at the exchange boundary, a leaf with **no table to scan**.

> do you think we need to add this or just a rewiring of split_connector would be adequate?

We need to add it. Rewiring means giving `split_connector` everything it lacks above. 

Thank you for noticing the existence of split_connector—it led to a real follow-up, now on a single `exec::batch_stream`: a single primitive (repository, stream state, and producer-error plane) shared by the source, the sink (#1321), and the upcoming streamming session. cc @aminaramoon

src/include/exec/batch_stream.hpp
```diff
@@ -0,0 +1,196 @@
...
+#include <functional>
+#include <memory>
+#include <mutex>
+#include <set>
+
+namespace sirius::exec {
+
+/// Identity of one producer feeding a stream. A fan-in stream (N remote senders into one
+/// source) is only at end-of-stream once every expected sender has closed, so closes must
+/// dedup by identity rather than by count.
```
dhruv9vats, line 33:
```suggestion
/// source) signals end-of-stream IFF every expected sender has closed, so closures must
/// be deduped by identity rather than by count.
```

src/include/exec/batch_stream.hpp
```diff
@@ -0,0 +1,196 @@
...
+/// `shared_data_repository`; consumers pull, poll, or block. The repository is the queue; this
+/// owns everything the repository lacks — who is still producing, whether "no batch right now"
+/// means "wait" or "the stream is over", how a starved consumer gets woken, and how a producer
+/// failure reaches the consumer.
+///
+/// The streaming source, the streaming sink, and the session all need this same pairing;
+/// written once here, each of them is left with task-protocol glue.
+///
+/// The repository is borrowed: the caller creates and registers it with the memory manager, so
+/// queued batches stay spillable. Hold the stream by `shared_ptr` — producer threads co-own it.
```
dhruv9vats, line 46:
```suggestion
/// queued batches stay spillable. Holds the stream by `shared_ptr` — producer threads co-own it.
```

src/include/exec/batch_stream.hpp
```diff
@@ -0,0 +1,196 @@
...
+///
+/// Admission, close, and every emptiness check share the one lock, which is what makes the races
+/// safe: no batch is admitted after the stream goes terminal, no queued batch is ever reported as
+/// end-of-stream, and a batch is in the repository before `on_data` announces it. The repository
+/// is touched only under that lock — except `try_pull()`'s pop, so wait-then-pop is not atomic
+/// and a blocking consumer loop must re-check after `wait()`.
```
dhruv9vats, line 52:
```suggestion
/// Admission, closure, and every emptiness check share the same lock, which is what makes this thread-safe
/// and prevents races: no batch is admitted after the stream is terminated, no queued batch is ever reported as
/// end-of-stream, and a batch is in the repository before `on_data` announces it. The repository
/// is touched only under that lock — except `try_pull()`'s pop, so wait-then-pop is not atomic
/// and a blocking consumer loop must re-check after `wait()`.
```

src/include/op/sirius_physical_streaming_source.hpp
```diff
@@ -16,58 +16,104 @@
...
+  /// Wires the two pipeline-facing stream hooks. Both fire on a producer thread, outside the
+  /// stream's lock, and both weak-capture the pipeline rather than `this`:
+  ///
+  /// - end-of-stream → `update_pipeline_status(false)`, so a stream that ends with no task in
+  ///   flight still finishes the pipeline *and* re-arms its downstream consumers;
+  /// - on-data → `task_creator::schedule(head)`, the self-nomination described above. Without it
+  ///   a batch that arrives after the source has been dropped is never looked at.
+  ///
+  /// A stream that has *already* ended fires the first hook inside this call.
   void set_pipeline(duckdb::shared_ptr<pipeline::sirius_pipeline> pipeline) override;
```
dhruv9vats, line 66: Question for clarification: Since we `schedule` the pipeline on every batch push and becasue the task creator also does `while (!node->all_ports_empty()) {`, wouldn't a single call to schedule be enough too? Since the call to `all_ports_empty()` is backed with `batch_stream::drained()` which also accounts for the empty-but-unfinished case by checking `terminal` ? Or maybe I'm missing something else.

dhruv9vats (approved):
Thanks @aocsa, this is fantastic! Really appreciate the explicitly stated + well-encoded invariants and the tests covering them!

Just a question and some nits.

src/include/op/sirius_physical_streaming_source.hpp
aocsa: Good question — the intuition would hold if the `while` loop waited, but it doesn't.

**The loop is a drain pass, not a poll.** `while (!node->all_ports_empty())` only processes
what is already queued. For `STREAMING_SOURCE`, `all_ports_empty()` = `batch_stream::drained()`
= `terminal && repo_empty && no_error`. When the stream is open-but-empty that is `false`, so
the loop enters — and immediately exits via `break` because `get_next_task_input_data()` returned
`nullptr`. It never re-polls, and it cannot block: the creator runs on a 2-slot bounded thread
pool, and `batch_stream::wait()` is explicitly "external threads only — never from a GPU worker."

Take as reference GPU_SCAN to contrast how different is this new operator **Why blocking is safe for `GPU_SCAN` but not here.** `GPU_SCAN` can block inside
`get_next_split()` because its producer — `load_balancing_scan_batch_coalescer` — runs on a
dedicated scan-manager thread pool that is completely independent of `task_creator`. Its
unblocking event never needs a task-creator slot to occur.

`STREAMING_SOURCE` is the opposite: its producer is the sink in another fragment — possibly
running on a **different compute node** in a distributed query. Even in the single-node case
that sink runs as a **GPU pipeline task** that `task_creator` must schedule. If the source
blocked waiting for data it would hold one of the 2 pool slots, preventing the sink's task from
ever being created → **deadlock**. In the multi-node case the producer is entirely off-process,
so blocking would mean occupying a local pool slot for an unbounded network round-trip.
Either way, `GPU_SCAN`'s guarantee — "my producer is local, on a disjoint pool, and always
makes progress independently" — simply does not hold here. So the source returns
`WAITING{nullptr}`, which drops the request with no retry.

**`set_on_data` is the re-nomination edge that closes the gap.** When the sink pushes a batch,
the hook fires `creator->schedule(head)` from the producer thread — outside the pool — and
re-enters the task-creation workflow at step ①, without ever occupying a slot while waiting.
A single upfront `schedule` would miss any batch that arrives after the first `WAITING{nullptr}`
drop and before a completing task re-nominates the source — there is no such completing task,
because the source has none in flight.

Issue #1310, opened 2026-07-28
Simpatico: support INT16/UINT16 element types in the fused codegen


kevkrist:
## Context

Simpatico's fused-codegen operators (`delta`, `rle`, `bitpack`, `for`, `zigzag` — the ops integer plan blocks use) exist only in the JIT path, and the JIT dtype tables (`dtype_to_cxx`, the encode/decode tables) cover `int8/int32/int64` plus unsigned/float reinterpretation — **`int16`/`uint16` are unsupported**. An unsupported element type is a hard `compress_with_plan` failure for the whole batch, and the pin-time `compression_failed` latch then pins the rest of the table uncompressed.

## Why it matters

Numeric width narrowing (#1260) narrows pinned numeric columns to the narrowest carrier their values fit, which naturally produces INT16/UINT16 columns. To avoid one narrowed column silently disabling compression for its whole table, #1260 floors 16-bit narrowing targets to INT32/UINT32 on pins that will compress. That floor costs decompressed-side width (working set, exchange, joins) on 16-bit-range columns and exists only because of this dtype gap.

## Ask

1. Extend the codegen dtype tables and kernel generation with `int16_t`/`uint16_t` (encode + decode, same reinterpretation treatment as the other widths).
2. Optionally: expose a per-block "encodable with dtype" predicate so callers can query support precisely instead of maintaining conservative floors.

## Acceptance

- INT16/UINT16 columns compress and round-trip through integer plan blocks (unit + round-trip tests).
- The carrier floor in Sirius's `narrow_pin_chunk` (#1260 integration) can be deleted outright.

Refs: #937, #1260.

joosthooz:
Closed by #1325

PR #1325, opened 2026-07-29
Add support for compressing 16-bit datatypes

### LINKED ISSUE #1326: 16-bit compression support

joosthooz:
Simpatico's fused codegen backend does not have support for 16-bit datatypes, this PR adds it (and tests).

Closes #1326

### DIFF
--- a/src/compression/simpatico_codegen/tests/test_fused_operator_sweep.cpp
+++ b/src/compression/simpatico_codegen/tests/test_fused_operator_sweep.cpp
@@ -37,12 +37,21 @@
 // shapes, plus boundary cases for leaf/passthrough forms and RLE nesting in
 // the runs branch. Set SIMPATICO_FUSED_SWEEP_DEPTH to override (default 4).
 //
-// Fixture
-// -------
-// int32_t throughout (matches the existing JIT-roundtrip tests).
-// Two synthetic columns: `synth_data` (generic, no-RLE shapes), and
-// `synth_rle_data` (chunk-varied RLE patterns).  Shapes containing
-// any Rle node use the RLE-friendly fixture.
+// Dtypes
+// ------
+// All four integer widths supported by the fused JIT path: int8_t, int16_t,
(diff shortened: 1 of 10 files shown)


kevkrist (approved):

kevkrist:
Also closes #1310

src/compression/simpatico_codegen/tests/test_fused_operator_sweep.cpp
```diff
@@ -311,51 +332,222 @@ int env_depth(int default_depth)
...
+
+    ++total_per_dtype[di];
+    if (runners[di](*t, s_tag, use_rle) == 0) {
+      std::printf("  %-60s OK\n", s_tag.c_str());
+      ++passed_per_dtype[di];
+    }
+  }
+
+  // Report results over the pipe.
+  std::string report;
```
felipeblazing, line 405: It feels like we should just be using our testing framework instead of writing out these reports to strings

felipeblazing: We can fix this in a follow up pr since I know this is how it was before

felipeblazing (approved):
This looks good but I think we need to make an issue to have these tests use our testing framework instead of outputting reports to make it easy to integrate with CI

src/compression/simpatico_codegen/tests/test_fused_operator_sweep.cpp
joosthooz: I made an issue for that here https://github.com/sirius-db/sirius/issues/1338.
It is a little more broad than just moving them over and not use these string reports, as is noted in that issue. The code you flag here is somewhat complex because it parallelizes JIT compilations across several processes (not threads, because JIT compiling appears to be locked to 1 per process). It parallelizes across a couple of levels (datatype fixture, and then 1 or 2 chain levels deep), to allow several dozen of them to run concurrently. That helps a lot in bringing the testing time down to a level where we can exhaustively test a good chain depth (like 4 or 5).

PR #1327, opened 2026-07-29
[Concurrency] Per query data repository manager registry, plus query_id and operator_id fixes

wmalpica:
This PR addresses 3 main problems in the context of concurrent query execution:

Problem 1. Sirius had a single shared_data_repository_manager owned by SiriusContext, shared by every in-flight query, and repositories were keyed by {operator_id, port_id}. At query end, the data_repository_manager would get cleared, which causes problems with concurrent query execution.
Solution: We how have a data_repository_manager_registry still owned by the SiriusContext, so that we have one shared_data_repository_manager per query. 

Problem 2. Operator ids were assigned at construction from a static inline counter, which was then reset for the next query. This means that concurrent query operator graphs would be using and incrementing the same counter.
Solution: Now we dont have a static inline counter. Instead, after all the operators are created. We iterate over them and assign them an id.

Problem 3. Query_ids were defined differently in different parts of the engine. This PR unifies some of them. NOTE: the Quent telemetry uses its own query id. If necessary this can be addressed later.
Solution: We now have a strongly typed unified query_id.

What changed

1. query_id_t, one query identity (src/include/query_id.hpp)

An enum class query_id_t : uint32_t minted once per execution window (SiriusContext::StandaloneQueryScope) and threaded through everything downstream: which repository manager the query owns, which repositories its cleanup drops, its scheduling priority, and its log/telemetry correlation key. planner::query no longer mints its own; it receives the window's id.

A distinct enum type rather than a uint64_t alias, precisely because operator ids, pipeline ids, connection ids and per-connection query ordinals are all plain integers here; making them silently interchangeable is what let the duplicate counters coexist. query_priority_bits() centralizes the priority-packing contract (masked to 31 bits so the id cannot shift into the sign bit of the signed queue_priority and invert scheduling order). A std::formatter specialization keeps log sites cast-free.

2. data_repository_manager_registry (src/include/data/data_repository_manager_registry.hpp)

One shared_data_repository_manager per in-flight query, keyed by query_id_t. Query end drops only that query's repositories. Managers are handed out as shared_ptr because a downgrade worker may be sweeping a manager while its query ends, and shared ownership keeps the manager object alive until the last borrower releases it.

get_all() returns a snapshot built under the lock, so the downgrade executors can iterate without holding the mutex across a long spill sweep (that would serialize query begin/end behind spilling). std::map rather than unordered_map so spill-candidate ordering is deterministic across runs. create_for_query throws on a duplicate id: window ids are monotonic, so a duplicate is a lifecycle bug, not recoverable state.

The downgrade executors now bind to the registry rather than a single manager, since memory pressure is a global condition and spill candidates must be drawn across all live queries.

3. Operator ids are per-query and stamped in one pass

next_operator_id is gone. Operators are constructed with an invalid_operator_id sentinel; pipeline::assign_operator_ids numbers the final pipeline set 0..N-1 in scheduling order (source, then operators, then sink; first visit wins, so an operator appearing in several pipelines keeps one id). It runs inside sirius_pipeline_converter::convert(), the one point every caller shares, including plan-inspection paths that never build an engine, so no consumer can observe an unnumbered plan.

The pipeline set, not the plan tree, is the traversal root on purpose: sirius_physical_delim_join owns its join and distinct_root subtrees outside children and does not override get_children(), so a tree walk would miss them. Every operator that executes belongs to a pipeline.

get_operator_id() now throws on the sentinel rather than returning it. An unassigned id would otherwise silently collide in the operator-id-keyed maps (repositories, task-creator global states) and mis-route data, so it fails loudly in every build, not just debug.

Tradeoffs and known limits

- Scheduling order wraps every 2^31 queries in one process, at which point a fresh query sorts ahead of older in-flight ones. This is inherent to packing the id into the priority at all, and is documented at query_priority_bits().
- registry::erase requires quiesced borrowers. Shared ownership protects the manager object, not the repositories inside it; the cleanup path drains the downgrade executors first. This precondition is documented, not enforced.
- The first window gets id 1, so 0 is never a live query id.

Tests

- test/cpp/data/test_data_repository_manager_registry.cpp: registration, per-query isolation, get_all() ordering, duplicate-id rejection, erase reporting.
- test/cpp/planner/test_query_id.cpp: id propagation and priority packing.
- New sirius::test::scoped_test_query helper. Tests that build a sirius_engine directly never open a real execution window, so nothing would create the manager the engine needs or clean it up. Leaving repositories behind would break the next real GPU query in the process, since operator ids now restart at 0 per plan and would collide on {operator_id, port_id}.
- Existing downgrade, pipeline, and planner tests updated for the new signatures.

### DIFF
--- a/src/downgrade/downgrade_executor.cpp
+++ b/src/downgrade/downgrade_executor.cpp
@@ -41,13 +41,13 @@ static std::string tier_to_string(cucascade::memory::Tier tier)
 
 downgrade_executor::downgrade_executor(
   exec::downgrade_executor_config config,
-  cucascade::shared_data_repository_manager& data_repo_mgr,
+  sirius::data::data_repository_manager_registry& data_repo_registry,
   cucascade::memory::memory_space_id space_id,
   cucascade::memory::memory_space* memory_space,
   sirius::memory::sirius_memory_reservation_manager& reservation_manager,
   sirius::exec::multi_index_priority_queue<sirius::parallel::itask>* pipeline_task_queue)
   : _config(std::move(config)),
-    _data_repo_mgr(data_repo_mgr),
+    _data_repo_registry(data_repo_registry),
     _space_id(space_id),
     _memory_space(memory_space),
     _source_label(tier_to_string(space_id.tier) + ":" + std::to_string(space_id.device_id)),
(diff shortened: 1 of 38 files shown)


src/downgrade/downgrade_executor.cpp
```diff
@@ -200,74 +201,93 @@ void downgrade_executor::processing_loop()
...
-    auto repos            = _data_repo_mgr.get_repositories();
-    for (auto* repo : repos) {
-      if (req->satisfied.load()) break;
-
-      convertible_data_batch_provider provider(repo);
-      auto candidates = provider.get_all_convertible(
-        source_space, /*front_to_back=*/false, /*ignore_subscribed=*/true);
-      for (auto& candidate : candidates) {
+    auto const managers   = _data_repo_registry.get_all();
+    for (auto const& manager : std::views::reverse(managers)) {
```
aminaramoon, line 224: Is it acceptable than downgrade request of a query would spill the data from a higher priority query to disk or out of gpu

src/include/data/data_repository_manager_registry.hpp
```diff
@@ -0,0 +1,169 @@
...
+  [[nodiscard]] size_t size() const
+  {
+    std::lock_guard<std::mutex> lock(_mutex);
+    return _managers.size();
+  }
+
+ private:
+  mutable std::mutex _mutex;
+  /// std::map (not unordered_map) so get_all() iteration is ascending by query id.
+  std::map<sirius::query_id_t, manager_ptr> _managers;
```
aminaramoon, line 166: Unordered map is better here

aminaramoon (approved):
Two minor comments

src/include/data/data_repository_manager_registry.hpp
wmalpica: Why do you say that? 
Like the comment says: std::map is used on purpose so that `get_all()` returns a list that is ordered by query id, and therefore is the FIFO list, which we can use to apply an implicit FIFO priority

src/downgrade/downgrade_executor.cpp
wmalpica: Here, we are iterating from last to first query. And we are currently assuming an implicit FIFO priority. So we would always be downgrading from last in query to first in query, meaning we are downgrading from the least priority query first. 

The alternative, which your comment may be alluding to, is that a lower priority query can never downgrade form a higher priority query. We could implement that. But I dont think we should, or at least not without other changes. Because if we are getting a lower priority query downgrade request, it means that the task executor is going to be executing a task from that lower priority query. That is already at the front of the queue. Not fullfilling that memory request via downgrading will just make it more likely to OOM, not necessarily "protect" the higher priority query

aminaramoon: the chances are you will be scheduling scan tasks for later queries and increasing memory pressure while your earlier queries are waiting for a hash table to be built and have no task to execute.

bwyogatama (approved):

PR #1334, opened 2026-07-29
Fix Simpatico multi-GPU stream affinity

### LINKED ISSUE #1330: Simpatico pin-table compression: `cudaErrorIllegalAddress` on multi-GPU decode (Q1 / TPC-H SF300)
# Simpatico pin-table compression: `cudaErrorIllegalAddress` on multi-GPU decode (Q1 / TPC-H SF300)

## Summary

With `sirius.compression.enable_pin_table_compression: true` and `num_gpus: 4`, TPC-H Q1 faults during Simpatico decode (`cudaErrorIllegalAddress` / cudaError 700). Sirius falls back to DuckDB for that iteration, then **hangs forever** on the next iteration because the CUDA context is poisoned.

**1-GPU is clean** under the same compression settings. Narrowing / compressed materialization is **not involved** (`enable_compressed_materialization: false`).

**Pre-existing on `dev`:** reproduced at merge-base `50352d1b` ("Add Simpatico as the compression subsystem #937") with a fresh build. Not introduced by the compressed-materialization branch.

Not related to the prefetch-cache race (`enable_prefetch_cache` was unset / default false).

## Environment

- Machine: `presto-gb200-gcn-17`
- 4× NVIDIA GB200, driver 580.105.08, CUDA 13.0
- Dataset: TPC-H SF300 parquet at `/tmp/sirius_tpch_parquet_sf300` (**must be absolute**)
- Plans: `<repo>/src/compression/simpatico_codegen/plans/tpch_sf1000`

## Reproduction

Minimal config (host-pinned, 4 GPU, compression on, narrowing off):

```yaml
sirius:
  topology:
    num_gpus: 4
  # ... rest of normal memory/executor settings for this machine ...
  operator_params:
    enable_compressed_materialization: false   # omit on commits before this key exists
  compression:
    enable_pin_table_compression: true
    input_plan_dir: <ABS>/src/compression/simpatico_codegen/plans/tpch_sf1000
```

```bash
export CUDA_DEVICE_ORDER=PCI_BUS_ID
export CUDA_VISIBLE_DEVICES=0,1,2,3
export SIRIUS_PIN_TIER=host

pixi run python test/tpch_performance/performance_test.py \
  --input /tmp/sirius_tpch_parquet_sf300 \
  --data-source parquet --mode grouped --iterations 2 \
  --engine gpu --pin host \
  --config <path-to-config-above> \
  --output /tmp/simpatico_repro --name host4_comp_q1 \
  --queries 1
```

Faults within ~1 minute of query start. Use `--iterations 1` to observe the fault without the hang.

## Observed failure

Pin succeeds:

```
[pin_table] 'lineitem' tier=host: compressing with plan for 7 column(s)
```

Then on q1 iter0 (merge-base `50352d1b` wording; tip may print a slightly different first line):

```
simpatico::codegen: launch_decode_fused_tree: transient synth failed: compute_bp_offsets failed (cudaError=700)
Error in Sirius GPU execution, fallback to DuckDB
```

Debug-level logs also show decode-path messages such as `nvcomp batched: decompress_async failed`, `codegen decompress: decode failed`, and illegal accesses on `D2H frame header` / `D2H comp sizes`, plus `OOM preparing batches for processing`. Subsequent `cudaErrorIllegalAddress` sites (`device_buffer.cpp`, `event.cpp`, `cuda_stream*.cpp`) are sticky-context fallout.

Pipeline at fault (info log): `GPU_SCAN -> PROJECTION -> PROJECTION -> HASH_GROUP_BY`.

## Scope checked

| Config | Result |
|---|---|
| 1 GPU + compression (`comp` / `both`), host or GPU pin | clean (full TPC-H 22 queries × 2 rounds) |
| 4 GPU + compression off / narrowing only | clean |
| 4 GPU + compression on, tip `1333d403` | **fault** |
| 4 GPU + compression on, merge-base `50352d1b` | **fault** (same) |

## Note on tooling

`compute-sanitizer --tool memcheck` is **not useful** for this on GB200 / CUDA 13: it dies during pin materialization with an unrelated CUB shared-memory OOB / `cuMemBatchDecompressAsync` invalid-value even with compression **disabled**. Don't burn time there.

## Ask

- Fix multi-GPU Simpatico decode (or gate pin-table compression to 1 GPU until fixed).
- Separately: fail fast after a poisoned CUDA context instead of hanging on retry.

kevkrist:
NOTE: this is admittedly a temporary fix, until Simpatico uses the Sirius stream pool.
Fixes #1330 

## What changed

- key Simpatico's process-lifetime internal stream-cache free lists by CUDA device
- capture the lease device once and use it consistently for checkout and return
- add a focused multi-GPU regression that uses serial compression and threaded decode in device order `0 -> 1 -> 0`
- register the regression as a standalone target and CTest

## Root cause

The `int column_threads` overloads leased worker streams from one process-wide free list. CUDA streams are device-bound, so after a threaded decode on one GPU, a later decode on another GPU could receive foreign stream handles. Kernel launches then fail with `invalid resource handle`, and surrounding asynchronous CUDA work can surface `cudaErrorIllegalAddress`.

`pin_table` compression itself uses a caller-owned stream pool; the affected cache is exercised by the subsequent per-device threaded decode path.

## Impact

Multi-GPU scans of compressed pinned tables keep internal decode streams on their owning CUDA device, preventing the observed GPU fallback and poisoned-context hang on the following iteration.

### DIFF
--- a/src/compression/simpatico_codegen/src/simpatico_codegen.cpp
+++ b/src/compression/simpatico_codegen/src/simpatico_codegen.cpp
@@ -15,6 +15,7 @@
 #include <nvtx3/nvtx3.hpp>
 
 #include <atomic>
+#include <map>
 #include <mutex>
 #include <span>
 #include <stdexcept>
@@ -108,17 +109,29 @@ void validate_column_names(std::vector<std::string> const& column_names, size_t 
 // result is safe to free by any stream (including the RMM default) with no
 // external rebinding. Streams are recycled between calls, so this also avoids
 // per-call stream create/destroy churn.
+//
+// The free lists are keyed by device. A CUDA stream belongs to the device that
+// was current when it was created, and driving work on a stream from another
+// device is invalid — kernel launches fail with "invalid resource handle" and
+// the async copies/codec calls around them can fault the context with
+// cudaErrorIllegalAddress. In the affected Sirius path, multi-GPU pin_table
+// compression uses a caller-owned pool; this cache is exercised by the
+// subsequent per-device threaded decode. Without the device key, one device's
+// completed decode can recycle its streams into the next device's decode.
 class stream_cache {
  public:
-  // Hand out n valid streams, reusing recycled ones and creating the rest.
-  std::vector<cudaStream_t> checkout(size_t n)
+  // Hand out n valid streams belonging to `device`, reusing ones recycled from
+  // that device and creating the rest. The caller must have `device` current:
+  // cudaStreamCreateWithFlags binds the new streams to the current device.
+  std::vector<cudaStream_t> checkout(int device, size_t n)
   {
     std::vector<cudaStream_t> out;
     out.reserve(n);
     std::lock_guard<std::mutex> lock(mu_);
-    while (out.size() < n && !free_.empty()) {
-      out.push_back(free_.back());
-      free_.pop_back();
+    auto& free_list = free_[device];
+    while (out.size() < n && !free_list.empty()) {
+      out.push_back(free_list.back());
+      free_list.pop_back();
     }
     while (out.size() < n) {
       cudaStream_t s{};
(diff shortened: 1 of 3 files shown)


src/compression/simpatico_codegen/src/simpatico_codegen.cpp
```diff
@@ -153,19 +170,30 @@ stream_cache& global_stream_cache()
...
+// synchronous lease lifetime, and run_column_workers binds each worker thread
+// to that device. Capture it once so checkout and return use the same free list.
 struct leased_pool {
   stream_pool pool;
+  int device = 0;
 
   explicit leased_pool(int column_threads)
   {
-    pool.streams = global_stream_cache().checkout(static_cast<size_t>(std::max(1, column_threads)));
+    if (cudaGetDevice(&device) != cudaSuccess)
```
felipeblazing, line 167: shouldnt we be calling cudaSetDevice here to ensure we are creating the stream below on the proper device?

felipeblazing (changes requested):

src/compression/simpatico_codegen/src/simpatico_codegen.cpp
kevkrist: My understanding is: the Sirius compute thread already has its CUDA device set when it calls Simpatico, and `checkout()` and `cudaStreamCreateWithFlags()` run on that thread, so newly created streams belong that device. Simpatico captures that device id, launches its own column-worker threads which each call `cudaSetDevice()` with the captured device id. The new cache key just ensures those workers only receive recycled streams that were created for that same device, and only if none are available, new device-local streams are created.

felipeblazing (approved):
Remove some of the more verbose unnecessary comments but otherwise looks good

PR #1335, opened 2026-07-30
[Concurrency] Make task_creator and task_executor more concurrency friendly

wmalpica:


src/sirius_context.cpp
```diff
@@ -322,15 +338,18 @@ void SiriusContext::run_mandatory_cleanup(std::string_view end_tag)
   } catch (...) {
   }
 
-  // Clear all data repositories between queries.
+  // Drop THIS query's data repositories, leaving any other in-flight query's untouched.
   // Any batches still present are leaked — operators should have popped everything.
-  if (data_repository_manager_) {
-    auto leaked = data_repository_manager_->clear_all_repositories();
+  // Safe to clear here because the downgrade executors were drained above, so nothing still
+  // holds a raw data_repository* borrowed from this query's manager.
```
ran-yuan-rui, line 343: `drain()` does not make this `erase()` safe. It restarts processing before returning (`downgrade_executor.cpp:128-141`) and does not stop the monitor.

```text
time    Cleanup              Executor / monitor            Sweep / repositories
  |
  |     drain() ----------> stop current work
  |                          reactivate queue + restart processing
  |     <------------------ return
  |                          enqueue under memory pressure
  |                          start sweep -----------------> snapshot repo*
  |     erase() ------------------------------------------> destroy repository
  |                                                       continue via repo*
  v                                                       dangling pointer
```

The `shared_ptr` keeps the manager alive, not the repositories returned as raw pointers:

```text
Sweep -- shared_ptr --> Manager                         (still alive)
Sweep -- raw repo* ---> Repository                      (not owned)
erase() -------------> Manager::clear_all_repositories()
                                      |
                                      +--> destroys Repository
```

Keep all downgrade producers and processors quiesced through `erase()`, or use sweep leases that `erase()` waits out before clearing the manager.

src/creator/task_creator.cpp
```diff
@@ -352,193 +459,223 @@ void task_creator::manager_loop()
...
+      continue;
+    }
+
     node = get_operator_for_next_task(node);
 
     if (node == nullptr) { continue; }
 
+    // Counted before dispatch so drain_pending_tasks(query_id) cannot observe zero in-flight
+    // while this task creation is still queued to run.
+    query_state->enter_in_flight();
```
ran-yuan-rui, line 499: `enter_in_flight()` is too late. The request leaves the queue at `:456`, and `node` is dereferenced at `:478`. Cleanup can therefore observe an empty queue and `in_flight == 0`, return from `reset(query_id)`, and destroy the plan before this increment.

The fix needs an atomic worker lease, not just an earlier increment:

```text
Worker                         Cleanup
try_acquire_lease(query)       close(query)
  lookup state                   reject new leases/requests
  require !closing               drain queued requests
  ++in_flight                    wait for leases == 0
dereference node               destroy plan
release lease
```

State lookup, the closing check, and the increment must be one operation. A request that cannot acquire the lease must be dropped without touching `node`.

src/sirius_context.cpp
```diff
@@ -300,6 +304,18 @@ void SiriusContext::run_mandatory_cleanup(std::string_view end_tag)
...
+  // pointers into the plan that query_.reset() below destroys, and in-flight creation lambdas
+  // dereference them. reset() drains those requests and joins that work before returning.
+  // Only this query's is touched; other in-flight queries keep creating tasks.
+  if (task_creator_) { task_creator_->reset(query_id); }
+
+  // With the producer stopped, drop whatever it already queued for this query. Ordering is
+  // deliberate: task_creator first so nothing new arrives, then the queues, then query_.reset()
+  // below — queued tasks reach operators through their pipeline, so they must be gone before the
+  // plan is destroyed. Other in-flight queries keep their queued work.
+  if (task_scheduler_) { task_scheduler_->drain_query_tasks(query_id); }
```
ran-yuan-rui, line 318: `drain_query_tasks(query_id)` misses tasks temporarily owned by Tier-2 downgrade. The provider removes a task from the scheduler queue; cleanup sees an empty queue and resets the query; the conversion wrapper can then return that task during the later downgrade drain.

```text
Scheduler queue       Tier-2 conversion        Cleanup
[task A]
    +---------------> take task A
[empty]                                        drain query -> empty
                                               query_.reset()
                      finish conversion
[task A] <---------- requeue
                                               erase repositories
```

Reordering the drains or adding another drain is still racy because `downgrade_executor::drain()` reactivates the queue and restarts processing before returning. Mark the query closing, reject new extractions, wait for its conversion/requeue leases, then drain the scheduler queue and destroy the plan. A wrapper must not requeue work for a closing query.

wmalpica:
This PR has been broken down into 3 stacked PRs: 
https://github.com/sirius-db/sirius/pull/1364
https://github.com/sirius-db/sirius/pull/1365
https://github.com/sirius-db/sirius/pull/1366

PR #1336, opened 2026-07-30
fix(join): null-safe mixed-key joins via cuDF mixed_join + NULL_EQUAL

### LINKED ISSUE #1293: mixed '=' + IS NOT DISTINCT FROM join returns wrong result when the null-safe key has NULLs (no fallback)
## Summary

A join whose keys **mix** plain equality with null-safe equality —
`... ON a = b AND c IS NOT DISTINCT FROM d` — silently returns a **wrong (under-counted) result** on the GPU when the null-safe column (`c`/`d`) actually contains NULLs that should match. The query runs on the GPU with **no fallback**.

This is a known, narrowed limitation left by the null-safe join fix (#1291): pure `IS NOT DISTINCT FROM` joins are now correct, but mixed joins are forced to `UNEQUAL` and are neither corrected nor caught.

## Reproduction

```sql
CREATE TABLE l (a INTEGER, b INTEGER);
CREATE TABLE r (a INTEGER, b INTEGER);
INSERT INTO l VALUES (1, NULL), (1, 10);
INSERT INTO r VALUES (1, NULL), (1, 20);

SELECT count(*)
FROM l JOIN r
  ON l.a = r.a AND l.b IS NOT DISTINCT FROM r.b;
-- CPU (correct): 1   -- l(1,NULL) ↔ r(1,NULL) matches null-safe
-- GPU:           0   -- forced UNEQUAL drops the NULL-to-NULL match
```

## Root cause

cuDF's hash-join APIs take a **single `null_equality` flag for all key columns**, so per-column semantics (null-safe on one key, plain on another) can't be expressed in one call. `sirius_physical_hash_join::compare_nulls()` therefore uses `EQUAL` only when *every* equi-key is `IS NOT DISTINCT FROM`; any plain `=` key forces `UNEQUAL` — which is wrong for the null-safe column(s) of a mixed join.

## Why we can't just reject mixed joins

An earlier attempt rejected mixed joins in `are_conditions_supported` to force a CPU fallback, but **delim joins** (for correlated subqueries) are legitimately mixed — DuckDB matches real keys with `=` and delim columns with `IS NOT DISTINCT FROM`. Rejecting them routed the join to the unsupported blockwise nested-loop path and threw (`plan tree shape - delim join ...`). So a blanket rejection isn't viable.

## Proposed fix

Distinguish **delim** joins from **user-authored** mixed joins and fall back to CPU only for the latter:

- A delim join carries non-empty `delim_types` (set by `sirius_plan_delim_join`); a user-written mixed join does not.
- For a user-authored join that mixes `=` with `IS NOT DISTINCT FROM`, have `are_conditions_supported` return `false` (→ CPU fallback, correct result) instead of running on the GPU with `UNEQUAL`.
- Delim joins keep running on the GPU (unchanged, pre-existing behavior).

Alternative (heavier): implement per-column null equality on the GPU via post-filtering (match with `EQUAL`, then drop rows where a plain-`=` key was NULL-matched-NULL). More work; only worth it if mixed null-safe joins prove common.

## Scope / severity

Silent wrong result, but narrow: requires an explicit mixed `=` + `IS NOT DISTINCT FROM` join **and** actual NULL-to-NULL matches on the null-safe key. Currently documented only in a code comment (`compare_nulls()` in `sirius_physical_hash_join.cpp`) and deliberately left untested (`MixedKeyJoinFixture` uses a NULL-free column). This issue tracks closing it.

Follow-up to #1291.

Jedi18:
## Summary

Fixes a silent GPU-vs-CPU result divergence for joins combining a plain equality key with a null-safe key:

```sql
... JOIN ... ON a.x = b.x AND a.y IS NOT DISTINCT FROM b.y
```

When the null-safe column contained NULLs on both sides, GPU execution dropped valid NULL-to-NULL matches without falling back or reporting an error.

## Root cause

cuDF applies one `null_equality` policy to every hash key. Mixed-key joins require two different policies:

- `a.x = b.x` requires `UNEQUAL` (`NULL` must not match `NULL`).
- `a.y IS NOT DISTINCT FROM b.y` requires null-safe equality (`NULL` must match `NULL`).

The earlier pure null-safe join fix (#1292) can select `EQUAL` when every hash key is null-safe. A mixed join still has a plain `=` key, so the shared policy remains `UNEQUAL` and incorrectly removes NULL-to-NULL matches from the null-safe key.

## Implementation

The plain `=` key(s) remain `UNEQUAL` hash keys. Null-safe key(s) are moved into the conditional predicate of a cuDF `mixed_*_join` and represented as AST `NULL_EQUAL` operations.

This selects Sirius's `MIXED_JOIN` mode and reuses the existing mixed-join dispatch for regular INNER, LEFT, RIGHT, FULL OUTER, SEMI, and ANTI joins.

- **`gpu_expression_translator`**: `add_join_condition()` now translates `not_distinct_from` to `NULL_EQUAL`. This is the specialized path used by `translate_join_conditions()` for mixed-join predicates; previously it returned `nullopt`, causing execution to throw.
- **`sirius_physical_hash_join`**: `has_mixed_null_safe_keys()` and `is_hash_equality_key()` distinguish hash keys from routed conditional keys. This classification is used consistently by support validation, condition reordering, and constructor key extraction/counting.
- **Null policy**: `compare_nulls()` remains `UNEQUAL` for the plain hash keys. Null-safe behavior is isolated in the `NULL_EQUAL` conditional predicate.

### Implicit casts on routed keys

Routing changes which code path applies an implicit cast on the null-safe key, and the two paths do not have the same capabilities:

- A **hash key** carries its cast as `key_cast_info` and is materialized at execute time by `prepare_join_keys()` using `cudf::cast`, which supports any target type.
- A **routed conditional key** is lowered into a cuDF AST, which only provides `CAST_TO_INT64`, `CAST_TO_UINT64`, and `CAST_TO_FLOAT64`.

DuckDB binds both sides of a comparison to a common type, so `l.b IS NOT DISTINCT FROM r.b` over `SMALLINT`/`INTEGER` columns arrives carrying a cast to `INTEGER`. Routed unchanged, that cast fails translation and the join throws at execute time — aborting the GPU pipeline mid-query and replaying the whole query on CPU, or failing outright when `enable_duckdb_fallback` is off or the query reads from S3. This affected such joins even with no NULLs present, where the pre-routing hash join was correct.

Two changes make routing safe:

- **`sirius_plan_comparison_join`**: `materialize_expression_join_keys()` now materializes a cast on a key that will be routed, when the cast target is outside `supported_ast_cast_types`. It reuses the existing projection-pushdown machinery for complex join keys, so the key reaches the predicate as a plain column reference and the join keeps the `MIXED_JOIN` fast path instead of degrading to a nested-loop join. The routing condition is mirrored by `routes_null_safe_keys_to_predicate()`. Casts the AST can express (`BIGINT`, `UBIGINT`, `DOUBLE`) remain inline as before.
- **`sirius_physical_hash_join`**: `are_conditions_supported()` previously validated only column disjointness. It now also rejects a mixed null-safe join whose routed key is not AST-translatable, covering key expressions that materialization cannot rewrite. It takes `join_type` so this check mirrors the constructor's routing gate exactly. The constructor throws at plan time — where the exception becomes a clean CPU fallback — rather than silently demoting the key back to an `UNEQUAL` hash key, which would reintroduce the undercount this PR fixes.

The cast allowlist is the existing `supported_ast_cast_types` / `supported_ast_cast_types_native` shared with the expression executor, so the planner, the hash join, and the translator cannot drift apart.

## Scope and limitation

Queries planned as **MARK joins** are intentionally not routed through this path. MARK joins require `BUILD_PROBE` mode to produce three-valued marks, while this fix requires `MIXED_JOIN`; the modes are incompatible.

Because MARK joins never route, their null-safe keys still resolve through `prepare_join_keys()`/`cudf::cast` and are unaffected by the AST cast restriction — hence the `join_type` gate on the new support check, which would otherwise demote a working MARK hash join to a nested-loop join.

Existing MARK null-safe behavior is therefore unchanged and remains a known limitation, tracked for a delim-aware CPU fallback.

## Validation

Added `MixedKeyNullSafeJoinFixture` to verify both sides of the mixed null semantics:

- NULL-to-NULL values in the null-safe key must match.
- NULL-to-NULL values in the plain `=` key must not match.
- INNER
- LEFT
- FULL OUTER
- RIGHT
- SEMI
- ANTI

`BUILD_SIDE_PROBE_SIDE` is disabled where necessary so DuckDB preserves the intended right-family or left SEMI/ANTI execution path.

Added `MixedTypeNullSafeJoinFixture` (`SMALLINT` vs `INTEGER` / `SMALLINT` vs `BIGINT` key columns) to cover the cast handling:

- Untranslatable cast (`INTEGER`) with NULLs on the null-safe key.
- The same join with NULLs filtered out — the null-semantics-independent case that regressed purely on the cast.
- Translatable cast (`BIGINT`), which stays inline in the predicate rather than being materialized.
- `SELECT *` over INNER and LEFT, pinning join output width and column order so a materialized key column cannot leak into the result.

`compare_gpu_vs_cpu` requires exactly one GPU execution and zero fallbacks, so these fail on an execute-time throw as well as on incorrect rows.

The existing null-free mixed-key regression test remains in place and verifies that supported mixed joins stay on GPU rather than entering the unsupported nested-loop path.

Fixes https://github.com/sirius-db/sirius/issues/1293 except for the mark join case.


Jedi18:
I have created an issue in cudf for a feature request for passing in independent null equality for each key column https://github.com/rapidsai/cudf/issues/23481

aocsa (approved):
LGTM

PR #1337, opened 2026-07-30
fix(config): defaults — memory fractions, thread pools, host capacity_fraction

bwyogatama:
## What

Defaults overhaul. Final values:

**Topology** (`6f71a3d`)

| Key | Old | New |
|---|---|---|
| `topology.num_gpus` | 1 | all visible GPUs (auto) |

Topology discovery (NVML, `CUDA_VISIBLE_DEVICES`-aware) already enumerates the machine; the default now uses everything it finds instead of one GPU. Explicit `num_gpus` / `gpu_ids` behave as before; discovery failure falls back to 1. **Note:** CI's main `sirius_unittest` step runs without `SIRIUS_CONFIG_FILE`, so on the 2×L4 runners the default suite now exercises 2-GPU execution — this PR's CI run is the first look at that.

**Memory** (`04f1a03`, `8b4480c`, `572edc5`)

| Key | Old | New |
|---|---|---|
| `gpu.reservation_limit_fraction` | 0.9 | 1.0 |
| `gpu.downgrade_trigger_fraction` | 1.0 | 0.8 |
| `gpu.downgrade_stop_fraction` | 0.7 | 0.6 |
| `host.capacity_fraction` *(new)* | — (8 GiB/node via `capacity_bytes`) | 0.9 of each NUMA node's RAM |
| `host.reservation_limit_fraction` | 0.9 | 1.0 |
| `host.downgrade_trigger_fraction` | 0.8 | 0.9 |
| `host.downgrade_stop_fraction` | 0.7 | 0.8 |

- The old GPU ordering `stop < reservation_limit < trigger` denied reservations at 90% before eviction could fire at 100% — GPU→host downgrade was unreachable on defaults. Both tiers now follow `stop < trigger < reservation_limit` (GPU 0.6 < 0.8 < 1.0, host 0.8 < 0.9 < 1.0), which is what the docs example already showed for the GPU.
- `capacity_fraction` (validated (0,1]) sizes each host space as a fraction of its backing NUMA node's RAM, via cuCascade's new per-NUMA memory discovery; `capacity_bytes` keeps the absolute per-node form and wins if both are set. Hosts exposing no NUMA meminfo fail init with a clear error — set `capacity_bytes` there.
- Host→disk eviction still requires `downgrade_root_dirs`; without one the downgrade executor logs a warning and skips.

**Thread pools** (`8a06e28`)

| Pool | Old | New |
|---|---|---|
| `downgrade` | 4 | 1 |
| `task_creator` | 2 | 1 |
| `pipeline` | 4 | 4 |
| `uring_n_reactors` | 1 | 1 |
| `scan_manager` | 8 | remaining cores, floor 4 |

`scan_manager` defaults to `max(4, hardware_concurrency − reserved)` where `reserved` is the compile-time sum of the other pools' named default constants.

**Operator params** (`e2f11ae`, `c6bd813`)

| Param | Old | New |
|---|---|---|
| `scan_task_batch_size` | 512 MiB | 2.5% of GPU mem, clamped [512 MiB, 5 GiB] |
| `hash_partition_bytes` | 512 MiB | same shared batch default |
| `concat_batch_bytes` | 512 MiB | same shared batch default |
| `sort_sample_bytes` | 512 MiB | same shared batch default |
| `max_build_hash_table_bytes` | 500 MiB | 2× the batch default |

The shared batch default is `config::derived_default_batch_size()`: 2.5% of the smallest visible GPU's total memory (one memoized CUDA query, `CUDA_VISIBLE_DEVICES`-aware), clamped to [512 MiB, 5 GiB], 800 MiB when no GPU is visible. `operator_params` derives its members in its initializers, so the engine config and the SET-variable registration defaults agree by construction (verified via `current_setting()`: 3.0 GiB batch / 6.0 GiB build cap on a GB10, i.e. exactly total/40 and 2×). `max_broadcast_join_size` stays 256 MiB. Examples: L4 24 GB → 600 MiB; A100-80G → 2 GiB; ceiling from 200 GB.

## Dependency

⚠️ `8b4480c` pins the `cucascade` submodule to the **unmerged** NVIDIA/cuCascade#173 head (`9a26463`, branch `numa_capacity` in the main cucascade repo, so it is fetchable). This PR should not merge until cuCascade#173 lands; the submodule then gets re-pinned to the merge commit.


src/include/sirius_config.hpp
```diff
@@ -31,12 +31,16 @@ namespace sirius {
...
 
-constexpr uint64_t DEFAULT_SCAN_TASK_BATCH_SIZE       = 512ULL * 1024 * 1024;  // 512 MB
-constexpr uint64_t DEFAULT_HASH_PARTITION_BYTES       = 512ULL * 1024 * 1024;  // 512 MB
-constexpr uint64_t DEFAULT_CONCAT_BATCH_BYTES         = 512ULL * 1024 * 1024;  // 512 MB
-constexpr uint64_t DEFAULT_SORT_SAMPLE_BYTES          = 512ULL * 1024 * 1024;  // 512 MB
-constexpr uint64_t DEFAULT_MAX_BUILD_HASH_TABLE_BYTES = 500ULL * 1024 * 1024;  // 500 MB
-constexpr uint64_t DEFAULT_MAX_BROADCAST_JOIN_SIZE    = 256ULL * 1024 * 1024;  // 256 MB
+/// Shared default for operator batch/partition sizing; the per-operator constants
+/// below alias it so one knob moves them together.
+constexpr uint64_t DEFAULT_BATCH_SIZE = 800ULL * 1024 * 1024;  // 800 MiB
```
wmalpica, line 37: What if we make this scale depending on GPU size but with a ceiling.
Like 0.025 x GPU size. For it would be 800MB for a 32GB GPU.
And then put a ceiling of 5GB for this default. 

THen if we do this, also have DEFAULT_MAX_BUILD_HASH_TABLE_BYTES = DEFAULT_BATCH_SIZE
And DEFAULT_MAX_BROADCAST_JOIN_SIZE = DEFAULT_BATCH_SIZE/2

bwyogatama: addressed

wmalpica (approved):

PR #1339, opened 2026-07-30
Reject unsupported expressions during GPU plan construction

ran-yuan-rui:
## Problem

Some unsupported expressions were added to GPU plans as null expression nodes.

`from_duckdb()` returns `nullptr` for expressions it cannot translate. Unmapped scalar functions are one source; the relevant struct/list/map extraction functions are not mapped either. Several plan builders did not check that result.

Seven reproduced shapes dereferenced a null translation and crashed the process. An ordinary filter above a join followed a different path: in the tested build, dereferencing the null expression propagated it into the expression evaluator, whose null check turned it into a mid-query failure and runtime CPU fallback. The dereference is still undefined behavior. This PR rejects the predicate during planning instead.

Affected shapes included:

- projections such as `round(x)`, `s.a`, and `lst[1]`
- sort keys computed by a projection, such as `ORDER BY round(x)`
- pushed-down and ordinary filters
- join conditions

## Fix

Plan construction now rejects expressions that cannot be translated in:

- projections
- pushed-down scan filters
- ordinary filters
- join conditions

Transparent execution then falls back to DuckDB CPU instead of installing an invalid GPU plan. Scan-side checks also distinguish an intentionally skipped filter from a failed translation, so an unsupported predicate cannot be silently dropped.

Nested sort keys are handled separately: STRUCT, LIST, and MAP keys are now rejected during planning instead of failing inside cuDF at runtime. Nested columns that only pass through the operator are unaffected, and supported operations such as `count(struct_col)` and `count(list_col)` are unchanged.

A defensive check also prevents partitioned-aggregate eligibility from dereferencing a malformed child projection.

The PR also fixes related diagnostics:

- DuckDB exception messages now use the correct printf-style placeholders.
- User-visible fallback errors no longer include DuckDB's serialized JSON envelope.
- Valid JSON that is not a DuckDB exception envelope is preserved unchanged.
- Plan-copy failures now all emit the same debug log; `NotImplementedException` was previously silent.

## Behavior

The affected queries now return the same result as `gpu_execution=false` instead of crashing or failing mid-query.

An unsupported expression is treated as a plan capability decision. If it remains in the logical plan, the query falls back even when the input is empty. Queries such as `LIMIT 0` or statistics-pruned scans can still stay on the GPU when DuckDB removes the unsupported expression before Sirius sees the plan.

Query recovery after a task-creation failure is handled separately and is not part of this PR.

## Verification

| Gate | Result |
|---|---|
| `make test` | 2221/2221 cases, 32,522,291 assertions |
| `make s3-test` | 90/90 cases, 31,383 assertions |
| `make s3-test-large` | 5/5 + 3/3 cases |
| `pre-commit run -a` | pass |
| `git diff --check` | pass |

Each production commit built independently, and the final branch passed a clean build.

Regression tests run the former crash shapes in isolated processes, compare results with DuckDB CPU execution, and assert plan-time fallback. Control cases verify that supported nested projections, aggregates, filters, and sort keys remain on the GPU.

### DIFF
--- a/src/op/sirius_physical_table_scan.cpp
+++ b/src/op/sirius_physical_table_scan.cpp
@@ -161,8 +161,6 @@ std::unique_ptr<operator_data> sirius_physical_table_scan::execute(const operato
   if (table_filters) {
     auto duckdb_filter = convert_table_filters_to_expression(
       *table_filters, column_ids, returned_types, batch_column_map);
-    // No converted filter means the predicates were handled elsewhere. A converted
-    // filter that cannot be translated must fail instead of passing rows through.
     if (duckdb_filter) {
       local_filter_expr = sirius::ast::from_duckdb(*duckdb_filter);
       if (local_filter_expr == nullptr) {
--- a/src/op/sirius_physical_table_scan.cpp
+++ b/src/op/sirius_physical_table_scan.cpp
@@ -161,7 +161,24 @@ std::unique_ptr<operator_data> sirius_physical_table_scan::execute(const operato
   if (table_filters) {
     auto duckdb_filter = convert_table_filters_to_expression(
       *table_filters, column_ids, returned_types, batch_column_map);
-    if (duckdb_filter) { local_filter_expr = sirius::ast::from_duckdb(*duckdb_filter); }
+    // A null converter result and a null translation mean different things and
+    // must not be collapsed. The converter returns nothing only when every
+    // predicate was already discharged — an advisory OPTIONAL_FILTER, an
+    // IS_NOT_NULL applied elsewhere, or a hive partition column pruned at the
+    // file-list level — in which case passing the batch through is correct. A
+    // translation that fails, on the other hand, means the filter cannot be
+    // applied at all; passing the batch through would silently emit rows the
+    // predicate should have removed. Fail closed instead. The planner refuses
+    // such plans before they get here (sirius_plan_get.cpp), so reaching this
+    // throw indicates a route that bypassed that gate.
+    if (duckdb_filter) {
+      local_filter_expr = sirius::ast::from_duckdb(*duckdb_filter);
+      if (local_filter_expr == nullptr) {
+        throw duckdb::InvalidInputException(
+          "TABLE_SCAN filter: cannot evaluate pushed-down predicate on GPU: %s",
+          duckdb_filter->ToString());
+      }
+    }
   }
 
   if (local_filter_expr != nullptr) {
--- a/src/op/sirius_physical_table_scan.cpp
+++ b/src/op/sirius_physical_table_scan.cpp
@@ -161,16 +161,8 @@ std::unique_ptr<operator_data> sirius_physical_table_scan::execute(const operato
   if (table_filters) {
     auto duckdb_filter = convert_table_filters_to_expression(
       *table_filters, column_ids, returned_types, batch_column_map);
-    // A null converter result and a null translation mean different things and
-    // must not be collapsed. The converter returns nothing only when every
-    // predicate was already discharged — an advisory OPTIONAL_FILTER, an
-    // IS_NOT_NULL applied elsewhere, or a hive partition column pruned at the
-    // file-list level — in which case passing the batch through is correct. A
-    // translation that fails, on the other hand, means the filter cannot be
-    // applied at all; passing the batch through would silently emit rows the
-    // predicate should have removed. Fail closed instead. The planner refuses
-    // such plans before they get here (sirius_plan_get.cpp), so reaching this
-    // throw indicates a route that bypassed that gate.
+    // No converted filter means the predicates were handled elsewhere. A converted
+    // filter that cannot be translated must fail instead of passing rows through.
     if (duckdb_filter) {
       local_filter_expr = sirius::ast::from_duckdb(*duckdb_filter);
       if (local_filter_expr == nullptr) {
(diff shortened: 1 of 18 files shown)


src/op/sirius_physical_table_scan.cpp
```diff
@@ -161,7 +161,16 @@ std::unique_ptr<operator_data> sirius_physical_table_scan::execute(const operato
   if (table_filters) {
     auto duckdb_filter = convert_table_filters_to_expression(
       *table_filters, column_ids, returned_types, batch_column_map);
-    if (duckdb_filter) { local_filter_expr = sirius::ast::from_duckdb(*duckdb_filter); }
+    // No converted filter means the predicates were handled elsewhere. A converted
```
felipeblazing: we dont need these comments

src/planner/sirius_plan_aggregate.cpp
```diff
@@ -163,6 +163,8 @@ static bool can_use_partitioned_aggregate(duckdb::ClientContext& context,
         for (auto& partition_col : partition_columns) {
           // we only support bound reference here
           auto const* expr = projection.select_list[partition_col].get();
+          // A partition key must reference a translated projection expression.
```
felipeblazing:
```suggestion
```

src/planner/sirius_plan_filter.cpp
```diff
@@ -91,11 +92,16 @@ sirius_physical_plan_generator::create_plan(duckdb::LogicalFilter& op)
       }
     }
 
-    auto filter =
-      duckdb::make_uniq<sirius::op::sirius_physical_filter>(std::move(filter_types),
-                                                            sirius::ast::from_duckdb(*combined),
-                                                            op.estimated_cardinality,
-                                                            std::move(output_indices));
+    // Reject unsupported predicates before the physical filter stores them.
```
felipeblazing:
```suggestion
```

src/planner/sirius_plan_get.cpp
```diff
@@ -62,6 +62,26 @@ duckdb::vector<std::unique_ptr<sirius::ast::node>> translate_expressions(
...
 }
 
+// Pushed-down filters bypass LogicalFilter. Match the runtime converter's skip
+// set, then validate the remaining predicate at plan time. The runtime translates
+// again because it resolves references against batch-relative positions.
+void reject_untranslatable_table_filter(duckdb::TableFilter const& filter,
+                                        duckdb::LogicalType const& column_type,
+                                        std::string const& column_name)
+{
+  if (filter.filter_type == duckdb::TableFilterType::OPTIONAL_FILTER ||
```
felipeblazing: why do we have this special check? can we wrap it in a function that is named something indicating why these are important is_blah_type_operator or osmething liek that?

src/planner/sirius_plan_get.cpp
```diff
@@ -62,6 +62,26 @@ duckdb::vector<std::unique_ptr<sirius::ast::node>> translate_expressions(
...
+      filter.filter_type == duckdb::TableFilterType::IS_NOT_NULL) {
+    return;
+  }
+  auto column_ref = duckdb::make_uniq<duckdb::BoundReferenceExpression>(column_type, 0);
+  auto expression = filter.ToExpression(*column_ref);
```
felipeblazing, line 84: can ToExpression fail and return  a nullptr? I don't think the next line will ever be false.

felipeblazing: it CAN throw an exception

felipeblazing (changes requested):

src/planner/sirius_plan_get.cpp
ran-yuan-rui: > can ToExpression fail and return a nullptr? I don't think the next line will ever be false.

You're right — removed

ran-yuan-rui: > it CAN throw an exception

Correct — StructFilter::ToExpression can throw InternalException (StructType::GetChildType goes through optional_ptr::CheckValid when the column type is not a struct, plus a checked NumericCast).

src/planner/sirius_plan_get.cpp
ran-yuan-rui: Done — wrapped as is_discharged_without_translation. The reason for the special-casing: those two filter types never reach the predicate the runtime translates (an OPTIONAL_FILTER is advisory, and IS_NOT_NULL is applied by the scan itself), so probing them would reject plans the scan actually handles correctly. The skip set mirrors convert_table_filters_to_expression in scan_utils.cpp, and the comment on the function records that they must stay in step.

src/planner/sirius_plan_get.cpp
ran-yuan-rui: It cannot fire from this call site, though: reject_nested_column_type runs on the same column type one line above and rejects STRUCT/LIST/MAP before the probe, and a StructFilter only appears on a struct column. And if one ever did throw, the exception leaves create_plan and the rebind path falls back to CPU — the same outcome this probe produces deliberately, so there is nothing to catch here.

src/planner/sirius_plan_aggregate.cpp
ran-yuan-rui: fixed

src/planner/sirius_plan_filter.cpp
ran-yuan-rui: fixed

felipeblazing (approved):

PR #1340, opened 2026-07-30
refactor: use cucascade io in sirius

aminaramoon:
replace sirius io with cucascade io


ran-yuan-rui:
Synced offline 

1. perf_instrumentation moves into cuCascade. The existing atomic counters (`perf_snapshot` and the per-context counters behind it) are ported over in full, not dropped — they are cheap enough for the production path, and the Sirius test suites and the s3-bench liveness check consume them. I am doing that port.

2. `ctrack` is a compile-time dev profiler only: coarse-grained scopes such as HEAD, LIST, and the parquet footer probe. It does not go into the per-GET production hot path — that is what the atomic counters are for.

3. Unit tests follow the code into cuCascade — not just the four files this PR currently deletes (sigv4, uri_parser, credentials, authorizer); the io unit-test set as a whole should live upstream next to the code.

I will open the counters PR against cuCascade first early next week

ran-yuan-rui:
@aminaramoon 
The agreed counters port is now NVIDIA/cuCascade#178, with 20 loopback conformance tests (`cucascade_io_tests` 83 → 103).

We rehearsed #1340 with dev `7b7be9c5`, the submodule on #178, and all deleted perf tests, YAML wiring, and benchmark JSON fields restored. All restored perf coverage passes; tests are green.

The only two failures are retry-log assertions. Retries and counters work, but `CUCASCADE_LOG_WARN` is currently a no-op. We should keep those tests as a documented carve-out until cuCascade#175 wires logging into Sirius.

After #178 merges, this PR only needs the new submodule pin, a dev update, and the rehearsed test restores. We already have those commits ready to cherry-pick.

ran-yuan-rui:
Merged dev (three conflicts in pin_table and sirius_scan_manager — took dev's new parameters and put your rename on top), 
bumped the pin to 416ba85 for thttps://github.com/NVIDIA/cuCascade/pull/178, restored the perf tests, fixed one dev-added test that includes a header this branch deletes, and corrected two docs.

Once the logging PR 175 lands, we could start to merge this one to Sirius

ran-yuan-rui:
The test failed at known logging gap: retries succeed and the counters are correct, but cuCascade `416ba85f` compiles `CUCASCADE_LOG_WARN` as a no-op, so no warning reaches the Sirius sink. The proper fix is the logging hook in cuCascade #175; until that lands, these logging assertions will temporary failed.

PR #1341, opened 2026-07-30
Allow some more dynamic filters to be published

joosthooz:
Two changes that widen dynamic-filter coverage without changing how any join executes.

  **1. Wire filters for unfiltered but *derived* build sides** (`aa9a80d9`). The "unfiltered
  build = whole key domain" refusal is only true for base tables. A build that is itself a
  join/aggregate output has already been reduced, so its filter can be highly selective.
  Those are now wired, and the runtime selectivity gate decides. Also adds a
  `dynamic filter NOT published` diagnostic for joins that are wired but never publish.

  **2. Publish from any single-partition folded build** (`d915715c`). The publisher was gated
  on `_join_mode == BUILD_PROBE`, but what it actually needs is a batch carrying the whole
  build side. The PARTITION operator now folds (`concat_all`) a single-partition build whose
  join publishes filters and reports it via `set_build_arrives_whole()`; the claim drops the
  mode test. This recovers e.g. q21 `id=51`, a MIXED_JOIN with an 855 MB single-partition
  build whose filter removes ~94% of q21's largest probe stream.

  ## Results

  TPC-H SF1000 vs `dev`, 3 iterations, grouped, pinned to host, best-of-3:

  | | dev | this branch |
  |---|---|---|
  | 22-query sum | 20.7602s | **20.1328s (−3.0%)** |
  | q21 | 2.6474s | **2.0941s (−20.9%)** |
  | filters published / iteration | 37 | **40** |

  No join changes execution mode; all 22 query outputs are byte-identical to `dev`.

### DIFF
--- a/src/op/sirius_physical_partition.cpp
+++ b/src/op/sirius_physical_partition.cpp
@@ -418,13 +418,12 @@ std::unique_ptr<operator_data> sirius_physical_partition::get_next_task_input_da
                                    "single batch (concat_all)");
         }
         build_arrives_whole = strategy.num_partitions == 1 || strategy.broadcast;
-      } else if (strategy.num_partitions == 1 && hash_join != nullptr &&
+      } else if (in.is_build_side && strategy.num_partitions == 1 && hash_join != nullptr &&
                  hash_join->publishes_dynamic_filters()) {
-        // Not BUILD_PROBE, but the whole build lands in one partition and the join publishes a
-        // dynamic filter from a single build batch. Folding that partition costs nothing in data
-        // volume — it is the same rows either way — and only moves a batch boundary, so the filter
-        // can be published from a key set that is provably complete. Unlike BUILD_PROBE this is an
-        // optimisation, not an invariant: if no build-side CONCAT exists, publication is skipped.
+        // Not BUILD_PROBE, but the build lands in one partition and this join publishes a filter
+        // from a single build batch, so folding it only moves a batch boundary. Best-effort: no
+        // build-side CONCAT means no publication. Build-side sizing is required — right-family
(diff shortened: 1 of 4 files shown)


wmalpica (approved):

src/op/sirius_physical_partition.cpp
```diff
@@ -415,7 +417,19 @@ std::unique_ptr<operator_data> sirius_physical_partition::get_next_task_input_da
                                    "to fold the build into a "
                                    "single batch (concat_all)");
         }
+        build_arrives_whole = strategy.num_partitions == 1 || strategy.broadcast;
+      } else if (strategy.num_partitions == 1 && hash_join != nullptr &&
+                 hash_join->publishes_dynamic_filters()) {
+        // Not BUILD_PROBE, but the whole build lands in one partition and the join publishes a
```
kevkrist: It looks like RIGHT joins size partitions by the probe side (sirius_pipeline_converter.cpp:461), in which case forcing the build to fully concatenate could be problematic. Also, I believe the probe side for RIGHT joins executes before the build, so the dynamic filter operator will execute `close_for_new_filters()` and disable filter publication, in which case build-side folding for dynamic filters is moot.

I think this can be fixed by simply to include `in.is_build_side` in the branch above.

kevkrist (changes requested):

src/op/sirius_physical_partition.cpp
joosthooz: Thanks, I've included your suggested change in ac43e0b and everything still works as before. We wouldn't have caught this by running tpch.

kevkrist (approved):

PR #1342, opened 2026-07-30
perf(join): Let max_build_hash_table_bytes decide BUILD_PROBE again

joosthooz:
BUILD_PROBE eligibility was only evaluated when the natural partition count fit
  (`proposed <= num_gpus`), so on a single GPU a build even one byte over
  `hash_partition_bytes` was refused. We added an override config parameter for that in #507; `max_build_hash_table_bytes` allows a join to use probing mode even if that means the (single) partition is larger than a normal batch. The reason is that probing mode is much faster and efficient.
  
This regressed in two steps. #1088 switched partition counts from floor to ceiling division, batches between `hash_partition_bytes` and `max_build_hash_table_bytes` silently used standard join mode instead of probing (#1088's own example, q21's 5.94 GB build vs a 5 GiB target, is exactly such a case). The multi-GPU rework in #1194 then rebuilt eligibility around that same count as proposed <= num_gpus (equivalent to the old num_partitions == 1 on a single GPU), cementing the coupling into the new structure.
 
  Eligibility is now evaluated at the count BUILD_PROBE actually runs with — one build table
  per GPU, capped at the natural count — and that count is returned when admitted. Multi-GPU
  cases that already fit are unchanged.
  
  ## Results

  TPC-H SF1000 vs `dev` (5 GB batch target, 10 GB build budget, grouped, pinned to host,
  best-of-3): q7 id=33 (6.00 GB), q19 id=8 (6.11 GB) and q21 id=18 (5.94 GB) return to
  single-partition BUILD_PROBE, each also regaining its dynamic filter (37 → 40 per 
  iteration). q7 −3.7%, q19 −3.8%, q14 −6.9%, q21 −1.9%; 22-query sum −1.1%
  (20.7602s → 20.5301s). All query outputs byte-identical to `dev`.
 
  q13 is the control: its 18 GB `RIGHT`-join build stays excluded by `is_right_family`, so
  #1088's probe-driven sizing keeps working.
  
  Tests: the case that encoded the coupling now asserts the corrected behaviour; added a
  both-limits-exceeded case (stays STANDARD) and a regression test with #1088's exact q21
  numbers.


kevkrist (approved):

bwyogatama:
Superseded by #1371

PR #1345, opened 2026-07-30
fix(scan): re-evaluate pipeline status when the split connector closes

felipeblazing:
## Problem

Pipeline completion is a conjunction — source exhausted AND first node drained AND `tasks_created == tasks_completed` — that `update_pipeline_status()` only re-evaluates at call sites paired with those facts. Whichever fact flips **last** must trigger a re-check, or the pipeline never finishes.

Task completion is paired via `mark_task_completed()`. Closing the split connector is the source-exhaustion edge, and on the scan path it was **unpaired**: the close happens on the scan-manager thread ([`load_balancing_scan_batch_coalescer.cpp`](../blob/dev/src/scan_manager/load_balancing_scan_batch_coalescer.cpp)) and pokes neither call site.

When the connector closes *after* the last task completed and *after* the last task-creation exit, the conjunction is never re-tested. The pipeline stays unfinished forever; a downstream hash join gated on its `probe_finished` parks with build batches still resident, and every engine thread goes idle with the query never completing.

## Fix

Pair the edge with a direct `update_pipeline_status()` call at each `connector->close()` site.

This mirrors existing precedent — `sirius_physical_streaming_source` already pairs its own channel close exactly this way:

```cpp
if (_input_channel->closed()) { pipeline->update_pipeline_status(false); }
```

The call must be **direct** rather than routed through `task_creator::schedule()`: a scheduled request is dropped whenever `get_operator_for_next_task()` returns null, so the re-evaluation inside the creation path would never run. (I tried the scheduled version first; it did not help.)

## Verification

SF1000, 4x GB200, scan-from-disk, dynamic filter pushdown on:

| build | hangs |
|---|---|
| before | 6 / 18 (~33%) |
| with this fix alone | 2 / 20 (~10%) |

## Relationship to #1344

This closes one unpaired edge but **not** the whole class. sirius-db/sirius#1344 fixes another one (the dropped-request path in `task_creator::manager_loop`) and is what takes the hang rate to zero.

Being straight about the evidence: an ablation running as I write this — #1344 alone, without this change — is clean so far (17/25 at time of writing), which suggests #1344 may be sufficient on its own and this change is not load-bearing. I'll update this PR with the final number.

I'm proposing it anyway because it is a genuine missing pairing on its own terms, consistent with how `streaming_source` already handles the same edge, and because it independently cut the failure rate by a factor of three. But if reviewers prefer the minimal change, #1344 is the one that matters and this can be closed without loss.

The two are independent (different files, no overlap) and can land in either order.

Based on `2f6192af`.

### DIFF
--- a/src/include/scan_manager/load_balancing_scan_batch_coalescer.hpp
+++ b/src/include/scan_manager/load_balancing_scan_batch_coalescer.hpp
@@ -108,6 +108,11 @@ class load_balancing_scan_batch_coalescer {
 
     std::size_t op_id{0};
     std::size_t pipeline_id{0};
+    /// The scan operator this slot feeds. Needed so closing the connector can
+    /// ask the task_creator to revisit the operator: pipeline completion is
+    /// only re-evaluated at a task-creation exit, and closing the source is the
+    /// last thing that can make the pipeline finishable. See notify_source_closed().
+    op::scan::sirius_gpu_scan_operator* scan_op{nullptr};
     using provider_value_t = exec::try_t<std::unique_ptr<op::scan::scan_info>>;
     duckdb_moodycamel::BlockingConcurrentQueue<provider_value_t> queue;
     std::shared_ptr<op::scan::batch_coalescer> coalescer;
--- a/src/scan_manager/load_balancing_scan_batch_coalescer.cpp
+++ b/src/scan_manager/load_balancing_scan_batch_coalescer.cpp
@@ -18,7 +18,37 @@
 
 #include "exec/try.hpp"
 #include "log/logging.hpp"
+#include "op/scan/sirius_gpu_scan_operator.hpp"
 #include "op/scan/sirius_gpu_scan_operator_data.hpp"
+#include "pipeline/sirius_pipeline.hpp"
+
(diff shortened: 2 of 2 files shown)


felipeblazing:
Ablation finished: **0 hangs / 25 runs with #1344 alone**, without this change.

So this PR is **not load-bearing** — #1344 is sufficient on its own.

Summary of the evidence:

| build | hangs |
|---|---|
| baseline (neither fix) | 6 / 18 (~33%) |
| this PR alone | 2 / 20 (~10%) |
| #1344 alone | **0 / 25** |
| both | **0 / 25** (q1-q8), 0 / 15 (all 22 queries) |

My recommendation: **land #1344 and close this one**, unless a maintainer judges the unpaired source-exhaustion edge worth closing on its own merits. It is a real missing pairing — the connector close flips a conjunct that nothing re-evaluates, and `sirius_physical_streaming_source` already pairs its equivalent edge the same way — but the data says #1344 alone already covers the observed failure, and a smaller diff in a concurrency path is worth more than a defensive extra edit.

Happy either way; deferring to your call.

PR #1349, opened 2026-07-31
perf(scan): memory prefetcher drain-handling tuning (stacked on #1181)

felipeblazing:
**Stacked on #1181** — review the top two commits only; the rest is the #1181 base. Draft until #1181 merges, then this rebases to a clean diff.

## What

Two behavioral refinements to the memory prefetcher's sweep, found by driving quent's batch-placement telemetry (#1187) against SF1000 host-pinned runs:

1. **Tail-first staging of draining connectors** (`f1de4ed9`): the drain guard previously skipped an actively-draining connector entirely, so the running scan's own splits always uploaded synchronously at task time. Contention with the scan's conversion threads only exists near the pop point — so convert the queue tail-first behind a 4-batch head margin.
2. **Parallel-safe probing + one-worker drain claim + gate diagnostics** (`0322119c`): the batch tier pre-check used a blocking `to_read_only()`, convoying N workers into effective concurrency 1 whenever a sibling was mid-conversion (telemetry-verified). With that fixed, a per-connector claim keeps tail-first work to a single worker so prefetch parallelism never stacks onto the scan's own conversion threads. Gate counters (headroom/reservation stops, lock/drain skips) log at teardown.

## Results (GB300, TPC-H SF1000 hot host-pinned, 3 iters, byte-identical results)

- On the #1181 base: suite 19.74s → 19.44s (-1.5% on top of the prefetcher's win); q9 -3.1%, q12 -2.5%, q21 -1.7%; worst query +0.5%. The originally guard-motivating queries q1/q13 improve as well.
- Guards validated: fully relaxing them (no claim/margin/tail-first) costs +3.5% suite (q15 +18%) — numbers in the gate counters made this cheap to establish.


felipeblazing:
Moving to a fork-internal PR based directly on feat/scan-prefetcher so the diff shows only this change's commits; will reopen here against dev once #1181 merges.

felipeblazing:
Reopening — keeping the stacked PR here against sirius-db; the diff includes #1181's commits until it merges, then this rebases clean.

felipeblazing:

PR #1351, opened 2026-07-31
bench: label each query execution for quent telemetry

felipeblazing:
`performance_test.py` never sets a query label, so quent captures show every execution as `unnamed_query` — a grouped-mode session with 22 queries × N iterations can only be attributed by execution order. This calls `sirius_set_query_label('q<N>_iter<i>')` before each GPU execution (non-fatal if the function is unavailable), making quent sessions navigable per query.

Used for all the quent DAG sessions behind #1349's analysis.


felipeblazing:

PR #1352, opened 2026-07-31
docs: tuned GB300/SF1000 host-pinned configuration profile

felipeblazing:
Adds a measured tuning profile to `configuration.md`: host `block_size: 64Mi`, `pipeline.num_threads: 8`, and the memory prefetcher enabled with 3 workers — worth **-13% suite hot** (20.8 → 18.1 s, q9 -19%) on GB300 at TPC-H SF1000 host-pinned, with per-knob attribution and the settings that measured neutral-or-worse.

Draft until its dependencies land: the prefetcher block needs #1181, part of the converter-side gain needs the cuCascade null-mask fix, and an 8–64 Mi `block_size` sweep is still pending to bound small-allocation fragmentation.


felipeblazing:

PR #1353, opened 2026-07-31
RFC: downgraded-task input prefetcher

felipeblazing:
Note: Not sure if this is a good approach yet or if we should find a way to fold this into the memory_prefetcher since they would almost always be leveraging the same i/o path, host to device.

## What

The memory-prefetcher concept (#1181) applied to the other host→GPU path: inputs of **queued pipeline tasks** that were downgraded to HOST/DISK get converted back to GPU tier ahead of dispatch, instead of synchronously inside `prepare_for_processing`. Workers walk the task queue in dispatch order (mirroring the scheduler's `pop_if(front_to_back=true)`) via a new `inspectable_mpsc::for_each_mutable` visitor, and upgrade through the downgrade path's own `convertible_data_batch` (non-blocking lock, reservation-backed, RAII idle-restore). Env-gated off (`SIRIUS_TASK_PREFETCH=1`); single-GPU scope; 6-case Catch2 suite.

Two safety mechanisms, both empirically required on GB300/SF1000:
- **Pressure-quiet hysteresis** (250 ms): without it, the prefetcher re-upgrades batches the downgrade executor just evicted and can wedge the engine at the memory budget under constrained GPU configs (reproduced and fixed during development).
- **Dispatch-order lookahead cap** (4 tasks): bounds the non-reclaimable prefetched footprint.

## Honest evaluation — why RFC/draft

- At the standard 0.95 GPU budget, q9/q21 at SF1000 never downgrade on current dev (zero downgrade events at debug level) — the prefetcher is a verified no-op there.
- Under an artificially constrained budget (0.4), it stages 15-20 GB/query correctly with byte-identical results and unchanged downgrade counts, but hot timings are **neutral**: the calm windows in which it may act don't overlap the downgrade traffic it would need to hide.
- The motivating regime is genuinely oversized workloads — disk-spill readback through the serialized `pipeline_io_backend` — which this box's HBM can't produce at SF1K. Parking as draft until such a workload validates (or falsifies) the win.

Note: branch is based on dev@3ee47f8b and needs a rebase before any serious review.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -333,6 +333,7 @@ set(EXTENSION_SOURCES
     src/op/sirius_physical_top_n.cpp
     src/op/sirius_physical_ungrouped_aggregate.cpp
     src/parallel/task_executor.cpp
+    src/pipeline/downgraded_task_prefetcher.cpp
     src/pipeline/gpu_pipeline_executor.cpp
     src/pipeline/gpu_pipeline_task.cpp
     src/pipeline/sirius_pipeline_itask.cpp
@@ -370,6 +371,7 @@ set(EXTENSION_SOURCES
     src/scan_manager/mvcc_mask_job.cpp
     src/scan_manager/pinned_chunk_stats.cpp
     src/scan_manager/round_robin_strategy.cpp
+    src/scan_manager/memory_prefetcher.cpp
     src/scan_manager/sirius_scan_manager.cpp
     src/scan_manager/split_connector.cpp
     src/scan_manager/split_provider.cpp
@@ -709,6 +711,7 @@ set(TEST_SOURCES
     test/cpp/planner/test_query_id.cpp
     test/cpp/planner/test_query_index.cpp
     test/cpp/pipeline/test_batch_lock_utils.cpp
+    test/cpp/pipeline/test_downgraded_task_prefetcher.cpp
     test/cpp/pipeline/test_gpu_pipeline_executor.cpp
     test/cpp/pipeline/test_oom_reschedule.cpp
     test/cpp/pipeline/test_pipeline_memory_history.cpp
--- a/src/include/exec/multi_index_priority_queue.hpp
+++ b/src/include/exec/multi_index_priority_queue.hpp
@@ -429,6 +429,30 @@ class multi_index_priority_queue {
     std::lock_guard<std::mutex> lock(_mutex);
     return _levels.empty();
   }
+
+  /// Visits every queued task in dispatch order (lowest priority level first,
+  /// FIFO within a level; reversed when @p front_to_back is false) without
+  /// removing anything. The visitor returns false to stop the walk early. Runs
+  /// under the queue mutex: the visitor must be lightweight and must not touch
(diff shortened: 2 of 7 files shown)


felipeblazing:
## Branch refreshed + full experimental record (2026-07-31)

The branch is rebuilt on current dev (`c4e8a10b`): the scheduler queue is now `multi_index_priority_queue`, so the prefetcher's dispatch-order snapshot moved to a new `for_each_mutable` visitor (locked walk of the priority spine, FIFO within level). Env knobs: `SIRIUS_TASK_PREFETCH=1`, `_THREADS`, `_MIN_FREE_FRACTION`, `_LOOKAHEAD`, `_QUIET_MS`, and the new `_DURING_PRESSURE`.

### Experimental verdict — measured on a workload that GENUINELY downgrades

Test bed: GB300, TPC-H SF1000, GPU-pinned Simpatico-compressed lineitem+orders (dims GPU raw, partsupp host), pipeline threads 8. q9 downgrades 41.8 GB of ~5 GB CONCAT intermediates at ~+479 ms and re-uploads them at ~+925 ms, every iteration (quent `data_batch` InTransit evidence; note downgrade activity is debug-level — info logs show nothing).

| variant | staged by prefetcher | q9 hot mean |
|---|---|---|
| OFF | — | 1.872 s |
| defaults (floor 0.4) | **0 bytes — deadlocked** | 1.879 s |
| floor 0.25, lookahead 8 | 10.6 GB | 1.868 s |
| floor 0.15, la 16, quiet 50 ms, 4 thr | 53.2 GB (saturated) | 1.869 s |
| floor 0.10, la 32, quiet 50 ms, 4 thr | 53.3 GB | 1.874 s |
| concurrent-during-pressure, orig evict order | 51.1 GB | 1.880 s |
| concurrent + order-disjoint eviction | 47.3 GB | 1.881 s |
| *(reference: no eviction at all — trigger 0.9/stop 0.85)* | — | **1.817 s** |

Findings:
1. **Design flaw in the defaults**: `min_free_fraction 0.4` deadlocks against `downgrade_stop_fraction 0.6` — eviction stops at exactly 40% free, so the prefetcher never sees headroom and silently stages nothing. Any future version must enforce `floor < 1 − downgrade_stop_fraction`.
2. **Flat across a 25x staged-bytes range**: with 8 pipeline threads the re-upload the prefetcher front-runs is already fully overlapped at task-prepare. Recovery-side prefetching does not move runtime.
3. **Prefetch-during-pressure is SAFE but not faster**: with order-disjoint victims (task-queue eviction is already back-to-front; an experimental `SIRIUS_EVICT_REPOS_DESC=1` flips repo eviction to latest-consumed-first) there is no bounce/wedge — but it is neutral-to-slightly-negative, because the eviction cost is the D2H work + batch locks contending with execution, which concurrent prefetch adds to.
4. **The only lever that recovered the time was not evicting**: trigger/stop tuning (see the compressed-pinning docs in #1356).

**Recommendation**: keep as RFC, do not merge for this regime. Revisit for spill-bound deployments (datasets exceeding GPU memory, small GPU budgets, multi-tenant slices) where eviction is unavoidable — with the floor/stop deadlock fixed first.

PR #1355, opened 2026-07-31
TPC-H harness: SIRIUS_PRE_SQL hook + per-table pin-tier overrides

felipeblazing:
Two ~4-line hooks in the TPC-H performance harness that make compressed and mixed-tier pinning experiments runnable without harness edits:

- **`SIRIUS_PRE_SQL`** (performance_test.py): semicolon-separated SQL executed right after `LOAD`, e.g. `SET pin_table_compression = true; SET pin_table_input_compression_plan_dir = '...'`.
- **`SIRIUS_PIN_TIER_<TABLE>`** (tpch_pin_columns.py): per-table override of the global `SIRIUS_PIN_TIER`, so one run can pin `lineitem` to `gpu` and `partsupp` to `host`.

Used to produce the GB300 SF1000 results in the compressed-pinning docs PR (baseline 20.74 s → 16.66 s hot with lineitem+orders GPU-pinned compressed).


felipeblazing:

PR #1357, opened 2026-07-31
fix: Addressing some open compression issues

### LINKED ISSUE #1211: gpu_pipeline_task::get_estimated_reservation_size_info should account for compressed pinned tabled properly
As a followup task after https://github.com/sirius-db/sirius/pull/937
We need to make sure that task reservation takes into account the memory requirements for decompression.

I think right now we are only assuming that the memory necessary is just the uncompressed size, but if you have a host table that is compressed the memory requirements are actually the compressed size moving to GPU, then the uncompressed size on GPU plus any necessary scratch space overhead the decompression needs.  This all somehow should get accounted for. Not sure what needs to come from `bytes_to_materialize` vs `peak_memory_estimate` we need to investigate.

Either way, we should have a unit test which gets an estimate from a table that is pinned on host and then we actually execute this task and we see how much memory was consumed and validate that our estimates are accurate.

I think its important to have some level of accuracy on scan memory estimates because it is one of the highest drivers of memory consumption and queries can actually OOM fairly easily when there is a very large scan at the beginning of a query.

### LINKED ISSUE #1306: Handle sliced views as input to compression API
Currently, passing sliced views is not supported and will result in an error.

### LINKED ISSUE #1307: Remove thread spawning from compression code
Instead, use a single thread that sends work to a couple of streams. This is just as fast as the current parallelized version but doesn't have the problem of memory reservation tracking (which is attached to the gpu executor thread, so will not work on those spawned threads).

joosthooz:
This PR aims to address some of the issues that were created from comments on the original compression subsystem PR (#937), for which we made follow-up issues.
Also fixes the following other issues:
- Increasing smem if needed, and checking if that succeeded
- Updating a test for bitpack that was changed in a previous PR #1325 
- Catch CUDA errors in ALP instead of silently corrupting
- Include fixed-point datatypes for exploring numerical operators, update default plans to use the much faster and more-compressing bitpack for decimal tpch columns

Closes #1307 
Closes #1211
Closes #1306


src/compression/compression_converters.cpp
```diff
@@ -46,25 +45,25 @@
...
+// Work is submitted from the calling thread so cuCascade memory-reservation
+// tracking (attached to the calling thread) sees all allocations.
+// 4 is not a configuration parameter — it matches the typical SM occupancy
+// sweet spot for column-parallel decode without thread-spawn overhead.
+simpatico::stream_pool& decode_pool()
 {
-  return g_decompress_column_threads.load(std::memory_order_relaxed);
+  thread_local simpatico::stream_pool pool;
+  if (pool.streams.empty()) {
+    if (!pool.init(4)) throw std::runtime_error("[compression_converters] stream_pool init failed");
```
kevkrist, line 58: It looks like you can still SET the number of threads as an extension option at sirius_extension.cpp:2292, which now I think does nothing? Or is it still doing something for compression?

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -219,6 +219,18 @@ std::unique_ptr<op::operator_data> run_one_operator(
...
   const cucascade::memory::memory_space* target_space) const
 {
+  // Peak device memory while making one representation GPU-resident.
+  // For uncompressed data the source lives in host memory; only the destination
+  // lands on device, so the peak equals the uncompressed size.
+  // For compressed data the encoded payload must first be staged on device
+  // before decompression produces the output, so both are alive simultaneously:
+  //   peak = compressed_bytes + uncompressed_bytes.
+  auto peak_materialization_bytes = [](const cucascade::idata_representation* data) -> std::size_t {
+    auto const compressed   = data->get_size_in_bytes();
```
kevkrist, line 247: I think this overestimates when there is a projection, right? I think this is the uncompressed/compressed bytes for *all* columns cached in the chunk, in which case the reservation will end up being excessively large. Or does the projection happen after materialization?

src/compression/compression_converters.cpp
joosthooz: Those options, and the config parameter, were still active but just for setting the nr of threads on the compression side. So currently only during pinning time. I don't like that the 2 paths behave differently, so I've converged them now and removed the option. It's possible that pinning became slower because of this, because the compression side of simpatico synchronizes after every operator. I've fixed that in my WIP branch with spill compression, but it might be a while before that gets merged. If that becomes a problem I can create a small PR just for that.

src/pipeline/gpu_pipeline_task.cpp
joosthooz: The projection does happen during the decompression already, so this is definitely an overestimation. 
I've added a very crude estimation of the projectedsize (just divides the nr of projected columns by the total columns). If we want more accurate sizes, we need to either parse the compressed header here, or store some extra info during pin time, that we can retrieve here. Let me know if you want either of those options.

kevkrist: Yeah, I think it'd be best to store exact per-column materialization sizes at pin time. `build_compressed_table_header()` already visits each buffer with its column index and exposes size_bytes/alloc_bytes, while the source table can provide each column’s decoded alloc_size(). The projected estimate can then simply sum the selected columns.

kevkrist (approved):

PR #1356, opened 2026-07-31
Docs: compressed pinning — tier choice, plan selection, GB300 SF1000 results

felipeblazing:
Adds `docs/super-sirius/compressed-pinning.md`: how `pin_table` compression works, and the measured guidance from the GB300 TPC-H SF1000 campaign:

- **Host-tier compressed pins lose** (+7.0% hot suite): per-batch payload fetch → sync → decode sits on the scan-prepare critical path and short queries can't hide it.
- **GPU-tier compressed pins win big** (-19.7% hot suite, 20.74 s → 16.66 s): decode-only replaces the ~370 GB/s C2C link, and compression is what makes hot-column GPU residency *fit* (top-7 lineitem columns: 264 GB raw vs ~93 GB compressed, ~2.4 of 2.9 TB suite pin traffic).
- Plan selection rule: prefer plans whose decompress throughput clears the H2D link rate; plan choice is tier-dependent (`o_comment` snappy for host pins, `identity` for GPU pins).
- `ans` >2 GiB per-batch limitation and why nvcomp codecs are size-safe.

Companion PRs: #1354 (plan fixes), #1355 (harness hooks used for the measurements).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1354, opened 2026-07-31
Faster TPC-H SF1000 compression plans (l_quantity bitpack, ps_comment identity, supplier ans)

felipeblazing:
Updates three checked-in SF1000 plan files with strictly better codecs, measured on GB300 with the `simpatico_cli` benchmark (verified roundtrips, part.0 samples):

| column | old plan | old ratio @ decomp | new plan | new ratio @ decomp |
|---|---|---|---|---|
| `l_quantity` | snappy | 3.96x @ 306 GB/s | bitpack | **4.885x @ 1693 GB/s** |
| `ps_comment` | dict cascade | 1.019x | identity | 1.0x @ memcpy speed |
| `s_name` | deep dict cascade | 3.20x @ 231 GB/s | str_split + chars ans | 2.02x @ 422 GB/s |
| `s_comment` | str_split cascade | 2.73x @ 235 GB/s | str_split + chars ans | 1.84x @ 493 GB/s |

Notes:
- `l_quantity`: the beam-search explorer never enumerates `bitpack` for DECIMAL64 (t26) columns, so the original Pareto pick missed a dominant plan. Explorer fix tracked separately.
- `ps_comment` is effectively incompressible; the old plan spent compress time to save ~2% of bytes.
- The supplier `ans` swaps trade some ratio for 2x decode speed; `ans` is size-safe there because supplier's total char payload is far below the 2 GiB cudf-column limit (`ans` on larger char channels crashes — also tracked separately).

These plans feed `pin_table` compression (`pin_table_input_compression_plan_dir`) and were validated end-to-end at SF1000: results identical to uncompressed pins across all 22 TPC-H queries.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1361, opened 2026-07-31
Zero-copy scan materialize for per-query converted pinned chunks

felipeblazing:
When a pinned-cache scan split needs conversion at prepare time — a compressed GPU/host pin decompressed on demand, or a host pin uploaded — the converted table exists only for this query (the wrapper batch is per-query and dropped after the scan). `materialize_table` nevertheless served it through a read-lock-backed `owning_table_view`, whose `release()` deep-copies every column: **~408 GB of device-to-device memcpy per 1.81 s hot q9 iteration** at SF1000 with compressed GPU pins (nsys, memcpy→NVTX correlation), plus second-order congestion (decompress workers waiting 4.4 ms per 0.3 ms decode).

This takes ownership instead: right after `prepare_for_processing`'s `convert_to` (same exclusive lock), move the fresh table out via `gpu_table_representation::release_table()` (a move for the owned alternative — no device copy), park it on the `scan_operator_input`, and leave the batch a valid empty placeholder. `materialize_table` moves the parked table into the scan output through the existing `no_alloc_materializable` path.

Safety invariants (each unit-tested):
- **Raw GPU pin storage is never stolen** — it never enters the conversion branch (`needs_upload == false`).
- Masked / row-filtered splits keep the copying view path (they need the source view alive), which also closes the OOM-retry window: an unmasked, unfiltered scan allocates nothing after the take.
- Re-running `prepare_for_processing` is a no-op once parked/consumed; size estimates answer from the recorded size; a materialize re-entry after consumption throws loudly instead of serving zero rows.

Validation:
- Full unit suite passes (2226 cases); 3 new `[cached_serving]` tests.
- SF10: q1,3,4,9,10,13,16,19 byte-identical across gpu-raw / host-raw / gpu-compressed / host-compressed pin configs.
- SF1000 same-session A/B (GPU-pinned compressed lineitem+orders, hot mean of 4 iters): **q9 1.825 → 1.734 s (−4.9%), q18 1.262 → 1.194 s (−5.4%)**, row counts and results identical.

Companion PRs: #1354 (plan fixes), #1355 (harness hooks used for validation), #1356 (compressed-pinning docs with the profiling evidence).

### DIFF
--- a/src/include/op/scan/sirius_gpu_scan_operator_data.hpp
+++ b/src/include/op/scan/sirius_gpu_scan_operator_data.hpp
@@ -146,6 +146,22 @@ class scan_operator_input : public op::operator_data {
   /// copy). Stamped by drain_cached_provider on resident splits; scan_info
   /// splits fold filter costs into their own estimates instead.
   bool row_filter_pending{false};
+  /// Per-query table taken out of the cached wrapper batch right after
+  /// prepare_for_processing's conversion produced it (decompressed or
+  /// uploaded fresh for this split) — never raw GPU pin storage, which is
+  /// served as a plain gpu_table_representation and never converted.
+  /// Consumed at most once by materialize_table, which moves it into the
+  /// scan output instead of deep-copying the batch. Mutable: the operator
+  /// only sees its input as const during execute.
+  mutable std::unique_ptr<cudf::table> stolen_table;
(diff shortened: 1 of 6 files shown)


felipeblazing:

PR #1360, opened 2026-07-31
docs: add issue templates

mike-wendt:
## Description

Using feedback from #1255 this PR adds three new issue templates (listed below) and changes 2 existing templates. Reasoning for these changes are detailed below and open for feedback and discussion.

### Live testing
🧪 Templates in this PR are live in my [fork](https://github.com/mike-wendt/sirius/issues/new/choose) for testing.

#### CI Bug report
- Removed unhelpful fields to shorten form
- Updated title prefix to `bug(ci): ` instead of `[CI-BUG] ` to match Conventional Commits usage in repo
#### CI Feature request
- Removed template
- Replaced by new `Feature request` template

#### Bug report
- Voting results
  - cuDF template voted 5x 👍 and winner
- Approach
  -  Use cuDF's template with minimal changes
    - Added mention of `SQL examples` to `Additional context` given that may be helpful
  - Added `print_env.sh` script in #1362
#### Feature request
- Voting results
  - cuDF template voted 2x 👍
  - Arrow template voted 2x 👍 
- Approach
  - Given the tie, I took a look at both templates and the existing CI Bug Report template
  - Compared the two projects in scale
    - Sirius is more similar to cuDF than Arrow in terms of maintainers
    - Maintainers in this repo need actionable information upfront so when they have bandwidth they take action and not need to ask for follow ups
    - Noted that the Feature Request from cuDF has been unchanged for 8 years, so the questions asked there are proven
  - Decided to take a hybrid approach with questions from cuDF but in a form (like Arrow) to enforce maintainers have information
    - Unlike with the Bug Report and the `print_env.sh` script, in this request we heavily rely on the reporter to provide all the information to determine if the suggest idea can be implemented  
#### Documentation request
- Voting results
  - cuDF template voted 5x 👍 and winner
- Approach
  - Use cuDF's template with minimal changes
  - Observe the first few issues and see if we need to make any adjustments

## Checklist
- [x] Read CONTRIBUTING.md
- [ ] Cover changes with new or existing tests
- [ ] Document configuration changes in code and summarize in the description above
- [ ] Update human and agent documentation (README.md, docs/, skills, CLAUDE.md)

## References

Closes #1255 
Depends #1362

Issue #1255, opened 2026-07-21
Discussion - Standard Issue Templates


mike-wendt:
⚠️ **RESPONSE DEADLINE - 7/24/26 Friday Morning 11AM ET**
See also #1252 for PR templates

As we continue to grow and develop Sirius, we should also add Issue template for common issue types to ensure we have the necessary information we need to reproduce a bug, implement a feature, or correct a documentation gap. Just like the PR template, I want to leverage work done by other open-source projects so we don't have to start from scratch. 

There are three issue types that I think Sirius could benefit from:
1. Reporting bugs
2. Feature request
3. Documentation issues

In the comments below I'll post an example of each of these from 3 other open-source projects. Please 👍 emoji react to the issue templates that you think would be most impactful for Sirius. From there we can discuss the specific content and start PRs for review and further input. If there is an issue template you would like to consider please add it to this thread for discussion.

mike-wendt:
## Reporting bugs - Option 1 - DuckDB
- Example of [bug report](https://github.com/duckdb/duckdb/issues/new?template=bug_report.yml)
- Template [file](https://github.com/duckdb/duckdb/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml?plain=1)
- Very detailed to get as much information as possible from users

mike-wendt:
## Reporting bugs - Option 2 - cuDF
- Example of [bug report](https://github.com/rapidsai/cudf/issues/new?template=bug_report.md)
- Template [file](https://github.com/rapidsai/cudf/blob/main/.github/ISSUE_TEMPLATE/bug_report.md?plain=1)
- No forms used, lighter weight because they rely on an in-repo script to get the environment data

mike-wendt:
## Reporting bugs - Option 3 - Apache Arrow
- Example of [bug report](https://github.com/apache/arrow/issues/new?template=bug_report.yaml)
- Template [file](https://github.com/apache/arrow/blob/main/.github/ISSUE_TEMPLATE/bug_report.yaml?plain=1)
- Very light weight single text area with component drop down

mike-wendt:
## Feature requests - Option 1 - DuckDB
- Example of [request](https://github.com/duckdb/duckdb/discussions/new?category=ideas&title=Feature%20Request:%20...&labels=feature&body=Why%20do%20you%20want%20this%20feature%3F)
[...]

PR #1362, opened 2026-07-31
docs: add print_env script for bug reports

mike-wendt:
## Description

Follows discussion in #1255, selecting cuDF as the Bug Report template. cuDF uses a [print_env.sh](https://github.com/rapidsai/cudf/blob/main/print_env.sh) script to

### Changes to script
- Add nvidia-smi topo to output
- Add error messages for missing tools
- Add clang version check
- Remove python and conda checks
- Add Rust, Cargo, Pixi, DuckDB checks
- Add Sirius build type
- Collect `pixi info` and `pixi list` at bottom to keep other data visible

## Checklist
- [x] Read CONTRIBUTING.md

## References

Blocks #1360
Refs #1255

### DIFF
--- a/print_env.sh
+++ b/print_env.sh
@@ -5,8 +5,8 @@
 # Reports relevant environment information useful for diagnosing and
 # debugging Sirius issues.
 # Usage:
-# "./scripts/print_env.sh" - prints to stdout
-# "./scripts/print_env.sh > env.txt" - prints to file "env.txt"
+# "bash print_env.sh" - prints to stdout
+# "bash print_env.sh > env.txt" - prints to file "env.txt"
 
 print_env() {
 echo "***git***"
@@ -79,6 +79,7 @@ if type "rustc" &> /dev/null; then
 else
     echo "rustc not found"
 fi
+echo
 if type "cargo" &> /dev/null; then
     which cargo && cargo --version
 else
--- a/print_env.sh
+++ b/print_env.sh
@@ -8,6 +8,22 @@
 # "bash print_env.sh" - prints to stdout
 # "bash print_env.sh > env.txt" - prints to file "env.txt"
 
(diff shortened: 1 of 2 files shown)


print_env.sh
```diff
@@ -0,0 +1,147 @@
...
+# "bash print_env.sh > env.txt" - prints to file "env.txt"
+
+print_env() {
+echo "***git***"
+if [ "$(git rev-parse --is-inside-work-tree 2>/dev/null)" == "true" ]; then
+    git log --decorate -n 1
+    echo "***git submodules***"
+    git submodule status --recursive
+else
+    echo "Not inside a git repository"
```
9prady9, line 34: early exit ? so that we don't need to ask for this info later again in case the bug turns out to be related to submodule.

9prady9: nvm, I see why you didn't do such early exits. this fn is just collection and you are dumping that into html styled page later.

For some reason I think if this command runs without any error/warning messages when something is not found - which is most likely required for reliable functioning of sirius - some users might just upload the html content without verifying.

It would be better if we collecting critical vs optional checks info and print a one line warning/summary of what is missing as per checks after the html output.

aocsa (approved):
LGTM

print_env.sh
mike-wendt: @9prady9 thanks for the feedback, I've added your suggestion please take a look and let me know what you think

mike-wendt:
Will merge once @9prady9 has been able to re-review

felipeblazing (approved):

mike-wendt:
Merging and we can iterate based on the bug reports we git

9prady9:
> Will merge once @9prady9 has been able to re-review

sorry @mike-wendt I was out with sickness.

PR #1364, opened 2026-07-31
feat: [Concurrency] PR 1 of ? Update task_creator for concurrent query support

wmalpica:
This is PR 1 from a stack of PRs which together will make Sirius internals be able to handle Concurrent query execution. This PR only targets a part of task_creator. It is purposefully kept small to make it easier to review.

## Summary

  Refactor task_creator and associated task queues to support concurrent
  queries: per-query state is now isolated in a keyed map so that cleanup,
  draining, and teardown target exactly one query without touching any other
  in-flight query.

NOTE: this does not fully address how query teardown works for the task_creator. That will be handled later. Additionally the following constructs were added, but should be removed later and replaced by something better:
   std::mutex in_flight_mutex;
  std::condition_variable in_flight_cv;
  std::size_t in_flight{0};
  void enter_in_flight();
  void leave_in_flight();
  void wait_for_in_flight();

  ### Motivation

  Previously, task_creator held a single global execution context and a
  single global pipeline-state map. Cleanup on query end called
  interrupt() + drain() on the whole task-creation queue -- stalling every
  other query's producers and consumers -- and then cleared all global
  states in the map. This made it impossible to run more than one query
  concurrently through the same SiriusContext.

  ### What changed

  --- task_creator: per-query state ---

  - Introduced query_task_global_state: a per-query bundle holding the
    client context, pipeline global states, lookahead queue, and an
    in_flight counter (replaces _bounded_pool->wait_all() for per-query
    join).
  - set_client_context(query_id, ctx) now registers an entry in a keyed
    map instead of overwriting a single pointer.
  - prepare_for_query, schedule_lookahead, and drain_pending_tasks all
    look up the correct entry by query_id instead of touching shared
    singletons.
  - drain_pending_tasks(query_id) drops only that query's queued creation
    requests (via multi_index_priority_queue::drain(query_index)), leaving
    every other query's requests in the queue. The queue itself stays open
    -- no interrupt()/reactivate() pair.
  - reset(query_id) drains pending tasks, waits for in-flight lambdas to
    finish, then removes the entry from the map.
  - reset_all() iterates every registered entry and calls reset() on each;
    called from SiriusContext::terminate() as a safety net.

  --- sirius_pipeline: stamped with query id and priority ---

  - Added set_query_id/get_query_id and set_priority/get_priority to
    sirius_pipeline so that schedule() can derive queue index keys from
    the pipeline without taking the task_creator per-query lock on the
    hot path.
  - planner::query::build_indices stamps every pipeline with its query_id_t
    so the key is available before task_creator sees it.

  --- Shared index_keys_for extractor ---

  - Moved the key extraction lambda out of task_scheduler's constructor
    into a free function pipeline::index_keys_for (declared in
    gpu_pipeline_task.hpp, defined in gpu_pipeline_task.cpp).
  - Both task_scheduler's queue and every gpu_pipeline_executor's queue now
    use this single extractor, so a per-query drain(query_index{...})
    targets the same tasks in both queues.
  - The extractor reads query_id from pipe->get_query_id() rather than
    unpacking the high bits of the scheduling priority; the priority only
    preserves 31 bits of the query id (sirius::query_priority_bits), so
    the unpacked value diverges from the real id once bit 31 is set,
    causing silent drain misses.

  --- itask_executor: per-query drain ---

  - Switched the internal queue from inspectable_mpsc to
    multi_index_priority_queue (using index_keys_for).
  - Added drain_query_tasks(query_id) which drops only that query's queued
    tasks without touching others.

  --- task_scheduler: per-query drain ---

  - Added drain_query_tasks(query_id) which calls drain_query_tasks on
    every gpu_pipeline_executor and on the scheduler's own queue.
  - Called from SiriusContext::run_mandatory_cleanup after
    task_creator::reset (producer stopped first, then queued work dropped,
    then query_.reset() destroys the plan).

  --- SiriusContext cleanup ordering ---

  - run_mandatory_cleanup now calls task_creator_->reset(query_id) then
    task_scheduler_->drain_query_tasks(query_id) before query_.reset(),
    ensuring no queued task or in-flight lambda holds a raw operator pointer
    into the plan that is about to be destroyed.
  - The same ordered pair runs in drop_query_runtime_state_best_effort
    (the noexcept error backstop).
  - terminate() calls reset_all() as a final safety net before destroying
    task_creator.

--- multi_index_priority_queue ---

- push() now returns bool (true = enqueued, false = dropped because
  interrupted).

### Tests

- test_task_creator_query_state.cpp (new): lifecycle correctness for the
  per-query entry map -- two queries coexist, cleanup targets exactly one,
  reset of an unknown query is a no-op, double-reset is safe.
- test_task_index_keys.cpp (new): index_keys_for derives query_id from
  the pipeline rather than from the priority's high bits; tasks with
  bit-31 set in their query id still key correctly; non-pipeline tasks
  get sentinel keys.
- test_multi_index_priority_queue.cpp: added a case verifying that
  drain(query_index{...}) leaves the queue open for subsequent pushes
  (unlike interrupt()).
- test_task_creator.cpp: updated fixture to pass query_id through
  set_client_context.

### DIFF
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -95,13 +95,13 @@ const std::vector<int>& task_creator::get_active_gpu_ids() const noexcept
   return _active_gpu_ids;
 }
 
-void task_creator::query_task_state::enter_in_flight()
+void task_creator::query_task_global_state::enter_in_flight()
 {
   std::lock_guard<std::mutex> lock(in_flight_mutex);
   ++in_flight;
 }
 
-void task_creator::query_task_state::leave_in_flight()
+void task_creator::query_task_global_state::leave_in_flight()
 {
   {
     std::lock_guard<std::mutex> lock(in_flight_mutex);
@@ -111,18 +111,18 @@ void task_creator::query_task_state::leave_in_flight()
   in_flight_cv.notify_all();
 }
 
-void task_creator::query_task_state::wait_for_in_flight()
+void task_creator::query_task_global_state::wait_for_in_flight()
 {
   std::unique_lock<std::mutex> lock(in_flight_mutex);
   in_flight_cv.wait(lock, [this] { return in_flight == 0; });
 }
 
-std::shared_ptr<task_creator::query_task_state> task_creator::get_query_state(
+std::shared_ptr<task_creator::query_task_global_state> task_creator::get_query_task_global_state(
   sirius::query_id_t query_id) const
 {
   std::lock_guard<std::mutex> lock(_global_state_mutex);
-  auto it = _query_states.find(query_id);
-  return it == _query_states.end() ? nullptr : it->second;
+  auto it = _query_task_global_states.find(query_id);
+  return it == _query_task_global_states.end() ? nullptr : it->second;
 }
 
 void task_creator::set_client_context(sirius::query_id_t query_id,
@@ -131,8 +131,8 @@ void task_creator::set_client_context(sirius::query_id_t query_id,
   std::lock_guard<std::mutex> lock(_global_state_mutex);
   // The window begins before the plan exists, so this is where the query's entry is created;
   // prepare_for_query then fills in the pipeline global states.
-  auto& state = _query_states[query_id];
-  if (!state) { state = std::make_shared<query_task_state>(); }
+  auto& state = _query_task_global_states[query_id];
+  if (!state) { state = std::make_shared<query_task_global_state>(); }
   state->client_context = std::addressof(client_context);
 }
 
@@ -146,7 +146,7 @@ void task_creator::prepare_for_query(const sirius::planner::query& query)
 {
   const auto query_id = query.query_id();
 
-  auto state = get_query_state(query_id);
+  auto state = get_query_task_global_state(query_id);
   if (!state) {
     throw sirius::internal_exception(
       "task_creator::prepare_for_query: no state registered for query {}; "
@@ -275,7 +275,7 @@ void task_creator::drain_pending_tasks(sirius::query_id_t query_id)
   _task_creation_queue.drain(
     exec::query_index{static_cast<exec::query_key>(sirius::value_of(query_id))});
 
-  auto state = get_query_state(query_id);
+  auto state = get_query_task_global_state(query_id);
   if (!state) { return; }
 
   // Wait out this query's in-flight creation lambdas — the per-query stand-in for
@@ -296,13 +296,13 @@ void task_creator::reset(sirius::query_id_t query_id)
   // dereference them, so both must be gone before the caller destroys planner::query.
   drain_pending_tasks(query_id);
 
-  std::shared_ptr<query_task_state> state;
+  std::shared_ptr<query_task_global_state> state;
   {
     std::lock_guard<std::mutex> lock(_global_state_mutex);
-    auto it = _query_states.find(query_id);
-    if (it == _query_states.end()) { return; }
+    auto it = _query_task_global_states.find(query_id);
+    if (it == _query_task_global_states.end()) { return; }
     state = std::move(it->second);
-    _query_states.erase(it);
+    _query_task_global_states.erase(it);
   }
   // `state` is released here, outside the lock. Any worker that resolved it before the erase
   // still holds its own shared_ptr and finishes safely against a map that no longer lists it.
@@ -313,8 +313,8 @@ void task_creator::reset_all()
   std::vector<sirius::query_id_t> query_ids;
   {
     std::lock_guard<std::mutex> lock(_global_state_mutex);
-    query_ids.reserve(_query_states.size());
-    for (const auto& [query_id, state] : _query_states) {
+    query_ids.reserve(_query_task_global_states.size());
+    for (const auto& [query_id, state] : _query_task_global_states) {
       query_ids.push_back(query_id);
     }
   }
@@ -427,7 +427,7 @@ void task_creator::schedule_lookahead(sirius::query_id_t query_id,
                                       std::optional<int> device_id_hint)
 {
   if (_config.strategy != request_type::lookahead) { return; }
-  auto state = get_query_state(query_id);
+  auto state = get_query_task_global_state(query_id);
   if (!state) { return; }
 
   std::lock_guard lock(state->lookahead_mutex);
@@ -481,7 +481,7 @@ void task_creator::manager_loop()
     // then reads global_states (immutable after prepare_for_query) with no lock and no risk of
     // the map being rehashed or erased underneath it: its own reference keeps the state alive
     // even if reset(query_id) removes the registry entry mid-flight.
-    auto query_state = get_query_state(query_id);
+    auto query_state = get_query_task_global_state(query_id);
     if (!query_state) {
       // The query was reset while this request sat in the queue (finished, or failed and was
       // cleaned up). Dropping it is the correct outcome — `node` may already be dangling.
@@ -515,7 +515,7 @@ void task_creator::manager_loop()
         // Released on every exit path, including the catch below, so a throwing creation can
         // never strand drain_pending_tasks() waiting forever.
         struct in_flight_guard {
-          const std::shared_ptr<query_task_state>& state;
+          const std::shared_ptr<query_task_global_state>& state;
           ~in_flight_guard() { state->leave_in_flight(); }
         } guard{query_state};
         try {
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -350,12 +350,13 @@ op::sirius_physical_operator* task_creator::get_operator_for_next_task(
 void task_creator::stop()
 {
   _task_creation_queue.interrupt();
+  std::lock_guard<std::mutex> lock(_thread_pool_mutex);
   do_stop_thread_pool();
 }
 
 void task_creator::start_thread_pool()
 {
-  std::lock_guard<std::mutex> lock(_global_state_mutex);
+  std::lock_guard<std::mutex> lock(_thread_pool_mutex);
   bool expected = false;
   if (!_running.compare_exchange_strong(expected, true)) { return; }
   // Re-arm the request queue. stop_thread_pool() calls
@@ -384,7 +385,7 @@ void task_creator::do_stop_thread_pool()
 
 void task_creator::stop_thread_pool()
 {
-  std::lock_guard<std::mutex> lock(_global_state_mutex);
+  std::lock_guard<std::mutex> lock(_thread_pool_mutex);
   do_stop_thread_pool();
 }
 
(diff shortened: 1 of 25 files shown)


src/creator/task_creator.cpp
```diff
@@ -214,43 +269,58 @@ task_creator::compute_pipeline_priorities(const sirius::planner::query& query) c
...
+  // the whole time, so other queries' producers and consumers are never stalled.
+  _task_creation_queue.drain(
+    exec::query_index{static_cast<exec::query_key>(sirius::value_of(query_id))});
+
+  auto state = get_query_task_global_state(query_id);
+  if (!state) { return; }
+
+  // Wait out this query's in-flight creation lambdas — the per-query stand-in for
+  // _bounded_pool->wait_all(), which would also wait on every other query's work. Workers
+  // decrement on exit (including by exception), so this cannot hang on a throwing lambda.
```
wmalpica, line 284:
```suggestion
  // Wait out this query's in-flight creation lambdas. Workers decrement 
  // on exit (including by exception), so this cannot hang on a throwing lambda.
```

ran-yuan-rui:
Please confirm the cross-query priority contract as bleow:

`query_priority_bits` puts query order above pipeline rank and documents the `2^31` ordering wrap. Once the single-flight gate is removed and multiple queries can occupy these queues concurrently, two points need a contract:

1. Within a device queue, a query that continually keeps work queued is always selected ahead of newer queries. Is strict oldest-first the intended policy, or will a later PR add fair sharing?
2. The documented wrap affects more than ordering once `drain(query)` is used. `qid=1` and `qid=0x80000001` have the same query priority bits. At the same pipeline rank, they produce the same level key. `push()` reuses that level but indexes it only under the first query ID. `drain(0x80000001)` then leaves the collided task queued, while `drain(1)` drops the shared level, including both queries’ tasks. But maybe this is by design , try to confrim

The current single-flight path makes this collision unreachable, and it remains operationally remote under concurrency, so it need not block this PR. At minimum, reject a push when an existing level has different query or operator keys. Longer term, include the full query ID in the level identity?

wmalpica:
> Please confirm the cross-query priority contract as bleow:
> 
> `query_priority_bits` puts query order above pipeline rank and documents the `2^31` ordering wrap. Once the single-flight gate is removed and multiple queries can occupy these queues concurrently, two points need a contract:
> 
> 1. Within a device queue, a query that continually keeps work queued is always selected ahead of newer queries. Is strict oldest-first the intended policy, or will a later PR add fair sharing?
> 2. The documented wrap affects more than ordering once `drain(query)` is used. `qid=1` and `qid=0x80000001` have the same query priority bits. At the same pipeline rank, they produce the same level key. `push()` reuses that level but indexes it only under the first query ID. `drain(0x80000001)` then leaves the collided task queued, while `drain(1)` drops the shared level, including both queries’ tasks. But maybe this is by design , try to confrim
> 
> The current single-flight path makes this collision unreachable, and it remains operationally remote under concurrency, so it need not block this PR. At minimum, reject a push when an existing level has different query or operator keys. Longer term, include the full query ID in the level identity?

Yes, strict oldest-first the intended policy is the indented policy for now. We can layer on some fair sharing policies later.

If we have two queries `qid=1` and `qid=0x80000001`  running at the same time, yes that can cause problems. But if we assume reasonably fast queries running in 0.1 sec. It will take almost 7 years of back to back queries to reach qid=0x80000001. Which is why i dont think we need to worry about adding code to handle this.

src/creator/task_creator.cpp
```diff
@@ -365,12 +488,23 @@ void task_creator::manager_loop()
...
+    auto request_kind   = request->type;
+    const auto query_id = request->query_id;
     if (node == nullptr) { continue; }
 
-    std::vector<duckdb::shared_ptr<pipeline::sirius_pipeline>> visited_pipelines;
+    // Resolve the query's state ONCE, here, and hand the shared_ptr to the worker. The worker
+    // then reads global_states (immutable after prepare_for_query) with no lock and no risk of
+    // the map being rehashed or erased underneath it: its own reference keeps the state alive
+    // even if reset(query_id) removes the registry entry mid-flight.
+    auto query_state = get_query_task_global_state(query_id);
```
kevkrist, line 508: This locks _global_state_mutex, or tries to, while the query thread might itself be holding the mutex (on, e.g., encountering an error) and invoking do_stop_thread_pool() and then waits on the manager thread here -> deadlock.

Maybe separate mutexes should be used, one for task creator thread pool and another for _global_state_mutex for the query-state map?

src/sirius_context.cpp
```diff
@@ -390,6 +391,18 @@ void SiriusContext::run_mandatory_cleanup(sirius::query_id_t query_id, std::stri
...
+  // pointers into the plan that query_.reset() below destroys, and in-flight creation lambdas
+  // dereference them. reset() drains those requests and joins that work before returning.
+  // Only this query's is touched; other in-flight queries keep creating tasks.
+  if (task_creator_) { task_creator_->reset(query_id); }
+
+  // With the producer stopped, drop whatever it already queued for this query. Ordering is
+  // deliberate: task_creator first so nothing new arrives, then the queues, then query_.reset()
+  // below — queued tasks reach operators through their pipeline, so they must be gone before the
+  // plan is destroyed. Other in-flight queries keep their queued work.
+  if (task_scheduler_) { task_scheduler_->drain_query_tasks(query_id); }
```
kevkrist, line 420: Is it possible that this drains the query's tasks, but a task has already been stolen by, e.g., the downgrade executor, which, before it is itself drained, returns the extracted task to the scheduler queue, which is still running and has no way to know that the query to which it belongs is in process of termination, so executes the task, which now has dangling data repository pointers? It seems that maybe there needs to be a way of marking a query is CLOSING and acknowledging this in the task queue consumers.

src/pipeline/sirius_pipeline.cpp
```diff
@@ -357,11 +357,17 @@ void sirius_pipeline::notify_downstream_pipelines(bool original_pipeline)
...
+      for (auto* consumer : get_output_consumers()) {
+        // If is possible to have a race condition here where one task finished and here it does to
+        // schedule a task right when the last task finished and marks the operator as finalized.
+        // That is ok. This check here is to minimize unnecessary scheduling of task creation.
+        if (!consumer->finalized.load()) { _task_creator->schedule(consumer); }
+      }
+    } catch (const std::exception& e) {
+      SIRIUS_LOG_ERROR(
+        "Pipeline {}: failed to schedule downstream consumers: {}", pipeline_id, e.what());
+      _task_creator->report_fatal_error(std::current_exception());
```
kevkrist, line 376: I think this gets called by a GPU worker thread (?), which eventually calls bounded_pool::wait_all() -> self-wait deadlock. It seems that you should request cancellation here rather than do it, and let the query thread observe the error and perform the cancellation.

kevkrist (changes requested):
[fistbump]

src/sirius_context.cpp
wmalpica: You are correct. There is a followup PR which implements something like you are describing. Queries will have a state, and all the various "executors" will check the query state before running tasks or before adding data to a repository

src/creator/task_creator.cpp
wmalpica: Good catch. Fixed

src/pipeline/sirius_pipeline.cpp
wmalpica: good catch. Fixed

src/sirius_context.cpp
ran-yuan-rui: The follow-up needs to cover work already removed from a queue, not just late enqueue attempts. There are two separate windows here.

**Creator manager.** In `task_creator.cpp`, the manager resolves the query state at line 508, dereferences operators in the hint walk at line 516 and the visited-pipeline update at line 528, and only enters `in_flight` at line 535. `reset()` can drain the queue, observe zero, erase the state, and return while that manager still holds the request. Its `shared_ptr` keeps the state alive, not the operator plan.

The current success/error paths stop and join the whole creator before `execute()` returns (`task_scheduler.cpp:275/236`). That covers requests already held at that stop, including one missed by the GPU worker's earlier drain at `gpu_pipeline_executor.cpp:477`. It does not make `reset()` a per-query lifetime barrier or cover work admitted after the creator restarts.

For that barrier, resolve the state and acquire an in-flight lease under the same `_global_state_mutex`. Transfer an owning, movable lease from the manager into the submitted callable so the hint walk, early exits, and dispatch failure are covered. `reset()` should unpublish under that lock, then drain and wait using the detached state. A manager that resolved before the erase is then counted; one arriving after it cannot dereference the operator.

**Scheduler manager.** The matcher holds a task locally between `task_scheduler.cpp:381/385` and `:410`. Neither queue drain at `:240/259` can see it, and `drain_after_error()` does not pause the matcher. Meanwhile, `drain_and_wait()` reopens each executor's queue and restarts its manager (`task_executor.cpp:135-137`):

```text
matcher:       pop T; pause before executor.schedule(T)
query thread:  drain scheduler queue; drain/restart executors;
               drain scheduler queue again; restart creator
matcher:       submit T to the restarted executor
```

T can now run after cleanup. If it fails after the creator restart, the catch at `gpu_pipeline_executor.cpp:425/430` stops the creator again; a subsequent GPU query can then hang because its initial creation request is dropped. Downgrade returns provide another source of late work: the wrapper destructor pushes directly back to the scheduler queue, and downgrade executors are not drained until `sirius_context.cpp:427`.

At #1369 `5da74a56`, the creator's pop-to-first-increment section is unchanged. The new creator, executor, and downgrade enqueue checks reject work observed while quiescing, but do not wait for work already in transit. The registry's `close()` documentation explicitly assigns that responsibility to the drains.

Can the follow-up cover both windows? Scheduler teardown needs to prevent new dispatch leases, wait for existing transfers, and then drain executor work, with late returns rejected for the closing query. A deterministic pause-after-pop test would distinguish that barrier from another queue drain.

src/creator/task_creator.cpp
```diff
@@ -388,209 +530,226 @@ void task_creator::manager_loop()
...
+                bool const names_excluded_device =
+                  preferred_device_id.has_value() &&
+                  std::find(_active_gpu_ids.begin(), _active_gpu_ids.end(), *preferred_device_id) ==
+                    _active_gpu_ids.end();
+                bool const unpinned_on_a_subset =
+                  !preferred_device_id.has_value() && _active_gpu_ids.size() < _full_gpu_count;
+                if (names_excluded_device || unpinned_on_a_subset) {
+                  auto const idx      = _admission_rr.fetch_add(1) % _active_gpu_ids.size();
+                  preferred_device_id = _active_gpu_ids[idx];
+                }
```
ran-yuan-rui, line 723: These fields are query-specific but still live on the shared creator. `sirius_engine::initialize_internal()` overwrites `_active_gpu_ids` and `_full_gpu_count` for each query, while creation workers read them here and in the partition-affinity branch without synchronization.

Normal execution is still serialized by the execution-window slot. Once two queries overlap, B's vector move-assignment races A's routing reads. Synchronizing access alone would not fix the isolation problem: A could still use B's admitted subset. For example, if A uses `[0,1]` and builds partition 1's hash table on GPU 1, then B replaces the list with `[0]`, A's next task for partition 1 routes to GPU 0 while its hash table remains on GPU 1. The shared list can also disagree with the query-local device-to-slot mapping used by broadcast partitioning, and the clamp can select a GPU outside A's admission.

Could both fields move into `query_task_global_state`, initialized once and read through the `query_state` already captured by each worker? The entry exists before admission runs: window begin registers it through `set_client_context()`. Initializing the admission snapshot there and keeping it immutable after `prepare_for_query()` fits the existing lifecycle.

`create_query()` also reads the subset through the id-less getter for `scan_manager` (`sirius_context.cpp:1119-1123`); that needs the same query-specific snapshot, either by query ID or passed explicitly. Should this move land here or in the follow-up before concurrent execution is enabled?

kevkrist (approved):

src/creator/task_creator.cpp
wmalpica: This is a good point. I will address this in a followup PR

wmalpica: I implemented your suggestion here https://github.com/sirius-db/sirius/pull/1365

PR #1365, opened 2026-07-31
feat: [Concurrency]  PR 2 of ?. Move completion_handler ownership from task_scheduler to sirius_engine; Removed SiriusContext::query_

wmalpica:
This is PR 2 from a stack of PRs (it goes on top of https://github.com/sirius-db/sirius/pull/1364)  which together will make Sirius internals be able to handle Concurrent query execution. There is more to come.

## Summary

This PR addresses:
- the `completion_handler`, creating one per query
- sets `active_gpu_ids` and `full_gpu_count` as part of the `_query_task_global_states`, instead of just being set globally for `task_creator`.
- deletes `SiriusContext::query_`, modifying code as needed
- deletes `_no_pref_rr_counter` (DEAD CODE)

This PR does not fully address everything including full query winddown.

### 1. Per-query completion handler (main change)

Previously, task_scheduler owned a single completion_handler and pushed a
raw pointer to it into every gpu_pipeline_executor. 

The handler is now a shared_ptr<completion_handler> created in
sirius_engine::execute() and propagated down to every pipeline's
`gpu_pipeline_task_global_state` via task_creator::prepare_for_query().
Each executor site reads the handler from the task it is currently
executing, so error reports and completion signals are always scoped to
the right query.

The shared_ptr lifetime is load-bearing: sirius_engine is destroyed in
sirius_interface::cleanup_internal() before run_mandatory_cleanup() drains
the task queues. A task still unwinding after
report without touching freed memory. Because
shared_ptr reference through its global state, the handler stays alive
until the last task releases it.

Changed interfaces:
  - task_scheduler: prepare_for_query() + start_query() -> start_query(query)
    (no return value; caller holds the future
  - task_scheduler: terminate_query(error) -> terminate_query(handler, error)
  - task_scheduler: drain_after_error() -> drain_after_error(query_id)
  - task_scheduler: wait_for_completion() ->
  - gpu_pipeline_executor: set_completion_handler() removed
  - SiriusContext::create_query() now returns the query (ownership to caller)
  - SiriusContext::get_query() removed
  - task_creator::prepare_for_query() now takr
  - task_creator::schedule_lookahead() no longer takes a query_id

### 2. Remove SiriusContext::query_

The query object (sirius::planner::query) was stored as SiriusContext::query_.
It is an index over the plan that sirius_engine owns, so storing it in a
shared context while the plan lived in the engine created an ownership
mismatch. The query is now owned by sirius_engine alongside the plan it
indexes.

The corresponding get_query() accessors on SiriusContext are removed; the
one call site in sirius_engine::execute() now reads from the engine's own
query_ field.

### 3. Remove _no_pref_rr_counter (DEAD CODE)

The round-robin counter (_no_pref_rr_counter and its reset in
prepare_for_query, plus set_no_pref_rr_counter_for_testing()) has been
removed.  It was dead code.

## Files changed

  src/sirius_engine.{hpp,cpp}         -- owns query_ and completion_handler_
  src/include/sirius_context.hpp       -- removed query_, get_query()
  src/sirius_context.cpp               -- cre
  src/include/pipeline/task_scheduler.hpp
  src/pipeline/task_scheduler.cpp      -- collapsed prepare+start, removed counter
  src/include/pipeline/gpu_pipeline_executor.hpp
  src/pipeline/gpu_pipeline_executor.cpp -- rield
  src/include/pipeline/sirius_pipeline_task_states.hpp
  src/include/pipeline/gpu_pipeline_task.hpp -- get_completion_handler()
  src/include/creator/task_creator.hpp
  src/creator/task_creator.cpp         -- takes handler in prepare_for_query
  docs/super-sirius/pipeline-execution.md -- updated for pull-signal dispatch
  test/cpp/pipeline/test_per_query_completion
  test/cpp/pipeline/test_oom_reschedule.cpp -r
  test/cpp/operator/test_mgpu_stress.cpp   -- removes counter injection

## New tests (test_per_query_completion_handl

  - completing one query leaves another query's future unset
  - one query's failure does not poison another's handler
  - the handler outlives the engine-side refe
  - every pipeline of one query shares its handler
  - a global state built without a query carries no handler (null-guard)

### DIFF
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -24,6 +24,7 @@
 #include "pipeline/task_scheduler.hpp"
 #include "planner/query.hpp"
 #include "planner/query_index.hpp"
+#include "sirius/exception.hpp"
 #include "sirius_context.hpp"
 
 #include <cucascade/cudf/gpu_data_representation.hpp>
@@ -390,14 +391,23 @@ void task_creator::stop_thread_pool()
 namespace {
 
 //! The query and scheduling priority of the pipeline `node` belongs to.
-//! An unplaced operator (no pipeline yet) yields query 0 / priority 0; such a request is
-//! droppable and will simply find no state when the worker resolves it.
 std::pair<sirius::query_id_t, exec::queue_priority> request_keys_for(
   const op::sirius_physical_operator* node)
 {
-  if (node == nullptr) { return {sirius::make_query_id(0), 0}; }
+  if (node == nullptr) {
+    throw sirius::internal_exception("task_creator::schedule: null operator");
+  }
   const auto pipe = node->get_pipeline();
-  if (!pipe) { return {sirius::make_query_id(0), 0}; }
+  if (!pipe) {
+    throw sirius::internal_exception(
+      "task_creator::schedule: operator {} (id {}) has no pipeline; it was never placed by "
+      "planner::query::build_indices",
+      node->get_name(),
+      // get_operator_id() itself throws on an unnumbered operator; report the sentinel instead so
+      // the pipeline-less diagnosis is not masked by an id-assignment one.
+      node->has_operator_id() ? std::to_string(node->get_operator_id())
+                              : std::string{"unassigned"});
+  }
   return {pipe->get_query_id(), pipe->get_priority()};
 }
 
@@ -421,6 +431,12 @@ void task_creator::schedule(op::sirius_physical_operator* node, sirius::query_id
   request->query_id        = query_id;
   request->priority        = priority;
   _task_creation_queue.push(std::move(request));
+}
+
+void task_creator::report_fatal_error(std::exception_ptr error)
+{
+  if (_task_scheduler != nullptr) { _task_scheduler->terminate_query(std::move(error)); }
+  stop();
 }
 
 void task_creator::schedule_lookahead(sirius::query_id_t query_id,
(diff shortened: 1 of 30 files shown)


src/creator/task_creator.cpp
```diff
@@ -70,30 +70,30 @@ task_creator::task_creator(task_creator_config config,
   // numa_node -1 ("unknown", per the Linux /sys/bus/pci/devices/*/numa_node
   // convention on non-NUMA / single-NUMA hosts) is the index's grouping key and
   // is queried verbatim at routing time.
-  //
-  // Materialize the active executor set sorted+deduped (topology_index preserves
-  // manager order, not sorted order) so partition affinity below stays inverse
-  // to sirius_physical_partition's device->slot mapping.
-  if (_topology_index) {
-    auto ids        = _topology_index->gpu_ids();
-    _active_gpu_ids = std::vector<int>(ids.begin(), ids.end());
```
wmalpica, line 79: Setting _active_gpu_ids here was never really necessary. In a query they were really being set  from sirius_ctx_ptr->get_memory_manager().get_memory_spaces_for_tier(Tier::GPU)

src/creator/task_creator.cpp
Jedi18, line 174: Could we also install handler on query-terminal pipelines here? 

`update_pipeline_status()` still uses the pipeline’s weak completion handler for finish transitions that occur outside the GPU epilogue for example, when source exhaustion is observed after the last task completes. The handler is currently attached only to gs, so that path gets an empty weak pointer and `sirius_engine::execute()` can remain blocked in `future.get()`.

```suggestion
    pipeline->set_task_creator(this);
    if (pipeline->is_query_terminal()) {
      pipeline->set_completion_handler(handler);
    }
```

Jedi18 (changes requested):

kevkrist:
lgtm aside from @Jedi18's issue.

kevkrist (approved):

Jedi18 (approved):

PR #1366, opened 2026-07-31
feat: [Concurrency]  PR 3 of ?. Per-query scan manager state (query_scan_manager_state)

wmalpica:
This is PR 3 from a stack of PRs (it goes on top of https://github.com/sirius-db/sirius/pull/1364 and https://github.com/sirius-db/sirius/pull/1365) which together will make Sirius internals be able to handle Concurrent query execution. There is more to come.

## Summary 
This PR addresses the scan_manager. It primarily takes all its internal structures that are query specific and puts them into a query_scan_manager_state. Then the scan_manager can hold a map of these states. 

## Background

The scan manager used to hold one query's worth of state in bare members
(_providers_by_op, _scan_op_order, _metadata_processor, _dispatcher,
_pending_mvcc_mask_jobs, _pending_insert_delta_jobs). reset() wiped all of
them. This was fine for a single serial query but breaks under concurrent
queries

### 1. query_scan_manager_state

All per-query state is now grouped in a new inner struct,
query_scan_manager_state, held as a shared_ptr in a map keyed by query_id:

    std::map<query_id_t, std::shared_ptr<query_scan_manager_state>> _query_states;
    mutable std::mutex _query_states_mutex;

State is published to the map AFTER prepare_for_query fully builds it, so
a failed prepare never leaves a half-built entry visible. The shared_ptr is
resolved under the lock and used outside it, so an erase racing a reader
cannot pull the state out from under the reader.

### 2. reset(query_id) / reset_all()

reset() now takes a query_id and only tears down that query's dispatcher,
coalescer, and providers. Other queries' state is not touched.

reset_all() is the teardown-only path used by stop() and the failed-query
backstop.

sirius_context::run_mandatory_cleanup now calls reset(query_id) instead of
reset(). sirius_context::drop_query_runtime_state_best_effort (the error
backstop) now also calls reset(query_id): prepare_for_query no longer
performs a global reset.

### 3. Shared-ownership pin table

_pinned_entries changed from:
    std::unordered_map<std::string, pinned_entry>
to:
    std::unordered_map<std::string, std::shared_ptr<pinned_entry>>

A query that matched a pinned entry now co-owns it for its whole duration
through the shared_ptr the cached_databatch_provider holds. A concurrent
UNPIN on another connection only drops the map slot; the data stays alive
until the last serving provider releases it.

find_pinned_entry_for_duckdb_table changed from returning a raw
pinned_entry const* to a shared_ptr<const pinned_entry> (owning), so
plan-time MVCC guards that read the entry hold it alive across the check.

make_provider_for_pinned_entry signature changed from
pinned_entry const& to std::shared_ptr<pinned_entry const> for the same
reason.

All pin-table mutations (insert, attach_mvcc_metadata, remove,
visit_pinned_entries) are now guarded by _pinned_entries_mutex.

try_match_cached_entry snapshots the pin table under _pinned_entries_mutex
and matches outside the lock: the match path (can_serve_with_columns,
validate, zone-map plan building, MVCC branch) is too slow to hold the
lock across.

### 4. Thread pool sizing

The pool is now sized num_threads + k_max_concurrent_queries (was
num_threads + 1). Each query's coalescer parks exactly one sequencer task
in queue.wait_dequeue, only unblocked by that query's own split_provider
tasks running on the same pool. With Q concurrent queries, Q threads are
parked in sequencers at all times, so the old +1 sizing would deadlock at
2 concurrent queries. k_max_concurrent_queries is left at 1 (matching the
old behavior for single-query runs) with a TODO to promote it to a real
config option.

### 5. Known gap documented

The prefetch cache epoch is a global counter (prefetching_cache::_ticker),
so a second query starting bumps the epoch and demotes the first query's
prefetched-but-unconsumed chunks to tier-0 in the eviction order.
Performance only (not correctness: a live reader holds pin > 0, blocking
mark_evicting). The fix belongs in prefetching_cache, not here. A comment
now calls this out clearly.

## Changed interfaces

  sirius_scan_manager::reset()     -> reset(query_id) + reset_all()
  sirius_scan_manager::find_pinned_entry_for_duckdb_table()
                                   raw pointer -> shared_ptr (owning)
  make_provider_for_pinned_entry() pinned_entry const& -> shared_ptr
  SiriusContext::run_mandatory_cleanup
                                   scan_manager_->reset() -> reset(query_id)
  drop_query_runtime_state_best_effort: scan_manager_->reset(query_id) added

## New test: test_scan_manager_query_state.cpp

  - two queries register independently and reset drops only one
      Pins that tearing A's dispatcher down does not kill B's sequencer,
      and B's split connector still yields splits after A resets.
  - concurrent queries do not collide on operator id
      Both queries get operator id 0. Asserts each operator drains splits
      for its own file only.
  - reset is a no-op for unknown and already-reset queries
      The failure backstop calls reset(query_id) unconditionally on every
      code path; an unknown id must not throw.
  - reset_all drops every query and stop() still returns
      A sequencer parked on a dequeue with nothing to feed it would hang
      stop() indefinitely.
  - a query with no GPU scan operators registers nothing
      The matching reset is a harmless no-op rather than touching the map.

## Files changed

  src/include/scan_manager/sirius_scan_manager.hpp   main API changes
  src/scan_manager/sirius_scan_manager.cpp           implementation
  src/sirius_context.cpp                             per-query reset calls
  src/planner/sirius_plan_get.cpp                    owning shared_ptr
  test/cpp/scan_manager/test_scan_manager_query_state.cpp  (new)
  test/cpp/scan_manager/test_cached_serving_hardening.cpp  shared_ptr entry
  test/cpp/scan_manager/test_s3_routing_cutover.cpp  distinct query ids
  CMakeLists.txt

### DIFF
--- a/src/scan_manager/sirius_scan_manager.cpp
+++ b/src/scan_manager/sirius_scan_manager.cpp
@@ -95,7 +95,17 @@ cudf::data_type host_column_carrier(cucascade::memory::column_metadata const& me
 }
 
 struct cached_databatch_provider : public databatch_provider {
-  cached_databatch_provider(pinned_entry const& entry,
+  /// Dereference the owned entry, refusing a null. The reference member below binds to the
+  /// result, so a null here would be an unnoticed UB rather than a loud failure.
+  static pinned_entry const& deref_or_throw(std::shared_ptr<pinned_entry const> const& entry)
+  {
+    if (!entry) {
+      throw std::invalid_argument("[cached_databatch_provider] pinned entry must be non-null");
+    }
+    return *entry;
+  }
+
+  cached_databatch_provider(std::shared_ptr<pinned_entry const> entry,
                             std::span<std::size_t const> selected_columns,
                             cached_scan_plan plan,
                             const telemetry::batch_telemetry_info& telemetry_info,
@@ -105,7 +115,8 @@ struct cached_databatch_provider : public databatch_provider {
                             bool has_physical_overrides,
                             sirius::decode_equality_pushdown equality_pushdown)
     : _plan(std::move(plan)),
-      _entry(entry),
+      _entry_owner(std::move(entry)),
+      _entry(deref_or_throw(_entry_owner)),
       _telemetry_info(telemetry_info),
       _mvcc_masks(std::move(mvcc_masks)),
       _delta_splits(std::move(delta_splits)),
@@ -349,11 +360,10 @@ struct cached_databatch_provider : public databatch_provider {
   cached_scan_plan _plan;
   std::vector<std::string> _column_names;
   std::vector<size_t> _column_indices;
-  /// Parallel to @c _column_indices; empty entries decompress normally. Only the
-  /// GPU-tier compressed path consumes it (the host path would have to parse the
-  /// chunk header to know whether the plan can exploit it, and that tier is not
-  /// where the decode gather costs anything).
-  sirius::decode_equality_pushdown _equality_pushdown;
+  /// Shared ownership of the served entry, keeping it alive for this provider's whole life.
+  /// A concurrent unpin drops the scan manager's map slot but not the data underneath us.
+  /// MUST stay declared before _entry, which binds to it.
+  std::shared_ptr<pinned_entry const> _entry_owner;
   const pinned_entry& _entry;
   telemetry::batch_telemetry_info _telemetry_info;
   /// This provider's own copy of the entry's per-chunk keep-masks (empty for
@@ -369,6 +379,11 @@ struct cached_databatch_provider : public databatch_provider {
   // Whether the scan carries an explicit plan sidecar. Normalization narrows a stored carrier
   // only with one installed, so it belongs in the conversion predicate.
   bool _has_physical_overrides{false};
+  /// Parallel to @c _column_indices; empty entries decompress normally. Only the
+  /// GPU-tier compressed path consumes it (the host path would have to parse the
+  /// chunk header to know whether the plan can exploit it, and that tier is not
+  /// where the decode gather costs anything).
+  sirius::decode_equality_pushdown _equality_pushdown;
   std::atomic<std::size_t> _index{0};
 };
 
@@ -442,11 +457,15 @@ sirius_scan_manager::sirius_scan_manager(
   : _config(config),
     _reservation_manager(reservation_manager),
     _topology_index(std::move(topology_index)),
-    _thread_pool(_config.thread_pool.num_threads + 1,
+    // num_threads + k_max_concurrent_queries, not num_threads + 1: each query's coalescer
+    // sequencer task BLOCKS (worker_loop -> process_provider_inputs -> queue.wait_dequeue)
+    // and is unblocked only by that query's own split_provider tasks, which run on this same
+    // pool. With Q concurrent queries, Q threads are parked in sequencers at all times, so
+    // sizing for a single sequencer would let Q queries consume the entire working budget
+    // and deadlock by starvation. See k_max_concurrent_queries for the bound and its TODO.
+    _thread_pool(_config.thread_pool.num_threads + k_max_concurrent_queries,
                  _config.thread_pool.thread_name_prefix,
                  _config.thread_pool.cpu_affinity_list),
-    _dispatcher(
-      std::make_unique<exec::scoped_dispatcher>(_thread_pool, _thread_pool.num_threads())),
     _ioctx_registry(config, reservation_manager)
 {
   if (!_topology_index) {
@@ -573,9 +592,32 @@ void sirius_scan_manager::prepare_for_query(const sirius::planner::query& query,
                                             bool enable_pinned_zone_map_pruning,
                                             const std::vector<int>& allocated_gpu_ids)
 {
-  _pruning_enabled = enable_pinned_zone_map_pruning;
-  reset();
+  auto const query_id = query.query_id();
 
+  // No global reset here. It used to be how the previous query's state was dropped; with
+  // concurrent queries that would tear down whatever is still running. Each query's state is
+  // dropped by its own reset(query_id) at window cleanup.
+  //
+  // A duplicate id means the previous window's cleanup never ran (a failed query on a
+  // latched-unavailable runtime). Drain and replace it rather than leaking a live dispatcher.
+  if (get_query_state(query_id) != nullptr) {
+    SIRIUS_LOG_WARN(
+      "[sirius_scan_manager::prepare_for_query] query {} was already registered; its cleanup "
+      "never ran. Draining and replacing the stale state.",
+      query_id);
+    reset(query_id);
+  }
+
+  // KNOWN GAP under concurrent queries: the prefetch cache's query epoch is a single GLOBAL
+  // generation counter (prefetching_cache::_ticker, bumped in
+  // prefetching_cache::prepare_for_query, src/io/cache/prefetching_cache.cpp).
+  // chunk_lifecycle::eviction_tier(query_tick) (src/include/io/cache/types.hpp) scores every
+  // chunk whose tick is older than the newest as tier 0 — evict first — so a second query
+  // starting here demotes all of the first query's prefetched-but-unconsumed chunks to the
+  // front of the eviction order. Performance only, never correctness: mark_evicting()
+  // succeeds only at pin == 0, so a chunk a live reader holds cannot be reclaimed; the query
+  // just re-reads on a miss. The fix belongs in prefetching_cache (track the set of live
+  // epochs rather than newest-wins), not here.
   if (_io_ctx && _io_ctx->cache()) {
     SIRIUS_LOG_INFO("[sirius_scan_manager] cache summary: {}", _io_ctx->cache()->summary());
     _io_ctx->cache()->prepare_for_query(query);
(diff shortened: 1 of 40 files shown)


src/sirius_extension.cpp
```diff
@@ -1885,8 +1885,12 @@ static void SiriusCreateAnnIndexFunction(ClientContext& context,
   rmm::cuda_set_device_raii device_guard{rmm::cuda_device_id{target_gpu}};
 
   auto& scan_mgr = sirius_ctx->get_scan_manager();
-  const auto* pin =
+  // OWNING: chunk_views below holds raw column views straight into pin's data for the whole
+  // build, so pin_owner must outlive it — a concurrent unpin on another connection must not
+  // invalidate the entry mid-build.
+  auto pin_owner =
     scan_mgr.find_pinned_entry_for_duckdb_table(entry_catalog, entry_schema, entry.name);
+  const auto* pin = pin_owner.get();
```
dhruv9vats: nit: explicit types for `pin_owner` `pin` and `chunk_views` instead of `auto` will help readability.

src/scan_manager/sirius_scan_manager.hpp
```diff
@@ -387,8 +388,13 @@ struct cached_scan_plan {
...
 /// itself (set_visibility_mask), so the decode compacts to visible survivors; a default
 /// slot gets the pushdown alone.
+///
+/// The provider CO-OWNS @p entry for its whole life: it reads the entry's chunks on every
+/// get_next_batch, and with concurrent queries another connection may unpin (or replace)
+/// the entry mid-scan. Shared ownership is what keeps that from dangling — an unpin drops
+/// the scan manager's map slot, and the data dies only once the last serving provider does.
 std::unique_ptr<databatch_provider> make_provider_for_pinned_entry(
-  pinned_entry const& entry,
+  std::shared_ptr<pinned_entry const> entry,
```
dhruv9vats: ❤️ the shared_ptr to ***const*** pinned_entry

src/scan_manager/sirius_scan_manager.hpp
```diff
@@ -724,17 +767,96 @@ class sirius_scan_manager {
...
+  /**
+   * @brief Everything the scan manager holds on behalf of ONE query.
+   *
+   * Handed out as a `shared_ptr` keyed by query id: a caller resolves it under the state
+   * mutex and uses it outside, so an erase racing a reader cannot pull the state out from
+   * under them. The manager's shared members (ioctxs, registry, pinned entries, thread pool)
+   * stay outside — they are process-wide and outlive every query. What lives here is exactly
+   * what the old global `reset()` used to wipe, and wiping it globally is what made
+   * finishing query A tear down query B's scan work.
+   *
```
dhruv9vats:
```suggestion
   * stay outside — they are process-wide and outlive every query.
```

Now that we are updating the implementation, do we want to keep documentation that references what the prior behaviour was? Describing just the current updated behaviour should be fine ?

src/scan_manager/sirius_scan_manager.hpp
```diff
@@ -724,17 +767,96 @@ class sirius_scan_manager {
...
+   * Operator ids restart at 0 for every query, so `metadata_processor`'s slot map (keyed by
+   * `scan_op->get_operator_id()`) is unique only *within* an entry; one shared coalescer
+   * would let two queries' scans collide on the same slot.
+   *
+   * Member order is teardown order: `dispatcher` is declared LAST so it is destroyed FIRST,
+   * while the coalescer and providers its tasks captured are still alive. The destructor's
+   * `drain()` makes that safe even if a member is later added below it.
+   */
```
dhruv9vats: >`dispatcher` is declared LAST

this no longer seems to be the case. But either way since we do have an explicit destructor, it should be OK to not mention this ?

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -1710,27 +1781,50 @@ void sirius_scan_manager::prepare_for_query(const sirius::planner::query& query,
...
-  _pending_insert_delta_jobs.clear();
+  state->pending_mvcc_mask_jobs.clear();
+  state->pending_insert_delta_jobs.clear();
 
-  start_metadata_processing();
+  // Publish only once the state is fully built: a prepare that threw above destroys `state`
+  // locally and never leaves a half-built entry visible to reset() or a concurrent prepare.
+  {
+    std::lock_guard lk{_query_states_mutex};
+    if (_query_states.size() >= static_cast<std::size_t>(k_max_concurrent_queries)) {
```
Jedi18: Could we enforce a limit here instead of only warning? The pool has `num_threads + 1` workers, but this branch allows more queries to start sequencers. If four concurrent prepares enqueue their sequencers before metadata tasks run (`num_threads = 3`), all four workers can block waiting for metadata, leaving no worker to produce it.

Jedi18: Given that this will only be an issue once concurrency is enabled, we can also add a todo to tackle this in a later concurrency PR

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -2573,36 +2756,40 @@ bool pinned_native_types_match_columns(pinned_entry const& entry,
...
   return covering_mismatch != nullptr ? covering_mismatch : identity_match;
 }
 
 pinned_entry const* sirius_scan_manager::find_pinned_entry_for_parquet_files(
   std::span<std::string const> resolved_file_paths) const
 {
+  std::lock_guard pin_lk{_pinned_entries_mutex};
   for (auto const& [name, entry] : _pinned_entries) {
-    if (entry.cache_info.matches_parquet_files(resolved_file_paths)) { return &entry; }
+    if (entry->cache_info.matches_parquet_files(resolved_file_paths)) { return entry.get(); }
```
Jedi18: Could this return a `shared_ptr<const pinned_entry>`, as the DuckDB lookup now does? The pin-table lock is released when this function returns, but `sirius_plan_get.cpp` keeps and dereferences the raw pointer afterward. Once concurrent planning and unpinning are enabled, an unpin can remove the last owner between those steps, leaving the planner with a dangling pointer.

src/scan_manager/sirius_scan_manager.hpp
```diff
@@ -724,17 +767,96 @@ class sirius_scan_manager {
...
-  /// pointer stays valid for the whole prepare (pin/unpin is
-  /// query-lifecycle-serialized).
+  /// provider takes its own copy of the entry's completed mask set. The
+  /// assignment SHARES OWNERSHIP of the entry across that gap: the mask and
+  /// insert-delta jobs block for as long as their IO takes, and with concurrent
+  /// queries another connection may unpin in that window.
   struct cached_assignment {
     op::scan::sirius_gpu_scan_operator* op{nullptr};
-    pinned_entry const* entry{nullptr};
+    std::shared_ptr<pinned_entry> entry;
```
dhruv9vats: There might be an opportunity to extend the concept of shared_ptr-to-const-obj here too:
```suggestion
    std::shared_ptr<const pinned_entry> entry;
```
by making this `mutable`  https://github.com/sirius-db/sirius/blob/6c41bf61c58b7c767d8fc423a9248a1d427fc676/src/scan_manager/sirius_scan_manager.hpp#L248 since this holds an internal mutex for thread safely: https://github.com/sirius-db/sirius/blob/6c41bf61c58b7c767d8fc423a9248a1d427fc676/src/scan_manager/mvcc_mask_cache.hpp#L74-L79 and because it is meant to be lazily evaluated.

This allows the read only cases to hold `const pinned_entry` through and through. We can also update `sirius_scan_manager::try_match_cached_entry` to hold snapshots of `std::shared_ptr<const pinned_entry>` instead.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -2507,6 +2685,9 @@ void sirius_scan_manager::attach_proven_unique_columns(
 void sirius_scan_manager::remove_pinned_entry(const std::string& name)
 {
   pin_registry_mutation_scope const registry_mutation{*this};
+  // Safe mid-query by construction: dropping the map slot only releases this map's reference.
+  // A query serving the entry holds its own, so its data outlives the unpin.
+  std::lock_guard pin_lk{_pinned_entries_mutex};
   retire_late_mat_handle(name);
```
Jedi18: If another connection unpins this table while a query is running, the query still holds the pinned data, but this invalidates the handle it needs to read deferred columns. Could we keep that handle valid until the query finishes?

Jedi18 (changes requested):

src/scan_manager/sirius_scan_manager.hpp
```diff
@@ -724,17 +767,96 @@ class sirius_scan_manager {
...
+    query_scan_manager_state& operator=(const query_scan_manager_state&) = delete;
+    query_scan_manager_state(query_scan_manager_state&&)                 = delete;
+    query_scan_manager_state& operator=(query_scan_manager_state&&)      = delete;
+
+    //! Stop and join every task this query put on the shared pool. Idempotent, and safe to
+    //! call outside the state mutex — it can block for as long as an in-flight read.
+    void drain() noexcept;
+
+    //! One scan operator and the disk-reading provider built for it (null when the operator
+    //! matched a pinned entry and is served from the cache instead). A vector rather than a
```
dhruv9vats: >(null when the operator matched a pinned entry and is served from the cache instead)

constraints like these would benefit from being encoded in the code itself, like using a `std::variant`, etc.

dhruv9vats (commented):
This is looking good @wmalpica. Mostly nits.

PR #1368, opened 2026-07-31
Support compressing nullable columns

### LINKED ISSUE #1308: Add support for nullable columns to all compression code
Currently, only string operators support nullable columns. We need to expand that and make sure it works throughout the compression subsystem.

joosthooz:
Adds transparent support for a null bitmask in a compressed batch representation. It is stored as a sidecar, so none of the compression code itself sees this data. The tradeoff here is that we can also not apply any compression to it. For that, we would have to make it an explicit channel and probably also an explicit operator to split that channel out. 

The code for handling nulls in str_split and dictionary, as well as the plumbing to make that work, is removed.
There are fastpaths for all nulls and zero nulls.
I'm not sure I'm happy with this approach, I think it adds a bit too much code, I'm open to opinions/suggestions.

Fixes #1308

### DIFF
--- a/src/compression/simpatico_codegen/CMakeLists.txt
+++ b/src/compression/simpatico_codegen/CMakeLists.txt
@@ -147,6 +147,7 @@ add_library(
   src/plan/operator_registry.cpp
   src/plan/representation_factory.cpp
   src/plan/bitjoin_layout.cpp
+  src/plan/validity.cpp
   src/plan/compress.cpp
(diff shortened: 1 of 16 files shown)


kevkrist:
This seems like the right approach to me, with minimal complexity. It must add less code than opening channels for NULLs, I assume, which to me is of dubious value (except heavyweight compression possibly?). At any rate, happy to review proposed approach when the PR is ready.

PR #1371, opened 2026-08-01
perf: TPC-H SF1000 campaign integration (15.99 s -> 8.04 s on GB300)

felipeblazing:


src/scan_manager/memory_prefetcher.cpp
```diff
@@ -0,0 +1,231 @@
...
+        _stops_reservation.fetch_add(1, std::memory_order_relaxed);
+        return converted;
+      }
+
+      try {
+        // Convert against the reservation so the conversion's device
+        // allocations draw from the arena reserved above.
+        mut->convert_to<cucascade::gpu_table_representation>(registry, *reservation, stream);
+      } catch (const rmm::out_of_memory&) {
+        SIRIUS_LOG_DEBUG("[memory_prefetcher] OOM converting batch {}; backing off",
```
aminaramoon, line 228: since this is not just data movement
1. we need to cache other exceptions here too.
2. future work: we need to think about a way to get this out of prefetchjer, sicne it is more than prefetching

aminaramoon (commented):
small comment in scan_manager

src/planner/sirius_plan_comparison_join.cpp
```diff
@@ -543,12 +586,31 @@ sirius_physical_plan_generator::plan_comparison_join(duckdb::LogicalComparisonJo
       // An unfiltered build is (for FK-shaped joins) the whole key domain — its filter keeps
       // every probe row by construction, so wiring a producer target for it only buys overhead.
       // DuckDB's flag covers the delim-join case where the effective build data is children[0].
-      if (!op.filter_pushdown->build_side_has_filter) {
+      //
+      // But "unfiltered" only implies "whole key domain" when the build side IS a base table.
+      // A build that is itself a join or aggregate output has already been reduced, so its key
+      // set is a strict subset of the domain and its filter can be highly selective even with
+      // no predicate anywhere in its subtree. TPC-H q3 is the case in point: the build of
```
kevkrist: I don't think q3 is the right example here -- every table has a local filter in that query. (I understand this is not this PRs comment)

src/planner/sirius_plan_comparison_join.cpp
```diff
@@ -358,6 +358,49 @@ std::vector<std::size_t> build_key_domain_cardinalities(duckdb::LogicalCompariso
...
+/// This is the distinction the "unfiltered build side" refusal actually needs.
+/// An unfiltered base relation is the whole key domain, so a dynamic filter from
+/// it keeps every probe row. An unfiltered *derived* relation is not — a join
+/// output carries only the keys that survived the join, and can be far smaller
+/// than the domain even though no predicate appears anywhere below it.
+///
+/// Descends through the pass-through operators (projection, filter, order) that
+/// sit between a join and its real input; stops at the first reducing operator.
+///
+/// A @c LOGICAL_DELIM_GET is a leaf with no children, so the recursion below can never reach the
```
kevkrist: Can we delete this justification paragraph? Or, can we at least simplify the entire comment for this recursion? It's accumulated some bloat.

kevkrist (commented):
Effectively review of #1373

src/cuda/sirius_dynamic_in_list_filter.cu
```diff
@@ -54,14 +55,48 @@
...
+// IN-list vs Bloom. `capacity_factor_for` may grow the *actual* capacity beyond it (see below);
+// that only ever makes an already-fitting set sparser, never flips the filter choice.
 constexpr std::size_t kCapacityFactor = 2;
 constexpr double kLoadFactor          = 1.0 / kCapacityFactor;
 
 // Threads per key probe.
 constexpr std::size_t kCgSize = 1;
 static_assert(kCgSize == 1, "cuco::static_set requires kCgSize==1 for device_for bulk iteration");
 
+// Keys per bucket. compute_mask is a pure miss-heavy membership probe, so its cost is dominated by
```
kevkrist:
```suggestion
// Keys per bucket
```

src/cuda/sirius_dynamic_in_list_filter.cu
```diff
@@ -54,14 +55,48 @@
 namespace sirius::op {
 
 namespace {
-// Match estimated_set_bytes' 0.5 load factor: capacity = 2 × keys.
+// Baseline load factor: capacity = 2 × keys. estimated_set_bytes reports this size, and the
```
kevkrist:
```suggestion
// Baseline load factor: capacity = 2 × keys.
```
I think the code speaks clearly for this comment.

src/cuda/sirius_dynamic_in_list_filter.cu
```diff
@@ -54,14 +55,48 @@
...
 
+// Keys per bucket. compute_mask is a pure miss-heavy membership probe, so its cost is dominated by
+// the number of *random* slot fetches, not by bytes compared. Packing 4 keys into one bucket lets a
+// single 16 B (INT32) / 32 B (INT64) fetch retire what bucket size 1 needed up to 4 dependent
+// fetches for. Measured at the SF1000 probe shapes this is worth 1.16-1.30x on its own, and it
+// leaves capacity (and therefore estimated_set_bytes and the replica byte count) unchanged.
+// Bucket size 8 regresses: the extra intra-bucket comparisons outweigh the fetches saved.
+constexpr std::int32_t kBucketSize = 4;
+
+// Largest capacity multiplier whose table still fits `budget`, else the baseline. A sparser table
```
kevkrist:
```suggestion
// Largest capacity multiplier whose table still fits `budget`, else the baseline.
```
The specific numbers and query references probably don't belong in the code comments.

src/cuda/sirius_dynamic_in_list_filter.cu
```diff
@@ -54,14 +55,48 @@
...
+// shortens probe chains, but only pays while the table stays cache-resident -- past that the extra
+// footprint costs more than the saved probes (measured: at 4x, q9's 41 MB of keys grow to 166 MB,
+// blow the 115 MB L2 and give back the entire win). Half of L2 is the budget so the streaming probe
+// column still has room; the remaining candidates are tried largest-first.
+constexpr std::size_t kGrowthCandidates[] = {4, 3};
+
+[[nodiscard]] std::size_t capacity_factor_for(std::size_t num_keys, std::size_t key_bytes) noexcept
```
kevkrist, line 76: Copying @Jedi18 comment from #1377 to make sure it lands here:

capacity_factor_for() uses only the producer GPU’s L2 cache size, while the IN-list/Bloom selection uses the minimum L2 cache size across all probe GPUs.

Our benchmarks currently use identical GPUs, so this is not an issue today. However, with different GPU models, the producer could build a 3×/4× table that does not fit another GPU’s cache. Performance would then depend on which GPU builds the filter.

Could we choose the growth factor using the smallest L2 cache among the planned replica devices?

src/cuda/sirius_dynamic_in_list_filter.cu
```diff
@@ -372,6 +418,11 @@ std::unique_ptr<cudf::column> sirius_dynamic_in_list_filter::compute_mask(
 
 std::size_t sirius_dynamic_in_list_filter::size() const noexcept { return _num_keys; }
 
+// Reports the *baseline* (kCapacityFactor) footprint, which is the smallest table this filter will
+// ever build. The constructor may pick a sparser table when one comfortably fits L2, so this is a
+// lower bound on the real allocation rather than an exact size -- deliberately so: it keeps the
```
kevkrist:
```suggestion
// lower bound on the real allocation rather than an exact size.
```

kevkrist (commented):
nits. Essentially review of #1377

src/cuda/sirius_dynamic_bloom_filter.cu
```diff
@@ -60,29 +62,104 @@ std::size_t blocks_for(std::size_t num_keys)
 
 using bloom_alloc = sirius::rmm_cuco_allocator<cuda::std::byte>;
 
+/**
+ * @brief Fingerprint policy for Sirius's dynamic Bloom filters.
+ *
+ * Byte-for-byte identical to @c cuco::default_filter_policy<xxhash_64<KeyT>,uint32_t,8> in hash
```
kevkrist: Can you condense/simplify this comment? Perf eval probably belongs more as a PR comment, and FPR should be highlighted alongside throughput (I see it there below -- it's just buried). Also, it's known that the default bloom filter policy in CuCo is not very good wrt FPR: https://github.com/NVIDIA/cuCollections/issues/682. This is why, for block counts that fit a uint32, we today use the Arrow policy -- FPR values should be compared against that policy in that regime. I understand that SF1000 doesn't hit that policy, but it still seems that Arrow policy should land in that regime, because it *does fully* decorrelate block index selection from bit pattern generation. For block counts > 2^20, the hash bits used for block index selection and bit pattern generation begin to overlap with this policy, resulting (I expect) in FPR degradation compared to Arrow.

src/cuda/sirius_dynamic_bloom_filter.cu
```diff
@@ -173,12 +250,10 @@ std::unique_ptr<bloom_replica> build_bloom_replica(int device_id,
                                                    rmm::device_async_resource_ref mr,
                                                    cuda::stream_ref stream)
 {
-  if (num_blocks <= arrow_policy<KeyT>::max_filter_blocks) {
-    return std::make_unique<bloom_replica>(
-      device_id, build_bloom<arrow_bloom<KeyT>>(keys, num_blocks, mr, stream));
-  }
+  // One policy for every size: unlike cuco's stock pair there is no cap to fall off, so there is
```
kevkrist: This explanation refers to previous policy so will not make much sense to future readers who aren't familiar with it.

kevkrist (changes requested):

Jedi18:
Can we have the PR rebased on dev and then split into reviewable layers since this covers a wide range of the codebase.

Codex found several issues (I've pasted the table below), I would recommend running the PR through codex and opus independently for review and fixes before we open this up for human reviews.

<details>
  <summary>Codex findings</summary>
  
| Severity | Issue | Primary location | Recommended correction |
|---|---|---|---|
| Critical | MIXED_JOIN can publish inequality keys as exact filters and drop valid rows | `src/op/sirius_physical_hash_join.cpp:1985–2002` | Publish membership filters only for equality conditions; disable MIXED_JOIN publication until fixed. |
| High | Task-hint recursion can leave an upstream pipeline incomplete | `src/creator/task_creator.cpp:359–367` | Re-evaluate every visited pipeline while recursion unwinds. |
| High | Compressed prefetch reserves encoded bytes, not peak encoded + decoded bytes | `src/scan_manager/memory_prefetcher.cpp:174–216` | Share the task materialization peak-size helper. |
| High | Parallel decompression discards asynchronous CUDA synchronization failures | `src/compression/simpatico_codegen/src/util/stream_pool.cpp:32–36` | Synchronize all streams and throw the first error. |
| High | BUILD_PROBE eligibility uses average size and ignores partition skew | `src/op/sirius_physical_hash_join.cpp:703–724` | Use measured partition sizes or require the whole build to fit. |
| High | Whole-build publication concatenates beyond normal streaming memory bounds | `src/op/sirius_physical_partition.cpp:421–429` | Respect concat and memory budgets or union filters incrementally. |
| High | Optional adaptive IN-list growth can fail the query under memory pressure | `src/cuda/sirius_dynamic_in_list_filter.cu:84–97, 210–250` | Reserve actual capacity, retry at 2×, then fall back to Bloom or no filter. |
| Medium | Prefetch workers are not bound to their target CUDA device | `src/scan_manager/memory_prefetcher.cpp:46–50, 81–88` | Install per-thread CUDA device affinity before stream use. |
| Medium | Prefetched tables bypass the new zero-copy ownership transfer | `src/op/scan/sirius_gpu_scan_operator_data.cpp:39–75` | Mark per-query prefetched representations as stealable. |
| Medium | Count-distinct fast path swallows OOM and CUDA exceptions | `src/op/aggregate/gpu_aggregate_impl.cpp:197–235` | Rethrow resource and cancellation errors; narrow fallback catches. |
| Medium | Optimizer pre-hook still changes plans when Sirius is disabled | `src/transparent/sirius_optimizer_extension.cpp:216–223` | Gate on an initialized Sirius context or skip registration. |
| Medium | Interrupt sentinels can survive queue reactivation | `src/include/exec/interruptible_mpmc.hpp:133–167` | Use a close-aware wait primitive or preserve work while draining all sentinels. |
| Medium | Derived-build publication gate uses row count instead of distinct key count | `src/op/dynamic_filter_publisher.cpp:175–189` | Use distinct cardinality or defer to the runtime keep-ratio gate. |
| Medium | IN-list capacity selection is source-GPU-specific on heterogeneous systems | `src/cuda/sirius_dynamic_in_list_filter.cu:84–97, 221–224` | Freeze capacity using the minimum probe-GPU L2 or rebuild per device. |
| Medium | Configured decompression parallelism is ignored | `src/compression/compression_converters.cpp:51–60` | Pass the configured degree into the decode pool or update the setting contract. |
| Medium | Prefetch shutdown can block for the full polling interval | `src/scan_manager/memory_prefetcher.cpp:61–78, 105–107` | Replace sleep_for with an interruptible wait. |
| Medium | OR-derived restrictions have no expression-size or execution-cost budget | `src/transparent/sirius_optimizer_extension.cpp:120–196` | Cap expression nodes and restrict derivation to cheap comparisons and IN predicates. |
</details>

bench/sf1000-repro/plans/lineitem.txt
```diff
@@ -0,0 +1,85 @@
+# Pareto-picked plans: max ratio with decompress >= 250 GB/s per column.
```
joosthooz, line 1: The defaults in `dev` have been updated to what's in these 2 files, so they can be removed.

felipeblazing:
> Can we have the PR rebased on dev and then split into reviewable layers since this covers a wide range of the codebase.
> 
> Codex found several issues (I've pasted the table below), I would recommend running the PR through codex and opus independently for review and fixes before we open this up for human reviews.
> 
> Codex findings

We decided to do them all together instead of separately to move faster but will definitely look into your comments

felipeblazing:
Triaged all 17 Codex findings — each one verified against the actual code paths rather than taken at face value. Fixes are in `15f00613` (comment cleanup) and `3cbe7297` (correctness fixes + tests). Full triage below.

**Fixed in `3cbe7297`:**

| Finding | What was actually wrong | Fix |
|---|---|---|
| Task-hint recursion leaves upstream pipeline incomplete (**High — confirmed**) | `get_next_task_hint()` is side-effecting at every recursion level (hash-join discard sweep drains ports), but only the *requesting* pipeline was re-evaluated. An upstream pipeline whose tasks had all completed earlier was never marked finished → its consumers wait forever. | The hint walk records every visited pipeline; the manager re-evaluates all of them on the no-task path. + regression test. |
| Interrupt sentinels survive queue reactivation (**High — confirmed, worse than rated**) | moodycamel's `try_dequeue` is not FIFO across producer sub-queues, so `reactivate()`'s stop-at-first-real-item scan could leave sentinels behind — a restarted `downgrade_executor` thread pops one and dies silently; `cancel_pending_requests` strands promises queued behind one. | `reactivate()` drains to exhaustion and re-enqueues survivors; `try_pop()` skips sentinels. + 2 tests that reproduce the cross-sub-queue ordering. |
| Parallel decompression discards CUDA sync failures | `stream_pool::sync_all()` ignored `cudaStreamSynchronize` returns — an async kernel failure returned garbage as success. | `sync_all()` reports the first failure (still syncing every stream, callable from destructors); `run_column_workers` throws on it. |
| Compressed prefetch reserves encoded bytes only | Confirmed — the prefetcher gated and reserved the resting (encoded) size while conversion peaks at encoded + decoded. | Now reserves the peak via the same sizing helper the task path uses (promoted from a local lambda to a shared utility). |
| Count-distinct fast path swallows OOM/CUDA errors | Confirmed — the fallback catch ate `rmm::out_of_memory` (bypassing task retry/downgrade) and sticky CUDA errors. | Rethrows `bad_alloc` and `cudf::cuda_error`; only logic errors fall back to multi-column keys. |
| Prefetch workers not device-bound | Real but narrower than stated: multi-GPU prefetch is refused at startup, so this only bites a single-space config with `device_id != 0` (compression converters allocate from the *current*-device resource). | Workers now bind to the space's device on entry. |
| Optimizer pre-hook runs when Sirius is disabled | Real gap / refuted impact: the derivation is provably row-preserving (F ⟹ derived restriction), but it perturbed EXPLAIN output and cost estimates on CPU-fallback runs. | Pre-hook now gated on an initialized SiriusContext + not-an-internal-query, mirroring the main hook. |
| Whole-build publication exceeds memory bounds | Partially stale (the build-side-sizing gate landed earlier in `ac43e0bf`), but the remaining fold consulted no byte budget — reachable when `hash_partition_bytes > max_build_hash_table_bytes`. | The fold now also requires the build to fit `max_build_hash_table_bytes`: a build refused BUILD_PROBE for its size is not concat-folded whole either. |

**Refuted, with evidence:**

- **MIXED_JOIN publishes inequality keys as exact filters (the Critical):** cannot fire. Only equality conditions populate `_right_key_col_indices` (`sirius_physical_hash_join.cpp`), and both DuckDB (`PhysicalComparisonJoin::ReorderConditions`) and Sirius (`reorder_join_conditions`) reorder equalities to the front with identical predicates, so every inequality ordinal fails the publisher's bounds check and publishes nothing. Since that protection was accidental, `3cbe7297` adds a regression-fence test (`dynamic-filter publisher never publishes for an inequality condition ordinal`) that turns it into a contract. Note the surrounding code predates this PR (#794).
- **OR-derivation has no budget:** it does — `kMaxDerivedDisjuncts = 16` is enforced before any work, and the derived expression is bounded by the size of the source OR.
- **Adaptive IN-list growth fails queries under memory pressure:** growth is hard-capped at L2/2 in absolute bytes (≈63 MB here), and an OOM would propagate to the task-level handler, which reschedules with a larger reservation. The un-reserved source-side build predates this PR and is unchanged by it.
- **Configured decompression parallelism ignored:** moot — #1357 removed the knob entirely; decode uses a fixed 4-stream pool. The one stale doc line naming the old setting is fixed.

**Real but deferred as follow-ups** (all pre-existing behavior this PR merely exposes more, and all perf-only):

- *BUILD_PROBE average-size/skew*: the averaging predates `9a9b7800`; on a single GPU the check is exact. The new exposure is multi-GPU with `natural > num_gpus`. A skew margin would un-do part of the measured q7/q19/q21 wins, so it needs measurement, not a blind fix.
- *Derived-build gate uses rows instead of distinct keys*: from #794; can only skip a useful filter, never drop rows. Fix is a `cudf::distinct_count` in the publisher — perf-sensitive, worth its own measured PR.
- *Prefetched tables bypass the zero-copy steal*: confirmed perf gap (a naive fix is unsafe — raw GPU pins must never be stolen; needs a provider-side wrapper marker, ~4 files).
- *Prefetch shutdown blocking*: real but ≤2 ms with the default 2 ms poll; the dominant teardown cost is an in-flight conversion, which an interruptible wait wouldn't fix.

**Validation:** full unit suite 2334/2334 (9 new tests, including the reinstated `test_task_creator.cpp`, which had been orphaned out of the build since May), pre-commit clean, and the TPC-H SF1000 suite re-measured at **8.042 s best-of-3** (vs 8.180 s before the rebase — no regression from any of the fixes; no query moved >5%), 22/22 results byte-identical against the stock-dev baseline.

The inline comment threads (in-list/bloom comment cleanup, `build_side_is_derived` comment bloat, prefetcher exception handling) are addressed in the two commits. @kevkrist on the Bloom FPR question — the fast-range policy draws the block index from the high hash bits and the 40 fingerprint bits from the low end, so the draws stay disjoint up to ~2^24 blocks; measured keep-ratios at the shapes that motivated the change were slightly *better* than the modulo's. Restoring the Arrow policy in its ≤2^22-block regime is doable if you want the decorrelation guarantee there — it re-introduces the size-dependent branch and a second filter type through the replica variant, so I'd rather take that as a follow-up with FPR measurements. Cross-GPU L2 sizing for the IN-list growth factor: agreed but deferred — we don't run heterogeneous GPU systems yet.

src/scan_manager/memory_prefetcher.cpp
felipeblazing: Done in `3cbe7297`: non-OOM conversion failures are now caught here (decode runs on this path), logged with the batch id, counted, and the batch is skipped — the scan task surfaces the error authoritatively if it persists. Same commit also fixes the reservation to cover the decode peak (encoded + decoded) and binds workers to the space's device. Moving the conversion out of the prefetcher is noted as follow-up work.

src/planner/sirius_plan_comparison_join.cpp
felipeblazing: Agreed — the q3 example was wrong (every table there has a local filter). The comment is condensed in `15f00613` and no longer cites a query.

src/planner/sirius_plan_comparison_join.cpp
felipeblazing: Condensed in `15f00613`: 23 lines down to 8, keeping only the derived-vs-base distinction and a one-liner on why DELIM_GET (childless but never a base table) is in the list.

src/cuda/sirius_dynamic_in_list_filter.cu
felipeblazing: Trimmed in `15f00613` — kept one line noting the value follows the probing scheme (double hashing) rather than the key type, since that's the invariant a future edit would trip over.

src/cuda/sirius_dynamic_in_list_filter.cu
felipeblazing: Applied verbatim in `15f00613`.

src/cuda/sirius_dynamic_in_list_filter.cu
felipeblazing: Trimmed in `15f00613` — dropped the measured numbers and query references, kept the one-line invariant that growth only sparsifies an already-chosen set (selection always evaluates the baseline footprint).

src/cuda/sirius_dynamic_in_list_filter.cu
felipeblazing: Acknowledged — the growth factor should use the minimum L2 across planned replica devices. Deferring for now since we don't run heterogeneous GPU systems yet; on identical GPUs producer L2 == replica L2. Tracked for a follow-up alongside the replica-capacity work.

src/cuda/sirius_dynamic_in_list_filter.cu
felipeblazing: Applied in `15f00613` — comment is now two lines: baseline footprint, lower bound by design so the IN-list/Bloom choice stays independent of probe-side tuning.

src/cuda/sirius_dynamic_bloom_filter.cu
felipeblazing: Comment condensed in `15f00613` (perf tables removed, FPR note surfaced next to the correctness invariant). On the substance: fast-range consumes the high hash bits and the 40 fingerprint bits come from the low end, so the two draws stay disjoint until ~2^24 blocks (128 MiB-cap filters are 2^22). Measured keep-ratios at the shapes that drove this change came out slightly better than the modulo's. Restoring the Arrow policy for the ≤2^22-block regime is a fair ask for the decorrelation guarantee — answered in more detail in the main thread; proposing it as a measured follow-up since it re-introduces the size-dependent branch and a second filter type in the replica variant.

src/cuda/sirius_dynamic_bloom_filter.cu
felipeblazing: Rewritten in `15f00613` — the comment now stands on its own without referencing the previous policy.

src/cuda/sirius_dynamic_bloom_filter.cu
felipeblazing: Filed the Arrow-regime restore as #1408 with the FPR-measurement acceptance criteria — proposing we take it there rather than growing this PR.

kevkrist:
Closed #1408 as based on my misunderstanding of Arrow policy -- see comment there.
/approve

kevkrist (approved):

aminaramoon (approved):

Jedi18 (approved):

aocsa (approved):
LGTM

bwyogatama (approved):

joosthooz (approved):

bench/sf1000-repro/plans/lineitem.txt
joosthooz: Oh now I see these 2 are the only tables that are being compressed by the run script.

bench/sf1000-repro/plans/orders.txt
```diff
@@ -0,0 +1,51 @@
...
+input -> bitpack -> chunk_min, chunk_count, chunk_bits, packed
+---
+# column: o_orderdate  dtype: t12  ratio: 2.651x  depth: 1
+# comp: 755.61 GB/s  decomp: 1668.67 GB/s
+input -> bitpack -> chunk_min, chunk_count, chunk_bits, packed
+
+---
+# column: o_orderpriority  dtype: str  ratio: 32.310x  depth: 3
+# comp: 181.10 GB/s  decomp: 247.82 GB/s
+input -> dictionary -> keys_offsets, keys_chars, indices
```
joosthooz, line 30: This codec makes compression fail because of some 1<<28 row limit in dictionary. You can change it to identity to make compression work for the other columns.

joosthooz: IIUC the branch already includes a fix to this problem, so no action required here

bench/sf1000-repro/sirius-sf1000.yaml
```diff
@@ -0,0 +1,58 @@
...
+        host:
+            capacity_bytes: 471200000000
+            initial_number_pools: 392
+            pool_size: 8
+            block_size: 67108864
+        disk:
+            disk_id: 0
+            capacity_bytes: 1000000000000
+            downgrade_root_dirs: "/localhome/local-faramburu/sirius_disk_memory"
+    executor:
```
aocsa, line 31: There is a hardcoded path here.

downgrade_root_dirs: "/localhome/local-faramburu/sirius_disk_memory"

PR #1373, opened 2026-08-01
perf: wire dynamic filters through LOGICAL_DELIM_GET build sides

felipeblazing:
build_side_is_derived() treats a childless leaf as an unfiltered base table, so a
LOGICAL_DELIM_GET build side fell through to that case and the dynamic filter was
never wired. DELIM_GET is the opposite: an already-joined, already-DISTINCTed key
set, which is exactly the shape a dynamic filter pays off on.

Compounded by DuckDB setting build_side_has_filter = IsFiltering(join.children[0]),
which inspects the PROBE side.

Measured on TPC-H SF1000 (GB300, 1 GPU, GPU-pinned compressed):
  suite -6.7%; q17 -65%, q21 -35%, q22 -53%. 22/22 byte-identical.

### DIFF
--- a/src/planner/sirius_plan_comparison_join.cpp
+++ b/src/planner/sirius_plan_comparison_join.cpp
@@ -358,6 +358,49 @@ std::vector<std::size_t> build_key_domain_cardinalities(duckdb::LogicalCompariso
 }  // namespace
 //===----------------------------------------------------------------------===//
 
+/// True when @p op is a *derived* relation rather than a base table read: its
+/// subtree contains a join, aggregate, or set operation that has already reduced
+/// its key set.
+///
+/// This is the distinction the "unfiltered build side" refusal actually needs.
+/// An unfiltered base relation is the whole key domain, so a dynamic filter from
+/// it keeps every probe row. An unfiltered *derived* relation is not — a join
+/// output carries only the keys that survived the join, and can be far smaller
+/// than the domain even though no predicate appears anywhere below it.
+///
+/// Descends through the pass-through operators (projection, filter, order) that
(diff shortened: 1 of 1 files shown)


kevkrist:
Amazing! #1277 partially overlaps -- it covers the `DELIM_GET` class (q17/q22) and also admits `LOGICAL_CTE_REF` builds. 

What this PR has that #1277 doesn't: derived builds on the zone-map-capable scan route, and any-descendant admission of join/aggregate builds. Noting here for #1277.

kevkrist (approved):

src/planner/sirius_plan_comparison_join.cpp
```diff
@@ -543,12 +586,31 @@ sirius_physical_plan_generator::plan_comparison_join(duckdb::LogicalComparisonJo
       // An unfiltered build is (for FK-shaped joins) the whole key domain — its filter keeps
       // every probe row by construction, so wiring a producer target for it only buys overhead.
       // DuckDB's flag covers the delim-join case where the effective build data is children[0].
-      if (!op.filter_pushdown->build_side_has_filter) {
+      //
+      // But "unfiltered" only implies "whole key domain" when the build side IS a base table.
+      // A build that is itself a join or aggregate output has already been reduced, so its key
+      // set is a strict subset of the domain and its filter can be highly selective even with
+      // no predicate anywhere in its subtree. TPC-H q3 is the case in point: the build of
```
kevkrist, line 593: I don't think q3 is the right example here. customer and orders each have local table filters, and the dynamic filter from the join of customer with orders is pushed to lineitem on dev.

felipeblazing:

PR #1374, opened 2026-08-01
perf: wake interruptible_mpmc consumers on interrupt

felipeblazing:
interrupt() set _is_active=false but never woke a waiting consumer, which sat in
wait_dequeue_timed(item, 10000) -- a 10 ms poll. The existing comment documented
the cost: 'Consumer threads will see this flag on their next loop cycle (max 10ms
delay)'. Mean 5 ms was burned per query teardown waiting for a flag nobody signalled.

interrupt() now enqueues null sentinels to wake waiters immediately. pop() already
mapped a null item to the interrupted return, so the semantics are unchanged;
reactivate() drains any leftover sentinels.

Measured on TPC-H SF1000 (GB300): suite -5.7%, 22/22 byte-identical. Found by
instrumenting run_mandatory_cleanup -> downgrade drain -> join_processing_thread,
which was 5.10 ms mean over 43 calls and now falls below a 200 us threshold.

### DIFF
--- a/src/include/exec/interruptible_mpmc.hpp
+++ b/src/include/exec/interruptible_mpmc.hpp
@@ -19,6 +19,7 @@
 #include <blockingconcurrentqueue.h>
 
 #include <atomic>
+#include <cstdint>
 #include <concepts>
 #include <memory>
 #include <optional>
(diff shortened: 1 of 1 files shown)


felipeblazing:

PR #1375, opened 2026-08-01
perf: label-encode group keys so count-distinct reaches the radix sort path

felipeblazing:
cudf's sorted groupby finds group boundaries via key_sort_order() ->
stable_sorted_order(keys), which only takes its radix fast path when
num_columns() == 1 (sort_impl.cuh:53). A multi-column group key therefore takes a
lexicographic merge sort -- on TPC-H q16 that is 118.8M rows, 308.7 ms, 92% of the
aggregate.

The pre-existing dictionary-encode optimisation fires here but does not help: it
narrows the comparator while leaving three key columns, so the single-column gate
is never reached. Narrowing is not enough; only collapsing to one column is.

cudf::encode the key table into one dense INT32 label, group by that, and recover
the keys with a gather at group cardinality (~18K rows).

Guarded: only with a COLLECT_SET aggregation present, only above 1M rows, never for
already-radix-sortable single keys or nested keys, and only when an HLL estimate puts
group cardinality below 1% of rows -- cudf::encode sorts the distinct rows, so on a
high-cardinality key it would cost as much as the sort it replaces.

Measured on TPC-H SF1000 (GB300): q16 0.490 -> 0.298 s (-39.0%), suite -2.01%,
22/22 byte-identical, q16 the only query moving >= 3%. Verified against DuckDB CPU
on 4 shapes including NULLs in both group keys and the count-distinct value, plus
the declined high-cardinality path.

### DIFF
--- a/src/op/aggregate/gpu_aggregate_impl.cpp
+++ b/src/op/aggregate/gpu_aggregate_impl.cpp
@@ -20,10 +20,15 @@
 #include "log/logging.hpp"
 
 #include <cudf/column/column_factories.hpp>
+#include <cudf/copying.hpp>
 #include <cudf/dictionary/dictionary_column_view.hpp>
 #include <cudf/dictionary/encode.hpp>
 #include <cudf/reduction/approx_distinct_count.hpp>
 #include <cudf/strings/strings_column_view.hpp>
+#include <cudf/transform.hpp>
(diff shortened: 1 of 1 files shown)


joosthooz (approved):
Great work, thanks, I see this works in addition to the old dict-encode string columns optimization

PR #1376, opened 2026-08-01
fix: guard empty _ready_devices in the lookahead hint

felipeblazing:
management_eventloop dereferenced *_ready_devices.begin() unconditionally. The
vector is legitimately empty whenever a loop iteration saw no non-scan device_ready
request, so this is UB on a live path. schedule_lookahead already takes
std::optional<int> defaulted to nullopt, so pass nothing instead.

The dereference happens at the call site, before schedule_lookahead can early-return,
so it is reachable regardless of the task_creator strategy setting.

### DIFF
--- a/src/pipeline/task_scheduler.cpp
+++ b/src/pipeline/task_scheduler.cpp
@@ -37,6 +37,7 @@
 #include <algorithm>
 #include <limits>
 #include <mutex>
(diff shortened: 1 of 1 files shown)


bwyogatama:
This bug is actually solved in dev

PR #1377, opened 2026-08-01
perf: widen the dynamic IN-list cuco set to 4-slot buckets

felipeblazing:
The IN-list membership set is double-hashed, so every probe step is a fresh random
sector and the probe is miss-heavy -- cost is the number of random fetches, not ALU.
Widening cuco::storage<1> to storage<4> lets one 16/32 B bucket fetch retire what
took up to 4 dependent fetches. Bucket 8 regresses; 4 is the knee.

Also raises the capacity factor to 3-4 while the table still fits half of L2.
estimated_set_bytes() deliberately still reports the 2x baseline, so the
IN-list/Bloom selection and the replica byte accounting are bit-identical.

Kernel-level at real build sizes: q18 1.64x, q21 1.50x, q9 1.30x/1.16x.
Measured on TPC-H SF1000 (GB300): suite -1.32%; q18 -3.4%, q9 -1.8%, q21 -1.7%,
q17 -12.3%. 22/22 byte-identical.

NOTE this is scheme-specific and does NOT generalise: cudf's groupby set uses linear
probing with step_size = BucketSize, so successive steps walk contiguous slots and a
wider bucket only adds load -- measured monotonically worse there, bucket 1 optimal.
The probing scheme decides it, not the constant.

### DIFF
--- a/src/cuda/sirius_dynamic_in_list_filter.cu
+++ b/src/cuda/sirius_dynamic_in_list_filter.cu
@@ -30,6 +30,7 @@
 #include <cub/device/device_for.cuh>
 #include <cuco/operator.hpp>
 #include <cuco/static_set.cuh>
+#include <cuco/storage.cuh>
 #include <cuda/sirius_rmm_cuco_allocator.cuh>
 #include <cuda/std/functional>
 #include <cuda/std/limits>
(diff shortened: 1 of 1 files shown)


src/cuda/sirius_dynamic_in_list_filter.cu
```diff
@@ -54,14 +55,48 @@
...
+constexpr std::int32_t kBucketSize = 4;
+
+// Largest capacity multiplier whose table still fits `budget`, else the baseline. A sparser table
+// shortens probe chains, but only pays while the table stays cache-resident -- past that the extra
+// footprint costs more than the saved probes (measured: at 4x, q9's 41 MB of keys grow to 166 MB,
+// blow the 115 MB L2 and give back the entire win). Half of L2 is the budget so the streaming probe
+// column still has room; the remaining candidates are tried largest-first.
+constexpr std::size_t kGrowthCandidates[] = {4, 3};
+
+[[nodiscard]] std::size_t capacity_factor_for(std::size_t num_keys, std::size_t key_bytes) noexcept
```
Jedi18, line 84: `capacity_factor_for()` uses only the producer GPU’s L2 cache size, while the IN-list/Bloom selection uses the minimum L2 cache size across all probe GPUs.

Our benchmarks currently use identical GPUs, so this is not an issue today. However, with different GPU models, the producer could build a 3×/4× table that does not fit another GPU’s cache. Performance would then depend on which GPU builds the filter.

Could we choose the growth factor using the smallest L2 cache among the planned replica devices?

Jedi18 (commented):
Thanks for the PR! LGTM except for one fix

felipeblazing:

PR #1378, opened 2026-08-01
perf: replace the Bloom filter's modulo block index with Lemire fast-range

felipeblazing:
blocks_for() sizes every Bloom above cuco::arrow_filter_policy's 128 MiB / 2^22-block
cap, so build_bloom_replica always fell through to cuco::default_filter_policy, whose
block_index is literally 'hash mod num_blocks'. GPUs have no integer-divide
instruction, so a 64-bit modulo by a runtime value is a long emulated sequence and it
dominated the probe kernel.

The miss is by a hair: the Arrow cap is 67.1M keys at 16 bits/key, and TPC-H q21's two
hot filters hold 73.2M and 70.6M -- over by 9% and 5%.

sirius_bloom_policy is identical to default_filter_policy<xxhash_64,uint32,8> in hash,
block geometry and fingerprint layout; only block_index changes, to
(hash * num_blocks) >> 64 -- one mul.hi.u64. Being uncapped, one policy now covers all
sizes, so the size-dependent branch is gone. kTargetBitsPerKey and estimated_bytes()
are untouched, so hash-set-vs-Bloom selection is bit-identical.

cuco::static_set already avoided this divide via cuco::utility::fast_int magic
reciprocals -- only the Bloom policy was raw, which is why the IN-list never showed
the symptom.

No false negatives: fast-range is a deterministic uniform map applied identically on
add and contains, so the filter stays a conjunctive superset and the join stays
authoritative. The false-positive set moved favourably -- fast-range takes the high
hash bits while the fingerprint takes the low 40, where the modulo drew both from the
low bits. Keep ratio 5.002% -> 4.878% against a 4.77% true-match rate.

Measured on TPC-H SF1000 (GB300): q21 -15.2%, q3 -8.9%, q8 -7.3%, suite -2.49%.
22/22 byte-identical. Zero-false-negative verified for INT32/INT64 at 1/12/1e3/1e6/
3e7/7.3e7 keys including 0, -1, INT_MIN, INT_MAX; 99 [dynamic_filter] tests pass.

### DIFF
--- a/docs/super-sirius/dynamic-filters.md
+++ b/docs/super-sirius/dynamic-filters.md
@@ -344,7 +344,7 @@ The zone-map captures only the build keys' `[min,max]` range, which is useless f
 
 - **`sirius_dynamic_small_in_list_filter`** — exact membership for 1–12 null-free INT32/INT64 build rows, counting duplicates. Each successfully materialized device-local replica owns a raw snapshot of the build values (the *needles*), and one CUB bulk kernel compares every probe value with every needle. It has no hash build, slot array, or reserved empty-key sentinel.
 - **`sirius_dynamic_in_list_filter`** — hash-based membership via a persistent `cuco::static_set`, with a device kernel probing its read-only set reference. It is exact for representable set keys and conservatively passes cuCO's reserved empty-key value, so it remains a safe join-filter superset.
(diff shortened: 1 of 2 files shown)


felipeblazing:

PR #1379, opened 2026-08-01
perf: derive single-table restrictions from OR branches before optimization

felipeblazing:
DuckDB has this rewrite (JoinDependentFilterRule) but declines unless the OR is the
filter ROOT, and DistributivityRule hoists common conjuncts out first, so the OR is
never re-offered in the shape the rule accepts. TPC-H q19's three-branch disjunction
is exactly that shape.

Registers a pre_optimize_function that, for every table restricted by ALL branches of
a disjunction, appends the OR of those per-branch restrictions. Soundness: each branch
D_i = AND_j c_ij implies its single-table part f_i^T, so OR_i D_i implies OR_i f_i^T --
the derived predicate is implied, never narrowing.

Conjuncts every branch carries verbatim are dropped, since DistributivityRule hoists
those itself and keeping them would bolt a duplicate predicate onto the scan.
Volatile / throwing / subquery / parameter expressions are left alone -- stricter than
DuckDB's own rule, which only checks volatility -- because the derived copy is
evaluated on rows the original disjunction rejects.

Measured on TPC-H SF1000 (GB300): suite -3.0%; q19 -27%, q3/q4/q5/q6 -8..-12%.
22/22 byte-identical.

### DIFF
--- a/src/include/transparent/sirius_optimizer_extension.hpp
+++ b/src/include/transparent/sirius_optimizer_extension.hpp
@@ -21,6 +21,27 @@
 
 namespace sirius::transparent {
 
+/// \brief Pre-optimization hook that derives single-table restrictions out of OR-ed filters.
+///
+/// For a filter `(a1 AND b1) OR (a2 AND b2) OR ...` where the branches span more than one table,
+/// every branch implies the AND of its own single-table conjuncts, so the disjunction implies the
+/// OR of those per-table restrictions. AND-ing that implied restriction onto the filter is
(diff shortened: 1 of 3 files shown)


felipeblazing:
Superseded by #1371, which carries the q19 OR-branch restriction derivation in commit cca384ec (kept by the wins-only pass at -3.0% suite). This standalone branch now only conflicts textually with #1371's evolved copy of the optimizer hook. Closing in favor of #1371.

PR #1380, opened 2026-08-01
perf: evaluate string equality on dictionary codes, skipping the decode gather

felipeblazing:
A dictionary-compressed column consumed only by an equality/IN filter no longer
materialises its decoded STRING column. The predicate is resolved against the
dictionary's key set -- four entries for TPC-H l_shipinstruct -- and mapped over the
already-decoded INT32 indices, producing a BOOL8 mask. cudf::dictionary::decode, the
~287 GB/s gather over the key chars, never runs.

Source to sink:
- simpatico::decode_predicate is a set-membership directive threaded through
  decompress(). DecodeWalk offers it to the node producing the column's final value;
  dictionary_compressed_representation::decompress_predicate answers it off the keys.
  Contract: an active directive ALWAYS yields BOOL8, including on the generic
  decode-then-compare fallback, so a caller that rewrote its filter is never surprised
  by a plan shape.
- parquet_gpu_ingestible advertises candidates: columns whose entire pushed-down filter
  is an equality/IN over string constants AND which are never projected (the mask
  replaces the values, so a projected column cannot survive it).
- sirius_scan_manager attaches the pushdown to the per-scan projection of a GPU-tier
  compressed chunk -- never the shared pinned chunk -- and only for dictionary-rooted
  plans, keeping this a strict no-op elsewhere.
- post_filter_and_project reads each batch's actual column types to decide which
  candidates were substituted. Deciding per column matters: q19 filters both
  l_shipinstruct (dictionary, substitutes) and l_shipmode (str_split, declines).

Measured on TPC-H SF1000 (GB300): q19 -21.4% in-suite, and 0.316 s vs 0.509 s standalone
against an uncompressed 'identity' plan for the same column -- i.e. it beats the
uncompressed arm outright while also keeping 78.6 GB of compression. 22/22 byte-identical.

Scoped to the GPU tier: a host pin would have to parse the chunk header to know whether
the plan can exploit the directive, and that tier is not where the decode gather costs
anything.

NOTE: includes scan-path changes from joosthooz's in-flight PRs (#1341/#1342/#1357),
which this builds on and which are not yet on dev.

### DIFF
--- a/src/compression/compressed_representation.hpp
+++ b/src/compression/compressed_representation.hpp
@@ -39,6 +39,25 @@ namespace sirius {
 
 struct compressed_device_blob;  // defined in device_compressed_blob.hpp
 
+/// Per-selected-column equality/IN pushdown for decompression.
+///
+/// Parallel to a representation's selected column list, so entry @c i describes
+/// the @c i-th column the converter will decompress. Entry @c i holds the string
+/// values that column is tested against; an empty entry decompresses normally.
+///
+/// A column with a non-empty entry comes back as a **BOOL8** column
+/// (`value ∈ values`, nulls propagated) rather than its declared type: a
+/// dictionary-compressed column answers the predicate from its key set and never
+/// gathers the decoded chars (see @c simpatico::decode_predicate). Consumers
+/// must therefore expect the type substitution — @c parquet_gpu_ingestible
+/// rewrites its filter expression to a bare boolean reference when it sees one.
+///
+/// Held as plain strings rather than @c simpatico::decode_predicate so this
+/// header stays free of the simpatico API (matching the forward-declared
+/// @c compressed_table above); @c compression_converters.cpp converts at the
+/// call boundary.
+using decode_equality_pushdown = std::vector<std::vector<std::string>>;
+
 /// A Simpatico-compressed chunk resident in pinned host memory.
 ///
 /// The (small) structural header is a flat byte vector; the (large) payload —
@@ -176,6 +195,22 @@ class compressed_host_representation : public cucascade::idata_representation {
     return _selected_indices;
   }
 
+  /// Attach an equality/IN pushdown, parallel to the selected column list.
+  ///
+  /// Call only on a freshly projected representation the caller owns outright
+  /// (as @ref select_columns returns): the pushdown is a property of one scan's
+  /// filter, never of the shared pinned chunk.
+  void set_equality_pushdown(decode_equality_pushdown pushdown)
+  {
+    _equality_pushdown = std::move(pushdown);
+  }
+
+  /// The attached pushdown; empty when the columns decompress normally.
+  [[nodiscard]] const decode_equality_pushdown& equality_pushdown() const noexcept
+  {
+    return _equality_pushdown;
+  }
+
  private:
   /// Construct a projection sharing the same backing blob.
   compressed_host_representation(cucascade::memory::memory_space& memory_space,
@@ -192,6 +227,7 @@ class compressed_host_representation : public cucascade::idata_representation {
   std::size_t _uncompressed_bytes;
   std::int64_t _num_rows;
   std::optional<std::vector<std::size_t>> _selected_indices;
+  decode_equality_pushdown _equality_pushdown;
 };
 
 /// A Simpatico-compressed chunk resident in GPU (device) memory.
@@ -258,6 +294,22 @@ class compressed_device_representation : public cucascade::idata_representation 
     return _selected_indices;
   }
 
+  /// Attach an equality/IN pushdown, parallel to the selected column list.
+  ///
+  /// Call only on a freshly projected representation the caller owns outright
+  /// (as @ref select_columns returns): the pushdown is a property of one scan's
+  /// filter, never of the shared pinned chunk.
+  void set_equality_pushdown(decode_equality_pushdown pushdown)
+  {
+    _equality_pushdown = std::move(pushdown);
+  }
+
+  /// The attached pushdown; empty when the columns decompress normally.
+  [[nodiscard]] const decode_equality_pushdown& equality_pushdown() const noexcept
+  {
+    return _equality_pushdown;
+  }
+
  private:
   compressed_device_representation(cucascade::memory::memory_space& memory_space,
                                    std::shared_ptr<compressed_device_blob> blob,
@@ -273,6 +325,7 @@ class compressed_device_representation : public cucascade::idata_representation 
   std::size_t _uncompressed_bytes;
   std::int64_t _num_rows;
   std::optional<std::vector<std::size_t>> _selected_indices;
+  decode_equality_pushdown _equality_pushdown;
 };
 
 }  // namespace sirius
--- a/src/compression/simpatico_codegen/include/api/simpatico_codegen.hpp
+++ b/src/compression/simpatico_codegen/include/api/simpatico_codegen.hpp
@@ -190,4 +190,35 @@ std::unique_ptr<cudf::table> decompress(
   simpatico::stream_pool& pool,
   rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
 
+// ── Predicate-pushdown decompression ─────────────────────────────────────────
+
+/// Decompress a column subset, answering a set-membership predicate on selected
+/// columns instead of reconstructing them.
+///
+/// @p predicates is parallel to @p selected_columns; an entry with an empty
+/// @c equals_any reconstructs that column normally. A column with an active
+/// directive comes back as BOOL8 of the same row count (`value ∈ equals_any`,
+/// nulls propagated) — never its declared dtype — so the caller must be prepared
+/// for the type change.
+///
+/// The point is to skip the decode entirely for dictionary-compressed columns
+/// consumed only by an equality / IN filter: the predicate is resolved against
+/// the key set and mapped over the indices, so the key chars are never gathered
+/// into a full-width column. Use @c column_supports_predicate_decode to check
+/// that a column's plan can actually do this before pushing a predicate into it.
+///
+/// @throws std::runtime_error if @p predicates and @p selected_columns differ in
+///         size, or on the usual decompression failures.
+std::unique_ptr<cudf::table> decompress(
+  const compressed_table& table,
+  std::span<const std::size_t> selected_columns,
+  std::span<const decode_predicate> predicates,
+  simpatico::stream_pool& pool,
+  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
+
+/// True iff column @p column_index of @p table resolves a predicate without
+/// materialising the column (i.e. its plan is dictionary-rooted). False for an
+/// out-of-range index or a column with no plan tree.
+bool column_supports_predicate_decode(const compressed_table& table, std::size_t column_index);
+
 }  // namespace simpatico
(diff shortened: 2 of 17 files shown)


src/compression/simpatico_codegen/include/api/simpatico_codegen.hpp
```diff
@@ -190,4 +190,35 @@ std::unique_ptr<cudf::table> decompress(
...
+/// Decompress a column subset, answering a set-membership predicate on selected
+/// columns instead of reconstructing them.
+///
+/// @p predicates is parallel to @p selected_columns; an entry with an empty
+/// @c equals_any reconstructs that column normally. A column with an active
+/// directive comes back as BOOL8 of the same row count (`value ∈ equals_any`,
+/// nulls propagated) — never its declared dtype — so the caller must be prepared
+/// for the type change.
+///
+/// The point is to skip the decode entirely for dictionary-compressed columns
```
joosthooz, line 204: Any plans for RLE? Also, I've got a WIP branch with FSST, that's also basically dictionay, so we can implement this for it too.

src/compression/simpatico_codegen/include/api/simpatico_codegen.hpp
```diff
@@ -190,4 +190,35 @@ std::unique_ptr<cudf::table> decompress(
...
   rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
 
+// ── Predicate-pushdown decompression ─────────────────────────────────────────
+
+/// Decompress a column subset, answering a set-membership predicate on selected
+/// columns instead of reconstructing them.
+///
+/// @p predicates is parallel to @p selected_columns; an entry with an empty
+/// @c equals_any reconstructs that column normally. A column with an active
+/// directive comes back as BOOL8 of the same row count (`value ∈ equals_any`,
```
joosthooz, line 200: Isn't it expensive to construct a whole column and populate that for each real entry, isn't it enough to just return a single bool or a bool per entry in `predicates`?

src/compression/simpatico_codegen/include/api/simpatico_codegen.hpp
```diff
@@ -190,4 +190,35 @@ std::unique_ptr<cudf::table> decompress(
...
+/// for the type change.
+///
+/// The point is to skip the decode entirely for dictionary-compressed columns
+/// consumed only by an equality / IN filter: the predicate is resolved against
+/// the key set and mapped over the indices, so the key chars are never gathered
+/// into a full-width column. Use @c column_supports_predicate_decode to check
+/// that a column's plan can actually do this before pushing a predicate into it.
+///
+/// @throws std::runtime_error if @p predicates and @p selected_columns differ in
+///         size, or on the usual decompression failures.
```
joosthooz, line 211: if it doesn't support predicates, will it also throw? Right now it doesn't specify

src/compression/simpatico_codegen/include/codegen/plan/representation.hpp
```diff
@@ -239,6 +261,19 @@ struct dictionary_compressed_representation : standalone_compressed_representati
   std::unique_ptr<cudf::column> decompress(rmm::cuda_stream_view stream,
                                            rmm::device_async_resource_ref mr) const override;
 
+  /// Evaluate @p pred against the dictionary *keys* and map the result over the
+  /// indices, returning a BOOL8 column of @c num_rows without ever gathering the
+  /// key chars into a decoded STRING column.
+  ///
+  /// The keys column is the distinct-value set (four entries for
+  /// `l_shipinstruct`), so the comparison is effectively free; the only
```
joosthooz, line 269: This is query/tpch-specific

src/compression/simpatico_codegen/include/codegen/plan/representation.hpp
```diff
@@ -77,6 +77,28 @@ struct compressible_output {
...
+/// than the reconstructed column itself. Nulls propagate as nulls, matching
+/// @c cudf::binary_operation(col, scalar, EQUAL).
+///
+/// This exists so a dictionary-compressed column consumed only by an equality /
+/// IN predicate never pays the decode gather: the predicate is resolved against
+/// the (tiny) key set and mapped over the indices. See
+/// @c dictionary_compressed_representation::decompress_predicate. Every other
+/// representation falls back to "decode, then compare", which is correct but
+/// buys nothing — so callers should gate the pushdown on
+/// @c plan_supports_predicate_decode.
```
joosthooz, line 94: Consider removing (lengthy and duplicated docs)

src/compression/simpatico_codegen/src/operators/dictionary_compressor.cu
```diff
@@ -285,6 +288,79 @@ std::unique_ptr<cudf::column> dictionary_compressed_representation::decompress(
...
+  if (indices.type().id() != cudf::type_id::INT32) { return nullptr; }
+  if (keys.type().id() != cudf::type_id::STRING) { return nullptr; }
+  // A null key would make the per-key comparison null rather than false, and the
+  // OR-accumulate below would then propagate it. cudf::dictionary::encode never
+  // produces one (nulls live on the parent's mask, not in the key set), so leave
+  // the shape to the generic path instead of carrying a tri-state accumulate.
+  if (keys.null_count() > 0) { return nullptr; }
+
+  // One bool per distinct value: keys ∈ equals_any. The key set is the column's
+  // whole distinct-value population (four entries for l_shipinstruct), so these
```
joosthooz, line 327: Another query-specific `l_shipinstruct`

src/compression/simpatico_codegen/src/operators/dictionary_compressor.cu
```diff
@@ -285,6 +288,79 @@ std::unique_ptr<cudf::column> dictionary_compressed_representation::decompress(
...
+  // kernels are noise next to the row-length pass below — which is the entire
+  // point: this is the work that replaces the decode gather.
+  std::unique_ptr<cudf::column> lut;
+  for (auto const& value : pred.equals_any) {
+    cudf::string_scalar const needle(value, true, stream);
+    auto hit =
+      cudf::binary_operation(keys, needle, cudf::binary_operator::EQUAL, bool_t, stream, mr);
+    lut = lut ? cudf::binary_operation(
+                  lut->view(), hit->view(), cudf::binary_operator::LOGICAL_OR, bool_t, stream, mr)
+              : std::move(hit);
```
joosthooz, line 337: Can't we use `cudf::contains` for this? These are 2 sequential kernel launches per predicate entry

src/compression/compressed_representation.hpp
```diff
@@ -39,6 +39,25 @@ namespace sirius {
...
+/// dictionary-compressed column answers the predicate from its key set and never
+/// gathers the decoded chars (see @c simpatico::decode_predicate). Consumers
+/// must therefore expect the type substitution — @c parquet_gpu_ingestible
+/// rewrites its filter expression to a bare boolean reference when it sees one.
+///
+/// Held as plain strings rather than @c simpatico::decode_predicate so this
+/// header stays free of the simpatico API (matching the forward-declared
+/// @c compressed_table above); @c compression_converters.cpp converts at the
+/// call boundary.
+using decode_equality_pushdown = std::vector<std::vector<std::string>>;
```
joosthooz, line 59: Too long & duplicated docs, and isn't it nicer to just use  `decode_predicate`? It's just a tiny include we're adding and it saves from having this duplicated type and conversion function

src/compression/simpatico_codegen/src/plan/decompress.cpp
```diff
@@ -794,7 +854,8 @@ std::unique_ptr<cudf::column> decode_fused_subtree(PlanTree const& tree,
 std::unique_ptr<cudf::column> decompress_column(PlanTree const& tree,
                                                 rmm::cuda_stream_view stream,
                                                 rmm::device_async_resource_ref mr,
-                                                std::string* error_out)
+                                                std::string* error_out,
+                                                decode_predicate const* pred)
```
joosthooz, line 858: I'm not strongly opposed to this, but maybe it would be nicer to have a separate function where the name shows we're doing a predicate, instead of having a default-null optional parameter.

src/compression/compressed_representation.hpp
```diff
@@ -176,6 +195,22 @@ class compressed_host_representation : public cucascade::idata_representation {
...
+  {
+    _equality_pushdown = std::move(pushdown);
+  }
+
+  /// The attached pushdown; empty when the columns decompress normally.
+  [[nodiscard]] const decode_equality_pushdown& equality_pushdown() const noexcept
+  {
+    return _equality_pushdown;
+  }
+
```
joosthooz, line 213: Looks like this could be dead code seeing you mentioned the feature only works for device-compressed data. If you want to keep them for symmetry or in case we want to implement it in the future I'm ok with that.

However, in general it looks a bit ugly that we're storing predicates (that's a query task item) on a representation (which should be task/query agnostic). It probably has something to do with cucascade plumbing, but it would be nice if we can find a way around it

src/op/scan/parquet_gpu_ingestible.cpp
```diff
@@ -376,14 +376,27 @@ parquet_gpu_ingestible::parquet_gpu_ingestible(std::unique_ptr<parquet_ingestibl
                                                       bind.returned_types,
                                                       _plan->batch_position_by_column_id,
                                                       _plan->partition_primary_indices);
-    if (duckdb_expression) {
-      // Validate before scan tasks retranslate and dereference the predicate.
-      if (sirius::ast::from_duckdb(*duckdb_expression) == nullptr) {
-        throw duckdb::InvalidInputException(
-          "parquet scan: cannot evaluate pushed-down predicate on GPU: %s",
-          duckdb_expression->ToString());
```
joosthooz, line 384: Why is this removed?

src/op/scan/parquet_gpu_ingestible.cpp
```diff
@@ -376,14 +376,27 @@ parquet_gpu_ingestible::parquet_gpu_ingestible(std::unique_ptr<parquet_ingestibl
...
+    if (duckdb_expression) { _duckdb_filter_expression = std::move(duckdb_expression); }
+
+    // Decode-time predicate pushdown: a pure-filter string column whose whole
+    // filter is an equality / IN can be delivered as a BOOL8 mask instead of
+    // values, letting a dictionary-compressed pin answer it off its key set
+    // rather than gathering the decoded chars. Restricted to pure-filter columns
+    // because the mask *replaces* the column — a projected one could not survive
+    // it. This only records what the scan is *willing* to accept; whether any
+    // given batch actually carries masks is decided per batch in
+    // post_filter_and_project.
```
joosthooz, line 388: Comment too lengthy

joosthooz (approved):
Great new feature thanks for this. I think maybe it can be expanded to FSST when we add that to simpatico. For RLE it depends if we need to return a full column with bools, maybe it is not worth doing. For an operator that just looks if a given value is present in a column it would allow to just look at the RLE's `values`.

I have some questions and suggestions, but no dealbreakers apart from the removed throw in `parquet_gpu_ingestible.cpp`. If that look ok to you I'm fine with merging this.

felipeblazing:
Superseded by #1371, which carries the dictionary predicate pushdown in commit feeeaa45 (q19 -21.4%). The only difference on this standalone branch is a mechanical rebase around #1339's predicate-validation guard; #1371's resolution keeps that guard. Closing in favor of #1371.

PR #1381, opened 2026-08-01
perf: stop deep-copying cudf::table at 14 construction sites

felipeblazing:
cudf exports no (vector&&, stream, mr) table constructor, so a three-argument call
builds an implicit temporary from the vector and then binds
explicit table(table const&, stream, mr) -- a full deep copy of data we already own.
Dropping stream/mr selects the move constructor.

Measured on TPC-H SF1000 (GB300): suite -0.8%, 22/22 byte-identical.

Deliberately NOT changed: the six sites where table(table_view, stream, mr) is the
intended materialisation -- top_n.cpp:87,133,289,303, sort_partition.cpp:148 and
host_keep_mask.cpp:80.

NOTE: sirius_physical_hash_join.cpp here also carries joosthooz's in-flight
#1341/#1342 changes, which are not yet on dev.

### DIFF
--- a/src/expression_evaluator/expression_evaluator.cpp
+++ b/src/expression_evaluator/expression_evaluator.cpp
@@ -331,7 +331,7 @@ std::unique_ptr<cudf::table> expression_evaluator::evaluate(cudf::table_view inp
     post_process(*ast_expr, std::move(result));
   }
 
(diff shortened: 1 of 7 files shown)


joosthooz (approved):

felipeblazing:

PR #1384, opened 2026-08-03
ci: validate pull request titles

johanpel:
## Description

PR titles are running a bit wild. Add a GitHub Actions workflow that enforces the Conventional Commits PR-title format already required by `CONTRIBUTING.md`.

The workflow runs when a pull request is opened, reopened, edited, or synchronized. It uses the pinned `amannn/action-semantic-pull-request` v6.1.1 action with read-only pull-request permissions and does not check out PR code.

Before merging, add `amannn/action-semantic-pull-request@*` to the repository Actions allowlist under Settings → Actions → General.

## Checklist

- [x] Read CONTRIBUTING.md
- [x] Validated the workflow with `pixi run pre-commit run --files .github/workflows/semantic-pr.yml`
- [x] No Sirius configuration changes
- [x] No documentation changes needed; this enforces the existing documented convention

## References

- https://github.com/marketplace/actions/semantic-pull-request

_Written by Codex_


.github/workflows/semantic-pr.yml
```diff
@@ -0,0 +1,16 @@
+name: Semantic PR title
+
+on:
+  pull_request_target:
+    types: [opened, reopened, edited, synchronize]
+
+jobs:
+  semantic-pr:
+    name: Validate PR title
```
mike-wendt:
```suggestion
    name: pr-title
```
This will now be a separate check from the existing ones due to the need to get a response to the user if their initial or attempt to fix the PR title are incorrect. I don't want to hide this in the existing `Check` workflow as it wouldn't be as obvious that it needs to be. The name change is to fit with the other required checks (see comment below with screenshot)

.github/workflows/semantic-pr.yml
```diff
@@ -0,0 +1,16 @@
+name: Semantic PR title
```
mike-wendt:
```suggestion
name: Validate
```

In addition to the name change, I would rename the file to `validate.yml` instead. This way this can act as a single workflow with parallel validation runs that we can expand on in the future. It also fits cleanly in the Actions tab and matches the names we use for the other workflows: Check, Distribution, Test

mike-wendt: I will call out that I agree with you this should not be in Check as check is code driven and the PR title is a validation for me. I think was we grow we'll likely have another validation so let's keep that separate from the code for now. If we need to we can combine them in the future. Hopefully that's helpful

mike-wendt (approved):
Approving so you can make the requested changes and we can update `dev` branch protections once merged to enforce this. Thanks for doing this, I was about to implement this to complement the doc changes but focused on those first, so I really appreciate the proactiveness to help. Thanks again @johanpel

mike-wendt:
Here are the current required checks, I will take a TODO to update this once merged
Screenshot 2026-08-03 at 14 39 04

.github/workflows/semantic-pr.yml
johanpel: >  I would rename the file to validate.yml instead.

Done in: https://github.com/sirius-db/sirius/pull/1384/commits/c21b8109d8bacd6d734279a81c8869d9c7bafae6

mike-wendt:
This is now required

PR #1401, opened 2026-08-04
fix(s3): surface object ETags so DuckDB 1.5.5 caches S3 reads again

ran-yuan-rui:
## Problem

DuckDB 1.5.5 no longer caches external files without validation metadata. `sirius_httpfs` returned neither a version tag nor a usable modification time, so S3 reads bypassed DuckDB's external file cache and went back to the network.

The existing footer-bind tests exposed this as three blocking GETs per glob scan instead of zero.

## Fix

Use the S3 object's ETag as DuckDB's version tag:

- Capture the ETag from the existing HEAD and validated 206 footer-probe responses.
- Keep retry attempts isolated. A 200, 416, or invalid 206 discards the probe tag; the fallback HEAD supplies both size and tag.
- Store the ETag on `rest_io_object` and return it through `sirius_httpfs::GetVersionTag`.
- Preserve the quoted ETag. If no ETag is available, DuckDB keeps its safe uncached behavior.
- Leave `GetLastModifiedTime` unchanged.

This adds no network requests and does not touch the asynchronous column-chunk read path.

## Tests

- The existing footer-bind regression checks pass unchanged.
- Covered HEAD, suffix-probe, missing-ETag, retry, and fallback behavior.
- Verified EFC invalidation by caching a range, overwriting the object with same-size different bytes, then checking for a new backend GET and the new contents. Sirius prefetch is disabled in this test to isolate DuckDB's EFC.

## Validation

- Full Catch2: 2,317 cases, 32,706,601 assertions
- `make s3-test`: 92/92, 31,433 assertions
- `s3-test-large`: 5/5 + 3/3
- `s3-bench`: passed
- `pre-commit --all-files`: passed
- `git diff --check`: passed

## Follow-up

The reactor and `io_object` parts must be ported to cuCascade before the #1340 pin update. `sirius_httpfs::GetVersionTag` remains in Sirius.

### DIFF
--- a/src/include/io/rest/rest_reactor.hpp
+++ b/src/include/io/rest/rest_reactor.hpp
@@ -62,6 +62,15 @@ struct footer_probe {
   std::size_t object_size{0};
   std::size_t window_lo{0};
   std::shared_ptr<const std::vector<std::uint8_t>> bytes;
+  // ETag from the verified 206, quotes preserved; empty otherwise.
+  std::string etag;
+};
+
+/// Result of a blocking HEAD: the object's size plus its ETag when the server
+/// sent one (quotes preserved, empty otherwise).
+struct head_object_result {
+  std::size_t object_size{0};
+  std::string etag;
 };
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


src/include/io/rest/rest_reactor.hpp
```diff
@@ -77,8 +86,13 @@ struct footer_probe {
...
  public:
-  rest_io_object(std::string path, std::string bucket, std::string key, size_t size)
-    : _path(std::move(path)), _bucket(std::move(bucket)), _key(std::move(key)), _file_size(size)
+  rest_io_object(
+    std::string path, std::string bucket, std::string key, size_t size, std::string etag = {})
+    : _path(std::move(path)),
+      _bucket(std::move(bucket)),
+      _key(std::move(key)),
+      _file_size(size),
+      _etag(std::move(etag))
```
aminaramoon, line 95: I think metadata might be a better place for this, don't you think so ?

aminaramoon (approved):

src/include/io/rest/rest_reactor.hpp
ran-yuan-rui: Discussed on slack - I considered the metadata store, but this ETag is per-open transport metadata rather than parsed Parquet metadata. DuckDB asks for GetVersionTag immediately after opening the handle, including for non-Parquet range reads, before the metadata store necessarily has an entry.

Issue #1338, opened 2026-07-30
Run some or all of the compression tests in CI


joosthooz:
Currently only integration-level unit tests are running in Sirius, that test the compression code end-to-end. The lower-level unit tests of the compression subsystem itself are not being invoked yet. We should run those in CI too.

Note:
1) There are 2 exhaustive tests that exercise every possible operator chain (1 only for fusable operators, the other for all operators), so for those we need to set it to a sensible chain depth to prevent the runtime from ballooning. There already exist env vars for that. We can even do larger runs in the weekend, for example. 

2) fusable chains are JIT-compiled and stored in an on-disk cache. That cache needs to hit between test runs, otherwise they will run for a very long time (even though those compilations have been parallelized in the test implementations). Then there's the question of when/how often do we want to clear that cache to test if the JITting itself still works.

PR #1403, opened 2026-08-04
feat: Add FSST compression operator

joosthooz:
## Description
Add a GPU-optimized FSST compressor operator to the simpatico compression subsystem. For typical string columns, it provides better compression ratio and much faster decode performance (like 4x higher, 900 - 1100 GB/s on a GB300) compared to libcudf's dictionary and nvcomp's LZ algorithms.

### Build integration

  The codec sources come from one pinned commit of
  [`joosthooz/fsst-gpu@c86c042`](https://github.com/joosthooz/fsst-gpu) (branch
  `gpu-decompressor`) — no local checkout needed:

  - **pixi / standalone-simpatico builds**: `FetchContent` of the pinned tarball
    (same pattern as cuco).
  - **vcpkg build**: a source-only overlay port (`vcpkg_ports/fsst-gpu`) that
    installs headers + `.cu` sources and exports `FSST_GPU_ROOT`.

  Either way simpatico compiles the sources itself into a `fsst_gpu` static
  library (rdc, device symbols resolved internally), so `CMAKE_CUDA_ARCHITECTURES`
  and the `GTSST_SPLIT_COUNT` knob apply uniformly across build flavors.

### Decode-path performance (from profiling q13/q19)

  - **Stream-scoped decompress** (fsst-gpu side): header reads, metadata upload
    and the decode kernel all run on the caller's stream — previously each decode
    did blocking memcpys, 4 cudaMalloc/cudaFree and a `cudaDeviceSynchronize`,
    serializing the whole device (114 device-wide syncs in one profiled q13
    execution → 1 after). 
  - **Zero-copy string reassembly** (`decompress_consume`): the decode walk moves
    freshly-decoded offsets/chars into `make_strings_column` instead of
    deep-copying them — removes an O(decoded bytes) D2D pass for *every*
    compressed string column (snappy/ans too, not just fsst). Profiled D2D volume
    on q13 dropped from 356GB to 120GB (baseline level).

### Results

  TPC-H SF1000 on GB300, compression enabled, `--mode grouped --pin gpu`,
  3-iteration medians, all 22 query results byte-identical to the baseline run:

  | | baseline plans | this PR |
  |---|---|---|
  | TPC-H total (22 queries) | 13.95s | **13.37s (-4.2%)** |

  Biggest wins: q19 **-32.1%**, q4 **-20.0%**, q2 **-16.2%**, q12 **-14.7%**,
  q10 -6.1%, q20 -5.3%, q8 -5.1%. Accepted trade: q21 +2.6% — its orders pin now
  actually compresses (fsst) instead of falling back to raw via the dictionary
  failure, costing decode time but saving pin footprint. No other query is more
  than ~1% from baseline (q15's apparent +4% did not reproduce in isolated A/B —
  session-sequence noise).

## Checklist
- [ ] Read CONTRIBUTING.md


Jedi18 (commented):
Looks good, though we should add FSST tests?

joosthooz:
> Looks good, though we should add FSST tests?

Thanks, the tests use sweeps across all operators, but I looked into it and this new operator was actually missing! So I added it and expanded the string tests a bit. Sirius has some integration tests that only do 1 operator, and simpatico has internal tests too that sweep through all possible chain combination up to a certain depth. Those do not run in sirius CI, and there's an issue to consider if we should add them #1338

joosthooz:
There's a note here with the latest commits that add a fix for dictionary-encoding very large columns. That crashes on current dev, making a couple of columns fail compression and revert to identity. The 8.1s branch actually relies on that, so merging this makes it a tiny bit slower. To fix that, we need to change o_orderpriority to identity.

joosthooz:
And an open question for this PR: I've chosen to pull in the kernel code from my fork of the fsst-gpu repo (and even a branch of that). Do we want to keep that or just copy the code into simpatico?

Jedi18:
> There's a note here with the latest commits that add a fix for dictionary-encoding very large columns. That crashes on current dev, making a couple of columns fail compression and revert to identity. The 8.1s branch actually relies on that, so merging this makes it a tiny bit slower. To fix that, we need to change o_orderpriority to identity.

This fix is also included in Felipe's branch right?

Jedi18:
> And an open question for this PR: I've chosen to pull in the kernel code from my fork of the fsst-gpu repo (and even a branch of that). Do we want to keep that or just copy the code into simpatico?

Copy the code into simpatico? Maybe you can ask on slack to get more people's opinions

src/compression/simpatico_codegen/src/operators/fsst_compressor.cu
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  auto const* base = static_cast<const std::uint8_t*>(payload_data());
+
+  // Stream-scoped header reads (a blocking cudaMemcpy here would serialize the whole device).
+  FrameHeader fh{};
+  check_cuda(cudaMemcpyAsync(&fh, base, sizeof(fh), cudaMemcpyDeviceToHost, stream.value()),
+             "read frame header");
+  stream.synchronize();
+  size_t const n_chunks     = fh.n_chunks;
```
Jedi18, line 227: Could we validate the frame metadata before decoding? 

Please check the header/directory bounds, `total_uncomp`, expected chunk count, each offset/size against `payload_size()`, and each decoded `out_size`. Corrupt metadata could otherwise cause out-of-bounds GPU access.

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -117,12 +117,68 @@ add_custom_command(
...
+    512
+    CACHE STRING "fsst-gpu splits (tiles) per block")
+if(VCPKG_BUILD)
+  # Provides FSST_GPU_ROOT (see vcpkg_ports/fsst-gpu).
+  find_package(fsst-gpu CONFIG REQUIRED)
+else()
+  include(FetchContent)
+  FetchContent_Declare(
+    fsst_gpu_src
+    URL https://github.com/joosthooz/fsst-gpu/archive/c86c0426a7fd2e3568ffcf6e6122d885867aaad4.tar.gz
```
Jedi18, line 138: In `fsst-gpu/src/compressors/compactionv5t/compaction-decompress.cu` (https://github.com/joosthooz/fsst-gpu/blob/c86c0426a7fd2e3568ffcf6e6122d885867aaad4/src/compressors/compactionv5t/compaction-decompress.cu#L252-L272)

Is this cache safe for multi-GPU use? The shared-memory limit and cudaFuncSetAttribute configuration are device-specific, but this initializer runs only once process-wide. Could we cache and initialize the ceiling per CUDA device?

joosthooz: This is a good point, and it happens in another place in simpatico too (if a long codegen chain can run with some more smem). We don't even know if this has a negative effect on the rest of query execution, so that's the least we should look into

joosthooz: BTW I just looked into this and this setting is specific to the kernel, not device-wide, so it should create problems

mbrobbel:
Marking as draft. There are some conflicts.

PR #1412, opened 2026-08-05
fix(config): honor loaded YAML defaults and reject zero hash partitions

infblueocean:
## Summary

- register DuckDB extension defaults from the Sirius operator and compression configuration already loaded from YAML
- preserve current hardware-derived and clamped defaults when YAML omits those values
- reject `hash_partition_bytes == 0` at YAML validation, DuckDB `SET`, and partition-strategy boundaries
- add direct reset, inheritance, invalid-YAML, and strategy-guard tests

## Why

The YAML loader accepts operator and compression defaults, but DuckDB setting registration currently substitutes compile-time constants. A later `RESET` can therefore discard the loaded configuration. Separately, a zero hash-partition target is not a usable configuration and reaches arithmetic that requires a positive partition size. This makes the loaded YAML the registered source of truth and rejects zero at each reachable boundary.

## Validation

Validated from a clean checkout based exactly on `dev` `53b1eb9891a474f63579310f6afaab9a3f24c7db`:

- `/home/ubuntu/.pixi/bin/pixi run make -j8 release` — passed all `1276/1276` actions
- `sirius_unittest "Sirius configuration rejects zero hash partition bytes"` — 1 assertion in 1 test case passed
- `sirius_unittest "DuckDB setting rejects zero hash partition bytes without a Sirius context"` — 3 assertions in 1 test case passed
- `sirius_unittest "YAML-backed operator and compression settings are DuckDB defaults"` — 57 assertions in 1 test case passed
- `sirius_unittest "[hash_join][build_probe][unit][config]"` — 1 assertion in 1 test case passed

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -352,7 +352,7 @@ batch default**. Each can still be overridden individually.
 | `scan_task_batch_size` | 2.5% of GPU mem (512 MiB – 5 GiB) | Target batch size for DuckDB scan tasks |
 | `enable_compressed_materialization` | true | Store eligible integer and fixed-point DECIMAL values in value-preserving narrower carriers when exact pin-time bounds permit it; restore native carriers at type-sensitive boundaries. |
 | `max_sort_partition_bytes` | 0 (auto) | Max bytes per sort partition. Auto = 33% of GPU memory. |
-| `hash_partition_bytes` | 2.5% of GPU mem (512 MiB – 5 GiB) | Target partition size for hash joins and group-bys |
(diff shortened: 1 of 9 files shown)


infblueocean:
@johanpel, could you review this configuration-default repair? It makes loaded YAML authoritative and rejects zero hash partitions; required checks are green.

johanpel:
This looks good to me but it also makes me realize there is an asymmetry in how the configs are validated, since YAML loaded defaults aren't checked with the same rules as the SET callbacks, but this can be addressed separately as it is a bit more involved: https://github.com/sirius-db/sirius/issues/1436

johanpel (approved):

PR #1285, opened 2026-07-24
CI: Surface build and test errors in GHA job summary

mike-wendt:
## Summary

Build failures produce long ninja output where the actual compiler error appears well above the `ninja: build stopped` line — engineers have to scroll to find it. This PR pipes build and test output through `tee` and writes the key error lines to the **GitHub Actions Job Summary** on failure, so they appear in the summary panel at the top of any run without scrolling.

### What's captured

| Step | Patterns surfaced |
|---|---|
| Build (`test.yml` + `check.yml` non-vcpkg) | `FAILED:`, `: error:`, `: fatal error:` |
| Build (`check.yml` vcpkg) | Same, from existing `/tmp/vcpkg.log` |
| C++ unit tests (`test.yml`) | Catch2 `FAILED:` block with 5 lines of context before (test name) and 10 after (assertion + expansion) |
| Linter | Catch failed linter errors and raise them |

Output is capped at 60/80 lines so a cascading header change doesn't flood the summary.

## Test runs

Within this PR I've made code changes to trigger build, test, and linter failures to test the visibility changes. Below are examples of the results:

### Build failures - [Check](https://github.com/sirius-db/sirius/actions/runs/30275270816) & [Test](https://github.com/sirius-db/sirius/actions/runs/30275270812/attempts/1#summary-90007623207) workflows
Screenshot 2026-07-27 at 15 35 15
Screenshot 2026-07-27 at 15 43 42

### Test failure - [Test](https://github.com/sirius-db/sirius/actions/runs/30275691950/attempts/1#summary-90010001289) workflow
Screenshot 2026-07-27 at 15 41 46

### Linter failure - [Check](https://github.com/sirius-db/sirius/actions/runs/30299490923/attempts/1#summary-90088641914) workflow
Screenshot 2026-07-27 at 15 48 51

## No behavior change

- `tee` is transparent — build/test behavior is identical.
- `set -o pipefail` ensures the step still fails when `make`/the test binary exits non-zero.
- The full scrollable log still exists in each step as before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1413, opened 2026-08-05
test(perf): make the thread sweep isolated and fail-closed

infblueocean:
## Problem

The thread-sweep runner rewrites the tracked integration configuration, publishes results into a shared directory, and can continue after missing queries, failed DuckDB executions, or missing timer output. A failed sweep can therefore leave repository state behind or publish incomplete timings as evidence.

The scan-thread parameter is also accepted by the runner but not written into the generated configuration, so the existing scan sweep does not actually vary scan threads.

## Changes

- Render every sweep configuration into a temporary file selected through `SIRIUS_CONFIG_FILE`; never modify the tracked integration YAML.
- Write `scan_manager.num_threads` and replace the invalid two-thread scan cell with the minimum accepted value of three.
- Validate all thread counts and required inputs before running.
- Fail on missing queries, DuckDB errors, or anything other than exactly two timer values.
- Write each CSV through a `.partial` file and publish it only after the configuration completes.
- Use a unique timestamped output directory by default, with explicit path overrides for automation.
- Remove the runner’s dependencies on GNU `grep -P`, `readarray`, and `bc`.
- Add a CPU-only shell regression covering all eight configurations, temporary-file isolation and cleanup, invalid scan-thread rejection, tracked-config preservation, and DuckDB failure behavior.
- Run that regression as the dependency-free first step of the `Check` workflow.
- Update the performance-test documentation to describe the isolated configuration and output behavior.

## Validation

Based on Sirius `dev` at `53b1eb9891a474f63579310f6afaab9a3f24c7db`.

```text
git diff --check
bash -n test/tpch_performance/sweep_threads.sh
bash -n test/tpch_performance/test_sweep_threads.sh
bash test/tpch_performance/test_sweep_threads.sh
  -> thread sweep harness tests passed
```

The updated `check.yml` also parses successfully as YAML. The regenerated full-index diff matches the reviewed patch at SHA-256 `3194b4270f2bebe4048ef884359125d6e898af5dc89c2361faa139af3f74fe2a`.

## Scope

This is harness reliability work. It does not run a GPU benchmark, select thread values, or change any Sirius runtime default. Full lint and build checks remain for PR CI.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -30,6 +30,9 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+      - name: Test thread sweep harness
+        run: bash test/tpch_performance/test_sweep_threads.sh
+
       - uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
         with:
           cache: false
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -31,7 +31,17 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Test thread sweep harness
-        run: bash test/tpch_performance/test_sweep_threads.sh
+        run: |
+          set -o pipefail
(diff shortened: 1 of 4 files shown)


mike-wendt (approved):
Thanks @infblueocean, I need to make two small changes for this to work with #1285 then we can merge

mike-wendt:
Changes, made testing changes and will merge when ✅

PR #1414, opened 2026-08-05
fix(testing): gate fallback injection behind explicit test opt-in

infblueocean:
## Summary

- omit `sirius_test_inject_transparent_gpu_error` from normal Sirius setting registration
- expose the injection hook only when the process explicitly sets `SIRIUS_ENABLE_TEST_OPTIONS=1`
- keep non-exact values such as `true` fail-closed
- opt the unittest process in before constructing its shared databases so runtime-fallback coverage remains intact
- directly test normal absence and `SET` rejection, non-exact absence, and exact opt-in presence

## Why

`sirius_test_inject_transparent_gpu_error` exists only to force a deterministic runtime GPU failure for transparent-fallback tests. Registering it in every normal database exposes an implementation test hook as a user-facing configuration choice. This keeps the hook available to the test harness without advertising or accepting it in production sessions.

## Validation

Validated from a clean isolated checkout based exactly on `dev` `53b1eb9891a474f63579310f6afaab9a3f24c7db`:

- `/home/ubuntu/.pixi/bin/pixi run make -j8 release` — passed all `1276/1276` actions
- `sirius_unittest "[sirius][config][test-settings][isolated_context]"` — 13 assertions in 1 test case passed
- `sirius_unittest "[transparent][fallback][integration]~[s3mix]"` — 89 assertions in 9 test cases passed

The surface test covers unset, non-exact `true`, and exact `1` process values. Both selectors exited zero on the existing AWS NVIDIA L4; no new instance was created.

### DIFF
--- a/test/cpp/config/test_context.cpp
+++ b/test/cpp/config/test_context.cpp
@@ -42,6 +42,7 @@
 #include <filesystem>
 #include <fstream>
 #include <iostream>
+#include <optional>
 #include <set>
 #include <source_location>
 #include <string>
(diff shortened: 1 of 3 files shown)


infblueocean:
@mike-wendt, could you review this test-surface repair? It gates fallback injection behind explicit test opt-in; required checks are green.

test/cpp/config/test_context.cpp
```diff
@@ -121,6 +122,60 @@ struct finally {
...
+  if (auto const* value = std::getenv("SIRIUS_ENABLE_TEST_OPTIONS")) {
+    original_test_options = value;
+  }
+  finally restore_env{[original_test_options]() {
+    if (original_test_options) {
+      setenv("SIRIUS_ENABLE_TEST_OPTIONS", original_test_options->c_str(), 1);
+    } else {
+      unsetenv("SIRIUS_ENABLE_TEST_OPTIONS");
+    }
+    setenv("SIRIUS_DISABLE", "1", 1);
```
mike-wendt, line 138: nit: this doesn't restore the value of `SIRIUS_DISABLE` and always sets it to 1. It may be worth restoring this like `SIRIUS_ENABLE_TEST_OPTIONS`

mike-wendt (approved):

infblueocean:
Thanks for the approval, @mike-wendt. All required checks are green and the PR is mergeable, but GitHub rejects my attempt to enqueue it because the `dev` merge queue requires collaborator permission. Could you enqueue/merge it when ready? #1418 overlaps only in `test/cpp/config/test_context.cpp`; whichever lands second, I will promptly rebase and resolve that focused test composition on our branch.

Issue #1416, opened 2026-08-05
SIGSEGV on `WHERE <col> IS NULL` over a parquet scan


Jedi18:
## Summary

Two predicate shapes crash the process with a SIGSEGV during parquet row-group
pruning: an `IS NULL` test, and a bare boolean column used as the whole
predicate. This is a hard crash, not an exception — it takes down the whole
DuckDB session (and, in CI, the entire test binary).

## Reproduction

```sql
COPY (SELECT i AS id,
             CASE WHEN i % 3 = 0 THEN NULL ELSE i END AS v,
             CASE WHEN i % 2 = 0 THEN (i % 4 = 0) END AS flag
      FROM range(1, 6001) t(i))
  TO '/tmp/repro.parquet' (FORMAT PARQUET, ROW_GROUP_SIZE 2048);

SELECT count(*) FROM read_parquet('/tmp/repro.parquet') WHERE v IS NULL;  -- SIGSEGV
SELECT count(*) FROM read_parquet('/tmp/repro.parquet') WHERE flag;       -- SIGSEGV
```

Both fault in `filter_row_groups_with_stats`, reached from
`parquet_gpu_ingestible::build_file_scan_info`. Note the process exits with
status **1, not 139** — Sirius catches the signal, prints a backtrace, and
exits normally.

PR #1417, opened 2026-08-05
fix(scan): don't push stats-filter-incompatible predicates into parquet pruning

Jedi18:
## Description

Two predicate shapes segfault during parquet row-group pruning:

```sql
SELECT count(*) FROM read_parquet('f.parquet') WHERE v IS NULL;   -- SIGSEGV
SELECT count(*) FROM read_parquet('f.parquet') WHERE flag;        -- SIGSEGV
```

This pushes only the safe part of the predicate, and adds regression tests.

**Cause.** cuDF's statistics filter evaluates a predicate against per-row-group stats by rewriting every column reference into that row group's `(min, max)` pair. Two shapes have no valid rewrite and fault rather than mis-prune:

- a **null test** — asks about the *absence* of values, not their range; answerable only from `null_count`;
- a **bare column reference used as the predicate** — the rewrite leaves two stats columns with no comparison over them.

Neither was filtered out before reaching `filter_row_groups_with_stats`. `convert_table_filters_to_expression` drops a top-level `IS_NOT_NULL` table filter but has no case for `IS_NULL`, and nothing looked at bare references at all.

**Fix.** Push only the top-level AND conjuncts that are safe. `v IS NULL AND id > 3000` still prunes on `id`; the unsafe conjunct is applied post-decode. Dropping conjuncts can only keep extra row groups, never drop wanted ones — nothing failing `id > 3000` can satisfy the full predicate.

The bare-reference rejection deliberately **does not recurse**; only a reference that *is* the whole predicate faults. Measured, not assumed:

| Predicate | Result |
|---|---|
| `WHERE flag` | crash |
| `WHERE flag_dense` (no NULLs) | crash |
| `WHERE flag = true` | ok |
| `WHERE flag AND v > 5000` | **ok** |
| `WHERE NOT flag` / `IN` / `BETWEEN` | ok |

So nullability is irrelevant and a nested bare reference is fine — rejecting those too would forfeit pruning that works. A test pins each of the passing shapes.

**Correctness.** When the push is partial the scan is reported `UNFILTERED` rather than `ROW_FILTERED`, so `post_filter_and_project` (or the inline hive branch) applies the full predicate. Reporting it as filtered would silently drop the stripped conjuncts — that bookkeeping is the part most worth reviewing.

Gated separately from `disable_filter_pushdown`, which also suppresses dynamic join filters; those build their own AST from join-key bounds and stay safe to push down.

**Cost.** Only the stripped conjuncts lose pruning. Verified on a 3-row-group file:

```
WHERE id > 5000                →  Row group pruning: 3 -> 1 row group(s)
WHERE v IS NULL AND id > 5000  →  Row group pruning: 3 -> 1 row group(s)
```

**Follow-up (not in this PR).** `null_count` can prune the null conjuncts too — skip a row group when `null_count == 0` for `IS NULL`, or `null_count == num_rows` for `IS NOT NULL`. The proper long-term fix is upstream: teach cuDF's `stats_expression_converter` to handle both shapes.

## Checklist

- [x] Read CONTRIBUTING.md

## References
- https://github.com/sirius-db/sirius/issues/1416


test/cpp/integration/test_parquet_null_predicate_pushdown.cpp
```diff
@@ -0,0 +1,314 @@
...
+}
+
+// A nested IS NOT NULL survives convert_table_filters_to_expression (which only
+// drops it as a column's top-level filter) and lowers to IS_NULL + NOT in the
+// cuDF AST, reaching the same stats-filter path.
+//
+// NOTE: any comparison on a column already implies it is non-null, so DuckDB's
+// filter combiner may fold the IS NOT NULL away before it ever reaches us. The
+// queries are correct either way; to confirm the branch was actually taken,
+// check for the "pushing only the null-free conjuncts" debug line.
```
johanpel, line 232: Nit: Can't we somehow assert whether this happened?

test/cpp/integration/test_parquet_null_predicate_pushdown.cpp
```diff
@@ -0,0 +1,314 @@
...
+        REQUIRE(r);
+        if (r->HasError()) { UNSCOPED_INFO("fixture setup error: " << r->GetError()); }
+        REQUIRE_FALSE(r->HasError());
+      }
+    }
+
+    scan_ = "read_parquet('" + pq_path + "')";
+  }
+
+  ~ParquetNullPredicateFixture() { fs::remove_all(dir_); }
```
johanpel: Nit: Perhaps use the non-throwing error_code overload here to avoid throwing from the destructor

test/cpp/integration/test_parquet_null_predicate_pushdown.cpp
```diff
@@ -0,0 +1,314 @@
...
+        REQUIRE(r);
+        if (r->HasError()) { UNSCOPED_INFO("hive fixture setup error: " << r->GetError()); }
+        REQUIRE_FALSE(r->HasError());
+      }
+    }
+
+    scan_ = "read_parquet('" + root + "/part=*/*.parquet', hive_partitioning=true)";
+  }
+
+  ~ParquetHiveNullPredicateFixture() { fs::remove_all(dir_); }
```
johanpel: Same here

johanpel (approved):
Looks good, just some nits:

test/cpp/integration/test_parquet_null_predicate_pushdown.cpp
Jedi18: We can but it's not a trivial change since I'll have to either add public accessors to parquet_gpu_ingestible for the sole purpose of unit tests or alternatively (preferably) moving the helpers behind a detail:: header makes them testable directly, but that's a bigger change.

There's a follow-up PR on null_count pruning that I'm working on, maybe I can do it as part of that

PR #1418, opened 2026-08-05
refactor(config): register legacy executor knobs only in legacy builds

infblueocean:
## Summary

- register ten process-global executor settings only when `SIRIUS_ENABLE_LEGACY` is compiled
- keep active Super Sirius settings present in both normal and legacy builds
- move the ten settings into an explicit legacy-release documentation section with their source defaults
- add one build-conditional surface test for normal absence/rejection versus legacy presence/`SET`/`RESET`

These ten settings only have runtime consumers on the legacy `gpu_processing` path. Normal Super Sirius builds currently expose them even though changing them cannot affect that path.

## Compatibility

This intentionally changes the normal-build configuration surface. The ten names previously registered successfully even though only `src/legacy/` consumed them. Normal builds now omit them from `duckdb_settings()` and reject `SET`; existing normal-build config or startup SQL using them must remove those inert names. Builds with `ENABLE_LEGACY_SIRIUS=ON` retain the settings and their existing defaults.

No files under `src/legacy/` are modified.

## Validation

Based on clean Sirius `dev` at `53b1eb9891a474f63579310f6afaab9a3f24c7db`.

- regenerated full-index diff matches the reviewed patch at SHA-256 `1d55991ba0c5f2dd0bd3e8af869b867a30a79e87f1e8716b1bb0d52e4a2af3d1`
- normal `release` build: `1276/1276` actions passed
- normal selector `Legacy-only settings follow the build surface`: 65 assertions in 1 test case passed
- guarded `src/sirius_extension.cpp` compiled successfully with `SIRIUS_ENABLE_LEGACY`
- shim-assisted legacy main target: final `611/611` actions passed
- shim-assisted legacy `sirius_unittest`: `181/181` actions passed
- legacy selector `Legacy-only settings follow the build surface`: 85 assertions in 1 test case passed

### Legacy-release baseline limitation

The clean canonical legacy build exposed a pre-existing stale include in untouched upstream source:

```text
src/legacy/expression_executor/specializations/gpu_execute_function.cpp:20:10:
fatal error: expression_executor/regex/regex_playground.hpp: No such file or directory
```

That file is byte-identical at base and review head, and this MR has zero `src/legacy/` diff. The current header lives under `expression_evaluator/regex/`. Because contributor rules prohibit editing `src/legacy/`, validation continued with an external temporary forwarding header supplied only through `CPLUS_INCLUDE_PATH`; the review checkout remained exact and clean. The legacy assertion result therefore validates this MR's registration surface under that disclosed compatibility seam, but does not claim live `dev` has a clean canonical `legacy-release` build.


johanpel (approved):
Thanks @infblueocean

infblueocean:
#1418 is now conflict-free at `407f860` after a normal merge of current `dev`
(`f93f599`); the conflict resolution keeps the new upstream defaults setup
immediately before this PR's legacy-only guard.

Scoped validation on the exact tree passed:

- normal release selector: 65 assertions
- legacy release selector: 85 assertions
- final review tree: `1b01b812218f8728fd2b73bc55d2cef1668a6be2`

The legacy validation used the previously disclosed external forwarding header
for the unchanged stale include path; this PR still changes nothing under
`src/legacy/`.

Check, Test, and Distribution are currently `action_required` at the
fork-workflow gate. Our account cannot approve them (GitHub returns the
repository-admin guardrail). @johanpel, could you approve those workflow runs
when convenient?

Issue #1415, opened 2026-08-05
Add parquet null materialization integration tests


Jedi18:
We don't have any tests for null coverage for the parquet reader

PR #1419, opened 2026-08-05
test: add parquet null materialization integration tests

Jedi18:
~Note that this is stacked on top of https://github.com/sirius-db/sirius/pull/1417, which should be merged before this PR.~ That PR is merged
## Description

Existing NULL coverage (`test_gpu_execution_allnull_scan`, `test_gpu_execution_filter_nulls`, and related tests) reads through the DuckDB-native scan. Parquet represents nullability using definition levels rather than a validity bitmap, leaving its decode paths untested.

Adds `test_gpu_execution_parquet_nulls.cpp` with 6 fixtures and 27 test cases:

| Fixture | Coverage |
|---|---|
| `ParquetNullFixture` | Wholly-NULL and partially-NULL columns across INT32, INT64, DOUBLE, FLOAT, BOOLEAN, BYTE_ARRAY, DATE, TIMESTAMP, and DECIMAL at INT32/INT64 widths |
| `ParquetDictNullFixture` | Dictionary-encoded columns containing NULL entries |
| `ParquetMultiRowGroupFixture` | NULLs spanning multiple row groups with a ragged tail |
| `ParquetDenseColFixture` | OPTIONAL columns containing no NULLs, guarding against spurious NULLs |
| `ParquetNullRunFixture` | Constant-valid run followed by NULLs, and a long NULL prefix followed by valid values |
| `ParquetFlbaDecimalNullFixture` | `FIXED_LEN_BYTE_ARRAY` decimal columns with NULLs and reader-side pushdown disabled |

Across the suite, queries cover:

- Projection of wholly and partially NULL columns
- `COUNT(*)` versus `COUNT(col)`
- `SUM`, `MIN`, and `MAX`
- `IS NULL` and `IS NOT NULL`
- Nullable comparisons and filters
- `COALESCE` and `CASE`
- `GROUP BY` with a NULL key
- Column-pruned scans projecting nullable columns alone
- FLBA decimal filtering through the post-decode path

**Fixture shapes are asserted, not assumed.** Construction validates physical types, logical annotations, decimal precision and scale, dictionary encoding, expected NULL populations, and exact row-group sizes using `parquet_schema()`, `parquet_metadata()`, and CPU-decoded checks. Changes to DuckDB’s writer behavior therefore fail loudly instead of silently exercising a different path.

Column types are declared on tables rather than inferred from a bare `COPY (SELECT ...)`. This matters because `range()` produces BIGINT and literals such as `1.5` are DECIMAL; inference would otherwise collapse INT32 onto INT64 and DOUBLE onto DECIMAL.

The multi-row-group fixture writes well over 2048 rows and asserts the exact layout of four 2048-row groups followed by a 203-row tail. `ROW_GROUP_SIZE` is not a hard cut because DuckDB flushes complete buffered DataChunks.

Files are generated by a `SIRIUS_DISABLE=1` DuckDB so the extension callback does not run during `COPY`. Queries use the standard `compare_gpu_vs_cpu` harness, which requires actual GPU execution with no fallback and compares results against DuckDB CPU execution.

### Known gaps

Both require fixtures from a non-DuckDB writer:

- INT96 timestamps
- REQUIRED (non-nullable) columns

DuckDB writes every top-level column as OPTIONAL regardless of `NOT NULL`, so a column without a definition-level stream cannot be generated by these fixtures.

## Checklist

## References

- https://github.com/sirius-db/sirius/issues/1415


test/cpp/integration/test_gpu_execution_parquet_nulls.cpp
```diff
@@ -0,0 +1,1182 @@
...
+  // Assert the file contains at least @p expected row groups.
+  //
+  // DuckDB's ROW_GROUP_SIZE is NOT a hard cut: ParquetWriteSink appends whole
+  // DataChunks (up to STANDARD_VECTOR_SIZE = 2048 rows) to a buffer and only
+  // then checks `buffer.Count() >= row_group_size`, and Flush writes the entire
+  // buffer as ONE row group. So the effective minimum row group is the arriving
+  // chunk size (2048) regardless of how small ROW_GROUP_SIZE is set -- asking
+  // for 50 on a 203-row table yields a single 203-row row group. Multi-row-group
+  // fixtures must therefore write well over 2048 rows AND assert the result.
+  void assert_min_row_groups(const std::string& pq_path, std::size_t expected) const
```
joosthooz: This function is dead, see if it can be removed

test/cpp/integration/test_gpu_execution_parquet_nulls.cpp
```diff
@@ -0,0 +1,1182 @@
...
+/// Sets SIRIUS_DISABLE=1 for the duration of a scope, then restores whatever
+/// was there before -- including the harness's own "1".
+///
+/// A bare setenv/unsetenv pair is wrong twice over: the harness deliberately
+/// keeps SIRIUS_DISABLE=1 so untagged tests' DuckDB instances do not
+/// auto-initialize a SiriusContext (see scoped_sirius_disable_clear in
+/// test_plan_printer.cpp), so unsetting it leaks a changed global into later
+/// tests; and the REQUIREs below throw on failure, which would skip the unset
+/// entirely. Restoring in a destructor fixes both.
+struct scoped_sirius_disable_set {
```
joosthooz: Duplicate in test_parquet_null_predicate_pushdown.cpp

test/cpp/integration/test_gpu_execution_parquet_nulls.cpp
```diff
@@ -0,0 +1,1182 @@
...
+// Fixture 6 — FLBA / BYTE_ARRAY decimal with nulls
+//
+// DuckDB picks a decimal's parquet physical type by precision: INT32 up to 9,
+// INT64 up to 18, and FIXED_LEN_BYTE_ARRAY beyond that. Sirius disables
+// reader-side row-group filter pushdown when an FLBA / BYTE_ARRAY decimal is
+// among the scanned columns, because cudf's stats filter cannot compare a
+// fixed_point_scalar literal against those stats -- the filter is instead
+// applied post-decode (parquet_gpu_ingestible.cpp).
+//
+// Every other fixture here uses DECIMAL(10,2), which lands on INT32/INT64 and
```
joosthooz: Nit: looks like 10,2 is not used, so those numbers are likely out of date

test/cpp/integration/test_gpu_execution_parquet_nulls.cpp
```diff
@@ -0,0 +1,1182 @@
...
+
+  // Convenience overload for a single statement.
+  void write(const std::string& stmt) const { write(std::vector<std::string>{stmt}); }
+
+  std::string path(const std::string& name) const { return (dir / name).string(); }
+
+  /// A single-quoted SQL literal with embedded quotes doubled. TMPDIR may
+  /// legally contain one, which would otherwise terminate the literal early and
+  /// break every generated statement.
+  static std::string sql_literal(std::string const& value)
```
joosthooz: This function can be deduplicated into a utils file, it looks like there are some other files that have very similar code

test/cpp/integration/test_gpu_execution_parquet_nulls.cpp
```diff
@@ -0,0 +1,1182 @@
...
+  ///
+  /// Catch2 rebuilds a fixture for every TEST_CASE_METHOD that uses it, and the
+  /// constructors here run ~20 metadata assertions each. Opening a database per
+  /// assertion meant several hundred in-memory DuckDB initialisations for a
+  /// single fixture; one connection, reused, makes it one per construction.
+  ///
+  /// SIRIUS_DISABLE only matters while the instance is being created — that is
+  /// when the extension callback would attach a SiriusContext — so the guard
+  /// does not need to outlive the constructor.
+  explicit ParquetFileGuard(std::string tag)
```
joosthooz: This would be nice to lift into a utils file, there's several other places/files where very similar code lives (Claude says: test_parquet_null_predicate_pushdown.cpp (twice), test_transparent_runtime_fallback.cpp:370 (S3MixFixture), test_gpu_execution_multi_format.cpp:499 (HivePartitionDataset), test_pin_table_zone_map_pruning.cpp:233, test_sirius_read_parquet_scan.cpp:108, test_duckdb_native_dynamic_filter_remap.cpp:157)

joosthooz (changes requested):
Thanks for this, I just noticed some deduplication opportunities. Let me know what you think. Also consider having a look at S3MixFixture to see if that could meet your needs for generating parquet, it could allow removing some more code

test/cpp/integration/test_gpu_execution_parquet_nulls.cpp
Jedi18: Thanks! I extracted the common stuff into `parquet_fixture_utils.hpp` and migrated the fixtures added by this PR. The similar helpers in existing unrelated tests I left for a follow-up since it's unrelated to this PR.

Also as for S3MixFixture, since it's built specifically for fallback tests and is only available within that test file, maybe it shouldn't be used here directly.

So in the follow-up I'm thinking I can address both the unrelated tests and also S3MixFixture to use these shared helpers.

joosthooz (approved):
Thank you!

Issue #1162, opened 2026-07-13
Support UPDATE with DuckDB format


bwyogatama:
## Summary

UPDATE is explicitly out of scope across the DuckDB DML work so far (#819 "write ops in scope", #1143, #1160): DuckDB versions updated VALUES in place via in-memory `UpdateSegment` chains, invisible to the `ChunkInfo`/`GetSelVector` visibility machinery the DELETE keep-masks ride. Neither the pinned cache nor the persistent disk image reflects updated values until a `CHECKPOINT` folds the chains into the base data.

## Current behavior

- **Pin time**: `pin_table` refuses a duckdb table that already carries update chains on the pinned columns (`any_update_chains`, `src/sirius_extension.cpp`); `CHECKPOINT` first, then pin.
- **Post-pin**: the plan-time update-chain guard was deferred with the other walk guards (#1160) — post-pin UPDATEs currently serve stale cached values from the pin.
- **Persistent-storage scans**: the duckdb-native read serves pre-update values for any uncheckpointed chain (formerly #1143 failure mode 3).

## Direction

- The real fix: walk `UpdateSegment` chains at serve time and materialize updated values into the served columns (per-column chain walk on the CPU side, merged like the delete masks).
- Interim contract: express UPDATE as DELETE + INSERT, or `CHECKPOINT` to fold chains away.
- Until support lands, #1160's execution-time fallback is the correctness backstop: detect chains during the metadata walk and hand the query back to DuckDB.

Refs: #1161 (parent), #819, #1143, #1160.

Issue #1160, opened 2026-07-13
Execution-time MVCC fallback for duckdb-native GPU scans (deferred plan-time walk guards)


bwyogatama:
## Summary

#1142 (#819 PR3) added plan-time MVCC guards in `create_plan(LogicalGet)` that walk every row group of the scanned table — serial, on the query thread, per query. That is planning overhead on hot paths: every unpinned duckdb-native seq_scan, and every query on an MVCC-pinned table (which walks the same row groups again at prepare to build the keep-masks). The walk-based guards were disabled (`#if 0` in `src/planner/sirius_plan_get.cpp`) before merge; this issue tracks their replacement: MVCC fallback driven from the scan manager at execution time.

## Disabled guards and current exposure

1. **Pinned, update-present** (`any_update_chains` over the scanned columns): post-pin UPDATEs serve stale cached values. UPDATE is documented out of scope for #819; `pin_table` still refuses tables that already carry update chains.
2. **Pinned, column-mismatch clean-table relaxation** (`check_native_read_mvcc_state`): a scan the pin cannot serve now declines to CPU unconditionally, instead of falling through to the disk-native read when the table provably matches its last-checkpointed image. Conservative — never wrong, just fewer GPU scans.
3. **Unpinned scans** (the whole guard arm): the duckdb-native read of unpinned tables is MVCC-blind again — #1143 re-exposed in full (silently served tombstones/updates, loud execution failure on uncheckpointed inserts).

Still live — the O(1) guards #819's fallback constraint requires at plan time: snapshot-too-old, insert-present, and transaction-local appends on MVCC-pinned tables. Without those, a stale-snapshot query hits `capture_mvcc_visibility_plan`'s prepare-time throw and a cache miss trips the mvcc-strict prepare backstop — `InternalException`, no fallback, database invalidated — and post-pin inserts would silently vanish from results.

## Prerequisite

Transparent DuckDB fallback **during query execution**. Today the only CPU-fallback gate is plan time (`OnFinalizePrepare`'s try/catch around `create_plan()`; see #819 → "Fallback constraint"). Any refusal after that point surfaces as an `InternalException` that invalidates the database, which is exactly why the disabled guards had to run at plan time in the first place.

## Direction

[...]

PR #1422, opened 2026-08-05
fix(pin): reject updates to pinned DuckDB tables

johanpel:
## Description

Prevent pinned DuckDB tables from silently serving stale values:

- reject `UPDATE`, `MERGE ... UPDATE`, and `INSERT ... ON CONFLICT DO UPDATE` while the target is pinned
- revalidate prepared statements after rebinding
- coordinate updates with concurrent pin/unpin operations
- preserve MVCC handling for inserts and deletes
- fall back or error when a checkpoint invalidates a live pin

### MRE

```sql
CREATE TABLE t AS SELECT 1 AS k, 10 AS v;
CHECKPOINT;
CALL pin_table(format='duckdb', name='t', tier='gpu');

UPDATE t SET v = 99 WHERE k = 1;
```

Previously, the update succeeded but pinned queries could still return 10. It now fails with instructions to run CALL unpin_table('t') before updating.

## Checklist

- [x] Read CONTRIBUTING.md
- [x] Update human and agent documentation

## Verification

- pixi run make
- pixi run build/release/extension/sirius/test/cpp/sirius_unittest "[pin_table_mvcc_update]"
- pixi run pre-commit run -a

## References

Refs #1160
Refs #1162

_Written by Codex._


Jedi18:
Thanks this is an important fix, but there is a possible race here

**Example**

Assume the pin contains balance = 100.

```
-- Initial setup
CREATE TABLE accounts (id INTEGER, balance INTEGER);
INSERT INTO accounts VALUES (1, 100);
CHECKPOINT;
CALL pin_table(format='duckdb', name='accounts', tier='gpu');
-- DuckDB now records an UPDATE chain, but the pin still contains 100
UPDATE accounts SET balance = 900 WHERE id = 1;
```

A concurrent race can then happen:

```
SELECT connection                 CHECKPOINT connection
-----------------                 ---------------------
Checks generation: unchanged
                                  Starts CHECKPOINT
                                  Writes 900 into new row groups
                                  Removes the UPDATE chain
Checks for UPDATE chains: none
                                  Increments generation
Serves pinned balance: 100
```

The correct result is 900, but Sirius can return the stale pinned value 100.

The important point is that both checks individually pass because CHECKPOINT changes the relevant state between them.

Here's Claude's review analysis after I asked it to independently verify if this is indeed a possible race

```
Blocking: the checkpoint/update guard isn't a sound fence against a concurrent CHECKPOINT.

Nothing couples the two prepare-time checks to the actual serving, and a read-only transaction holds no checkpoint lock (DuckTransaction::SetModifications only takes the shared lock on write/catalog changes), so another connection's CHECKPOINT can land in between:

UPDATE commits — chains exist, pin still holds pre-update values.
Our SELECT passes the GetCheckpointIteration() check.
Another connection runs CHECKPOINT. It folds the chains into a new row-group tree (row_group_collection.cpp:1867 SetRowGroups) — the new RowGroups get fresh ColumnData with no UpdateSegment, values merged in by ColumnData::CheckpointScan.
WriteHeader (and the iteration_count bump) only happens at the very end of the checkpoint, after every remaining table and the metadata flush (checkpoint_manager.cpp:224).
In that window any_update_chains returns false and the iteration still matches the pin, so we serve stale cached values. Re-checking the iteration after the update walk narrows the window but doesn't close it, and swapping the order of the two checks doesn't either.

Two related notes:

Likely use-after-free, not just stale values. mvcc_row_group_slice keeps a raw RowGroup* and capture_mvcc_visibility_plan drops its GetRowGroups() shared_ptr on return. SegmentNode owns shared_ptr<RowGroup>, and a checkpointed row group is a new object, so the swap can free the ones the mask job later calls GetSelVector on. The comment on mvcc_row_group_slice justifies the raw pointer with "checkpoint compaction suppressed while pinned" — the assumption this PR is fixing. Holding the tree's shared_ptr in the plan looks worth doing either way.
GetCheckpointIteration() is an unlocked read of a plain uint64_t that WriteHeader mutates under single_file_block_lock — a data race. Making it atomic would quiet TSan but wouldn't fix the above.
What I think closes it: hold a shared checkpoint lock across prepare → serve (DuckTransactionManager::SharedCheckpointLock), or otherwise refuse CHECKPOINT while these pins are live. Heads-up on the tradeoff so you can scope it: a concurrent plain CHECKPOINT would then throw "there are other write transactions active" (misleading for a reader), and FORCE CHECKPOINT busy-spins in GetExclusiveLock while holding start_transaction_lock, stalling new transactions for the length of the GPU query.

The rest — prepare-time reject → transparent fallback, column scoping, the tests — looks solid, and this is a clear improvement over silently serving stale values. 
```

Jedi18:
Thanks the rejection resolves the race, but there is one remaining issue which is that a prepared UPDATE can bypass the pinned-table check after rebinding.

Example:

```
1. Prepare UPDATE t while t refers to an unpinned table.
2. Later, DuckDB rebinds the statement so t refers to a different, pinned table.
3. The old target is checked and the update guard is acquired.
4. The rebound target is not checked because has_pinned_update_guard() causes an early return.
5. The pinned table can then be updated, making its cached values stale.
```

Holding the guard should skip reacquiring the lock, but it should not skip validating the current physical plan’s target. Could we re-check the target while reusing the existing guard?

Jedi18 (approved):

PR #1426, opened 2026-08-05
fix(test): pin tpch_performance duckdb to 1.5.5 to match the built extension

roaramburu:
## Problem

`test/tpch_performance/` is its own pixi workspace (`name = "tpch-parquet-tools"`) and declared `duckdb = "*"`, so its environment drifts independently of the workspace root, which pins `duckdb = "=1.5.5"`. Its lock had resolved to **duckdb 1.5.2**.

The Sirius extension is built against an exact DuckDB version and `LOAD` hard-rejects any other. So the most natural invocation of the harness — from its own directory — fails on every query:

```
cd test/tpch_performance && pixi run python performance_test.py ...

_duckdb.InvalidInputException: Invalid Input Error: Failed to load
'.../sirius.duckdb_extension', The file was built specifically for DuckDB version
'v1.5.5' and can only be loaded with that version of DuckDB. (this version of
DuckDB is 'v1.5.2')
```

## Why this is worth fixing rather than documenting

It fails in a way that is easy to misread as a Sirius bug rather than an environment one:

- the run **exits 0**
- it still creates the benchmark directory and `csv/runtimes.csv`, but the CSV contains **only the header**
- with `--engine both` you get DuckDB rows, **no** sirius rows, and no `sirius/` output directory at all

Nothing in the output says the GPU never ran. Working around it by pinning `--manifest-path` to the repo root also works, but the root cause is the unpinned dependency in the second manifest.

## Change

Pin `duckdb = "=1.5.5"` in `test/tpch_performance/pixi.toml` to match the root manifest, with a comment explaining the coupling, and relock.

⚠️ The relock also upgrades the lock format **v6 → v7**, which accounts for nearly all of the diff. Happy to split that out or regenerate differently if you'd prefer the format bump in its own change.

Found while benchmarking `dev` @ `53b1eb98` across several GPU SKUs.

### DIFF
--- a/test/tpch_performance/pixi.toml
+++ b/test/tpch_performance/pixi.toml
@@ -6,6 +6,11 @@ platforms = ["linux-64", "linux-aarch64"]
 [dependencies]
 python = "3.12.*"
 rust = ">=1.89"
-duckdb = "*"
+# MUST match the `duckdb` pin in the workspace-root pixi.toml. The Sirius
+# extension is built against that exact DuckDB, and `LOAD` hard-rejects any
+# other version. Leaving this unpinned lets this environment drift: it resolved
+# to duckdb 1.5.2 against a 1.5.5 build, so every run from this directory died
+# with "The file was built specifically for DuckDB version 'v1.5.5'".
+duckdb = "=1.5.5"
 fastavro = "*"
 pyarrow = ">=14"
--- a/test/tpch_performance/pixi.lock
+++ b/test/tpch_performance/pixi.lock
@@ -1,10 +1,21 @@
-version: 6
+version: 7
+platforms:
+- name: linux-64
+  virtual-packages:
+  - __unix=0=0
+  - __linux=4.18
+  - __glibc=2.28
+  - __archspec=0=x86_64
+- name: linux-aarch64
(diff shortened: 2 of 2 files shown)


test/tpch_performance/pixi.toml
```diff
@@ -6,6 +6,11 @@ platforms = ["linux-64", "linux-aarch64"]
 [dependencies]
 python = "3.12.*"
 rust = ">=1.89"
-duckdb = "*"
+# MUST match the `duckdb` pin in the workspace-root pixi.toml. The Sirius
+# extension is built against that exact DuckDB, and `LOAD` hard-rejects any
+# other version. Leaving this unpinned lets this environment drift: it resolved
+# to duckdb 1.5.2 against a 1.5.5 build, so every run from this directory died
+# with "The file was built specifically for DuckDB version 'v1.5.5'".
```
johanpel, line 13: Nit:
```suggestion
# MUST match the `duckdb` pin in the workspace-root pixi.toml.
```

There's git blame for the rest.

johanpel (approved):

mbrobbel:
This has some conflicts, marking as draft.

PR #1428, opened 2026-08-05
perf(narrowing): Add DATE type to narrowing

kevkrist:
## Description
Feature:
- narrowing_rep_type / narrowing_rep_view / cast_through_rep: DATE (TIMESTAMP_DAYS) narrows through its int32 epoch-day representation; conversions tunnel via zero-copy bit_cast because cuDF forbids timestamp<->numeric casts. One kernel each way, same as integers.
- DATE literals fold to epoch days so date filters compare at the narrow width; disjoint semantic domains (narrow_domain_of) keep epoch days from ever pairing with plain integers at a shared carrier.
- narrow_domain_reference_pair_eligible: col-vs-col comparisons at one carrier stay narrow (planner policy and evaluator share the predicate).

Performance:
- TPC-H SF1000 (GB300) vs dev: -4.18% HOST-pinned uncompressed
- Perf-neutral on optimized config

## Checklist
- [x] Read CONTRIBUTING.md


src/op/scan/sirius_gpu_scan_operator_data.cpp
```diff
@@ -58,10 +58,14 @@ void scan_operator_input::prepare_for_processing(
...
-      // OOM retry can never re-enter materialize on a consumed split.
-      if (!mvcc_keep_mask.has_mask() && !row_filter_pending) {
+      // view path: they filter by copy and need the source view alive. Splits
+      // pending a carrier conversion keep it too: normalize_physical_schema
+      // allocates a cast output AFTER materialize, and if that cast OOMs the
+      // rescheduled task must be able to re-enter materialize — impossible once
+      // the wrapper batch has been emptied by the take. Skipping all three
+      // means the scan allocates nothing after the take, so an OOM retry can
+      // never re-enter materialize on a consumed split.
+      if (!mvcc_keep_mask.has_mask() && !row_filter_pending && !needs_carrier_conversion) {
```
Jedi18, line 68: I know that this makes OOM retries safe, but it also prevents stealing the uploaded/decompressed table. For a wide HOST/compressed batch, one column requiring carrier conversion can therefore cause every selected column to be copied.

We should maybe create a follow-up issue to see if there's a zero-copy approach to optimize this in the future?

Jedi18 (approved):
LGTM

One remaining edge case is that pinned DuckDB tables are identified by qualified name and schema. Dropping and recreating a table with the same name and schema can therefore match the old pin and return stale rows.

This issue predates the DATE narrowing changes, so it should not block this PR. I’ve opened [#1591](https://github.com/sirius-db/sirius/pull/1591) to fix it by including the DuckDB table incarnation in the cache identity.

src/op/scan/sirius_gpu_scan_operator_data.cpp
kevkrist: Yes good call -- I am fixing this in #1584.

PR #1430, opened 2026-08-06
perf(scan): prune parquet row groups from null_count statistics

Jedi18:
## Description

Follow-up to [#1417](https://github.com/sirius-db/sirius/pull/1417), which prevented null predicates from reaching cuDF’s row-group statistics filter because cuDF faults on them. Those predicates still run after decoding, but no longer contributed to row-group pruning, so queries such as `WHERE v IS NULL` read every row group.

Parquet footer statistics provide the necessary information:

- `IS NULL` cannot match when `null_count == 0`.
- `IS NOT NULL` cannot match when `null_count == num_rows`.

Sirius now applies these checks alongside cuDF’s pruning of other conjuncts. Only row groups proven unable to contain a match are skipped, so query results are unchanged.

### Safety constraints

- **Missing statistics:** `null_count` is optional. An absent value means unknown, so the row group is retained.
- **Scalar columns only:** Eligibility is determined from the declared type and validated against the Parquet leaf mapping. Nested-column leaf statistics describe repeated values or leaf-level nullness, not top-level nullness. This also excludes legacy top-level `REPEATED` primitives that DuckDB exposes as `LIST`.
- **Conjunctive predicates only:** Null tests in conjunctive positions can prune independently. Null tests inside `OR` expressions cannot, because another branch may still match.

Predicates are collected directly from `TableFilterSet`. This is necessary because `convert_table_filters_to_expression` removes top-level `IS_NOT_NULL` filters before constructing the expression; collecting afterward would miss ordinary `WHERE v IS NOT NULL` predicates.

### Testing

Result comparison alone cannot prove pruning occurred because the full predicate is still applied after decoding. The new tests therefore drive the ingestible directly and assert the surviving row-group count.

The fixture also validates its physical layout—including per-group row counts, null counts, and the presence of statistics—before testing pruning.

## Checklist

- [x] No configuration changes
- [x] No user-facing documentation changes required

## References

Follow-up to [#1417](https://github.com/sirius-db/sirius/pull/1417).

### DIFF
--- a/src/op/scan/parquet_gpu_ingestible.cpp
+++ b/src/op/scan/parquet_gpu_ingestible.cpp
@@ -50,7 +50,11 @@
 // duckdb
 #include <duckdb/common/hive_partitioning.hpp>
 #include <duckdb/planner/expression/bound_conjunction_expression.hpp>
+#include <duckdb/planner/expression/bound_operator_expression.hpp>
+#include <duckdb/planner/expression/bound_reference_expression.hpp>
 #include <duckdb/planner/expression_iterator.hpp>
+#include <duckdb/planner/filter/conjunction_filter.hpp>
+#include <duckdb/planner/filter/null_filter.hpp>
 
 // uring_reactor MUST be included last among sirius headers: liburing.h,
 // pulled in transitively, defines a BLOCK_SIZE macro that collides with the
@@ -67,6 +71,7 @@
 #include <stdexcept>
 #include <string>
 #include <system_error>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


src/op/scan/parquet_gpu_ingestible.cpp
```diff
@@ -714,6 +776,84 @@ std::unique_ptr<scan_info> parquet_gpu_ingestible::build_file_scan_info(
...
+      // neither does a one-component path: legacy parquet allows a top-level
+      // REPEATED primitive, whose path is just the column name, and DuckDB
+      // surfaces that as a LIST.
+      auto const batch_index = static_cast<std::size_t>(pred.batch_index);
+      if (batch_index >= _plan->data_columns.size()) { continue; }
+      auto const primary_idx = _plan->data_columns[batch_index].primary_idx;
+      if (primary_idx >= _info->returned_types.size()) { continue; }
+      auto const& column_type = _info->returned_types[primary_idx];
+      if (column_type.id() == sirius::type_id::LIST || column_type.id() == sirius::type_id::ARRAY ||
+          column_type.id() == sirius::type_id::STRUCT) {
```
joosthooz, line 811: Maybe it is intentional, but this check does not include MAP and UNION

test/cpp/scan/test_parquet_null_count_pruning.cpp
```diff
@@ -0,0 +1,247 @@
...
+{
+  REQUIRE(surviving_row_groups(std::nullopt) == 3);
+}
+
+// `v IS NULL` cannot match the all-non-NULL row group (null_count == 0).
+TEST_CASE_METHOD(NullCountFixture,
+                 "null_count pruning - IS NULL drops row groups with no nulls",
+                 "[scan][parquet][pruning][null_count]")
+{
+  CHECK(surviving_row_groups(/*filter_expects_null=*/true) == 2);
```
joosthooz: Consider using REQUIRE for consistency with the other tests

test/cpp/scan/test_parquet_null_count_pruning.cpp
```diff
@@ -0,0 +1,247 @@
...
+// (null_count == num_rows).
+//
+// A BARE filter on purpose: with no sibling comparison there is nothing for
+// cuDF's statistics filter to push, so null_count pruning is the only thing
+// that can drop a row group and the count below is evidence about it alone.
+TEST_CASE_METHOD(NullCountFixture,
+                 "null_count pruning - IS NOT NULL drops wholly-null row groups",
+                 "[scan][parquet][pruning][null_count]")
+{
+  CHECK(surviving_row_groups(/*filter_expects_null=*/false) == 2);
```
joosthooz: same

joosthooz (approved):
Great, thanks! Note that this new code also has a bunch of deduplication opportunities, some code in the other PR is repeated here so it would be nice to make shared helpers for those things

Jedi18:
Thanks! I’ve updated this PR to use the same `parquet_fixture_utils.hpp` helpers. I’ll merge this PR first, then rebase the other PR

src/op/scan/parquet_gpu_ingestible.cpp
Jedi18: added comments explaining why map and union aren't included

PR #1431, opened 2026-08-06
test: replace obsolete multi-GPU scheduler stress test

johanpel:
## Description

Replace the obsolete multi-GPU counter-offset stress test with a deterministic test of the scheduler’s active ready-device matching behavior.

The previous test repeated end-to-end queries across roughly 100 counter offsets, but the active matcher no longer used that counter.

This change:

- verifies that preferred tasks execute only on their selected GPU;
- verifies that a preference-less task can be claimed by a ready GPU;
- covers matching with two ready devices when two GPUs are available;
- removes the unused round-robin counter and test-only setter;
- documents `SCHED-RR` as the established scheduling contract name, distinct from the current ready-device matching algorithm;
- renames the test to reflect its routing coverage.

This removes the approximately 53-second CI stress loop. The replacement completed in 0.093 seconds on the available single-GPU system.

### Verification

- `pixi run pre-commit run --files ...`
- `pixi run make release MAIN_BUILD_TARGETS= TEST_BUILD_TARGET=sirius_unittest`
- `pixi run build/release/extension/sirius/test/cpp/sirius_unittest 'task_scheduler matches tasks to ready devices' --durations yes`
  - Passed with 7 assertions.
  - One GPU was visible, so assertions requiring two GPUs were skipped with an explicit warning.

## Checklist

- [x] Read CONTRIBUTING.md
- [x] Document configuration changes in code and summarize in the description above — not applicable; no configuration changes

## References

None.

_Written by Codex._


test/cpp/operator/test_task_scheduler_routing.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+  int device_count = 0;
+  REQUIRE(cudaGetDeviceCount(&device_count) == cudaSuccess);
+  if (device_count == 0) {
+    WARN("Task scheduler routing test requires a GPU; skipping");
+    return;
+  }
+  if (device_count == 1) {
+    WARN("Only one GPU is visible; multi-GPU routing assertions will not run");
+  }
```
mbrobbel: I think we could fold these, use INFO, and shouldn't we return in the second case too?

mbrobbel (approved):

test/cpp/operator/test_task_scheduler_routing.cpp
johanpel: https://github.com/sirius-db/sirius/pull/1431/commits/84e93e844c6422970d186eeb8e4fb7f5e09f2f76

joosthooz (approved):
Thanks

Issue #838, opened 2026-05-29
Streaming sink partitioning support


mbrobbel:
Add partitioning support to the streaming sink operator #837. This allows us to efficiently partition in the GPU and split out the different partitions across multiple output streams to avoid head-of-line blocking by slow receivers.

PR #1432, opened 2026-08-06
feat(exec): streaming execution — source, sink, session, and fragment (#836, #837, #838, #839)

### LINKED ISSUE #836: Streaming source operator

aocsa:
Closes #836, #837, #838, #839.

**What.** The complete streaming execution layer: a GPU plan whose leaf is a `STREAMING_SOURCE`
(batches pushed in by remote producers) and whose root is a `STREAMING_SINK` (batches pulled out
by an external consumer), addressed by numeric stream ids, built from a declarative spec, and
exposed to the Rust wrapper through `sirius::ffi::Fragment`. Thirteen commits in dependency order.

**Replaces the prototype from #1094.** `exec::exchange_channel` and its 528-line test file are
deleted, not ported. `sirius_physical_streaming_source` is rebuilt on `exec::batch_stream`. The
queue, EOS rule, and error plane all changed — read the new source fresh rather than as a diff.

**Core design.** A scan owns its input; "empty" means exhausted. A streaming source owns nothing
and its senders are remote: "empty" means *wait*, waiting must not block an engine thread, and the
stream ends only when every expected sender has closed. So **the repository IS the queue** —
`batch_stream` borrows a `cucascade::shared_data_repository` so queued batches stay visible and
spillable to the downgrade executor for their whole queued life. The stream adds only what the
repository lacks: sender-aware EOS, consumer wake-up, and an error plane.

Key invariants:
- **EOS is a set, not a counter.** Two closes from sender 0 cannot stand in for senders {0,1}.
- **`on_data` fires on every `push()`.** The scheduler drops a WAITING source; the hook is the
  only re-nomination edge. It is persistent so no push can slip in between fire and re-arm.
- **`fail_input(error)` poisons the stream.** Consumer wakes as if data arrived; next pull
  rethrows. Without this, a dead producer looks like a finished one.
- **The sink must sit in `operators[]`.** `on_finalize_operator()` is the sole EOS path. A sink
  reachable only through the `sink` member never finalizes; consumers block in `wait()` forever.
- **One `batch_stream` per destination.** A slow receiver's backlog cannot head-of-line-block siblings.
- **Fragment repositories never register with `data_repository_manager_`.** A sender's output
  survives its own `StandaloneQueryScope::finish()` and is still pullable when the receiver runs.
- **No channel-level backpressure, by design.** Memory pressure is relieved by the downgrade
  executor, which can only see batches that live in a repository (#1320 removed the bounded channel).

**Contracts.** S1: batch in repo before `on_data` fires. S2: `fail()` wakes a WAITING consumer.
S3: errored stream never reports `END_OF_STREAM` or `drained()`. S4: `try_pull()` rethrows before
popping. S5: `wait()` then `try_pull()` is not atomic. P1–P4: error-plane details in `batch_stream`.

---

**Reading order** (bottom-up by file).

- `src/include/exec/batch_stream.hpp` — class comment defines S1–S5 / P1–P4; `.cpp` for lock discipline (`try_pull` pops outside `_mutex`). Tests: `[batch_stream]`, no GPU.
- `src/include/op/sirius_physical_streaming_source.hpp` — class comment is the design note; `.cpp` for `set_pipeline()` hooks and task-protocol. Five lines in `src/planner/query.cpp` make it schedulable. Tests: `[streaming_source]`.
- `src/include/op/sirius_physical_streaming_sink.hpp` — `partition_spec` with `partition_mode`; `.cpp` for three `sink()` branches (N=1, broadcast, hash). `on_finalize_operator()` is the sole EOS path. Tests: `[streaming_sink]`.
- `src/include/exec/stream_session.hpp` — six forwarding verbs; router only. Tests: `[stream_session]`.
- `src/include/exec/streaming_fragment.hpp` — `build()` derives `key_cast_types`, roots in `STREAMING_SINK`, wires session; `sink_types()` captures output types before the move; `run()` must share the caller's query window. Tests: `[streaming_fragment]`, `[streaming_sink_root]`.
- `src/include/sirius_ffi.hpp` — `Fragment` public surface; `sirius_ffi.cpp` for `Fragment::Impl` (heap-allocated `StandaloneQueryScope`, schema validation in `relay_from()`). Not ported: `export_packed` / `StagingArena`.
- `docs/super-sirius/streaming-sessions.md` — consolidated reference for the whole layer.

---

**Review focus.**

- **Lock discipline.** `classify()`/`drained()`/`wait()` read under `_mutex`; `try_pull()` pops
  outside it. Confirm consumers tolerate S5 non-atomicity.
- **`on_data` on every push.** Is `schedule()` cheap and safe enough for a fast producer?
- **Null `memory_space` guard in partitioned sink.** Throws `internal_exception` (= null
  representation), not "not GPU-resident" — a spilled batch passes. Does `hash_partition` handle
  non-GPU-tier input?
- **Warn-and-drop on refused push.** Acceptable for a shuffle, or should a late push fail the fragment?
- **Broadcast clone allocation.** N−1 `clone()` calls per batch under GPU memory pressure. OOM
  leaves output 0 not yet pushed; is the partial state safe?
- **Lifetime.** Session holds raw non-owning pointers; `stream_input_binding::built` points into
  the engine-owned plan tree. Verify nothing reads either after engine destruction.
- **Repositories outside the manager.** The memory-pressure sweep cannot see in-flight stream data.
  Acceptable at the intended batch sizes?

**Testing.** All suites need a GPU except `[batch_stream]`. Tags:
`[batch_stream]`, `[streaming_source]`, `[streaming_sink]`, `[stream_session]`,
`[stream_bind_catalog]`, `[streaming_fragment]`, `[streaming_fragment_control]`,
`[streaming_sink_root]`, `[streaming_sink_root_exec]`. `pixi run pre-commit run -a` is clean.

**Base:** `dev`. **Depends on:** nothing.


aocsa:
Superseded by a three-PR stack, so each layer can be reviewed on its own:

- #1479 — `STREAMING_SINK` + partitioning (#837, #838), plus the post-#1320 source follow-up
- #1480 — `exec::stream_session`, the id-addressed router (#839, addressing)
- #1481 — `streaming_fragment` + Fragment FFI (#839, execution)

Same content: the top of the stack is tree-identical to this branch on every streaming path,
verified with a diff gate rather than by inspection. The branch `stream/engine` is left in place.

Closing in favour of the stack.

PR #1439, opened 2026-08-07
feat(telemetry): capture libcudf NVTX spans and serve NVTX routes

9prady9:
## Summary

- Captures Sirius, libcudf, and CCCL NVTX events in both supported deployments without a sidecar injection DSO:
  - **Vanilla DuckDB + `LOAD`** exports Quent's initializer from the Sirius extension and points NVTX at that DSO before the first NVTX call.
  - **Statically linked Sirius** exports the same initializer from DuckDB and maps a Sirius-private NVTX injection token to the running executable, while forwarding every other `dlopen` unchanged.
- Honors `SIRIUS_DISABLE` and `enable_quent: false` by skipping automatic injection; otherwise preserves configuration precedence: `NVTX_INJECTION64_PATH`, then `sirius.telemetry.nvtx_injection_lib`, then automatic discovery.
- Mounts Quent's NVTX catalog/viewport routes and pins Quent to `381e1be0`, which includes [quent#544](https://github.com/rapidsai/quent/pull/544) and the `nvtx-sys` build fix from [quent#697](https://github.com/rapidsai/quent/pull/697). Generalized schema work remains tracked by [quent#618](https://github.com/rapidsai/quent/issues/618).

## Verification

- `pixi run make`
- `pixi run make test` — 3,080 test cases and 32,833,137 assertions passed
- TPC-H SF10 Q1 produced the exact same NVTX range/count signature in both modes: 812 complete ranges across Sirius/default, libcudf, and CCCL, with zero incomplete ranges or reconstruction anomalies.

### DIFF
--- a/src/sirius_extension.cpp
+++ b/src/sirius_extension.cpp
@@ -111,6 +111,7 @@ extern "C" int cudaProfilerStop();
 #include "sirius_extension.hpp"
 #include "sirius_interface.hpp"
 #include "sirius_sql_rewrite.hpp"
+#include "telemetry/nvtx_injection.hpp"
 #include "util/segfault_backtrace.hpp"
 #include "vss/cuvs_index_cache.hpp"
 #include "vss/distance_metric.hpp"
@@ -161,6 +162,28 @@ extern "C" __attribute__((visibility("default"))) int InitializeInjectionNvtx2(
 {
   return quent_InitializeInjectionNvtx2(get_export_table);
 }
+
+#ifndef DUCKDB_BUILD_LOADABLE_EXTENSION
+// NVTX v3 discovers an injector independently in each ELF image. libcudf's
+// injection pointer is local to libcudf.so and its process-global preinjection
+// lookup is compiled out, so it can only reach Quent through its dlopen path.
+// A statically linked Sirius has no DSO to name. Interpose just our private
+// sentinel and turn that request into dlopen(NULL), whose handle exposes the
+// initializer exported by the running DuckDB executable. Every other request
+// is forwarded unchanged to libc.
+extern "C" __attribute__((visibility("default"))) void* dlopen(const char* filename, int flags)
+{
+  using dlopen_fn   = void* (*)(const char*, int);
+  auto* real_dlopen = reinterpret_cast<dlopen_fn>(::dlsym(RTLD_NEXT, "dlopen"));
+  if (real_dlopen == nullptr) { return nullptr; }
+
+  if (filename != nullptr &&
+      std::string_view{filename} == sirius::telemetry::detail::static_injection_path) {
+    return real_dlopen(nullptr, flags);
+  }
+  return real_dlopen(filename, flags);
+}
+#endif
 
 namespace duckdb {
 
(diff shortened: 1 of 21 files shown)


src/sirius_extension.cpp
```diff
@@ -142,6 +145,17 @@ extern "C" int cudaProfilerStop();
...
+// extension link (`--exclude-libs,ALL`), so this regular C++ object supplies the
+// public NVTX entry point and forwards it into the same Quent hook state used by
+// Sirius's static-injection pointer.
+extern "C" int quent_InitializeInjectionNvtx2(void* get_export_table);
+extern "C" __attribute__((visibility("default"))) int InitializeInjectionNvtx2(
+  void* get_export_table)
+{
+  return quent_InitializeInjectionNvtx2(get_export_table);
+}
+
```
9prady9, line 164: Most likely I will move this into a generated header from quent so that it can be just a inclusion.

9prady9:
nvtx events in both workflows: (1) vanilla-duckdb+LOAD extension (2) duckdb with statically linked sirius a.k.a executable.

@dhruv9vats thanks for reporting the missing path for (2) - that also works now.

Screenshot from 2026-09-08 11-11-22


Screenshot from 2026-09-08 11-10-57

src/sirius_extension.cpp
```diff
@@ -3455,8 +3492,104 @@ static void publish_transparent_optimizer_mask(DBConfig& config)
...
+    // would be premature here (logging is not yet configured).
+  }
+}
+
 static void LoadInternal(ExtensionLoader& loader)
 {
+  // libcudf is mapped before this point, but its NVTX state is still fresh
+  // because its constructors do not call NVTX. Configure discovery before any
+  // Sirius initialization can make the first NVTX call in an image.
+  maybe_set_nvtx_injection_path();
```
johanpel: This runs before `SIRIUS_DISABLE` so IIUC it can route all NVTX events through Quent even if you're not enabling Sirius, is that what we want?

johanpel (commented):
One high-level question, if DuckDB spawns multiple `DatabaseInstance`s, will this still work?

9prady9:
> One high-level question, if DuckDB spawns multiple `DatabaseInstance`s, will this still work?

Clarifying the multiple-DuckDB-instance scope: here, `DatabaseInstance` means an independent DuckDB object inside one OS process. Separate DuckDB executables have independent process-global NVTX/Quent state.

| Scenario | Current support | NVTX result |
|---|---|---|
| Multiple DuckDB processes | Supported | Works independently |
| Multiple connections/cursors to one `DatabaseInstance` | Supported | Works; all share one `SiriusContext` |
| `ATTACH` databases within one instance | Supported | Works |
| Two concurrent Sirius-enabled `DatabaseInstance` objects in one process | Sirius currently assumes at most one live context per process | The first context receives process-wide NVTX events |
| Destroy instance A, then create instance B in the same process | Not explicitly unsupported; Sirius tests exercise context recreation | B loses NVTX because Quent's one-shot hook still references A |

The two single-instance deployment paths validated in this PR are unaffected. Sequential in-process recreation remains a known NVTX lifecycle gap.

src/sirius_extension.cpp
9prady9: you are right, that's an oversight on my part. I didn't realize it was to disable instrumentation but something else. Let me check what is it about actually.

johanpel:
> The first context receives process-wide NVTX events

While this seems undesirable since:

> Sirius currently assumes at most one live context per process

I suppose this is fine?

> B loses NVTX because Quent's one-shot hook still references A

This doesn't seem desirable, but I suppose this needs a fix in Quent first?

9prady9:
> > The first context receives process-wide NVTX events
> 
> While this seems undesirable since:
> 
> > Sirius currently assumes at most one live context per process
> 
> I suppose this is fine?

It may seem fine given how sirius is right now, but should there ever be two instances then instance B's events will be wrongly attributed to instance A. 

> 
> > B loses NVTX because Quent's one-shot hook still references A
> 
> This doesn't seem desirable, but I suppose this needs a fix in Quent first?

This is totally missing.

I am looking into whats the best way forward, nvtx is generally process wide, so how this needs to handled in nvtxmodel and UI also changes accordingly.

Having said that, I don't think we need to block this PR for that follow up work. I don't see row 4 and 5 use cases right now, so we can handle those as follow ups.

johanpel (approved):
Agreed, this is already very useful in its current state. Let's document the limitations and follow-up on the rest.

9prady9:
> Agreed, this is already very useful in its current state. Let's document the limitations and follow-up on the rest.

👍🏾 Yeah, let me create the issues, link them here and then we can merge this in.

9prady9:
> > Agreed, this is already very useful in its current state. Let's document the limitations and follow-up on the rest.
> 
> 👍🏾 Yeah, let me create the issues, link them here and then we can merge this in.

Created https://github.com/rapidsai/quent/issues/696

9prady9:
The merge queue run exposed a build issue, I am looking into it.

9prady9:
> The merge queue run exposed a build issue, I am looking into it.

Need this https://github.com/rapidsai/quent/pull/697

PR #1442, opened 2026-08-07
feat(gpu): per-query GPU subset allocation

Jedi18:
## Summary

Removes the hard assumption that every query uses all GPUs. Each query receives an explicit GPU allocation at admission time via a new `topology.gpus_per_query` config knob. This is the foundational plumbing for concurrent query execution, where a plan-derived estimate (follow-up PR) will drive how many GPUs each query gets.

Default behavior (`gpus_per_query = 0`) is unchanged.

## What changed

- **`sirius_engine::initialize_internal`** — computes the allocated subset (first k entries of the sorted active-GPU list) before any pipelines are built, so every downstream consumer derives from one list.
- **`task_creator`** — `set_active_gpu_ids()` / `get_active_gpu_ids()` replace the topology-derived list built in the constructor, so partition→GPU routing and broadcast-join device→slot mapping both use the allocated subset.
- **`scan_manager`** — `prepare_for_query()` now accepts the allocated IDs and uses them for scan round-robin and insert-delta routing instead of querying `_topology_index` directly.
- **`pipeline_build_context`** — built from the same list, so `num_gpus` matches what `task_creator` routes across. 
- **Observability** — a `[gpu_alloc]` log line at query admission records which GPUs were assigned.

## Keeping work on the admitted GPUs

Choosing a subset isn't enough on its own — three paths could still place work on an excluded device. All three are closed in `task_creator`:

- **Preferences derived from data location.** A task's target GPU can come from the operator's own hint, from where the most bytes already sit, from NUMA locality via `gpus_of()`, or from a cached chunk's home device. None of those know about the admitted subset, and the scheduler treats a preference as *binding*. These are now folded back into the subset.
- **Tasks with no preference at all.** The scheduler hands those to whichever executor asks first, excluded ones included — `SELECT 42` is the everyday case, since a `GPU_VALUES` task holds no device-bound state. Now pinned round-robin over the subset, but only when the query is actually on a subset: a pin costs the scheduler's freedom to place work wherever frees up first, and there is nothing to confine when nothing was narrowed away. The default path dispatches exactly as before.
- **A negative `gpus_per_query`.** The admission path treats any non-positive value as unset, so `-1` silently meant "use all GPUs". Now rejected at config load.

## Tests and docs

`apply_gpu_cap` moves into `planner/gpu_admission.hpp` as a pure function with 7 cases: prefix semantics, no cap, cap at and above the fleet size, non-contiguous device ids (`CUDA_VISIBLE_DEVICES` can leave gaps), negative, and an empty fleet.

`task-creator.md`, `multi-gpu-architecture.md` and `scan.md` all described `_active_gpu_ids` as constructor-only and topology-derived, and `scan.md` still showed the two-argument `prepare_for_query`; all three are updated, and `configuration.md` documents the new knob.

## What this does NOT do

True concurrency (multiple queries running in parallel on disjoint GPU subsets) still requires `task_creator`, `scan_manager`, and `task_scheduler` to become per-query rather than shared. The lifecycle mutex is untouched here.

Allocation is also not NUMA-aware — it takes the first k GPUs of the sorted list rather than k that share a NUMA domain. Noted in the header as future allocator work.

## Verification

- Full C++ suite green.
- Observed on 4×GB200 with a reduced allocation: the excluded GPU stays at ~84 MB (context and pool baseline) against ~170 GB on the admitted ones, with results identical to the all-GPU run.

## Follow-up

- https://github.com/sirius-db/sirius/pull/1455


9prady9 (commented):
took a first pass, most of it looks fine. I would like to just think through once more while others look at it in the meanwhile.

src/creator/task_creator.cpp
```diff
@@ -528,6 +536,24 @@ void task_creator::manager_loop()
...
+            // executor asks first. Pin those as well, but only on a real subset, since a pin
+            // costs the scheduler's freedom to place them wherever frees up first.
+            if (!_active_gpu_ids.empty()) {
+              bool const names_excluded_device =
+                preferred_device_id.has_value() &&
+                std::find(_active_gpu_ids.begin(), _active_gpu_ids.end(), *preferred_device_id) ==
+                  _active_gpu_ids.end();
+              bool const unpinned_on_a_subset =
+                !preferred_device_id.has_value() && _active_gpu_ids.size() < _full_gpu_count;
+              if (names_excluded_device || unpinned_on_a_subset) {
```
9prady9, line 557: Should residency-derived preferences (`:469`, `:515`) be exempt from the clamp? Nothing downstream migrates a resident cached chunk or accounts for the copy.

src/creator/task_creator.cpp
```diff
@@ -528,6 +536,24 @@ void task_creator::manager_loop()
...
+                auto const idx      = _numa_affinity_rr.fetch_add(1) % _active_gpu_ids.size();
```
9prady9: Should the clamp get its own counter? Sharing `_numa_affinity_rr` with `:487`/`:525` double-bumps it and stride-locks the clamp on even subset sizes right

9prady9 (commented):
Two questions on the new clamp block.

src/creator/task_creator.cpp
Jedi18: They can't be exempt but yes as a follow-up issue we should choose gpus accounting for NUMA residency (this however should happen where gpu admission takes place in `initialize_internal`). I will create an issue.

src/creator/task_creator.cpp
Jedi18: Thanks you're right, fixed the bug

src/creator/task_creator.cpp
Jedi18: Created the issue - https://github.com/sirius-db/sirius/issues/1471

src/creator/task_creator.cpp
```diff
@@ -528,6 +536,26 @@ void task_creator::manager_loop()
...
+            // escapes too: the scheduler gives those to whichever executor asks first. Pin
+            // those as well, but only on a real subset, since a pin costs the scheduler's
+            // freedom to place them wherever frees up first.
+            if (!_active_gpu_ids.empty()) {
+              bool const names_excluded_device =
+                preferred_device_id.has_value() &&
+                std::find(_active_gpu_ids.begin(), _active_gpu_ids.end(), *preferred_device_id) ==
+                  _active_gpu_ids.end();
+              bool const unpinned_on_a_subset =
+                !preferred_device_id.has_value() && _active_gpu_ids.size() < _full_gpu_count;
```
9prady9, line 556: As far as I can tell `set_active_gpu_ids()` never updates `_full_gpu_count` — it's assigned only in the constructor, and only inside `if (_topology_index)`. The two also come from different sources: the ctor reads `_topology_index->gpu_ids()`, while `initialize_internal` builds its list from the memory manager's `Tier::GPU` spaces.

Does that leave this gate able to fail open? With a null `topology_index` — permitted by the ctor, and what `test/cpp/creator/test_task_creator.cpp:162` constructs — `_full_gpu_count` stays 0, so a 2-of-4 subset evaluates `2 < 0 == false`. If I'm reading it right, confinement never fires and every no-preference task goes to whichever executor asks first, excluded GPUs included. Curious whether that's reachable in practice or whether something upstream rules it out.

src/sirius_engine.cpp
```diff
@@ -204,23 +205,35 @@ void sirius_engine::initialize_internal(op::sirius_physical_operator& plan)
...
-    if (space != nullptr) { active_gpu_ids.push_back(space->get_device_id()); }
+    if (space != nullptr) { gpu_ids.push_back(space->get_device_id()); }
+  }
+  std::sort(gpu_ids.begin(), gpu_ids.end());
+  gpu_ids.erase(std::unique(gpu_ids.begin(), gpu_ids.end()), gpu_ids.end());
+
+  // Admit the query here, before anything downstream is built, so that the build context,
+  // partition->GPU routing and the scan round-robin all derive from this one list. Order
+  // matters: task_creator holds it, and create_query later reads it back for scan_manager.
+  std::vector<int> active_gpu_ids =
```
9prady9, line 220: Admission runs here in `initialize_internal`, i.e. after physical plan generation, so device bindings decided at plan time escape the subset right ?

Concretely: `src/planner/sirius_plan_comparison_join.cpp:643` builds `filter_replica_spaces` from *every* GPU memory space, and `dynamic_filter_publisher.cpp:301` calls `replicate_to_devices()` over that set. With `gpus_per_query: 1` on a 4-GPU box, a hash join with dynamic-filter pushdown still allocates and copies filter replicas onto the three excluded GPUs.

src/sirius_config.cpp
```diff
@@ -507,6 +513,7 @@ void sirius_config::load_from_file(const std::filesystem::path& config_path)
     // Topology
     topology topo;
     r.optional("topology", topo);
+    _gpus_per_query = topo.gpus_per_query;
```
9prady9, line 596: The YAML plumbing for `gpus_per_query` is untested — `test_gpu_admission.cpp` only covers the pure `apply_gpu_cap` helper. Nothing asserts that `load_from_file` parses `topology.gpus_per_query`, assigns `_gpus_per_query`, or throws on `gpus_per_query: -1` (the rejection called out in the PR description). `test/cpp/config/test_object_store_config.cpp:205` has the established `CHECK_THROWS(cfg.load_from_file(path))` pattern for this.

As it stands, a refactor that drops this line would silently restore the "negative means use all GPUs" behavior with the suite green.

src/creator/task_creator.cpp
Jedi18: Thanks not reachable in practice today, but you're right that it's a bad shape.

Fixed, `set_active_gpu_ids()` now receives the pre-narrowing GPU count from `initialize_internal`, so both values come from the same memory-manager GPU list. This also removes the null `topology_index` fail-open behaviour.

src/sirius_engine.cpp
Jedi18: Thanks fixed it

After admission, the pipeline converter restricts every hash join’s dynamic-filter replica spaces to the admitted GPU IDs, including joins fused as intermediate pipeline operators. `replicate_to_devices()` therefore no longer targets excluded GPUs.

src/sirius_config.cpp
Jedi18: Thanks added tests

9prady9 (approved):
lgtm

PR #1448, opened 2026-08-08
fix(config): reject ambiguous memory budget forms

infblueocean:
## Summary

- reject simultaneous non-null byte and fraction forms for the high-level GPU usage, GPU reservation, host capacity, and host reservation budgets
- preserve the existing YAML convention that an explicit `null` is absent
- update the configuration reference so it no longer documents silent byte precedence
- add focused rejection and preservation fixtures

## Why

The source described these key pairs as mutually exclusive but previously parsed both and silently preferred the byte form. A configuration could therefore state two different budgets while giving the user no direct indication that one had been ignored. Rejecting the ambiguity makes the active budget explicit without changing valid single-form or omitted defaults.

## Validation

Source-level validation:

- clang-format dry-run with warnings as errors
- `git diff --check`
- clean-base apply and reverse-apply checks
- YAML fixture parse and test-registration checks
- independent source review against Sirius `dev` `4d88a0fed6782d37c9753aa1179cbe9f47816d5a`

Compiled validation at exact PR head `ec2143a433089cd0d618acf512312151c5b8ed9f`:

- isolated `sirius_unittest` build completed all 929 actions
- `Sirius configuration rejects conflicting memory budget forms`: 4/4 assertions passed
- `Sirius configuration accepts one form of each memory budget`: 1/1 assertion passed
- `Sirius configuration treats null memory budget forms as absent`: 1/1 assertion passed
- recursive submodules matched their pins and the source tree was clean before and after

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -118,10 +118,10 @@ Controls how much GPU VRAM Sirius claims and when it starts evicting data to hos
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
-| `usage_limit_fraction` | double | 0.95 | Fraction of total VRAM to use as Sirius's GPU memory capacity. The remaining 5% is left for the CUDA runtime, cuDF temporaries, and other GPU consumers. |
-| `usage_limit_bytes` | bytes | — | Absolute VRAM limit. Mutually exclusive with `usage_limit_fraction`. |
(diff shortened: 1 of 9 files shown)


infblueocean:
The source diff and focused fixtures have completed independent source review. This remains draft until the three named config selectors compile and run. Could a maintainer approve the fork workflows so Check/Test can establish that boundary?

infblueocean:
Compiled validation is complete at the exact PR head and recorded in the PR body; this PR is now ready for review. @mike-wendt, could you review when convenient and approve the remaining fork workflows?

bwyogatama (approved):

PR #1447, opened 2026-08-08
fix(config): derive batch defaults from effective GPU capacity

infblueocean:
## Summary

- keep the existing physical-memory-derived operator defaults when no effective GPU cap is explicitly configured
- when YAML explicitly caps the active GPU memory path, derive the shared scan, hash-partition, CONCAT, and sort-sampling batch default from the smallest resolved GPU capacity while preserving the existing 2.5% policy
- keep `max_build_hash_table_bytes` at twice that shared value
- treat missing and YAML-null GPU limits as absent; explicit `operator_params` still win
- keep DuckDB `current_setting`, `SET`, and `RESET` synchronized with the resolved values

This fixes the formula-class failure reported in #1434: a default derived only from physical GPU memory can exceed a deliberately smaller effective managed capacity.

## Compatibility

- uncapped startup is unchanged: `clamp(physical GPU memory / 40, 512 MiB, 5 GiB)`
- explicit-cap startup uses `min(physical default, max(1 byte, smallest resolved GPU capacity / 40))`
- both high-level `sirius.memory.gpu.usage_limit_*` and low-level `sirius.space.gpu[].memory_capacity` follow the active memory path
- explicit values, repeated loads, null handling, and `apply_defaults()` retain their existing precedence without stale derived state

## Integration order

Author: @infblueocean.

PRs #1448 and #1449 are already merged into `dev` and are preserved in this branch. This PR should land before #1450. After it lands, #1450 will be rebased and its deliberately conflicting mixed-path fixture reconciled under #1450's new rejection contract; the combined focused selectors will then rerun.

## Exact-current-head validation

- current head: `b12bdb7727d4284bc15f5187f0890fc16eb9c2db`
- tree: `296f9dd6ed4ba56288f6b9dcefbfd1651c66c3e9`
- reconciled with `dev` `ca41d15eb83efd31be2d5a2a77f3f4127dcc68d8`; the only content conflict was adjacent #1447/#1449 tests in `test/cpp/config/test_context.cpp`, resolved by retaining both complete test sets
- source clean before and after validation
- all 21 pre-commit hooks passed; the broken-symlink hook skipped because no files matched
- clean Brev L4 CUDA 13 release build passed
- `[sirius][config][operator_defaults]`: 72 assertions across 8 test cases passed
- `[sirius][config]`: 255 assertions across 18 test cases passed, including #1447's operator-default cases and #1449's downgrade-hysteresis cases
[...]

PR #1449, opened 2026-08-08
fix(config): reject invalid downgrade hysteresis

infblueocean:
## Summary

- reject GPU and host downgrade hysteresis pairs unless `0 < stop < trigger <= 1`
- enforce the same contract on high-level `sirius.memory` and low-level `sirius.space` YAML surfaces
- preserve omitted, explicit-null, and one-sided values by validating the resolved pair against that surface's defaults
- document the enforced relationship and correct stale recovery and pinned-host defaults

This turns contradictory user-visible settings into a fail-fast configuration error and makes the accepted value domain consistent with cuCascade.

## Compatibility

- current default pairs and valid explicit pairs remain valid
- missing and explicit-null members retain initialized defaults
- disk configuration and the numeric defaults themselves are unchanged
- 12 invalid fixtures cover zero, equal, and reversed pairs on all four GPU/host surfaces; four valid fixtures cover omitted, null, and one-sided forms on both configuration paths

## Integration order

Author: @infblueocean.

This PR is based directly on `dev` `ba76b628d5dfe9c4a3a6df8b6973e544da4d0e04` and has no semantic dependency on #1447. To preserve the agreed #1447-first order and avoid resolving the same test insertion twice, rebase this PR after #1447 lands; the current branches overlap only in `test/cpp/config/test_context.cpp`.

## Exact-head validation

- head: `77eafce9ffaa74798a9d7aa429fd1b9b81739b1b`
- tree: `c5d6a1a1601bfcd1e9944f020e537b3c1099af44`
- source clean before and after validation
- all 21 pre-commit hooks passed; the broken-symlink hook skipped because no files matched
- clean Brev L4 release build at the exact head
- `[sirius][config]`: 177 assertions across 10 test cases passed, including the invalid-hysteresis and omitted/null/one-sided cases
- test binary SHA-256: `9c94d4a615de8035e131e24af18f4b78c57f10a24c082bcf3f1ccfe659b34a8b`

GitHub reports this head `MERGEABLE` and `APPROVED`. Fresh hosted validation is running; completed jobs are green and the only skips are the expected CUDA 12/13 distribution placeholders.


infblueocean:
CI status at exact head `0a66e75bf95b6bd5d27a56f6b137ee0fb33bdac1`: Test, Check, and Distribution are `action_required` with zero jobs created (fork-workflow approval), not source failures. Could a maintainer approve those workflows? This will remain draft until the focused compiled selectors and CI pass.

infblueocean:
Compiled validation is complete at the exact PR head and recorded in the PR body; this PR is now ready for review. @mike-wendt, could you review when convenient and approve the remaining fork workflows?

src/sirius_config.cpp
```diff
@@ -62,6 +64,15 @@ uint64_t derived_default_batch_size()
 
 // ================ from_yaml for external types ================= //
 
+static void validate_downgrade_fractions(std::string_view scope, double trigger, double stop)
+{
+  if (stop <= 0.0 || stop >= trigger) {
+    throw std::runtime_error(std::string(scope) +
+                             ": downgrade_stop_fraction must be greater than zero and less than "
+                             "downgrade_trigger_fraction");
+  }
```
mike-wendt:
```suggestion
  if (stop <= 0.0) {
    throw std::runtime_error(std::string(scope) +
                             ": downgrade_stop_fraction must be greater than zero");
  if (stop >= trigger) {
    throw std::runtime_error(std::string(scope) +
                             ": downgrade_stop_function must be less than downgrade_trigger_function");
  }
```
nit: split these so you have separate error messages for both conditions

mike-wendt (approved):
@infblueocean thanks for this, one small NIT

Let me know if you want to change it or keep it AS-IS and we can merge

src/sirius_config.cpp
infblueocean: Addressed in 9a30274f: the zero and stop-at/above-trigger cases now have distinct diagnostics, and each invalid fixture asserts the matching constraint-specific message. Full CI is rerunning on the pushed branch.

bwyogatama (approved):

PR #1454, opened 2026-08-10
fix: make standard debug builds non-sanitized

dhruv9vats:
Makes standard debug builds non-sanitized (the sanitizer flags were transitively included via duckdb configs). This allows debug builds to run certain unit tests that test memory management behaviors without crashing with OOM now.

### DIFF
--- a/cmake/CMakePresets.json
+++ b/cmake/CMakePresets.json
@@ -123,7 +123,9 @@
         "CMAKE_BUILD_TYPE": "Debug",
         "CMAKE_CUDA_FLAGS_DEBUG": "-g -lineinfo -O0",
         "CMAKE_CXX_FLAGS_DEBUG": "-g -O0",
-        "CMAKE_C_FLAGS_DEBUG": "-g -O0"
+        "CMAKE_C_FLAGS_DEBUG": "-g -O0",
+        "ENABLE_SANITIZER": "FALSE",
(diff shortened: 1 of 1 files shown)


9prady9 (approved):

cmake/CMakePresets.json
```diff
@@ -123,7 +123,9 @@
         "CMAKE_BUILD_TYPE": "Debug",
         "CMAKE_CUDA_FLAGS_DEBUG": "-g -lineinfo -O0",
         "CMAKE_CXX_FLAGS_DEBUG": "-g -O0",
-        "CMAKE_C_FLAGS_DEBUG": "-g -O0"
+        "CMAKE_C_FLAGS_DEBUG": "-g -O0",
+        "ENABLE_SANITIZER": "FALSE",
+        "ENABLE_UBSAN": "0"
       },
       "displayName": "Clang Debug",
```
9prady9, line 130: nit be worth adding a preset with `with sanitizer`

dhruv9vats: they exist: https://github.com/sirius-db/sirius/blob/c6c51bc0373b8accaec9940add26c4823e3484f5/cmake/CMakePresets.json#L152-L182

9prady9: ah interesting why was this added to debug build then

PR #1455, opened 2026-08-10
feat(gpu): size the GPU allocation from the plan

Jedi18:
## Summary

Replaces the fixed `gpus_per_query` count with an estimate derived from the query's own plan. Before any pipelines are built, the engine walks the physical plan for scan operators and sizes the allocation from their projected output.

Default behavior (`admission_bytes_per_gpu = 0`) is unchanged.

## What changed

- **Plan walk** — `collect_gpu_scans` mirrors `sirius_physical_plan_generator::set_parent_ops`, the canonical full-tree descent. Two operators keep subtrees outside `children[]`: `RESULT_COLLECTOR` (the engine-injected root) holds its child in `plan`, and `DELIM JOIN` holds `join` and `distinct_root`. `get_children()` is not sufficient — only the collector overrides it — so a walk built on it silently misses every scan inside a delim join, under-counting bytes and admitting the query onto too few GPUs. It lives in `src/planner/gpu_admission.cpp` so the header needs only a forward declaration rather than four operator headers.
- **Estimate** — total bytes = sum over scans of `estimated_cardinality × per-row width`. GPU count = `ceil(total / admission_bytes_per_gpu)`, clamped to `[1, active_gpus]`.
- **Column widths** — fixed-width types defer to `logical_type::fixed_width_byte_size()`, which already tracks the cuDF carrier each one lands in: DECIMAL follows its precision to DECIMAL32/64/128, and `HUGEINT`/`UHUGEINT` cost 8 bytes rather than 16 since cuDF has no 128-bit integer and narrows both to INT64/UINT64. Variable-width types have no static width and cost `avg_variable_column_bytes`.
- **Unsized scans** — a scan reporting zero rows is ambiguous: provably empty, or simply unestimated. Sirius already reads zero as "cannot size" (`sirius_physical_sort_sample` gates on `estimated_cardinality > 0`), so one unsized scan makes the whole estimate unavailable and admission keeps the full capped fleet, rather than sizing a possibly-large query off whichever scans happened to carry numbers. `avg_variable_column_bytes` is likewise rejected at zero, which would zero out every variable-width column.
- **Overflow** — `estimated_cardinality` is an unbounded planner estimate, so the per-scan sum saturates rather than wrapping, and the GPU count uses `1 + (total-1)/per_gpu` clamped in 64-bit before narrowing. Both guard the same failure: a wrapped total reads as a *small* query, which would admit the largest possible query onto one GPU.
- **Composition** — `topology.gpus_per_query` still applies and bounds the estimate, so the two compose as `min(estimated, cap)`.

## Config

Under `operator_params`:

```yaml
admission_bytes_per_gpu: 0      # 0 disables estimation
avg_variable_column_bytes: 32   # fallback width for VARCHAR/LIST/...
```

Both are also registered as DuckDB `SET` variables and take effect from the next query, since admission runs per query. `topology.gpus_per_query` stays YAML-only — it is read when the GPU memory spaces are configured, so a `SET` would silently not apply.

## What this knob is (and isn't)

`admission_bytes_per_gpu` is a parallelism dial, not a memory budget. Measured on 4×GB200 against TPC-H SF500, per-GPU peak memory was unchanged between a 3-GPU and a 4-GPU admission (~163–170 GiB either way, within the inter-GPU spread of a single run), because peak is bounded by partition sizing rather than by data volume divided across devices. Fewer GPUs means each processes more partitions sequentially, so the cost shows up as wall-clock: 23.16 s on 3 GPUs against 21.96 s on 4. Tune it against how much of the fleet a query should occupy, not against VRAM.

## Tests

The width calculation, the GPU-count arithmetic (including both overflow paths) and the unsized-scan rule live in `planner/gpu_admission.hpp` as pure functions, free of operator types so they can be exercised directly — 21 new cases. Two more cover the config path: `avg_variable_column_bytes` rejected at zero, `admission_bytes_per_gpu` accepted at zero as the off switch.

The plan walk is cross-checked against `test_plan_tree_shape`'s own independently written tree visitor, over a real `RIGHT_DELIM_JOIN` plan generated from SQL: both descents must find the same scans. Two separately authored walks agreeing is a stronger property than a hard-coded count, and it is the check that catches a missed subtree.

## Verification

- Full C++ suite green.
- On TPC-H SF500 Q9: estimates 129534 MiB over 6 scans and admits onto 3 of 4 GPUs, with results identical to the all-GPU run. With `gpus_per_query: 2` set alongside, the cap binds and the query admits onto 2.

## What this does NOT do

The estimate is static — it uses DuckDB's pre-execution cardinalities, not runtime measurements. It also reads *logical* types, so it misses the narrower physical carriers compressed materialization can install, over-charging those columns and biasing admission slightly upward.

Dynamic tapering (releasing GPUs mid-query as joins reduce volume) is future work and would build on the runtime estimator from the data-size-estimation branch.

### DIFF
--- a/src/include/planner/gpu_admission.hpp
+++ b/src/include/planner/gpu_admission.hpp
@@ -16,10 +16,30 @@
 
 #pragma once
 
+#include "helper/logical_type.hpp"
+
+#include <algorithm>
 #include <cstddef>
+#include <cstdint>
+#include <limits>
+#include <optional>
 #include <vector>
 
+namespace sirius::op {
+class sirius_physical_operator;
+}  // namespace sirius::op
+
 namespace sirius::planner {
+
+/**
+ * @brief Collect every GPU_SCAN operator reachable from @p node.
+ *
+ * Mirrors sirius_physical_plan_generator::set_parent_ops, the canonical full-tree descent.
+ * RESULT_COLLECTOR holds its child in `plan` and DELIM JOIN holds `join`/`distinct_root`,
+ * both outside `children[]`; missing either under-counts scans and admits too few GPUs.
+ */
+void collect_gpu_scans(const op::sirius_physical_operator& node,
+                       std::vector<const op::sirius_physical_operator*>& out);
 
 /**
  * @brief Narrow an active-GPU list to the first @p cap entries.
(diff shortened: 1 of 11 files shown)


src/sirius_config.cpp
```diff
@@ -274,6 +274,8 @@ static void from_yaml(const YAML::Node& node, operator_params& opt)
   r.optional("dynamic_filter_keep_threshold", opt.dynamic_filter_keep_threshold);
   r.optional("enable_pinned_zone_map_pruning", opt.enable_pinned_zone_map_pruning);
   r.optional("enable_compressed_materialization", opt.enable_compressed_materialization);
+  r.optional("admission_bytes_per_gpu", yaml::bytes(opt.admission_bytes_per_gpu));
```
kevkrist, line 283: Can you verify that these are nonnegative -- it looks like read_yaml() at yaml_reader.hpp:255 will cast this as signed integral type during the parse which will then get cast to uint64_t. e.g. -1 -> infinity forcing 1 gpu if I understand the logic correctly. Obviously this isn't done for other unsigned values but now's as good a time to start as any. And avg_variable_column_bytes should probably be strictly positive, right?

src/include/planner/gpu_admission.hpp
```diff
@@ -16,11 +16,51 @@
...
 
 namespace sirius::planner {
 
+/**
+ * @brief Collect every GPU_SCAN operator reachable from @p node.
+ *
+ * Mirrors sirius_physical_plan_generator::set_parent_ops, the canonical full-tree descent.
+ * RESULT_COLLECTOR holds its child in `plan` and DELIM JOIN holds `join`/`distinct_root`,
+ * both outside `children[]`; missing either under-counts scans and admits too few GPUs.
+ * get_children() is not enough — only RESULT_COLLECTOR overrides it, and delim join's
```
kevkrist: Can you delete just this last sentence?

src/include/planner/gpu_admission.hpp
```diff
@@ -16,11 +16,51 @@
...
+ * declaration is commented out.
+ */
+inline void collect_gpu_scans(op::sirius_physical_operator& node,
```
kevkrist: I think at some point we should think about merging some of our tree walks -- not necessary now I don't think, but you could probably fold the scan bytes accumulation into the set_parent_ops walk.

Also, can sirius_physical_operator be const?

src/include/planner/gpu_admission.hpp
```diff
@@ -38,4 +78,117 @@ inline std::vector<int> apply_gpu_cap(std::vector<int> gpu_ids, int cap)
...
+ *
+ * Reads *logical* types, so it misses the narrower physical carriers compressed
+ * materialization can install, over-charging those columns and biasing admission upward.
+ */
+inline uint64_t estimate_bytes_per_row(const std::vector<sirius::logical_type>& types,
+                                       uint64_t avg_var_bytes)
+{
+  uint64_t total = 0;
+  for (const auto& t : types) {
+    switch (t.id()) {
```
kevkrist: Does this duplicate logical_type::fixed_width_byte_size()?

src/include/planner/gpu_admission.hpp
```diff
@@ -16,11 +16,51 @@
 
 #pragma once
 
+#include "helper/logical_type.hpp"
+#include "op/sirius_physical_delim_join.hpp"
+#include "op/sirius_physical_operator.hpp"
+#include "op/sirius_physical_operator_type.hpp"
+#include "op/sirius_physical_result_collector.hpp"
```
kevkrist: Consider putting implementation in a source file to avoid some of these heavy headers in this header file.

src/sirius_config.cpp
kevkrist, line 283: Should these be SET-able? I think it would be an easy add, and it would allow users to tune these values without recompiling.

kevkrist (commented):
Some nits and small requests

src/sirius_config.cpp
Jedi18: Thanks, this got fixed on dev so I've rebased it. And yes `avg_variable_column_bytes` should be strictly positive

src/include/planner/gpu_admission.hpp
Jedi18: Thanks yes, made it const and as for the tree walk I agree, I'll create an issue for it

src/include/planner/gpu_admission.hpp
Jedi18: Yep, thanks!

src/sirius_config.cpp
Jedi18: Yep added, except for `topology.gpus_per_query` which should be YAML-only since it's read when the GPU memory spaces are configured at startup

kevkrist (approved):

bwyogatama (approved):

PR #1443, opened 2026-08-07
build: Fix ``simpatico_codegen`` cccl cmake for CUDA 12

jrbourbeau:
I was attempting to build the latest `dev` branch of sirius on CUDA 12 but ran into this compilation error

```
CMake Error at /home/shadeform/sirius/src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake:18 (message):
  embed_cccl_headers: CCCL_DIR
  '/home/shadeform/sirius/.pixi/envs/cuda12/targets/x86_64-linux/include/cccl'
  is not a directory
```

<details>
<summary>Full traceback:</summary>

```
cd duckdb && cmake --preset release
-- Found Python3: /home/shadeform/sirius/.pixi/envs/cuda12/bin/python3.14 (found version "3.14.6") found components: Interpreter
-- The C compiler identification is GNU 14.3.0
-- The CXX compiler identification is GNU 14.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /home/shadeform/sirius/.pixi/envs/cuda12/bin/x86_64-conda-linux-gnu-cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/shadeform/sirius/.pixi/envs/cuda12/bin/x86_64-conda-linux-gnu-c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
-- Found Git: /usr/bin/git (found version "2.43.0")
fatal: No names found, cannot describe anything.
CMake Warning at CMakeLists.txt:372 (message):
  git is available (at /usr/bin/git) but has failed to execute 'describe
  --tags --long', likely due to shallow clone.  Consider providing explicit
  OVERRIDE_GIT_DESCRIBE or clone with tags.  Continuing with dummy version
  v0.0.1


-- GIT_COMMIT_HASH has length 8 different than the expected 10
-- git hash d8cdaa33, version v0.0.1, extension folder v0.0.1
[...]

PR #1464, opened 2026-08-10
ci: update ``run-on-pr`` CI job exclusion logic

jrbourbeau:
While working on https://github.com/sirius-db/sirius/pull/1443, I noticed there's actually CI coverage for cuda 12 pixi builds but it's excluded on PRs with 

https://github.com/sirius-db/sirius/blob/e9f30e427f3ee76ae77fb4bc1471f87fd04fc1dd/.github/workflows/test.yml#L47

and this GHA `exclude:`

https://github.com/sirius-db/sirius/blob/e9f30e427f3ee76ae77fb4bc1471f87fd04fc1dd/.github/workflows/test.yml#L48-L51

meant to make it so it's not run on PRs but is run otherwise (like on the `dev` branch).

This (and other places with the same logic like `check.yml`) end up getting unintentionally skipped on non-PRs because GHA uses loose equality, where false and null both coerce to 0, so rows marked `run-on-pr: false` end up get excluded by accident. 

This PR updates things to use `'no-match'` instead of `'null'` as the fallback value instead. Unrelated, I also switched to using the `case()` function https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#case, which has the same result as the existing logic but IMO is easy to reason about (happy to stick with the existing logic if folks prefer).

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -112,7 +112,7 @@ jobs:
         exclude:
           # On PRs, drop rows marked merge/dev-only; on merge_group/dev, match nothing.
           - build:
-              run-on-pr: ${{ fromJSON(github.event_name == 'pull_request' && 'false' || 'null') }}
+              run-on-pr: ${{ case(github.event_name == 'pull_request', false, 'no-match') }}
       fail-fast: false
     name: build (${{ matrix.build.arch }}, ${{ matrix.build.compiler }}, ${{ matrix.build.build-type }}, ${{ matrix.build.environment }}, ${{ matrix.build.cudf-channel }})
     runs-on: >-
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,7 +48,7 @@ jobs:
         exclude:
           # On PRs, drop CUDA variants marked merge/dev-only; on merge_group/dev, match nothing.
           - cuda:
-              run-on-pr: ${{ fromJSON(github.event_name == 'pull_request' && 'false' || 'null') }}
+              run-on-pr: ${{ case(github.event_name == 'pull_request', false, 'no-match') }}
       fail-fast: false
     runs-on: ${{ fromJSON(format('["self-hosted", "{0}", "cpu-xl"]', matrix.cuda.runner)) }}
     timeout-minutes: 60
@@ -118,7 +118,7 @@ jobs:
         exclude:
           # On PRs, drop CUDA variants marked merge/dev-only; on merge_group/dev, match nothing.
           - cuda:
-              run-on-pr: ${{ fromJSON(github.event_name == 'pull_request' && 'false' || 'null') }}
+              run-on-pr: ${{ case(github.event_name == 'pull_request', false, 'no-match') }}
       fail-fast: false
     runs-on: ${{ fromJSON(format('["self-hosted", "{0}", "gpu-2xt4"]', matrix.cuda.runner)) }}
     timeout-minutes: 60
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,6 +62,7 @@ jobs:
         with:
           cache: false
           pixi-version: v0.71.1
+          environments: ${{ matrix.cuda.pixi-environment }}
           activate-environment: ${{ matrix.cuda.pixi-environment }}
 
       - name: Configure sccache backend
@@ -132,6 +133,7 @@ jobs:
         with:
           cache: false
           pixi-version: v0.71.1
+          environments: ${{ matrix.cuda.pixi-environment }}
           activate-environment: ${{ matrix.cuda.pixi-environment }}
 
       - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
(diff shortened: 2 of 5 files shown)


mike-wendt:
Hi @jrbourbeau I appreciate the PR; however, the current implementation is intentional and by design. We use merge queues in this repo to merge changes. During the merge queue tests both Check and Test workflows run CUDA 12 & 13 builds to ensure there are no issues prior to merge. If there is an issue it's rejected from the queue and we fix the issue. The results of these merge queue tests are visible in the PR as well for debugging.

Right now, I'm not inclined to change this process in your proposed way due to the time and resource costs to run CUDA 12 builds and testing on every PR commit. 

I do agree we should have an override for Check and Test triggering similar to #1331. The challenge with `workflow_dispatch` is they can only runs on branches in this repo and not on forks. For now, let's rely on your reviewer to queue #1443 and see what the outcome is first before making either change.

jrbourbeau:
Thanks for the context @mike-wendt, I'm still getting familiar with the dev workflow here

I could definitely be missing something, but my goal here was to make it so nothing changes on PRs (e.g. cuda 12 builds are still excluded) but non-PR CI jobs do run the cuda 12 build (and other checks that use `run-on-pr: false`). From what I could tell it looked like non-PR CI jobs were also excluding those `run-on-pr: false` builds -- am I missing something? 

EDIT: Like specifically looking at this build https://github.com/sirius-db/sirius/actions/runs/31405261822 (which I think is a merge group), I don't see cuda 12 builds

mike-wendt:
My apologies @jrbourbeau I am incorrect and misread your PR. You are fixing an active issue where we are not testing CUDA 12 at all. Thanks for catching this, as it appears we've been missing these builds for over a month. I will review this and get it into the merge queue. We can then see what other issues unrelated to your PR are there, hopefully there are none.

Really thank you for explaining and showing evidence, my apologies once again

mike-wendt (approved):
Thanks @jrbourbeau this looks good and addresses the current CUDA 12 testing gap

mike-wendt:
@jrbourbeau sorry today is a very off day. We do test with Distribution [workflow](https://github.com/sirius-db/sirius/actions/runs/31405260682) in the merge queue that builds CUDA 12 and 13 as well. Let me check if that is full coverage as in Check/Test just to see if we are covered by Distribution. The developer that made these changes is on leave so I'm unable to ask them about the coverage. I'll research this and come back to you when I can

jrbourbeau:
No worries! My understanding (which could be wrong) is that the distribution workflow do the vcpkg builds, but not the pixi/conda builds. But you should definitely look -- I'm not as aware of all the ins and outs as you : )

mike-wendt:
Confirmed that gap, we are in the queue and will see if it passes or if we find any untested issues that we need to fix.

jrbourbeau:
@mike-wendt the cuda 12 pixi build is failing, but during the pixi environment setup step before the actual sirius build happens https://github.com/sirius-db/sirius/actions/runs/31431974152/job/93597302040. Pushed a commit to hopefully resolve that issue and make sure the right environment is installed for each CI job https://github.com/sirius-db/sirius/pull/1464/commits/607ae3fa79323343e96bd3e54ef6ba59035d30c9

mike-wendt (approved):
Thanks for the quick fix

jrbourbeau:
Okay, now we're seeing the 

```
CMake Error at /home/runner/actions-runner/_work/sirius/sirius/src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake:18 (message):
  embed_cccl_headers: CCCL_DIR
  '/home/runner/actions-runner/_work/sirius/sirius/.pixi/envs/cuda12/targets/x86_64-linux/include/cccl'
  is not a directory
[32/1283] Building CXX object cucascade/CMakeFiles/cucascade_objects.dir/src/memory/reservation_manager_configurator.cpp.o
```

error in CI https://github.com/sirius-db/sirius/actions/runs/31435289905/job/93607941721 that https://github.com/sirius-db/sirius/pull/1443 fixes

mike-wendt:
@jrbourbeau let's merge #1443 now and once on `dev` we can add this to the PR to the merge queue again to see if it clears. An alternative would be to pull these changes into #1443 and test the combined changes together. 

I'm open to either, how would you like to proceed?

jrbourbeau:
No strong thoughts from me. I went ahead and cherry picked the one commit from https://github.com/sirius-db/sirius/pull/1443 over to here. Hopefully all the CI builds should be green now

jrbourbeau:
@mike-wendt there were cuda 12 unittest failures due to thrust headers not being included

```
thrust/iterator/detail/device_system_tag.h(31): catastrophic error: could not open source file "thrust/system/cuda/detail/execution_policy.h" (no directories in search list)
  #include __THRUST_DEVICE_SYSTEM_TAG_HEADER
                                            ^

1 catastrophic error detected in the compilation of "codegen_jit.cu".
Compilation terminated.
```

Added this commit https://github.com/sirius-db/sirius/pull/1464/commits/3f0d216ea885c987144dbf7f3cabb26249cdb08d which resolves that (also confirmed that unit tests pass on an H100 with the cuda 12 build)

mike-wendt:
Thanks @jrbourbeau for the quick patching 🥷 skills here, I've queued this again so 🤞 this is the last untested unittest failure

PR #1466, opened 2026-08-10
perf(join): runtime distinct-build probe for BUILD_PROBE hash joins

kevkrist:
Performance:
TPC-H SF1000 (GB300, best-of-3, measured stacked on the #1371 perf branch under its reproduction config): 22-query suite (-4.8%); q11 -48%, q9 -18%, q5 -13%, q4 -7%, q3 -6%, q10 -3%; no per-query regression above run-to-run noise; results byte-identical across 7 consecutive suite runs. Behind enable_runtime_distinct_build_probe (default on).

## Description
cudf's general hash_join probes twice (a count pass to size the output, then a retrieve pass) while cudf::distinct_hash_join probes once, but the single-pass path was gated on a planner proof of build-key uniqueness -- a declared PRIMARY KEY, which costs an ART index -- memory and insert/update/delete cost. Test distinctness at runtime instead: one cudf::distinct_count pass over the cached build, in BUILD_PROBE mode only, where a single test amortises across every probe batch. 

## Checklist
- [x] Read CONTRIBUTING.md
- [x] Update human and agent documentation (README.md, docs/, skills, CLAUDE.md

### DIFF
--- a/src/include/sirius_config.hpp
+++ b/src/include/sirius_config.hpp
@@ -72,6 +72,15 @@ constexpr double DEFAULT_MAX_SORT_PARTITION_MEMORY_FRACTION = 0.33;
 /// disable (always use filtered_join).
 constexpr double DEFAULT_MARK_JOIN_BUILD_SWITCH_RATIO = 8.0;
 
+/// Test build-key uniqueness at runtime when the planner could not prove it statically.
+///
+/// cudf's general hash join probes twice — a count pass to size the output, then a retrieve pass —
+/// while cudf::distinct_hash_join probes once, because a distinct build bounds the output by the
+/// probe row count. Sirius already implements both, but the distinct path is gated on a *proof* of
+/// uniqueness, which only a declared PRIMARY KEY on a catalog table can supply. The runtime test is
+/// one cudf::distinct_count pass over the build keys, taken only in BUILD_PROBE mode.
+constexpr bool DEFAULT_ENABLE_RUNTIME_DISTINCT_BUILD_PROBE = true;
+
 }  // namespace config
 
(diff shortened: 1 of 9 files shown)


src/include/sirius_config.hpp
```diff
@@ -72,6 +72,15 @@ constexpr double DEFAULT_MAX_SORT_PARTITION_MEMORY_FRACTION = 0.33;
...
 constexpr double DEFAULT_MARK_JOIN_BUILD_SWITCH_RATIO = 8.0;
 
+/// Test build-key uniqueness at runtime when the planner could not prove it statically.
+///
+/// cudf's general hash join probes twice — a count pass to size the output, then a retrieve pass —
+/// while cudf::distinct_hash_join probes once, because a distinct build bounds the output by the
+/// probe row count. Sirius already implements both, but the distinct path is gated on a *proof* of
+/// uniqueness, which only a declared PRIMARY KEY on a catalog table can supply. The runtime test is
+/// one cudf::distinct_count pass over the build keys, taken only in BUILD_PROBE mode.
+constexpr bool DEFAULT_ENABLE_RUNTIME_DISTINCT_BUILD_PROBE = true;
```
Jedi18, line 82: The new option works in the YAML config, but it cannot be changed with DuckDB’s SET command. We should let the users change it at runtime:
```
SET enable_runtime_distinct_build_probe = false;
```
Could we register this option in `sirius_extension.cpp` and add it to the configuration docs and settings test?

Jedi18 (commented):
Looks good! We should add tests and update the relevant docs though

Jedi18 (approved):

PR #1391, opened 2026-08-03
perf: fused scan-filter — decode-time masks + selection-consuming decompression (SF1000 8.180 s -> 6.918 s)

felipeblazing:
Stacked on #1371 (`repro/sf1000-8.5s`, the 8.180 s integration stack) — the first 61 commits are that PR's history; **review from `13b4b612` onward** for this PR's delta. Like #1371 this is the full-stack experiment branch; the same work exists as it lands in focused PRs.

## What this adds

**Fused scan-filter**: predicate evaluation and row selection pushed *into* simpatico decompression, so filtered rows are never decompressed into full-width columns and payload columns are decoded directly into compacted survivor-sized output. Everything is gated behind `SIRIUS_EXP_FUSED_SCAN_FILTER` (default **off**; gate off is byte-identical to the base stack — re-proven at every milestone).

**Result: 8.180 s → 6.918 s (−15.4%) on TPC-H SF1000 / GB300**, 22/22 byte-identical, best-of-3 with 0.01% spread. Campaign total from 15.99 s: **−56.7%**.

| query | classic | fused | Δ | mechanism |
|---|---|---|---|---|
| q6 | 0.158 | 0.059 | −63% | 3-col range mask → K4 index decode |
| q15 | 0.149 | 0.083 | −44% | date mask + write-skip payloads |
| q14 | 0.156 | 0.091 | −42% | same |
| q1 | 0.756 | 0.467 | −38% | mask + dict-K5 group keys at 52.6% sel |
| q12 | 0.287 | 0.184 | −36% | constant hulls + K3-delta + K4 + K6 |
| q19 | 0.323 | 0.242 | −25% | dual-delivery dict-code masks |
| q17 | 0.224 | 0.178 | −21% | dynamic in-list membership mask |
| q20 | 0.205 | 0.163 | −21% | date mask |
| q8 | 0.252 | 0.205 | −19% | dynamic membership (capped, ordered) |
| q21 | 0.722 | 0.649 | −10% | dynamic suppkey in-list mask |
| q9 | 0.937 | 0.870 | −7% | dynamic Bloom mask |

## Architecture (two-wave decode, all NVRTC-JIT per plan shape)

[...]

PR #1474, opened 2026-08-11
perf: Decompression pushdown

joosthooz:
## Description
This PR is a an attempt to create a cleaner API for pushing down filters or other work into simpatico decompression, based on #1391 and the already merged (via the big perf PR) #1380.

## Checklist
- [x] Read CONTRIBUTING.md


src/compression/simpatico_codegen/include/codegen/selection/decompression_pushdown_policy.hpp
```diff
@@ -0,0 +1,78 @@
...
+// One reader per knob, shared by everything that has to agree on it: the scan
+// (does it bother analysing its filter?), the memory estimator (will the batch
+// come back smaller?) and the decode itself. They used to be copied per
+// translation unit, in three different parse styles, and had already drifted —
+// one reader accepted only "1" where the decode accepts anything but "0", so a
+// value like "true" turned the feature on in one layer and off in another.
+//
+// All are read once and cached: the gate check sits on the per-batch path.
+//
+// TODO: these should be Sirius parameters, not environment variables. As env
```
joosthooz, line 14: We need to decide on what to do with these env vars (e.g. `SIRIUS_EXP_FUSED_SCAN_FILTER`), if we want to keep them or turn into proper Sirius config params (which does require plumbing)

.claude/settings.json
```diff
@@ -1,19 +0,0 @@
-{
```
joosthooz, line 1: Will revert this

DECODE_PUSHDOWN_PLAN.md
joosthooz, line 1: I don't want to merge this into dev, but it would be a waste to lose all the experiments we tried and the results. Where would we want to store development history like this?

It includes lots of things that didn't work, those are also important

docs/super-sirius/compressed-pinning.md
```diff
@@ -56,6 +56,59 @@ Rules of thumb:
   `scan_task_batch_size` of the *pinned subset*, so a narrow pin concentrates a column's
   chars in each batch.
 
+## Filtering while decompressing (experimental)
```
joosthooz, line 79: I've added docs to this file, because the fused filtering we add here works on the pinned compressed tables we document in this file. Let me know if you think it should be in a separate file or elsewhere

src/op/scan/scan_filter_analysis.cpp
```diff
@@ -0,0 +1,481 @@
...
+
+  // DATE decodes to its stored day count, which is not a numeric literal domain
+  // constant_numeric_range covers.
+  if (col_type.id() == sirius::type_id::DATE) {
+    if (value.type().id() != duckdb::LogicalTypeId::DATE) { return std::nullopt; }
+    auto const days = static_cast<int128>(duckdb::DateValue::Get(value).days);
+    return sirius::numeric_range{sirius::numeric_range_domain::SIGNED_INTEGER, days, days, 0};
+  }
+
+  if (!col_type.is_decimal() && !col_type.is_integer()) { return std::nullopt; }
```
9prady9, line 151: `is_integer()` includes `UBIGINT`/`UHUGEINT` (`logical_type.hpp:150-166`), but the decoded domain is signed end to end — the ballot does `static_cast<int64_t>(v)` before comparing (`renderer.cpp:836-837`), and nothing in between gates on signedness.

A `UBIGINT` column with values ≥ 2^63 and `WHERE col >= 100`: those rows cast negative, fail the ballot, and are dropped from the mask. The residual can't restore them — the decode is over-filtering here, not under-filtering. Silently returns too few rows.

Likely latent today (TPC-H has no unsigned columns), but nothing guards it. Suggest refusing `UBIGINT`/`UHUGEINT` here and clearing coverage, like the INT128 refusal at `:160`.

src/compression/compressed_scan.cpp
```diff
@@ -0,0 +1,534 @@
...
+    ++selecting_columns;
+  }
+
+  if (selecting_columns == 0 && !has_external_selection) {
+    SIRIUS_DECOMPRESSION_PUSHDOWN_DIAG(
+      "[decode-filter] plan: no row-selecting source survived — plain decode");
+    return {};
+  }
+  plan.enabled             = true;
+  plan.covers_whole_filter = request.ranges_cover_whole_filter && !dropped_conjunct &&
```
9prady9: `has_external_selection` covers two different things: a dictionary-answered equality, and a dynamic join filter. Only the second should clear coverage — the first is part of the scan's static filter and is enforced on an applied decode (all `bool8_filters` are validated up front at `simpatico_codegen.cpp:458-465`, and any failure refuses the whole decode).

Splitting the flag looks like it recovers most of the 674/928 batches §7 measures. Performance only, so not a blocker.

src/compression/simpatico_codegen/src/selection/selection_wave.cu
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// Up-to-8 source masks, passed by value as kernel params.
+struct mask_src_ptrs {
+  uint32_t const* p[8];
+};
+
+// dst = AND of n srcs, uint4-vectorized (num_words is a multiple of 4: masks
+// are 32 words per chunk). Grid-stride.
+__global__ void mask_and_combine_kernel(uint32_t* __restrict__ dst,
```
9prady9: `dst` is `__restrict__` but aliases `src[0]` by design — `combine_masks_and(combined, mask_ptrs.data(), ...)` with `mask_ptrs[0] == combined` (`simpatico_codegen.cpp:553,563,648`). Formally UB; works today only because each thread reads before it stores.

`test_selection_wave.cpp:187-196` covers the aliased shape, which keeps this low severity. Simplest fix is to drop `__restrict__` from `dst`.

src/compression/simpatico_codegen/tests/test_masked_decode_variants.cpp
```diff
@@ -0,0 +1,943 @@
...
+
+  std::printf("PASS: render contract checks\n");
+  return true;
+}
+
+}  // namespace
+
+int main()
+{
+  if (cudaSetDevice(0) != cudaSuccess) {
```
9prady9, line 917: Returns 1 (FAIL) on a machine without a GPU; the sibling tests skip with exit 0 (`test_selection_wave.cpp:330`, `test_multi_gpu_stream_affinity.cpp:84-96`).

CMakeLists.txt
```diff
@@ -827,6 +831,9 @@ target_include_directories(
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/test/cpp>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/include>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
+    # Fused scan-filter: scan headers include codegen/selection/selection.hpp,
+    # so every target compiling them needs the simpatico include root.
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/compression/simpatico_codegen/include>
```
9prady9: This comment no longer matches the tree: no header under `src/include/op/scan/` includes `codegen/`, and no test under `test/cpp/` does either — the facade work removed that dependency. Looks like a dead include directory now.

.gitignore
```diff
@@ -69,3 +69,6 @@ tools/log_analyzer/metrics/__pycache__/
 test/tpch_performance/__pycache__/
 test/tpch_performance/output/
 test/tpch_performance/log/
+
+# local debug log
+rmm_log.txt
```
9prady9: Looks like a personal debug artifact rather than something the project emits — `.git/info/exclude` may be a better home.

9prady9 (commented):
**CI covers very little of this.** `cuda-12` and `cuda-13` both skip, how do we want to cover this once merged ?

**The env-var gate is why the feature is untested.** `decompression_pushdown_enabled()` caches its value in a function-local static, so it's read once per process. No test can flip it: all Catch2 cases share one process, and SQL can't set env vars at all. As a `SET` option it'd be a two-line test — same query with the gate on and off, assert the rows match. That's the PR's core claim, currently unverified.

Doc gap: `scan.md` gets nothing despite ~950 lines of scan-side change, and its Key Files table (`:497-529`) omits both new files. For the plan doc, `experimental/` already holds one.

Static review — I didn't build or run anything.

src/compression/compressed_scan.cpp
```diff
@@ -0,0 +1,534 @@
...
+{
+  column_capability capability;
+  if (column_index >= table.columns.size()) { return capability; }
+  auto const& column = table.columns[column_index];
+  switch (column.dtype.id()) {
+    case cudf::type_id::INT8:
+    case cudf::type_id::INT16:
+    case cudf::type_id::INT32:
+    case cudf::type_id::INT64:
+    case cudf::type_id::UINT8:
```
kevkrist: The uints are decoded through signed int JIT lanes, and the range ballot sign-extends the decoded value to int64 (renderer.cpp:823), which could erroneously carry the sign bit from the signed integer types which are supposed to represent unsigned values.

src/op/scan/sirius_gpu_scan_operator_data.cpp
```diff
@@ -15,17 +15,102 @@
...
+        kind_rank = 0;
+        num_keys  = small->size();
+      } else if (auto const* set =
+                   dynamic_cast<sirius::op::sirius_dynamic_in_list_filter const*>(filter.get())) {
+        kind_rank = 1;
+        num_keys  = set->size();
+      } else if (filter->kind() == sirius::op::sirius_dynamic_filter_kind::BLOOM) {
+        kind_rank = 2;
+      }
+      // The closure co-owns the filter and binds device 0 (GPU-tier pinned
```
kevkrist: I don't think I understand the bind to GPU 0 here. Wouldn't this cause an illegal memory access if the task is launched on a different GPU? resolve_dynamic_filter_device_id() will just pass device id 0 through if I remember correctly.

src/op/scan/sirius_gpu_scan_operator_data.cpp
```diff
@@ -52,16 +137,102 @@ void scan_operator_input::prepare_for_processing(
...
+            snap.attached_probes,
+            snap.generation,
+            snap.skipped_non_mask);
+          if (snap.attached_probes == 0) { return; }
+          auto const base =
+            rep->decode_scan()
+              ? rep->decode_scan()
+              : std::make_shared<const ::sirius::compressed_scan>(::sirius::scan_decode_request{});
+          rep->set_decode_scan(
+            base->with_membership_probes(std::move(snap.probes), snap.generation));
```
kevkrist, line 192: Membership probes are attached here after the pipeline has already acquired the reservation from get_estimated_working_set_size_in_bytes(). Each probe fully decodes its key column and allocates a boolean result mask plus packed-mask storage, so even a key-only scan can exceed its reservation. Is there a way to have better scratch reservation in the pre-prepare working set estimate?

kevkrist (changes requested):

CMakeLists.txt
joosthooz: Removed, thanks

.gitignore
joosthooz: reverted

src/op/scan/scan_filter_analysis.cpp
joosthooz: Thanks, this has been addressed now

src/compression/simpatico_codegen/src/selection/selection_wave.cu
joosthooz: Fixed! Thanks

src/compression/simpatico_codegen/tests/test_masked_decode_variants.cpp
joosthooz: This should now be consistent

src/compression/compressed_scan.cpp
joosthooz: Thanks for this, these should be rejected now (also in the spot where Pradeep pointed it out)

src/op/scan/sirius_gpu_scan_operator_data.cpp
joosthooz: Should be fixed here [5729339](https://github.com/sirius-db/sirius/pull/1474/commits/57293391b8a93043eee4c5bee023dfd8aff123fd)

src/op/scan/sirius_gpu_scan_operator_data.cpp
joosthooz: Now reserving more memory when a dynamic filter is possible [4066317](https://github.com/sirius-db/sirius/pull/1474/commits/4066317164c7aad0f81ffecfc6310daa9db8d9e1)

src/compression/compressed_scan.cpp
joosthooz: Adressed in [377c5a3](https://github.com/sirius-db/sirius/pull/1474/commits/377c5a3efffd8a09acfe6cae22769303e6224329)

9prady9 (approved):

src/compression/compressed_scan.cpp
```diff
@@ -0,0 +1,565 @@
...
+// decompression_pushdown_scan
+//===----------------------------------------------------------------------===//
+
+std::shared_ptr<const decompression_pushdown_scan>
+decompression_pushdown_scan::without_row_selection() const
+{
+  pushdown_request narrowed;
+  narrowed.columns.reserve(_request.columns.size());
+  for (auto const& column : _request.columns) {
+    narrowed.columns.push_back({column.equals_any, std::nullopt, {}});
```
kevkrist, line 476: It looks like this will trigger decompress_with_pushdown with equality (line 335), which will trigger compaction and cause row ids for MVCC to not align -- which is the intention of this function, right?

src/op/scan/sirius_gpu_scan_operator_data.cpp
```diff
@@ -136,7 +313,51 @@ std::size_t scan_operator_input::get_estimated_working_set_size_in_bytes() const
...
+  // conservative envelope as a known static filter rather than falling
+  // through to the zero-copy view estimate below.
+  bool const dynamic_filter_possible = sirius::decompression_pushdown_enabled() &&
+                                       dynamic_filters != nullptr && !mvcc_keep_mask.has_mask();
+  if (row_filter_pending || dynamic_filter_possible) {
+    // Once compaction has been measured unprofitable, later batches drop the
+    // row selection and decode full width — keep the full-width envelope.
+    bool const unprofitable = pushdown_selection_unprofitable &&
+                              pushdown_selection_unprofitable->load(std::memory_order_relaxed);
+    if (auto const forecast = pushdown_compaction_forecast(*this);
```
kevkrist, line 345: I think this is still under-reserving for when a dynamic filter is attached? Which requires the full key decode + mask budget, right? Maybe this is ok but noting here.

src/compression/compressed_scan.cpp
```diff
@@ -0,0 +1,565 @@
...
+  // full envelope. A column decoded through a dictionary gather also lifts the
+  // selectivity ceiling — that shape wins at every selectivity, so the decode
+  // never gives compaction up for it and the surviving row count is unbounded.
+  bool any_unbounded = false;
+  for (auto const index : selected) {
+    if (index >= chunk.columns.size()) { return forecast; }
+    auto const& plan_tree = chunk.columns[index].plan_tree;
+    if (!plan_tree) { return forecast; }
+    auto const route = simpatico::probe_column(*plan_tree).compact_route;
+    if (route == sirius::codegen::decode_route::full) { return forecast; }
```
kevkrist, line 561: I think a delta INT32 has a non-full `delta_mask` route but `can_produce_mask() == false`, so it can receive but not produce a mask, but it will fall through here to `compacts == true`. Is that intended?

src/compression/compressed_scan.cpp
```diff
@@ -0,0 +1,565 @@
...
+    any_unbounded = any_unbounded || route == sirius::codegen::decode_route::dict_codes;
```
kevkrist: `str_split` memory is not bounded by row selectivity, right? E.g., pathological case, a single row with a huge string that passes filters. It looks like this falls through to `survivors_bounded == true`.

src/compression/simpatico_codegen/src/simpatico_codegen.cpp
```diff
@@ -306,6 +317,508 @@ std::unique_ptr<cudf::table> decompress_columns_parallel(compressed_table const&
...
+    return refuse("num_rows > INT32_MAX (int32 row indices)");
+  for (auto const idx : selected) {
+    if (idx >= table.columns.size()) return refuse("selected column index out of range");
+    auto const& col = table.columns[idx];
+    if (!col.plan_tree || col.num_rows != num_rows)
+      return refuse("selected column missing plan_tree or row-count mismatch");
+  }
+  for (auto const& f : request.filters) {
+    if (f.column >= selected.size()) return refuse("filter directive column out of range");
+    if (f.pred.lo > f.pred.hi) return refuse("empty predicate range (lo > hi)");
```
kevkrist, line 451: Wouldn't this be valid SQL even though it makes no sense -- should return empty result?

src/compression/simpatico_codegen/src/plan/decompress.cpp
```diff
@@ -514,39 +543,118 @@ std::unique_ptr<cudf::column> decode_fused_subtree_impl(PlanTree const& tree,
...
+    // (the index walk rejects it at render), as does the dictionary codes
+    // region by contract. Any anomaly — indices absent, or a count that
+    // disagrees with the mask — silently keeps the mask walk, since the pick is
+    // an optimization and the mask walk is always renderable here.
+    bool const use_index_decode =
+      sel->enumerate_by_index && sel->route == sirius::codegen::decode_route::bitpack_mask &&
+      sel->survivor_count > 0 && root_nid < tree.nodes.size() &&
+      tree.nodes[root_nid].op == "bitpack" &&
+      static_cast<std::int64_t>(sel->survivor_indices.size()) == sel->survivor_count;
+    bool const masked_ok = launch_decode_fused_tree_compacted(
```
kevkrist, line 644: You may want to guard this when survivor_count == 0, since then the data pointer will be null, and the launcher will reject, causing a valid zero result to fall back to full decode.

src/op/scan/sirius_gpu_scan_operator_data.cpp
```diff
@@ -52,8 +140,94 @@ void scan_operator_input::prepare_for_processing(
...
+        }
+      }
+      // Decode-time join filter snapshot: the scan-manager drain runs at query
+      // PREPARE, before any join build has published, so a drain-time snapshot
+      // is empty for the whole scan. Executor tasks prepare right before decode
+      // — by then upstream builds have published — so refresh the projected rep
+      // with a fresh per-batch snapshot here, replacing the (typically empty)
+      // drain-time one. The mapping invariant lives in
+      // snapshot_membership_probes; same mvcc guard as the row selection.
+      if (sirius::decompression_pushdown_enabled() && dynamic_filters &&
```
kevkrist, line 170: It looks like this code block will reattach a filter after it's been potentially dropped above?

src/compression/compressed_scan.cpp
```diff
@@ -0,0 +1,565 @@
...
+    any_asked = any_asked || !column.empty();
+  }
+  if (!any_asked) { return nullptr; }
+  return std::make_shared<const decompression_pushdown_scan>(std::move(narrowed));
+}
+
+decompression_pushdown_scan::compaction_forecast decompression_pushdown_scan::forecast_compaction(
+  simpatico::compressed_table const& chunk, std::span<const std::size_t> selected) const
+{
+  compaction_forecast forecast{};
```
kevkrist, line 520: Should the forecast here also be aware of `kMaxSelectionSources`? Otherwise the reservation will under-reserve when, say, k = 9.

src/op/scan/sirius_gpu_scan_operator_data.cpp
```diff
@@ -136,7 +313,51 @@ std::size_t scan_operator_input::get_estimated_working_set_size_in_bytes() const
...
+      // filtered column — <= batch/4 across the source limit at >= 4 B/row
+      // columns) plus the compacted outputs, and such splits steal their table,
+      // so there is no second output copy. The surviving row count is bounded
+      // by the selectivity ceiling unless a column is exempt from it, in which
+      // case size for up to full width. A decode that gives compaction up
+      // re-runs the full-width path and over-allocates into the adaptor's
+      // over-reservation handling; by policy that only happens where the
+      // forecast was wrong. Replaces a ~5x over-reservation on
+      // highly-selective batches.
+      auto const cap =
```
kevkrist, line 363: This estimate does not inlcude the INT32 survivor-index buffer allocated for the low-selectivity bitpack index walk. FOr narrow columns (INT8), that can be larger than the compacted output and exceed the reservation. Probably fine. Noting here in case you want to include 4 8 survivor_count when the index-walk route could be selected.

kevkrist (approved):
Just a few questions you may want to address.

joosthooz:
@kevkrist all of your comments should have been addressed in [a86c55f](https://github.com/sirius-db/sirius/pull/1474/commits/a86c55fa865e4dc6b7288d8e2cab74f25e2b52a5), thanks for the detailed reviews

PR #1476, opened 2026-08-11
WIP: updated io stack for better caching, and readahead scans

aminaramoon:
## Checklist
- [ ] Read CONTRIBUTING.md


infblueocean:
Foxglove's current configuration-surface audit found one literalness issue relevant to this backend rename.

On current `dev`, `use_sirius_datasource: false` is meaningful for a single GPU, but a realized multi-GPU topology silently warns and rewrites it to `true`; the programmatic scan-manager constructor instead rejects the same conflict. This PR retains the silent-rewrite behavior for `backend: kvikio`.

Could the explicit incompatible choice fail configuration loading instead? Suggested contract:

- omitted/default or explicit `backend: sirius` + multi-GPU: pass
- explicit `backend: kvikio` + realized topology greater than one GPU: throw a field-scoped diagnostic naming the resolved GPU count and required backend
- cover both `topology.num_gpus: 2` and two explicit `space.gpu` entries, so the check binds realized topology rather than only high-level syntax
- keep the scan-manager construction-time rejection as defense in depth

That preserves zero-choice safety while making an impossible user choice visible instead of claiming the requested backend was applied. The single-GPU Kvikio and S3-routing changes in this draft are separate semantics and should retain their own coverage.

infblueocean:
Reactor-count follow-up from the Foxglove configuration census (audited against dev `0d7f98d9` and this PR head `57f84da`):

The current docs describe `uring_n_reactors` and `rest_n_reactors` as thread counts, but the actual resource multiplier is worth making explicit in this PR:

- every uring reactor starts one worker, one ring, and 64 pinned host bounce blocks;
- every REST reactor starts one epoll/libcurl worker and reserves `rest.max_connections` pinned bounce blocks;
- therefore total REST connection/staging capacity is `rest_n_reactors * max_connections`, not merely `max_connections`;
- `templated_ioctx::next_reactor()` currently chooses one round-robin reactor per logical request, so a larger pool raises concurrency across independent requests rather than splitting one read across all reactors.

Please document those multipliers and emit the effective pool totals when each backend starts: reactor count, worker count, block size, total bounce slots, and total bounce bytes. Also make explicit what happens when `uring_n_reactors` is supplied under `backend: kvikio`: the value is inactive and should not look applied. REST needs separate care because this PR still uses REST for LIST/glob control traffic even when kvikio serves S3 data.

I do not think source alone supports changing the shipped `1`/`2` defaults. Existing reference values are inconsistent rather than comparative: production uring defaults to 1, `parquet_benchmark` defaults to 2, the GB300 SF1000 config uses 4 without listing reactor count among its measured sweeps, and the S3 bench fixes REST at 2 while silently using `max_connections=8` instead of the shipped 16.

A bounded screen can settle this without a new framework:

1. Local/NVMe: existing parquet benchmark, reactor counts `{1,2,4}`, one-large-file and 32+-file shapes, balanced five reps on x86 NVMe plus GB300 local storage. Choose the smallest count within 5% of best on both shapes and both hosts; otherwise keep it expert-visible.
2. REST: existing S3 perf path, one-large-object and 32-file glob on same-region AWS S3 plus MinIO. Hold total connection capacity constant with `(reactors,max_connections)={(1,16),(2,8),(4,4)}`, then include shipped `(2,16)` and historical `(1,32)`. Record throughput, queue wait, TTFB, retries/failures, CPU, and staging bytes; use the same smallest-within-5% rule.

All cells remain oracle-correct, fresh-process, balanced-order, and timing-admissible by within-cell spread. Until that screen exists, these are expert/empirical knobs rather than safely hardware-derived defaults.

infblueocean:
Foxglove completed the bounded AWS-S3/L4 max-connections screen discussed above, pinned to Sirius test-harness source `7c47cac9` and binary SHA-256 `9d7eaac6...`.

All 25 cells were oracle-correct, both REST reactors were active and slot-saturated, configured/effective values matched, and no retries, terminal failures, fallback, OOM, or unadmitted diagnostics occurred. The apparent medians favored 32/64 over current `dev` default 16 (aggregate runtime ratios were roughly 0.85/0.82), but every candidate failed the predeclared <5% timing-spread gate (8.0%-17.3%). The only honest outcome is therefore `TIMING_INADMISSIBLE_MECHANISM_ONLY`: the run proves the setting binds and the tested path is saturated, but it does not validate a new default.

This draft's commit `9c897bc9` changes `rest::config::max_connections` from 16 to 64, while the draft documentation still says 16 and its benchmark helper defaults to 128. Is 64 intended as the shipped general default or only as a benchmark candidate? What endpoint/hardware evidence supports that choice, and should the source default remain 16 until a timing-admissible cross-endpoint comparison exists?

We are running a narrower 16/32/64 paired follow-up with direct GPU state telemetry; I will post only if it produces admissible evidence.

ran-yuan-rui:
@aminaramoon 

The gatekeeper borrow model and the split producer/consumer state are clean. One issue I think is worth fixing/discuss before the scheduler API settles: admission is query-wide, while IO capacity is backend-specific.

## Why

Today `prepare_for_query` takes the largest `n_max_concurrent_scans()` from all active ioctxs and uses that backend's strategy for the whole query:

```cpp
for (auto const& [_, ctx] : _routed_io_ctxs) {
  if (ctx->n_max_concurrent_scans() > backend_budget) {
    widest = ctx.get();
    backend_budget = ctx->n_max_concurrent_scans();
  }
}

auto plan = config.resolve_readahead(
  backend_budget, backend_prefetch_strategy(widest->type()));

_readahead = std::make_shared<readahead_scan_manager>(plan.budget);
```

This works for a query using one backend, but mixed queries couple unrelated transports. For example, with local parquet joined to S3, REST supplies the larger budget and eager strategy. Local NVMe scans then inherit a queue depth intended to hide S3 latency, which can create unnecessary pinned-buffer and reservation pressure.

The shared ticket pool also couples demand reads. A burst of cold S3 reads can borrow the shared gate into debt and stop local readahead even while the disk is idle.

The current S3RDMA path is cache-unarmed, so it would not issue readahead IO, but its cold reads can still be charged against a gate sized for another backend. More generally, a backend whose concurrency limit comes from registered memory or queue depth should not inherit a connection-count budget from REST.

## Suggested Changes

I think the least invasive ownership is one `readahead_scan_manager` per active ioctx, with `scan_manager` holding the small map. Each manager keeps the current gatekeeper, per-operator FIFOs, cursor, worker loop and borrow behavior.

The manager should be chosen after per-path routing:

```cpp
auto io_ctx = ioctx_for_path(file_path);
auto readahead = readahead_for(io_ctx);  // budget and strategy from this ioctx

readahead->register_scan_task(split, operator_id);
split->set_readahead_domain(readahead);
```

This needs to be per split rather than per operator. One scan operator can contain files routed to different ioctxs, and routed contexts may be created lazily after `prepare_for_query` begins.

Cold-read accounting then follows the split's domain:

```cpp
if (auto domain = split->readahead_domain()) {
  domain->update_scan_state(operator_id, split, stage);
}
```

Cursor and queue rules remain unchanged inside each backend. Query-stage idle signals such as `on_task_queue_empty` can be broadcast to every active manager, since executor idleness is shared, while admission remains transport-local.

One config detail needs an explicit decision. If `max_readahead_scans` becomes a per-backend cap, total query concurrency can become the sum of those caps rather than the old single value. That may be the intended behavior, but it should be documented. If it must remain a query-wide memory bound, it can stay as a separate outer cap while each ioctx keeps its own admission gate.

No new backend concepts should otherwise be needed: `n_max_concurrent_scans()` and the backend strategy already exist.

ran-yuan-rui:
@aminaramoon 
One more boundary question in the same spirit as the readahead one: who owns LIST.

## Why

With `backend: kvikio`, data reads leave REST but listing cannot, since KvikIO has no LIST capability. The scan manager therefore reaches for REST explicitly through `rest_ioctx_for_list()`. Glob expansion calls `list_objects_paged()` and `s3_list_max_matches()` through that path, and the resulting object sizes are carried into the later opens.

That behavior is right for KvikIO today. The concern is the ownership shape. The scan manager depends on the concrete REST ioctx for an object-store capability, and the pairing "KvikIO serves data, REST serves LIST" exists only inside those call sites.

A future S3RDMA transport with its own control plane would still list through REST. Its LIST requests would bypass that transport's admission and failure domain and could use a different endpoint or credential set from the data path. The data transport could appear healthy while listing runs outside its own error handling.

The open cuCascade PR NVIDIA/cuCascade#181 proposes the boundary I think we want here: a concrete `rest::object_store_lister` owns pagination, parsing, caps and anti-loop checks, while each backend injects the operation that fetches one page. Keeping the same ownership model in Sirius would also avoid another seam that has to be reconciled by hand when IO code moves between the trees.

## Suggested change

Use an injected lister component instead of casting the selected context to `rest_ioctx`.

- REST composes the lister with its existing page fetch.
- KvikIO is explicitly paired with a REST-backed lister. This preserves today's behavior, but makes the split control/data ownership visible.
- S3RDMA composes the same lister with its control-plane page fetch, including its admission and failure handling.

Conceptually, backend selection returns both responsibilities:

```cpp
struct object_store_route {
  std::shared_ptr<ioctx> data;
  object_store_lister* lister;
};
```

The two scan-manager LIST call sites use the selected lister:

```cpp
auto route = object_store_route_for(s3_prefix_uri);
route.lister->list_objects_paged(bucket, prefix, page_size, sink, max_scanned);
```

The exact ownership does not have to follow this sketch, but the data backend and lister should be selected together. Known-size propagation remains unchanged, and connection warmup remains the responsibility of the data ioctx.

There is also one config combination worth rejecting explicitly. #1476 now parses both `backend` and `object_store.s3_transport`; `backend: kvikio` together with `s3_transport: RDMA` should fail during config or registry construction rather than silently letting KvikIO win. `AUTO` or `HTTP` can continue to mean KvikIO data with REST LIST.

PR #1478, opened 2026-08-11
refactor(scheduler): two-sleep-site management loop that hears direct queue pushes

bwyogatama:
## Problem

Same deadlock as #1467: the management event loop sleeps only on the task-request channel, and the downgrade executor's TIER-2 pass returns extracted pipeline tasks via `convertible_gpu_pipeline_task`'s RAII destructor — a direct `_task_queue.push()` that emits no channel event. If every executor drained and parked while the tasks were out for GPU→host conversion, the matcher never runs again: runnable tasks sit in the queue next to parked ready devices forever. Reproduced deterministically at SF1000 TPC-H q18 with all tables host-pinned (hung 7/7); see #1467 for the full interleaving and gdb evidence.

## Approach (alternative to #1467)

#1467 keeps the single-sleep-site loop and makes the silent producer emit the missing `task_available` event. This PR instead restructures the loop so it can hear direct queue pushes **by construction**, choosing its sleep site from queue state:

Each pass tops up the two things the matcher needs, sleeping at most once per missing ingredient:

| Step | Action | Why it's safe |
|---|---|---|
| 1. Devices | collect channel events non-blockingly; **block on `get()` only when no device is parked** | no parked device means every executor is busy and will post `device_ready` |
| 2. Work | **sleep on the new `_task_queue.wait()` whenever the queue is empty** (issuing `schedule_lookahead` first if a device waits) | the queue's own condvar hears *every* push — `schedule()` or the downgrade's silent RAII return |

When a parked device coexists with tasks pinned to busy devices, both guards fall through and the loop busy-waits by design (see trade-off 1).

Supporting changes:
- `multi_index_priority_queue::wait()` — blocks until non-empty **without extracting**, so the matcher keeps selecting tasks by device preference via `try_pop_from`. Returns `false` on `interrupt()`, which `task_scheduler::stop()` already calls, so shutdown wakes both sleep sites.
- `push()` now uses `notify_all` — with mixed waiter kinds (`pop()` vs `wait()`) on one condvar, `notify_one` could wake the non-consuming waiter and strand a `pop()` despite an available task.
- The matcher is factored into `run_matcher()`, returning the dispatch count that drives the sleep-site decision (this is what prevents a hot spin when the queue holds only tasks pinned to busy devices).
- The lookahead call is preserved: with an empty queue and a parked device, `schedule_lookahead()` runs *before* sleeping on the queue; anything it creates arrives through `schedule()` and wakes the sleep.
- Unit tests for `wait` (wake-on-push without extraction, interrupt returns false, mixed-waiter wake).

## Known trade-offs vs #1467

1. **Dispatch-latency bubble**: while blocked on the channel (non-empty queue, nothing dispatchable), a silent downgrade return is not noticed until the next channel event. Bounded by the busy device's current task — a stall, not a deadlock. #1467's explicit notify dispatches immediately in this case.
2. **Lookahead deafness while asleep on the queue**: `device_ready` events arriving during a queue sleep buffer instead of triggering `schedule_lookahead` immediately, and the queue sleep can be entered with no parked device (after consuming a non-device event with an empty queue). No-op under the default `active` creation strategy; under the non-default `lookahead` strategy this delays — and in the buffered-device case could stall — lookahead until the next queue push. Flagging for reviewer scrutiny.
3. In exchange, the loop tolerates **any** future producer that pushes into the queue without going through `schedule()` — the class of bug in #1467 can't recur.

## Validation status

- **Deterministic regression test** (`[deadlock-1467]` in `test_task_scheduler.cpp`): stages the extract-while-parked interleaving with full fidelity — tasks extracted with `mutable_pop_if` and returned via `convertible_gpu_pipeline_task`'s RAII destructor while every executor is parked. **Times out (deadlock) against dev's loop and passes with this PR's loop** — verified in both directions on hardware, so the trigger no longer depends on benchmark timing luck.
- Unit: all 30 `[task_scheduler]`/`[pipeline_queue]`/`[multi_index_priority_queue]`/`[deadlock-1467]` Catch2 cases pass.
- **SF1000 official run, #1467 reproducer configuration** (unsorted `tpch_sf1000.duckdb` fresh copy, fixed predicates, host pin + Simpatico lineitem/orders, GB300): power run completed all phases, **q18 1.20 / 1.30 / 1.29 s across clean / post-RF1 / post-RF2 — no hang**; GPU-vs-DuckDB validation **PASS** after RF1 and after RF2; Power@Size = 4,141,672.
- SF1000 power + 7-stream throughput (sorted dataset, qgen-varied predicates): all phases completed, no hangs under concurrency; QphH@Size = 3,247,838.
- Note: neither SF1000 run shows downgrade-executor activity in its logs, so they establish no-regression under the reproducer's nominal conditions; the positive re-test of the deadlock interleaving is the regression-test pair above.


src/include/exec/multi_index_priority_queue.hpp
```diff
@@ -240,6 +242,23 @@ class multi_index_priority_queue {
...
+  /// try_pop_from() by device preference). Returns true when a task is present,
+  /// false if the queue was interrupted while empty (shutdown).
+  ///
+  /// The waiter shares _cv with pop()/pop_back(). push() must therefore use
+  /// notify_all(): with notify_one(), a push could wake only a wait_non_empty()
+  /// caller (which takes nothing) while a pop() caller stays blocked, or vice
+  /// versa a pop() could consume the task and leave this waiter correctly
+  /// asleep on an empty queue.
+  [[nodiscard]] bool wait_non_empty()
+  {
```
aminaramoon, line 249: this is a bad name, also make it const,

src/pipeline/task_scheduler.cpp
```diff
@@ -319,36 +319,69 @@ void task_scheduler::management_eventloop()
...
-    // Drain any further events that are already queued, so a single matcher
-    // pass handles a burst of ready signals plus task pushes together.
-    while (auto more = _task_request_channel.try_get()) {
-      if (more->kind == task_request_kind::device_ready && !more->is_scan) {
-        _ready_devices.emplace_back(more->device_id);
+    auto evt = _task_request_channel.try_get();
+    if (evt) {
+      // Drain the pending burst so one matcher pass sees all of it.
+      while (evt) {
+        if (evt->kind == task_request_kind::device_ready && !evt->is_scan) {
```
aminaramoon: oh this is bug!!!! did we just throw out a request from an executor ?

aminaramoon: you can safely remove the scan check, it is never used, maybe even remove it from the reqeust type altogether

src/pipeline/task_scheduler.cpp
```diff
@@ -319,36 +319,69 @@ void task_scheduler::management_eventloop()
...
+          _ready_devices.emplace_back(evt->device_id);
+        }
+        evt = _task_request_channel.try_get();
+      }
+    } else if (!_ready_devices.empty()) {
```
aminaramoon: I think this is all we need to here:

```cpp
auto ev == _channel.try_get();
if (!ev && _read_device.empty()) {
ev = _channel.get();
}

if (ev)  {  update_ready_devices(); } 

/// 

if (_task_queue.empty()) {
_task_queue.wait();
}
```

aminaramoon (changes requested):

aminaramoon (approved):

PR #1479, opened 2026-08-12
feat(op): streaming sink over an output repository, partitioned across N destinations (#837, #838)

aocsa:
Implements #837 and #838. Part 1 of a three-PR stack that replaces #1432.

**What.** The output boundary of a streaming plan. Where `STREAMING_SOURCE` (landed in #1320)
stands in for a scan that has no table, `STREAMING_SINK` stands in for a result that has no
client: it writes each finished batch into one or more `cucascade::shared_data_repository`
handles, so a downstream fragment can pick them up.

**Design: the sink is a writer, not a buffer.** A regular sink accumulates a result and hands it
back at the end. This one owns no storage — each batch goes straight into a repository the caller
supplied, and the operator's only state is which destination the next batch belongs to. That is
what makes a fragment boundary cheap: nothing is copied to hand a batch across it.

**Partitioning (#838).** A sink may fan a batch out across N destinations:

- *hash* — partition on a key column set. Key types are normalized first, so two senders that
  disagree on integer width still hash to the same destination. Without that, a plan whose
  fragments were built from different sub-plans silently splits one logical key across two
  partitions.
- *broadcast* — every destination receives every batch.

**Also folded in.** The `batch_stream` / `STREAMING_SOURCE` follow-up work that accumulated after
#1320 merged: corrections to the `on_data` hook documentation, the availability state machine, and
the sender-aware EOS description. These are doc and comment changes to files #1320 already landed;
they are here rather than in their own PR because they are too small to review alone.

---

**Reviewing.** One commit. `src/op/sirius_physical_streaming_sink.cpp` and its header are the
substance; the `src/pipeline/` changes are the wiring that lets a pipeline be rooted in a sink
rather than a result collector.

**Tests.** `test_physical_streaming_sink.cpp` covers the operator directly (write, multi-batch,
per-destination independence, the error plane, hash fan-out in `PART-1..7`, and broadcast in
`PART-8/9` — every destination receives every batch, destination 0 forwards the original handle,
and the rest are independent clones with distinct batch ids). `test_streaming_sink_root.cpp`
covers a pipeline whose plan-tree root is a sink, including the partitioned cases.

**Refused pushes poison the stream.** A push is refused only once the output is terminal, which
means the fragment produced data after its own output closed. Rather than dropping the batch, the
sink fails that output so the consumer learns from `try_pull()` instead of silently receiving a
short result. `SNK-5` pins this.

---

**Stack.** This PR is the base.

| Part | PR | Branch | Scope |
|---|---|---|---|
| 1 | **#1479 (this)** | `stream/13-sink` | `STREAMING_SINK` + partitioning (#837, #838) |
| 2 | #1480 | `stream/14-session` | `exec::stream_session` (#839, addressing) |
| 3 | #1481 | `stream/15-fragment` | `streaming_fragment` + FFI (#839, execution) |

Supersedes #1432, which carried all three layers as one 17-commit PR. The three parts together
reproduce that branch exactly — verified tree-for-tree, not by inspection.

### DIFF
--- a/docs/super-sirius/streaming-sessions.md
+++ b/docs/super-sirius/streaming-sessions.md
@@ -1,0 +1,298 @@
+# Stream Sessions
+
+A **stream session** is the boundary object for one Sirius plan fragment in a distributed query.
+It gives the wrapper above the engine a single, id-addressed handle to feed batches into the
+fragment and collect batches out of it:
+
+```
+push(stream_id, batch)              // feed an input stream
+close_input(stream_id, sender_id)   // one remote sender is done producing
+pull(stream_id) -> optional<batch>  // collect from an output stream (non-blocking)
+wait(stream_id)                     // block until data arrives or the stream ends
+drained(stream_id) -> bool          // stream ended cleanly and nothing is left
+```
+
+Sirius itself stays **fragment-blind**: it never learns that it is distributed, which compute
+node a partition ships to, or how many nodes exist. One session models one fragment; pairing a
+leaf fragment's output id with a root fragment's input id — across sessions and nodes — is the
+wrapper's routing table, never the engine's.
+
+A session is built from three pieces:
+
+| Piece | Role |
+|---|---|
+| `STREAMING_SOURCE` operator | The fragment's input boundary — remote senders push batches in |
+| `STREAMING_SINK` operator | The fragment's output boundary — external consumers pull batches out |
+| `exec::batch_stream` | The one-directional stream primitive both operators are built on |
+
+The session (`exec::stream_session`) is the id-addressed router over these operators.
+
+## The model: the repository is the queue
+
+A `cucascade::shared_data_repository` already *is* a thread-safe queue of `data_batch`es, and it
+is what the downgrade executor sweeps for spill candidates. What it lacks is the **lifecycle** of
+a stream: who is still producing, whether "nothing right now" means *wait* or *the stream is
+over*, and how a starved consumer gets woken. So each streaming operator owns both:
+
+| Concern | Owned by |
+|---|---|
+| The queue of batches | `cucascade::shared_data_repository` |
+| End-of-stream, availability, waking | `exec::batch_stream` |
+
+Batches cross the boundary **natively**, as `cucascade::data_batch`, in whatever tier they
+currently sit. Nothing is materialized to Arrow on the way in or out, so a queued batch stays
+spillable (GPU → host → disk) right up until it is pulled, and `pull()` hands it back in its
+current tier without forcing an upgrade.
+
+There is no bounded channel and no channel-level backpressure.
+
+## `exec::batch_stream`
+
+**Files:** `src/include/exec/batch_stream.hpp`, `src/exec/batch_stream.cpp`
+
+One direction of batch flow: N declared senders push into one repository; consumers pull, poll,
+or block.
+
+```cpp
+class batch_stream {
+ public:
+  enum class availability { HAS_DATA, WAITING, END_OF_STREAM };
+
+  batch_stream(shared_ptr<shared_data_repository> repo, set<sender_id_t> expected);
+
+  // Producer
+  [[nodiscard]] bool push(shared_ptr<data_batch>);  // false once terminal
+  void close(sender_id_t);       // idempotent per sender; set-based fan-in
+  void fail(exception_ptr);      // immediate, first failure wins
+
+  // Consumer
+  shared_ptr<data_batch> try_pull();  // rethrows a pending error before popping
+  availability classify() const;
+  bool drained() const;               // clean end only
+  void wait();                        // not atomic with try_pull — loop and re-check
+
+  // Hooks (single slot; fire after unlocking; registering on an ended stream fires immediately)
+  void set_on_data(function<void()>);           // persistent — fires on every push and on fail()
+  void set_on_end_of_stream(function<void()>);
+};
+```
+
+Three behaviors are load-bearing:
+
+- **End-of-stream is a set, not a counter.** A fan-in stream fed by N remote senders is over only
+  when all N *distinct* senders have closed. A repeated close is a no-op; an unexpected sender id
+  is a defined error. A counter could not tell "both senders closed once" from "one closed twice".
+- **Push, close, and every emptiness check share one lock.** No batch is ever admitted after
+  end-of-stream, and every batch is in the repository before the wake that announces it. Hooks
+  fire after the lock is released, so a hook may safely re-enter the scheduler.
+- **`classify()` separates "not yet" from "never".** Queued data outranks terminal: EOS is never
+  reported while an accepted batch is still pullable. A pending error reads as `HAS_DATA` even
+  over an empty queue — the only way out is the rethrow from `try_pull()`, never a clean finish
+  that would let a failed query succeed silently.
+
+### Contracts S1-S5
+
+Named observable contracts cited by call sites and tests. This section is the source of truth;
+headers keep only the tokens.
+
+- **S1 — admission ordering.** `push()` puts the batch in the repository before firing `on_data`,
+  and returns false once the stream is terminal. A consumer that saw EOS can never be raced by a
+  batch that was not yet visible when `on_data` fired.
+- **S2 — poison dominates.** `fail()` ends the stream and wakes a consumer parked on `WAITING`:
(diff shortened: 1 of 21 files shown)


src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,258 @@
...
+  if (_spec.mode == partition_mode::broadcast) {
+    // Output 0 gets the original handle (zero-copy); outputs 1..N-1 each get an independent
+    // deep copy in the batch's current memory space so destinations cannot race over one handle's
+    // residency. Clones first so output 0's push does not advance the stream's terminal state
+    // before the copies are made.
+    const auto& batches  = input.get_data_batches();
+    const auto read_only = input.get_read_only_batches();
+    for (std::size_t b = 0; b < batches.size(); ++b) {
+      for (std::size_t i = 1; i < _outputs.size(); ++i) {
+        auto copy = read_only[b].clone(sirius::get_next_batch_id(), stream);
```
dhruv9vats: so the cloned batch is also tracked by quent
```suggestion
        auto copy = read_only[b].clone(sirius::get_next_batch_id(), stream, batch_telemetry());
```

probably need to make this harder to miss.

src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,258 @@
...
+                                                     _spec.key_cast_types,
+                                                     num_partitions,
+                                                     stream,
+                                                     *space,
+                                                     batch_telemetry());
+
+    for (std::size_t i = 0; i < slices.size(); ++i) {
+      // Empty partition stays WAITING until finalize.
+      if (sirius::get_cudf_table_view(*slices[i]).num_rows() == 0) { continue; }
+      // Refused push = silent data loss.
```
dhruv9vats: Is there a reason we are not bubbling up the push errors here? Or, not poisoning the batch_stream?

test/cpp/pipeline/test_streaming_sink_root.cpp
```diff
@@ -0,0 +1,218 @@
...
+{
+  auto repos = make_repos(1);
+
+  sirius::test::with_initialized_streaming_fragment(
+    *con,
+    "SELECT n_regionkey FROM nation",
+    repos,
+    std::nullopt,
+    [&](sirius::sirius_engine& engine, sirius_physical_streaming_sink& sink) {
+      // The sink is the plan root and its subtree hangs off children[], so the plan generator
```
dhruv9vats: > `root`

I seem to keep getting confused between what a `root` vs `leaf` vs `children` mean here? `root` for someone uninitiated would mean a place/thing where something starts from/is initiated, but here the meaning seems to be flipped and refers to the end of the query plan DAG (the sink)? Even `children` is confusing because, usually, it would be thought of as something as is fed by the current entity, rather, here is refers to entities that feed the _current_ operator. I feel we should look into establishing some unambiguous nomenclature.

dhruv9vats (approved):
Thanks @aocsa.

Just a small suggestion for quent data batch tracking, otherwise LGTM.

test/cpp/pipeline/test_streaming_sink_root.cpp
aocsa: You're right that everyday English and this comment disagree.

This is the Volcano / DuckDB plan-tree convention. The tree is drawn from the result, not the scan:
```
plan-tree root = the sink (get_parent_op() == nullptr)
children[] = producers / inputs (who feed this operator)
Data still runs leaf → root. Execution starts at the pipeline head (the scan)
```

"The sink is the plan root" means "top of the operator tree," not "where work starts."

plan-tree root, producer children, pipeline head. 

Happy to reword this test comment to "plan-tree root."

src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,258 @@
...
+  }
+}
+
+std::size_t sirius_physical_streaming_sink::no_history_peak_memory_estimate(
+  const input_stats& stats) const
+{
+  // 0 if N==1; ~2× when partitioned (hash_partition holds reorder + slices).
+  if (_outputs.size() == 1) { return 0; }
+  return stats.bytes * 2;
+}
```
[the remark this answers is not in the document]
aocsa: Good catch — this is a real under-estimate. The 2× only ever described the hash path (`hash_partition` holding its reorder buffer alongside the slices); broadcast keeps the original plus `N-1` clones resident across `sink()`, so its peak scales with `N`.

Split the two modes, and `SNK-9` now pins all three cases (single = 0, hash = 2×, broadcast = N× for N ∈ {2,3,8}).

src/op/sirius_physical_streaming_sink.cpp
```diff
@@ -0,0 +1,258 @@
...
+        }
+      }
+      if (!_outputs[0]->push(batches[b])) {
+        SIRIUS_LOG_WARN(
+          "sirius_physical_streaming_sink: broadcast batch refused after end-of-stream and "
+          "dropped");
+      }
+    }
+    return;
+  }
```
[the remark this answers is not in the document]
aocsa: Agreed — this was a known gap called out in the PR description, and it shouldn't ship uncovered.

Added `PART-8` and `PART-9` covering exactly the properties you named: every destination receives every batch (the union is N copies of the input, not a partition of it), destination 0 forwards the original handle, and destinations `1..N-1` are independent clones with distinct `batch_id`s.

docs/super-sirius/streaming-sessions.md
```diff
@@ -0,0 +1,297 @@
...
+                  push()                       close(last sender) + repo empty
+WAITING ───────────────────────► HAS_DATA ──────────────────────────────────► END_OF_STREAM
+  ▲                                  │
+  └──────────────────────────────────┘
+        try_pull() drains repo; stream still open
+
+fail() (any state) ──► HAS_DATA  ──try_pull()──► rethrow
+                       ^^^^^^^^
+                       error is invisible to classify(); it is only revealed by try_pull()'s rethrow.
+                       The stream stays in HAS_DATA — never reaches END_OF_STREAM (S3).
```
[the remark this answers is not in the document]
aocsa: You're right, that sentence contradicts P4 and the table twenty lines above it — the diagram's own `fail() → HAS_DATA` arrow says so too.

What it meant is that `classify()` cannot *distinguish* poison from data (both read `HAS_DATA`), not that the error is hidden. Reworded so the diagram and P4 agree.

src/op/sirius_physical_streaming_sink.cpp
aocsa: Fixed, thanks. Matched the existing idiom from `sirius_physical_partition.cpp:232` and `sirius_physical_grouped_aggregate_merge.cpp:222` — hoist the id, then pass `telemetry::quent_data_batch_probe::create(batch_telemetry(), clone_batch_id)`, since `clone()` wants the probe rather than the telemetry handle itself.

On "harder to miss" — agreed, and it's worse than it looks: `clone()`'s third parameter defaults to a no-op `idata_batch_probe`, so omitting it compiles clean and silently produces an untracked clone. No warning, no `[[nodiscard]]` help. That's exactly how this one slipped through.

I left the helper out of this PR because the natural home (`data_batch_utils.hpp`, next to `get_next_batch_id`) and the two other call sites are all outside this stack's file set, and I didn't want to widen an approved PR. Worth noting the root cause is really upstream — cucascade defaulting the probe instead of requiring one — so fixing it there would cover every future caller too. Happy to file either as a follow-up, or pull the Sirius-side helper in here if you'd rather not wait.

src/op/sirius_physical_streaming_sink.cpp
aocsa: No good reason — poisoning is right, changed.

Worth recording what I found while checking, because my first answer was wrong: `push()` returns false only when the stream is already terminal, which happens via `close()` once every expected sender has closed, or via `fail()` (P3).  

It now calls `fail()` on the refusing output, so the consumer learns through `try_pull()`'s rethrow instead of quietly getting a short result. `fail()` is first-failure-wins (P2), so this never displaces a real cause.

One consequence worth flagging: `fail()` on an already cleanly-closed stream sets `_error`, so `drained()` flips to false and a late sink surfaces as an error rather than a no-op. `SNK-5` is rewritten to assert the poison instead of the silent drop.

Issue #1283, opened 2026-07-24
Create data size estimation propagation system


wmalpica:
Today, for joins we base our number of partitions decision making purely on the size of the build only (and a little bit on the planner's cardinality estimates). It would be nice to have some APIs which allow us to estimate at runtime how much data will arrive at an operator input port based on the historical information of what has already progressed in the query.

This could look like as follows:
Assume that a join operator wants to get an estimate of the number of bytes or rows that will arrive at its probe input port.
The API would look at the chain of operators which are upstream and look at each of their input to output ratio. And follow the chain of operators until it finds the first one which is complete (or until it reaches a scan (leaf node)). Given that first operator which is complete, you can get how much its total output and then use that and the chain of input to output ratios back to the calling operator to get an estimate of how much data it has received. 
We do currently have historical information that the operators use to estimate their memory usage. We might have to see if that is enough to determine input to output ratio.
Next, we would have to see if scan operators can give us an estimate of the total scan size. Not all scan or leaf node operators may be able to provide this. For example the streaming input node may not be able to provide this. 
Then we also have to consider that depending on when this API is called, we may not have enough information, so this API has to be able to return a nullopt response. For example of one operator has not received any input yet, then it has no input to output ratio to propagate (unless we allow a default in/out ratio of 1, which actually would be a convenient API flag). 

This issue then describes a feature which would then be used elsewhere, probably first for doing partition estimation.  The simplest place to go use this feature first is for grouped aggregations. Right now the partition operator used in a grouped aggregation is a hard barrier because we wait until all the data has arrived before we know how many partitions to create. With this feature, we would be able to make an estimate as soon as a single batch arrives at the partition operator and decide how many partitions to create. Which would allow us to change the barrier to a partial barrier and therefore make the performance better by improving parallelism and reducing latency in that operator chain. 
When we implement this, we can test the efficacy of the solution by looking at grouped aggregation performance, but also by comparing the estimated total size of the data arriving at the partition operator vs the actual total size (we would look at this post query completion by analyzing the logs or quent)

PR #1482, opened 2026-08-12
feat(pipeline): runtime data size estimation API

Jedi18:
## Description

Adds a runtime estimator that answers: **"how many bytes will ultimately arrive at this
operator's input port over the whole query?"** — the API issue #1283 asks for. Today,
decisions like a join's partition count are made from the build side and planner guesses
only; this API lets downstream operators size themselves from what the query has actually
done so far.

Two free functions over the pipeline graph (no engine state):

- `estimate_port_total_input_bytes(op, port_id)` — projected total for one input port
- `estimate_pipeline_total_output_bytes(pipeline)` — projected total a pipeline will emit

**How an estimate is formed** — walk upstream to the first known total, then chain each
pipeline's measured output/input ratio back down. Four terminating cases:

1. **Finished pipeline** — its recorded output total, reported `exact`
2. **Fan-in (join)** — follow only the source's nominated primary (probe) port, scaled by
   output-per-probe-byte; joins that stream the build instead (RIGHT-family, OUTER) nominate none and get no estimate
3. **Leaf** — anchor on the source's own total (scan split discovery, VALUES plan size)
4. **Single producer** — recurse, then apply this pipeline's learned ratio

**Design stance:** the chaining is measurement-derived — every ratio comes from completed
tasks, and any unknown link yields `nullopt` rather than a guess (opt-in
`assume_unit_ratio` relaxes this for single-input pipelines only). There is exactly one
planner-derived number: the scan's output-side anchor (`estimated_cardinality × measured
bytes/row`), used only while split discovery is open and identifiable via the result's
diagnostics. Corollary: the API cannot answer before a query starts running — there is no
pre-execution mode.

**Supporting changes:**

- `pipeline_memory_history` gains monotonic totals that survive ring-buffer eviction, split
  into *ratio terms* (tasks with a usable pipeline-input basis) and *output terms* (every
  successful task) — so zero-basis scan splits still count toward a finished pipeline's
  total, and tasks resumed mid-pipeline after an OOM/CUDA reschedule can't skew the ratio
  with intermediate-data units
- Four operator hooks on `sirius_physical_operator`, all defaulting to "cannot answer";
  implemented by `GPU_SCAN`, `GPU_VALUES` (leaf totals) and `HASH_JOIN` (probe-side
  nomination + probe-byte counting, weighted by how far through its pairings each batch is so the ratio holds mid-run rather than only once the join drains)
- `split_connector` tallies discovered bytes, making a scan's total knowable before it drains
- `try_get_port` (const + non-const): non-throwing port lookup; `get_port` delegates to it

**Consumers:** none in-tree yet. The first — sizing the group-by partition count from a
projected total, turning that hard barrier into a partial one — follows in a stacked PR.
Expected next: join partition/broadcast strategy and per-query GPU memory tapering.

Start reviewing at `docs/super-sirius/data-size-estimation.md`, which covers the mechanism,
the fan-in bias and its sample floor, and every limitation called out above.

## Checklist
- [x] Read CONTRIBUTING.md
- [x] Cover changes with new or existing tests — 36 unit cases drive the arithmetic
      against a synthetic pipeline DAG and the join's probe accounting (`[data_size_estimator]`, `[pipeline_memory_history]`, `[size_estimation]`)
- [x] Document configuration changes in code and summarize in the description above — none
      (this PR adds no settings; tuning lives in per-call `size_estimate_options`)
- [x] Update human and agent documentation (README.md, docs/, skills, CLAUDE.md) — new
      `docs/super-sirius/data-size-estimation.md`, indexed in the Super Sirius README

## References
Refs #1283 (the API half; the issue's partition-sizing consumer lands in the stacked
follow-up PR)

### DIFF
--- a/src/include/pipeline/data_size_estimator.hpp
+++ b/src/include/pipeline/data_size_estimator.hpp
@@ -104,9 +104,7 @@ struct size_estimate_options {
 /**
  * @brief `bytes * ratio`, or nullopt if the product would not survive narrowing to std::size_t.
  *
- * Shared with the source hooks that feed the estimator. Narrowing an out-of-range double is UB,
- * and on x86-64 yields a huge positive value rather than anything obviously wrong — an unguarded
- * cast turns "we cannot answer" into a confident exabyte-scale total.
+ * Shared with the source hooks that feed the estimator.
  */
 [[nodiscard]] std::optional<std::size_t> scale_bytes_checked(std::size_t bytes, double ratio);
 
--- a/src/include/pipeline/data_size_estimator.hpp
+++ b/src/include/pipeline/data_size_estimator.hpp
@@ -37,7 +37,9 @@ struct data_size_estimate {
   std::size_t bytes = 0;
   /// True when anchored on an exact total with no learned ratio applied.
   bool exact = false;
-  /// Pipelines traversed; ratio error compounds per hop.
+  /// Ratios applied between the anchor and this answer; error compounds per hop, and exact
+  /// implies zero. Distinct from the recursion depth @ref size_estimate_options::max_hops
+  /// bounds — a leaf applies its own ratio without recursing.
   std::size_t hops = 0;
   /// Completed tasks behind the weakest measured ratio; zero means no measured ratio applies.
   std::size_t ratio_samples = 0;
--- a/src/include/pipeline/data_size_estimator.hpp
+++ b/src/include/pipeline/data_size_estimator.hpp
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2026, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstddef>
+#include <optional>
+#include <string_view>
+
+namespace sirius {
+namespace op {
+class sirius_physical_operator;
+}  // namespace op
+
+namespace pipeline {
+
+class sirius_pipeline;
+
+/**
+ * @brief Projected total bytes that will flow through a point in the plan.
+ *
+ * A total for the whole query, not a so-far figure: "how many bytes will this port have
+ * received once its producer is done?". See docs/super-sirius/data-size-estimation.md.
+ */
+struct data_size_estimate {
+  std::size_t bytes = 0;
+  /// Measured rather than projected: the walk anchored on a finished pipeline or an exactly
+  /// known source total, with no learned ratio applied anywhere along the chain.
+  bool exact = false;
+  /// Pipelines traversed. Diagnostic — ratio error compounds per hop.
+  std::size_t hops = 0;
+  /// Completed tasks behind the *weakest measured* ratio in the chain. Separates "we sampled too
+  /// early" from "the model is wrong" when an estimate misses. Zero means no measured ratio backs
+  /// the result: an exact estimate, or — under @ref size_estimate_options::assume_unit_ratio — a
+  /// chain that substituted 1:1 at every link. @ref exact separates the two.
+  std::size_t ratio_samples = 0;
+  /// The walk anchored on a planner guess rather than a measurement — in practice
+  /// `GPU_SCAN::total_source_output_bytes()`, built from DuckDB's `estimated_cardinality`.
+  /// Sticky: every estimate chained on top of such an anchor inherits it. Not inferable from
+  /// `ratio_samples == 0`, which holds only for the anchor's own pipeline — one downstream hop
+  /// with a trusted ratio overwrites that zero (see weaker_sample_count in the .cpp).
+  bool planner_derived = false;
+};
+
+/// Tuning for a single estimation call.
+struct size_estimate_options {
+  /// Use a 1:1 ratio where the measured one is unusable, instead of returning nullopt. Never
+  /// marks the result exact. Applies to *pipeline* ratios only: 1:1 reads as "assume
+  /// pass-through", which a join — free to multiply or divide its input volume — cannot claim.
+  bool assume_unit_ratio = false;
+  /// Recursion guard against pathological plan depth, and defensively against graph cycles.
+  std::size_t max_hops = 16;
+  /// Sample floor for a single-input pipeline ratio. Below it the ratio is treated as absent, so
+  /// @ref assume_unit_ratio still applies.
+  std::size_t min_ratio_samples = 4;
+  /// Sample floor for a fan-in ratio. Far higher, because that ratio is systematically biased low
+  /// while tasks are in flight where a single-input one is merely noisy — and a **hard gate**:
+  /// nullopt below it even under @ref assume_unit_ratio. See data-size-estimation.md#fan-in.
+  std::size_t min_fan_in_ratio_samples = 16;
+};
+
+/**
+ * @brief Project the total bytes arriving at @p op's @p port_id input port.
+ *
+ * @return nullopt for a missing port, a dependency-only port (null repo), a port with no
+ *         producer, or when the upstream walk cannot produce an estimate.
+ */
+[[nodiscard]] std::optional<data_size_estimate> estimate_port_total_input_bytes(
+  op::sirius_physical_operator& op, std::string_view port_id, size_estimate_options options = {});
+
+/**
+ * @brief Project the total bytes @p pipeline will emit over the whole query.
+ *
+ * Walks upstream to the first known total, then chains each intervening pipeline's measured
+ * output/input ratio back down. Four terminating cases, in the order the implementation tries
+ * them (see data_size_estimator.cpp for why each is shaped as it is):
+ *
+ *  1. finished pipeline — its recorded output total, exactly;
+ *  2. fan-in — follow only the source's nominated primary port;
+ *  3. leaf — anchor on the source's own total;
+ *  4. single producer — recurse, then apply this pipeline's ratio.
+ *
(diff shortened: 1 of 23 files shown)


src/include/pipeline/data_size_estimator.hpp
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * @return nullopt whenever any link in the chain is unknown.
+ */
+[[nodiscard]] std::optional<data_size_estimate> estimate_pipeline_total_output_bytes(
+  sirius_pipeline& pipeline, size_estimate_options options = {});
+
+/**
+ * @brief `bytes * ratio`, or nullopt if the product would not survive narrowing to std::size_t.
+ *
+ * Shared with the source hooks that feed the estimator. Narrowing an out-of-range double is UB,
```
kevkrist: I don't think you need this elaboration comment here.

src/pipeline/data_size_estimator.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+/**
+ * @brief Count the data-carrying ports feeding @p pipeline and identify the first producer.
+ *
+ * Mirrors sirius_pipeline::get_ingress_ports_info, which returns the producer *operator* where
+ * we need the producer *pipeline*. `operators` spans source through sink after is_ready(), so
+ * walking it reaches build-side ports on the sink too. Dependency-only ports (null repo) and
+ * ports with no producer carry no bytes and are skipped.
+ *
+ * Deduplication by port pointer is load-bearing: `operators` normally already contains the
```
kevkrist: I don't understand this comment. What's wrong with a set here?

src/scan_manager/split_connector.cpp
```diff
@@ -34,6 +34,9 @@ void split_connector::push_split(std::unique_ptr<op::operator_data> split)
   {
     std::lock_guard<std::mutex> lock(_mutex);
     assert(!_closed && "push_split after close() is forbidden");
+    // Accumulate before the move: this is the one choke point every split passes through,
+    // so it is where the scan's total input basis is tallied for the data size estimator.
+    _discovered_bytes += split->get_estimated_size_in_bytes();
```
kevkrist: This can return 0 when a split has no a priori estimate, I believe, in which case `total_source_input_bytes()` returns a partial sum, and mixed known/unknown scans can be underestimated. or even a zero estimate when the unit-ration fallback is enabled. Conservative fix is to track whether any zero split is seen and return nullopt in that case -- I can't tell if that's too aggressive though.

kevkrist (commented):
Still in progress. Very cool! -- excited to see where this goes.

I would shrink/delete some of the source-file comments -- there were times when I just found it easier to read the code. Specifically comments on rationale behind a decision and some 'this is safe because...'-like comments.

src/pipeline/data_size_estimator.cpp
Jedi18: Nothing wrong with a set, just that for small N vector would be faster.

I'll fix and make the comments easy to read

src/pipeline/data_size_estimator.cpp
```diff
@@ -0,0 +1,244 @@
...
+                                                       data_size_estimate input,
+                                                       const size_estimate_options& options)
+{
+  bool const trust_measured = ratio_is_trusted(totals, options);
+  if (!trust_measured && !options.assume_unit_ratio) { return std::nullopt; }
+
+  auto const scaled = scale_bytes_checked(input.bytes, trust_measured ? *totals.ratio() : 1.0);
+  if (!scaled.has_value()) { return std::nullopt; }
+  return data_size_estimate{
+    .bytes = *scaled,
```
kevkrist: Should this be `std::max(*scaled, totals.output_bytes)`, in case the ration under-estimates what's already been observed? Same with the fan-in projection at line 162?

src/pipeline/data_size_estimator.cpp
```diff
@@ -0,0 +1,244 @@
...
+    auto* source = pipeline.get_source().get();
+    if (source == nullptr) { return std::nullopt; }
+
+    auto const port_name = source->primary_input_port();
+    if (!port_name.has_value()) { return std::nullopt; }
+
+    // Read output before consumed input. A concurrent task can then only bias the ratio low;
+    // reversing the order could omit its input while including its output.
+    auto const totals = pipeline.get_memory_history().totals();
+    if (totals.output_records < options.min_fan_in_ratio_samples) { return std::nullopt; }
```
kevkrist: You acknowledge that this can go wrong in the docs, I think (data-size-estimation.md:204), for STANDARD/MIXED joins, where things can get pathologically biased under build/partition skew. Is there a way we can protect ourselves from this? Perhaps we should disable the estimation path for STANDARD/MIXED until we have a robust solution to this?

src/pipeline/data_size_estimator.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  // Leaf: anchor on the source's total.
+  if (producers.count == 0) {
+    auto* source = pipeline.get_source().get();
+    if (source == nullptr) { return std::nullopt; }
+
+    if (auto input_bytes = source->total_source_input_bytes()) {
+      return apply_pipeline_ratio(
+        pipeline.get_memory_history().totals(),
+        data_size_estimate{.bytes = *input_bytes, .exact = true, .hops = hops},
```
kevkrist: Should this be hops + 1. Same question for leaf and multi-hops `apply_pipeline_ration()`.

src/scan_manager/split_connector.cpp
```diff
@@ -34,6 +34,10 @@ void split_connector::push_split(std::unique_ptr<op::operator_data> split)
   {
     std::lock_guard<std::mutex> lock(_mutex);
     assert(!_closed && "push_split after close() is forbidden");
+    // A zero estimate means unknown, not empty; latch it so the total remains conservative.
+    auto const split_bytes = split->get_estimated_size_in_bytes();
```
kevkrist: `get_estimated_size_in_bytes()` on a cached-batch split does `batch->to_read_only()` which blocks under this lock, and the lock will be held during downgrading. I think you can just hoist this estimate above the lock?

src/op/sirius_physical_hash_join.cpp
```diff
@@ -1185,6 +1269,12 @@ std::unique_ptr<operator_data> sirius_physical_hash_join::get_next_task_input_da
         std::to_string(step.partition) + " of operator " + std::to_string(this->get_operator_id()));
     }
 
+    // Retry the non-blocking size read at every pairing; batch-id dedup prevents double counting.
+    note_probe_bytes(probe_batch);
+
+    // Republish after the paired-count increment so the denominator leads task output.
+    publish_weighted_probe_bytes(finished.second);
```
kevkrist: The probe bytes denominator is recomputed here under the `op_state_mutex` lock even though usually only one term has changed since the last publish. For STANDARD joins this feels lock it could prolong the lock.

Recommended fix: we know the partition, the probe index, and the frozen build count, so we can apply the ceiling arithmetic cheaply, with one full pass the first time `build_finished` flips to set the baseline.

test/cpp/pipeline/test_data_size_estimator.cpp
```diff
@@ -0,0 +1,854 @@
+/*
```
kevkrist, line 1: Can you have more extensive testing of the join's byte accounting (I don't see any -- but could be missing something)? You may want to try to include pathological STANDARD joins as discussed above.

src/pipeline/data_size_estimator.cpp
```diff
@@ -0,0 +1,244 @@
...
+producer_scan scan_producer_pipelines(sirius_pipeline& pipeline)
+{
+  producer_scan result;
+  std::vector<const op::sirius_physical_operator::port*> seen;
+  seen.reserve(4);
+
+  auto collect_from = [&](op::sirius_physical_operator* candidate) {
+    if (candidate == nullptr) { return; }
+    result.output_capped |= candidate->caps_pipeline_output();
+    for (auto port_id : candidate->get_port_ids()) {
```
kevkrist: nit: this allocates and returns a vector of strings per call -- this seems to happen pretty frequently. Consider providing an iterator accessor into `_ports_list`.

src/op/scan/sirius_gpu_scan_operator.cpp
```diff
@@ -351,6 +373,17 @@ std::unique_ptr<op::operator_data> sirius_gpu_scan_operator::execute(
                                                mem_space->get_default_allocator());
     }
   }
+
+  // Read the size before moving the table into a batch to avoid locking the batch. Publish rows
+  // and bytes together because they form one sample.
+  auto const emitted_rows  = static_cast<std::size_t>(output_table->num_rows());
+  auto const emitted_bytes = output_table->alloc_size();
```
kevkrist, line 380: nit: It looks like if `make_data_batch` OOMs, these emission counts will be double counted. Is this as easy as moving this update below `make_data_batch`?

kevkrist (changes requested):
Looks good. 1 real issue regarding batch lock taken under mutex and some nits.

src/pipeline/data_size_estimator.cpp
Jedi18: Fixed and added tests

src/pipeline/data_size_estimator.cpp
Jedi18: Agreed, fixed so only BUILD_PROBE mode does estimation now, STANDARD/MIXED we should do properly in a follow-up

src/pipeline/data_size_estimator.cpp
Jedi18: Fixed, hops now counts ratios actually applied rather than recursion depth

src/scan_manager/split_connector.cpp
Jedi18: Fixed thanks

src/op/sirius_physical_hash_join.cpp
Jedi18: Since STANDARD/MIXED estimation is now disabled, the weighted denominator is no longer computed for those modes. This removes the full partition walk from the scheduler lock entirely.

test/cpp/pipeline/test_data_size_estimator.cpp
Jedi18: Added tests

kevkrist (approved):
[fistbump]

PR #1483, opened 2026-08-12
fix(aggregate): record COUNT DISTINCT local schema

### LINKED ISSUE #1475: TPC-H Q16 emits datatype-mismatch diagnostics with oracle-correct results
At Sirius `37ab8bff1b439641e3fd50bcacfceb8c5ef61716`, TPC-H SF100 Q16
completes with exit code 0 and matches the CPU oracle, but emits these
diagnostics from `src/pipeline/gpu_pipeline_task.cpp:92`:

```text
HASH_GROUP_BY ... col3 got cudf::list_view expected int64_t
PARTITION ... col3 got cudf::list_view expected int64_t
```

We reproduced both messages on NVIDIA H100 80 GB and L40S 48 GB GPUs:

| GPU | Configuration | Incidence | Counts per trial | Result |
| --- | --- | ---: | --- | --- |
| H100 | no Sirius settings | 5/5 | 1 + 1 | CPU-oracle equal |
| H100 | Sirius integration-reference values, full memory | 5/5 | 7--9 + 7--9 | CPU-oracle equal |
| L40S | no Sirius settings | 1/1 | 1 + 1 | CPU-oracle equal |
| L40S | Sirius integration-reference values, half-memory cap | 2/2 | 9 + 9 | CPU-oracle equal |

Every H100 trial also had zero fallback, spill, managed recovery, unmanaged
OOM, counter violations, comparator output, and other unclassified
diagnostics. The L40S zero-settings and reference trials likewise exited zero,
matched the CPU oracle, and had none of those recovery or error signals.

The payload identity is stable while multiplicity varies with configuration
and execution scheduling. The H100 reference trials observed 7, 7, 8, 9, and
7 occurrences of each payload; all five zero-settings trials observed one of
each. The L40S reproductions rule out a device-specific H100 artifact, and the
zero-settings trials rule out the earlier hypothesis that the warning was
caused only by the half-memory reference configuration.

Unlike the adjacent column-count warning, the pinned source does not label
this datatype mismatch a known non-bug. We therefore treat it as a platform
defect and make no timing claim from an affected trial. It is either a
spurious invariant check or a real internal type-invariant violation masked
before the correct final result; neither is a user-tuning decision.

Questions:

1. Is `cudf::list_view` an expected internal delimiter-join representation at
   these `HASH_GROUP_BY` and `PARTITION` boundaries?
2. If yes, should the diagnostic compare a normalized logical type or be
   scoped away from these internal operators?
3. If no, what invariant is violated even though the final Q16 result remains
   oracle-correct?

The complete runs and exact per-trial logs are retained privately. We can
provide a minimized reproducer or additional hashes/counts if useful.

infblueocean:
## Description

Fix the datatype diagnostics reported by TPC-H Q16 for `HASH_GROUP_BY` and its pass-through `PARTITION` stage.

`COUNT(DISTINCT)` lowers to a local `COLLECT_SET`, so those stages carry LIST values. `MERGE_GROUP_BY` later counts the set elements and returns the declared BIGINT result. This change constructs the merge while the final SQL schema is still present, then records LIST only on the supported local one-slot-per-aggregate COUNT(DISTINCT) shape and its partition.

AVG mixtures, grouping-function layouts, and physical-sidecar shapes retain their existing behavior instead of guessing an intermediate schema.

## Validation

- Sirius `dev` base: `f107ba88f1d473223169d6d2d9dd0c9e3bf75af0`
- Clean isolated NVIDIA L4 release build: 1,285/1,285 actions completed
- Focused planner selector: 29/29 assertions passed
- Stored source diff SHA-256: `dc464a7291444bd392274decb7333cd8932891ca4de1d656bea4d78bc15b59ee`
- Clean-base apply, reverse-apply, and `git diff --check` passed

## Checklist

- [x] Read `CONTRIBUTING.md`
- [x] Covered the repair with a focused planner regression
- [x] No Sirius configuration option changes
- [x] No user-facing documentation change required

## References

Closes #1475

### DIFF
--- a/src/include/op/sirius_physical_grouped_aggregate.hpp
+++ b/src/include/op/sirius_physical_grouped_aggregate.hpp
@@ -96,8 +96,7 @@ class sirius_physical_grouped_aggregate : public sirius_physical_operator {
 
   //! Runtime schema of the local COUNT(DISTINCT) accumulator. The local aggregate and PARTITION
   //! carry LIST sets; MERGE_GROUP_BY later converts those sets to the declared BIGINT count.
-  [[nodiscard]] duckdb::vector<sirius::logical_type>
-  get_count_distinct_local_output_types() const;
+  [[nodiscard]] duckdb::vector<sirius::logical_type> get_count_distinct_local_output_types() const;
 
   // Source interface
   bool is_source() const override { return true; }
--- a/src/include/op/sirius_physical_grouped_aggregate.hpp
+++ b/src/include/op/sirius_physical_grouped_aggregate.hpp
@@ -94,6 +94,11 @@ class sirius_physical_grouped_aggregate : public sirius_physical_operator {
     return indices;
   }
 
+  //! Runtime schema of the local COUNT(DISTINCT) accumulator. The local aggregate and PARTITION
+  //! carry LIST sets; MERGE_GROUP_BY later converts those sets to the declared BIGINT count.
+  [[nodiscard]] duckdb::vector<sirius::logical_type>
+  get_count_distinct_local_output_types() const;
+
   // Source interface
   bool is_source() const override { return true; }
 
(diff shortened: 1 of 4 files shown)


Jedi18 (approved):
LGTM, thanks!

PR #1485, opened 2026-08-12
fix(aggregate): record AVG local accumulator schema

infblueocean:
## Summary

- record the local two-column AVG accumulator schema below `MERGE_AGGREGATE`
- keep the local sum carrier aligned with execution, including integer widening and DECIMAL preservation
- add planner coverage for integer and DECIMAL AVG plus a DECIMAL64 execution case

## Why

The ungrouped aggregate child emits partial AVG state as `SUM + COUNT`, while the merge operator exposes the final one-column SQL schema. The planner previously copied the final schema onto the local child, so task-level validation compared a two-column runtime batch with a one-column declaration.

## Validation

Validated on Sirius `f107ba88f1d473223169d6d2d9dd0c9e3bf75af0` using an NVIDIA L4:

- affected Release build completed successfully
- planner selector: 33 assertions in 1 test case
- AVG execution wildcard: 26 assertions in 5 test cases
- the five executed template instances were exactly `int32_t`, `int64_t`, `float`, `double`, and `decimal64_tag`
- `clang-format` 20.1.4 was idempotent; focused codespell and orphan-test checks passed

## Validation history

The first implementation attempt exposed two compile defects before any result was accepted: a case-local variable crossed switch labels, and a `std::vector` was passed to a helper requiring `duckdb::vector`. This revision braces the AVG case and performs the type conversion explicitly. A later selector used an exact Catch2 name and matched zero template instances; that run was rejected. The final wildcard run enumerated and executed all five instances above, including DECIMAL64.

### DIFF
--- a/src/include/op/sirius_physical_ungrouped_aggregate.hpp
+++ b/src/include/op/sirius_physical_ungrouped_aggregate.hpp
@@ -44,6 +44,10 @@ class sirius_physical_ungrouped_aggregate : public sirius_physical_operator {
   //! The aggregates that have to be computed
   duckdb::vector<std::unique_ptr<sirius::ast::node>> aggregates;
 
+  //! Schema emitted by the per-task partial aggregate. AVG expands to SUM + COUNT here; the
+  //! MERGE_AGGREGATE above this operator collapses those carriers to the declared SQL schema.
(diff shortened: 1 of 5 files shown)


kevkrist (approved):
Thx!

PR #1487, opened 2026-08-12
perf(partition): hybrid zero-copy hash partitioning

kevkrist:
Add hybrid zero-copy hash-partition output for eligible fixed-width columns, with per-partition fallback copies for unsupported layouts. Rather than copy slices of the the partitioned buffer make views into the unified buffer to pass downstream. This is a memory efficiency optimization, not a throughput optimization that shows itself at SF 1K GB300.


src/op/partition/gpu_partition_impl.cpp
```diff
@@ -41,69 +116,166 @@ std::vector<std::shared_ptr<cucascade::data_batch>> gpu_partition_impl::hash_par
...
+          checked_add(logical_size,
+                      checked_multiply(checked_size(end - begin), cudf::size_of(combined.type())));
+      } else {
+        output_columns.push_back(copied_columns->view().column(plan.compact_index));
+      }
+    }
+    if (copied_columns) { logical_size = checked_add(logical_size, copied_columns->alloc_size()); }
+
+    partition_view_owner owner{shared_fixed_columns, copied_columns};
+    output_batches.push_back(make_data_batch_from_view(cudf::table_view{output_columns},
```
9prady9, line 275: `release_table()` starts an asynchronous copy, then immediately destroys this view owner. That can free the source buffers before the copy finishes. Please use cuCascade's reader event to keep the owner alive until the copy completes, and add a cross-stream test where it is the last owner.

9prady9 (changes requested):

src/op/partition/gpu_partition_impl.cpp
9prady9: Also I wonder if we can use event semantics of cuCascade to make this more unblocking 🤔

kevkrist: Claude:

Confirmed, and it predates this PR: `release_table()`'s view branch enqueues the deep copy on the release stream, then destroys the `owning_table_view` and its `std::any` owner in the same statement (`gpu_data_representation.cpp:107`) — if that was the last reference, the buffers free on the owner's stream with the copy still pending. The existing view-branch test can't catch it: `test_release_table_stream.cpp:335` keeps a co-owning `shared_ptr` and asserts `use_count() == 1` at :356, so the owner drop frees nothing.

The docs made this worse, and that's on me: `operators.md` claimed consumers materialize view-backed batches through `release_table()`. No consumer does, and it's unsafe as a general contract. Corrected there and in `data-management.md`, with the constraint now recorded in code at `make_data_batch_from_view` (aa526358).

A reader event can't be the primitive here, though. They're `data_batch` members, and `mutable_data_batch` — what both callers hold — exposes none; batch-shaping the owner doesn't help either, since `synchronize_reader_events_no_throw()` early-returns on an empty pool (`data_batch.cpp:175`) and `release_table()` records nothing; and cuCascade can't reach into the owner to fix that (`std::any`, no `any_cast` anywhere). The ordering has to be established inside `release_table()`: hold the view in a local, enqueue the copy, then either `stream.synchronize()` or record a `cuda_event` and defer the owner drop. The blocking form already exists one layer up in `install_converted_representation` (`data_batch.hpp:687-697`), so I'd rather fix it once there than work around it per-producer.

PR 182 needs a rebase first — #184 rewrote `release_table()` to add `validate_stream_device` and the `rebind_stream` branch, so 182 conflicts in that function, and its tests only cover `clone()`'s writer ordering. I'll add the last-owner cross-stream case there. No live UAF today (both call sites take owned representations by provenance), but happy to gate this PR on the bump if you'd rather the guard land first.

On unblocking: agreed in principle, but it's bigger than it looks. The blocking call is the unconditional `stream.synchronize()` after every operator `execute()` (`gpu_pipeline_task.cpp:214`), which reads as timing-only but is currently load-bearing — it makes partition's missing `record_reader_event` harmless, retires consumer reads before the last owner drops, and is why this hazard is latent rather than live. Doing it properly needs reader events at every reader, `cudaStreamWaitEvent` on `get_writer_event()` at consumers (only `sirius_physical_hash_join.cpp:2142` does today), and timing moved onto event pairs. Separate issue — want me to open it?

kevkrist (commented):
Thanks @9prady9 but I think the danger is not live in production and latent in cuCascade. Let me know what you think.

PR #1488, opened 2026-08-13
fix(config): reject inactive prefetch-cache settings

abelatnvidia:
## Summary

- Reject a non-null `scan_manager.cache` subkey when
  `enable_prefetch_cache` is false or omitted.
- Preserve absent-value behavior for a null cache block, an empty cache map,
  and null-valued subkeys.
- Reject before parsing the inert value and verify all four subkeys, so a
  refused load does not apply the ignored cache value.
- Keep the enabled path unchanged and verify readback of all four settings.

## Why

Sirius currently accepts the four cache settings even while both cache
construction sites are disabled by `enable_prefetch_cache`. The values look
valid but have no effect. Rejecting that inconsistent configuration makes the
dependency visible without selecting a new cache default or claiming the
cache should always be enabled.

## Validation

Current identity:

- base: `4872cd3c490baf59f2591c527f8fbb6833416e0f`
- head: `2b524020564329e76bbdf297f14190bbaa6b0923`
- tree: `d5c4f4bc2222d8b5100b82f30eefdc157e3a0b28`

The retained zero-context diff replays on the exact base to the exact head
tree. Exact-head lint, x64 GCC release, x64 Clang debug, CUDA 13 build, the
20-minute CUDA 13 runtime suite, distribution, title, and aggregate gates are
terminal green.

Earlier compiled L4 validation of the same activation rule passed on Sirius
`ee6b3d80` and `a4f31124`:

- clean release build: 1285/1285 actions;
- warm incremental build: 245/245 actions;
- `[scan_manager][config][prefetching_cache]`: 21/21 assertions on both;
- `[config_opt]`: 93/93 on both;
- `[object_store_config][s3][config]`: 26/26 on both;
- `[sirius][config]`: 272/272 and 277/277 respectively.

The current-head regression additionally covers every cache subkey under
implicit and explicit disablement and verifies default preservation after
refusal.

## Scope

This validates configuration dependencies only. It does not change the
prefetch-cache default or make a performance claim.

### DIFF
--- a/src/sirius_config.cpp
+++ b/src/sirius_config.cpp
@@ -249,10 +249,10 @@ static void from_yaml(const YAML::Node& node, scan_manager::scan_manager_config&
                                    cache_reader.has_value("dispose_after_use") ||
                                    cache_reader.has_value("min_prefetching_budget_fraction") ||
                                    cache_reader.has_value("eviction_threshold_fraction");
-    sirius::from_yaml(*n, opt.cache);
     if (!opt.enable_prefetch_cache && has_cache_setting) {
       throw std::runtime_error("scan_manager.cache: requires enable_prefetch_cache: true");
     }
+    sirius::from_yaml(*n, opt.cache);
   }
   if (auto n = r.optional_node("object_store")) sirius::from_yaml(*n, opt.object_store);
   if (auto n = r.optional_node("memory_prefetcher")) from_yaml(*n, opt.memory_prefetcher);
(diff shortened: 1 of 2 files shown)


infblueocean:
Superseded by #1527, which now carries this activation contract together with validation of the same cache object’s budget ordering. The exact composed tree is 1e42ef6e1689d237ba0874f34334ab5dafafbc77.

PR #1489, opened 2026-08-13
fix(config): validate dynamic filter thresholds

abelatnvidia:
## Summary

- validate `dynamic_filter_domain_coverage_threshold` as greater than zero in YAML
- validate `dynamic_filter_keep_threshold` as a fraction in `[0, 1]` in YAML
- reject NaN through both YAML and DuckDB `SET` without mutating the active value
- retain the documented valid boundaries: domain values above one and keep values at zero or one

## Why

The same threshold bounds were enforced through DuckDB `SET` but not through
`sirius.yaml`, so invalid file values could survive until dynamic-filter
planning or execution. The prior `SET` comparisons also admitted NaN because
every ordered comparison against NaN is false. This aligns both entry paths and
makes the existing contracts fail closed.

## Rebase coordination

This candidate is rebased onto the protected merge of #1490. The only conflict
was additive test placement in `test/cpp/config/test_context.cpp`. Its union
retains the merged negative-host-capacity and MARK-ratio tests plus this PR's
complete dynamic-filter threshold coverage. Production semantics from the
reviewed predecessor are preserved, with the additional NaN-closed `SET`
comparisons.

## Validation

- YAML rejects domain zero and NaN; accepts a domain value above one
- YAML rejects keep values below zero, above one, and NaN; accepts zero
- invalid YAML and DuckDB `SET` both preserve the prior threshold values
- DuckDB `SET` rejects NaN for both thresholds before assignment
- DuckDB `SET` accepts domain 1.5 and keep boundaries 0 and 1; RESET remains valid
- independent exact-current semantic and conflict-union review: CLEAR
- exact patch replay from current `dev` reproduces the candidate tree
- `git diff --check` and orphan-test validation: pass
- hosted workflow [31711145481](https://github.com/sirius-db/sirius/actions/runs/31711145481): runtime job `94484972199` passed in 20m09s; aggregate job `94513537302` passed

Exact base: `e64b8ad79ea3cee50149d7e3039dbbb92ed291a9`  
Candidate: `8ef3ed9c67b34800b7d08dbf15a2c995a87b218b`  
Tree: `f158f5217eb29e689b7a09514df7d05ae71385c1`  
Patch SHA-256: `5dc8cac9c320b24a7dd2150b09aa606e45150b1491a2ed23e443e4df42f97dfb`

The predecessor head `ef4dffca8b6bb0faa3e161f8bdb24c3f438f6011`
passed hosted lint, GCC release, Clang debug, CUDA 13 build, full CUDA runtime,
and terminal aggregate checks. Those receipts are historical and do not
transfer to this rebased candidate.

### DIFF
--- a/src/sirius_config.cpp
+++ b/src/sirius_config.cpp
@@ -273,8 +273,11 @@ static void from_yaml(const YAML::Node& node, operator_params& opt)
   r.optional("enable_dynamic_filter_pushdown", opt.enable_dynamic_filter_pushdown);
   r.optional("enable_dynamic_zone_map_filter", opt.enable_dynamic_zone_map_filter);
   r.optional("dynamic_filter_domain_coverage_threshold",
-             opt.dynamic_filter_domain_coverage_threshold);
-  r.optional("dynamic_filter_keep_threshold", opt.dynamic_filter_keep_threshold);
+             opt.dynamic_filter_domain_coverage_threshold,
+             yaml::greater_than<double>{0.0});
+  r.optional("dynamic_filter_keep_threshold",
+             opt.dynamic_filter_keep_threshold,
+             yaml::fraction<double>{});
   r.optional("enable_pinned_zone_map_pruning", opt.enable_pinned_zone_map_pruning);
   r.optional("enable_compressed_materialization", opt.enable_compressed_materialization);
   r.reject_unknown();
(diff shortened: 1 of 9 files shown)


infblueocean:
Superseded by #1536, which now carries this complete threshold-validation contract together with the same dynamic-filter object’s session-surface simplification. The exact composed tree is 5d49259a4978ff36f55cb91a657147145b5cc07d.

PR #1490, opened 2026-08-13
fix(config): reject negative MARK join switch ratios

infblueocean:
## Summary

- reject negative and NaN `mark_join_build_switch_ratio` values on both YAML and DuckDB `SET` surfaces
- preserve `0` as the documented way to disable the MARK join build-side switch
- preserve the prior configured value when validation fails
- retain existing positive-infinity and negative-zero semantics; finite-only validation is not introduced here

Previously, YAML accepted negatives that silently behaved like `0`. The first
repair aligned YAML with the existing `SET ratio < 0.0` check. Maintainer review
then caught that YAML rejected NaN while `SET` accepted it because ordered
comparison with NaN is false. The setter now uses `!(ratio >= 0.0)`, matching
the YAML range validator without broadening the contract.

## Validation

- exact-base NVIDIA L4 release build and focused suites passed for the negative/zero candidate: `[sirius][config]` 290 assertions / 24 test cases; `[sirius][context][config][isolated_context]` 102 assertions / 3 test cases
- test binary SHA-256: `6d6853f13ac476bdf58874b291fcf4087599f355ced67c631b67f8c09f29adb8`
- maintainer-requested YAML `.nan` and DuckDB `SET ... = 'NaN'` regressions assert refusal and destination preservation
- current exact-head lint, Clang debug, and CUDA 13 build pass; GCC release and CUDA runtime are running/queued

Exact base: `e967464a90bbad8ce49a4b1d1c24d5038f83d1b1`  
Current candidate: `943dde68ef8318a18f38f32d40eed44955236831`  
Current tree: `7e926e2cc547349abd253c9b6490505a43d06702`

### DIFF
--- a/src/sirius_config.cpp
+++ b/src/sirius_config.cpp
@@ -28,6 +28,7 @@
 
 #include <algorithm>
 #include <exception>
+#include <limits>
 #include <stdexcept>
 #include <string_view>
 #include <variant>
@@ -265,7 +266,9 @@ static void from_yaml(const YAML::Node& node, operator_params& opt)
   r.optional("sort_sample_bytes", yaml::bytes(opt.sort_sample_bytes));
   r.optional("max_build_hash_table_bytes", yaml::bytes(opt.max_build_hash_table_bytes));
   r.optional("max_broadcast_join_size", yaml::bytes(opt.max_broadcast_join_size));
-  r.optional("mark_join_build_switch_ratio", opt.mark_join_build_switch_ratio);
+  r.optional("mark_join_build_switch_ratio",
+             opt.mark_join_build_switch_ratio,
+             yaml::between<double>{0.0, std::numeric_limits<double>::infinity()});
   r.optional("enable_runtime_distinct_build_probe", opt.enable_runtime_distinct_build_probe);
   r.optional("enable_dynamic_filter_pushdown", opt.enable_dynamic_filter_pushdown);
   r.optional("enable_dynamic_zone_map_filter", opt.enable_dynamic_zone_map_filter);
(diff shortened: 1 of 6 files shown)


src/sirius_config.cpp
```diff
@@ -261,7 +262,9 @@ static void from_yaml(const YAML::Node& node, operator_params& opt)
   r.optional("sort_sample_bytes", yaml::bytes(opt.sort_sample_bytes));
   r.optional("max_build_hash_table_bytes", yaml::bytes(opt.max_build_hash_table_bytes));
   r.optional("max_broadcast_join_size", yaml::bytes(opt.max_broadcast_join_size));
-  r.optional("mark_join_build_switch_ratio", opt.mark_join_build_switch_ratio);
+  r.optional("mark_join_build_switch_ratio",
+             opt.mark_join_build_switch_ratio,
+             yaml::between<double>{0.0, std::numeric_limits<double>::infinity()});
```
Jedi18, line 271: This rejects `NaN` in YAML, but the DuckDB setter still accepts it because `ratio < 0.0` is false for NaN. Could the setter use `!(ratio >= 0.0)` instead of the current `if (ratio < 0.0) {` and add NaN regression tests so both configuration surfaces remain consistent?

Jedi18 (changes requested):

src/sirius_config.cpp
infblueocean: Addressed in 943dde68: the DuckDB setter now uses `!(ratio >= 0.0)`, matching the YAML range validator for NaN while preserving zero, +infinity, and negative-zero semantics. Added YAML `.nan` and DuckDB `SET ... = 'NaN'` regressions, including destination-preservation checks. Fresh exact-head CI is running.

Jedi18 (approved):

infblueocean:
The exact PR head `7f24b196470bb26c590babcae2d64f9688933797` remains approved and all PR-head checks are green. It was placed through the protected merge queue twice; the merge-group CUDA 12 run hit the unrelated existing `mgpu_stress - SCHED-RR counter offset rotation` flake, while the PR-scoped MARK-join runtime passed. I have not re-enqueued it a third time. Please merge/requeue when maintainers judge the unrelated queue flake acceptable; no source change is proposed for that flake.

PR #1495, opened 2026-08-13
fix(config): validate positive REST counts

infblueocean:
## Summary

- require positive `max_connections`, `max_read_split`, `max_retry_attempts`, and `max_auth_retry_attempts` in startup YAML
- bound `max_connections` to `1..1024`, matching its per-reactor slot and optional pinned-bounce allocation cost
- check the bounce-storage multiplication before allocation
- parse through a signed temporary so negative inputs cannot wrap through `size_t`
- preserve omitted/null defaults and every valid value in each field's supported range
- document the count contract; no DuckDB `SET` options are added

`max_auth_retry_attempts` is the total-attempt bound for HTTP 403 recovery. A
value of zero already behaves like one attempt at runtime, so rejecting zero
removes an accidental alias without removing the intentional no-retry value
(`1`).

The `max_connections` ceiling is 64x the production default and 32x the
largest committed integration-test value. It keeps substantial tuning
headroom while bounding the slot, curl-handle, event, and pinned-bounce
resources allocated per reactor.

## Review follow-up

- `max_connections=1024` is accepted and `1025` is rejected without mutation
- programmatic reactor construction rejects values above the same ceiling
- bounce-storage sizing rejects `size_t` multiplication overflow before the allocator is called
- zero-valid listing caps are handled separately in #1506
- the generic negative-to-unsigned wrap class, including reactor counts, is consolidated in #1529 (predecessor #1507)
- negative byte wrappers are already repaired by merged #1491

## Validation

- focused config regression covers all four fields: omitted/null default preservation, exact readback for `1` and `7`, and rejection of `0`, `-1`, and one-past-`LLONG_MAX` without mutation
- focused max-connection regressions cover both bound values and multiplication overflow
- independent source review of the auth-retry extension: clear
- local clang-format, `git diff --check`, and orphan-test detection: pass on exact head
- exact-head hosted workflow [31713645617](https://github.com/sirius-db/sirius/actions/runs/31713645617) is terminal green: lint, GCC release, Clang debug, CUDA 13 release build, full CUDA runtime suite plus TPC-H snapshot (job `94494884488`, 20m00s), and aggregate job `94501723069` all pass
- predecessor `97969b53` also passed exact-head lint, GCC release, Clang debug, CUDA 13 release build, full CUDA runtime suite (20m09s), and terminal aggregate check before this review repair

Exact base: `e64b8ad79ea3cee50149d7e3039dbbb92ed291a9`  
Current candidate: `705854f3eda534940fa9af049efa7349f659c415`  
Current tree: `ace9d5d7d963b9e65a5cb3b0a201df630dcd8766`

The earlier three-field candidate `caaa1b8e71b604fde8d8313ac92be79e0f0e6c2d`
was independently built and tested on an exact-base AWS L4 release build. Its
raw receipts are retained in Foxglove commit `f6a2b8e`; those receipts do not
claim execution of the later extensions or this review repair.

### DIFF
--- a/src/sirius_config.cpp
+++ b/src/sirius_config.cpp
@@ -170,6 +170,17 @@ static void from_yaml(const YAML::Node& node, sirius::io::object_store_config& o
   r.reject_unknown();
 }
 
+static void read_positive_count(yaml::reader& reader, const char* key, std::size_t& out)
+{
+  // Parse through a signed temporary so a negative YAML scalar cannot wrap to
+  // a large size_t before validation.  Missing and null values retain the
+  // struct's explicit default.
+  auto const has_value = reader.has_value(key);
+  long long value      = 1;
+  reader.optional(key, value, yaml::greater_than<long long>{0});
+  if (has_value) { out = static_cast<std::size_t>(value); }
+}
+
 static void from_yaml(const YAML::Node& node, sirius::io::rest::config& opt)
 {
   yaml::reader r(node, "rest");
@@ -181,15 +192,15 @@ static void from_yaml(const YAML::Node& node, sirius::io::rest::config& opt)
     }
   }
   r.optional("request_timeout_s", opt.request_timeout_s);
-  r.optional("max_connections", opt.max_connections);
+  read_positive_count(r, "max_connections", opt.max_connections);
   r.optional("chunk_size", yaml::bytes(opt.chunk_size));
   r.optional("max_n_chunks", opt.max_n_chunks);
-  r.optional("max_read_split", opt.max_read_split);
+  read_positive_count(r, "max_read_split", opt.max_read_split);
   r.optional("upkeep_interval_ms", opt.upkeep_interval);
   r.optional("conn_max_age_s", opt.conn_max_age);
   r.optional("retry_backoff_base_ms", opt.retry_backoff_base);
   r.optional("retry_jitter_ms", opt.retry_jitter);
-  r.optional("max_retry_attempts", opt.max_retry_attempts);
+  read_positive_count(r, "max_retry_attempts", opt.max_retry_attempts);
   r.optional("max_auth_retry_attempts", opt.max_auth_retry_attempts);
   r.optional("honor_retry_after", opt.honor_retry_after);
   r.optional("perf_instrumentation", opt.perf_instrumentation);
--- a/src/sirius_config.cpp
+++ b/src/sirius_config.cpp
@@ -201,7 +201,7 @@ static void from_yaml(const YAML::Node& node, sirius::io::rest::config& opt)
   r.optional("retry_backoff_base_ms", opt.retry_backoff_base);
   r.optional("retry_jitter_ms", opt.retry_jitter);
   read_positive_count(r, "max_retry_attempts", opt.max_retry_attempts);
-  r.optional("max_auth_retry_attempts", opt.max_auth_retry_attempts);
+  read_positive_count(r, "max_auth_retry_attempts", opt.max_auth_retry_attempts);
   r.optional("honor_retry_after", opt.honor_retry_after);
   r.optional("perf_instrumentation", opt.perf_instrumentation);
   r.optional("footer_probe_bytes", yaml::bytes(opt.footer_probe_bytes));
--- a/src/sirius_config.cpp
+++ b/src/sirius_config.cpp
@@ -170,14 +170,17 @@ static void from_yaml(const YAML::Node& node, sirius::io::object_store_config& o
   r.reject_unknown();
 }
 
-static void read_positive_count(yaml::reader& reader, const char* key, std::size_t& out)
+static void read_positive_count(yaml::reader& reader,
+                                const char* key,
+                                std::size_t& out,
+                                long long upper = std::numeric_limits<long long>::max())
 {
   // Parse through a signed temporary so a negative YAML scalar cannot wrap to
   // a large size_t before validation.  Missing and null values retain the
   // struct's explicit default.
   auto const has_value = reader.has_value(key);
   long long value      = 1;
-  reader.optional(key, value, yaml::greater_than<long long>{0});
+  reader.optional(key, value, yaml::between<long long>{1, upper});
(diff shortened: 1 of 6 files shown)


ran-yuan-rui:
Heads-up on a moving piece this PR sits next to: #1340 is in flight and flips the S3/REST backend to the cuCascade copy. It deletes src/io/rest/* (including config.hpp and rest_reactor.cpp) and re-points from_yaml in sirius_config.cpp at cucascade::io::rest::config — the exact function this PR modifies.

flagging the double touch so it doesn't surprise anyone at merge time. I suggest we wait for 1340 completed first then change IO config files

src/sirius_config.cpp
```diff
@@ -169,22 +169,33 @@ static void from_yaml(const YAML::Node& node, sirius::io::object_store_config& o
...
+}
+
 static void from_yaml(const YAML::Node& node, sirius::io::rest::config& opt)
 {
   yaml::reader r(node, "rest");
   r.optional("request_timeout_s", opt.request_timeout_s);
   r.optional("ca_bundle_path", opt.ca_bundle_path);
   r.optional("tls_verify", opt.tls_verify);
-  r.optional("max_connections", opt.max_connections);
+  read_positive_count(r, "max_connections", opt.max_connections);
```
ran-yuan-rui: max_connections` needs an upper bound.

The runtime multiplies it by `bounce_slot_size` without an overflow check and allocates one slot per connection; `slot_pool` also represents slot indices as `int`. A large accepted value can overflow the allocation size, terminate the worker, and leave later reads waiting indefinitely. Define a bounded range, check the multiplication, and test both boundaries.

src/sirius_config.cpp
```diff
@@ -169,22 +169,33 @@ static void from_yaml(const YAML::Node& node, sirius::io::object_store_config& o
...
-  r.optional("max_read_split", opt.max_read_split);
+  read_positive_count(r, "max_read_split", opt.max_read_split);
   r.optional("bounce_block_size", yaml::bytes(opt.bounce_block_size));
   r.optional("upkeep_interval_ms", opt.upkeep_interval);
   r.optional("conn_max_age_s", opt.conn_max_age);
   r.optional("retry_backoff_base_ms", opt.retry_backoff_base);
   r.optional("retry_jitter_ms", opt.retry_jitter);
-  r.optional("max_retry_attempts", opt.max_retry_attempts);
+  read_positive_count(r, "max_retry_attempts", opt.max_retry_attempts);
   r.optional("max_auth_retry_attempts", opt.max_auth_retry_attempts);
```
ran-yuan-rui: Negative values still wrap for zero-valid REST counts.

`max_auth_retry_attempts`, `footer_probe_bytes`, `list_max_matches`, and `list_max_scanned` are still parsed directly into unsigned storage. `-1` becomes `SIZE_MAX`, effectively disabling retry/listing limits or requesting an unbounded footer range. Parse through a signed temporary, require `>= 0`, and test negative and overflow inputs.

src/sirius_config.cpp
ran-yuan-rui, line 1: Reactor count validation happens after unsigned conversion.

For `rest_n_reactors` and `uring_n_reactors`, `-1` becomes `SIZE_MAX` and then passes the `> 0` validator, causing an oversized allocation at runtime. Parse and validate the signed value before converting to `size_t`, with boundary tests for both fields.

infblueocean:
Rebase follow-up is complete at exact head `97969b53`: lint, GCC release, Clang debug, CUDA 13 build, full CUDA runtime suite (20m09s), and terminal aggregate check all pass. The retained patch replays to tree `910892071b428c1f6805d40db32f8229bd349ecf`. Ready for fresh review.

src/sirius_config.cpp
infblueocean: Addressed at exact head `705854f`: `max_connections` is now bounded to 1..1024 in YAML and again at reactor construction; the bounce-storage multiplication is checked before allocation; focused regressions cover 1024/1025, constructor rejection, and multiplication overflow. The ceiling is 64x the production default and 32x the largest committed integration-test value.

src/sirius_config.cpp
infblueocean: Split by contract rather than broadening this positive-count PR: zero-valid listing caps are covered by #1506; the generic unsigned YAML conversion fix in #1507 covers zero-valid unsigned fields and the remaining negative-wrap class; merged #1491 covers byte wrappers such as `footer_probe_bytes`. `max_auth_retry_attempts` remains intentionally positive here because 1 is the no-retry total-attempt value and 0 is only an accidental runtime alias.

src/sirius_config.cpp
infblueocean: Handled in the generic parser repair #1507, which rejects negative YAML scalars before assignment to any unsigned integral destination while retaining each field validator’s existing zero policy. That includes `rest_n_reactors` and `uring_n_reactors`; its exact-current hosted suite is green.

infblueocean:
Agreed. #1340 deletes/replaces the REST and object-store configuration types and tests this family targets, so reviewing these heads now would create disposable work. I have marked #1495, #1500, #1505, #1506, #1509, and #1515 draft and withdrawn the reviewer request. After #1340 lands, I will re-audit the surviving cuCascade surface, consolidate the still-relevant contracts into the smallest current-base review unit(s), and retain the existing green runs only as predecessor evidence.

infblueocean:
Bookkeeping update only while this remains correctly parked behind #1340: the generic unsigned YAML repair formerly referenced as #1507 is now preserved in consolidated parser PR #1529 at head 058a26917a82e3c57db2706fa04fbc09c2da7a4f; #1507 is closed as superseded with its terminal hosted receipt retained. This does not change the accepted wait-for-#1340 routing or make any REST claim current.

infblueocean:
Closing this draft back into the local validated patch queue per the maintainer guidance on #1340: that change replaces the S3/REST config and reactor types this branch validates. The exact head and evidence remain retained; after #1340 lands we will re-audit the new types and reopen only the surviving behavior, rather than leave a knowingly stale review item open.

PR #1497, opened 2026-08-13
fix(config): use one REST TLS source

infblueocean:
## Summary

- remove the two REST-local TLS YAML inputs that were silently overwritten before the REST reactor was constructed
- make `object_store.ca_bundle_path` and `object_store.tls_verify` the single documented TLS source
- reject the removed REST keys with their fully qualified replacements while preserving unrelated REST YAML fields
- repair `yaml::reader::has()` so missing keys no longer look present merely because the yaml-cpp sentinel is defined/null

## Validation

- independent exact-base AWS L4 release build: pass
- `[config_opt][has]`: pass (3 assertions / 1 test case)
- `[scan_manager][config][s3][rest][tls]`: pass (4 assertions / 1 test case)
- `[object_store_config]`: pass (55 assertions / 9 test cases)
- `sirius_config still loads unrelated REST YAML fields`: pass (4 assertions / 1 test case)
- independent test binary SHA-256: `4d8ac7aa0a03efd308a1db2d377bb5b86b928673b41ba708dc5a7dca5c548bad`
- hosted CUDA 13 build: pass (3m13s)
- hosted CUDA 13 full `test-run`: pass (20m12s), followed by the terminal
  `test` job: pass (5s)
- hosted GCC release build: pass (5m53s)
- hosted Clang debug build: pass (3m42s)
- hosted lint/check: pass (59s / 5s)

Exact base: `0c5af9aee02f36d003141b432d799216f164a6c6`  
Independently validated candidate: `5c2415a0122c05e05a56070484bc204b246db3e2`  
Independently validated tree: `61c16c2bcdbe0849a8e49d582f5d6203d0c168f4`

The first candidate (`0096a199`) built but failed its focused selector: one expectation omitted the public filename wrapper, and the other exposed the `reader::has()` defect. It is retained as a failed candidate and carries no PASS claim. The validated successor fixes both defects in the stricter direction.

Independent raw receipts: Foxglove commit `54b8819`, under `patches/current-dev/evidence/rest-tls-source-of-truth-0c5af9ae/aws-results/`.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -314,11 +314,13 @@ single-GPU configurations only (logs a warning and disables itself otherwise).
 
 ### `scan_manager.rest` — REST / S3 backend (`io/rest/config.hpp`)
 
+TLS verification policy and the CA bundle are configured only under
+`scan_manager.object_store`. The REST reactor consumes those values so signing
+and transport use one trust policy; there are no separate REST YAML controls.
+
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | `request_timeout_s` | int (seconds) | 30 | Whole-request timeout and presigned-URL TTL (0 = no limit). |
-| `ca_bundle_path` | string | "" | PEM CA bundle for TLS verification. |
-| `tls_verify` | bool | true | Verify the endpoint's TLS certificate (peer + host). |
 | `max_connections` | int | 16 | Max concurrent in-flight connections per reactor. |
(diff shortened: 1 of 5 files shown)


ran-yuan-rui:
Using `has()` rather than `has_value()` for the rejection is the right call — `tls_verify: null` still names the key and should get the migration message.

One logistics note: this PR and #1495 overlap textually in all three shared files (the same `from_yaml` block, the same docs table, the same test-file tail), so whichever merges second needs a small rebase — worth picking an order.

ran-yuan-rui (approved):

PR #1505, opened 2026-08-13
fix(config): reject zero REST chunk size

infblueocean:
## Summary

- reject zero `scan_manager.rest.chunk_size` values during YAML load
- preserve omitted/null defaults and every positive raw or suffixed byte value
- reuse #1491's generic negative-byte rejection and validate through a temporary, so rejected input cannot mutate live config
- document the startup contract while retaining the runtime defensive clamp for programmatic configs

## Why

`chunk_size` is a target byte count for ranged GETs. It has no documented or implemented zero-as-disable policy. Clean Sirius instead coerces zero to one in `chunk_host_segments`, which can split an oversized host segment into one ranged GET per byte. Rejecting zero at the user surface removes that accidental alias without inventing a tuning minimum.

## Validation

- omitted and `null` preserve the 8 MiB default
- `1` and `1MiB` parse and read back exactly
- `0`, `0MiB`, `-1`, and `-1MiB` reject with `rest.chunk_size` diagnostics
- rejection leaves the existing value unchanged
- rebased onto merged #1491 and collapsed to one commit; negative parsing now delegates to the generic `yaml::bytes` contract while this PR adds only the field-specific nonzero rule
- exact `format-patch` replay onto the base reproduces the candidate tree
- `git diff --check` passes
- exact-head hosted lint, GCC release, Clang debug, CUDA 13 build, full runtime, and terminal aggregate checks pass
- hosted workflow [31692860146](https://github.com/sirius-db/sirius/actions/runs/31692860146): runtime job `94424648296` passed in 20m02s; aggregate job `94447223404` passed

Exact base: `4872cd3c490baf59f2591c527f8fbb6833416e0f`  
Candidate: `639e5995c29091ca43228b84711deaf9a9d93bc9`  
Tree: `5e192247aa9d70af5a9494232f9a2606e6aaf5cb`  
Patch SHA-256: `bbe09958725dfe797ecde982f57d4937950d15806c7064584567b797723433f4`

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -323,7 +323,7 @@ and transport use one trust policy; there are no separate REST YAML controls.
 |-----|------|---------|-------------|
 | `request_timeout_s` | int (seconds) | 30 | Whole-request timeout and presigned-URL TTL (0 = no limit). |
 | `max_connections` | int | 16 | Max concurrent in-flight connections per reactor. |
-| `chunk_size` | bytes | 8Mi | Target bytes per ranged GET (scatter/device-staging paths). |
+| `chunk_size` | bytes (**> 0**) | 8Mi | Target bytes per ranged GET (scatter/device-staging paths). Zero is rejected because it degenerates split reads into one-byte ranged GETs. |
 | `max_n_chunks` | int | 16 | Max file-adjacent segments fused into one scatter GET. |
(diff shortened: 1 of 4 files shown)


infblueocean:
Closing this draft back into the local validated patch queue while #1340 replaces the S3/REST config types and parser boundary it targets. The exact head and evidence remain retained; after #1340 lands we will re-audit the new types and reopen only the surviving behavior, rather than leave a knowingly stale review item open.

PR #1509, opened 2026-08-13
fix(config): reject partial object-store activation

infblueocean:
## What changed

- Keep an empty `scan_manager.object_store` block disabled.
- Once any member of the built-in REST backend quartet is supplied, require all four: `endpoint`, `region`, `access_key`, and `secret_key`.
- Reject incomplete activation during YAML load instead of accepting it and later disabling REST.
- List the exact missing quartet members and leave the previous config value unchanged after refusal.
- Document the all-or-none contract.

## Validation

- Focused config tests cover the empty block, both complete long-lived and temporary credentials, each required field alone, and each required field omitted from an otherwise complete activation.
- An explicit dormant-policy test proves `session_token`, signing, transport, CA, and TLS options alone remain accepted without activating the backend.
- `git diff --check` passes and changed C++ files are clang-formatted.
- Exact patch replay onto the rebased base reproduces the candidate tree.
- Exact-head hosted lint, GCC release, Clang debug, CUDA 13 build, full runtime, and terminal aggregate checks pass.
- Hosted workflow [31692800790](https://github.com/sirius-db/sirius/actions/runs/31692800790): runtime job `94424479329` passed in 20m00s; aggregate job `94430838875` passed.

This is intentionally YAML-scoped: ancillary session/transport/TLS/signing options alone do not activate the object store, and `session_token` remains optional once the required quartet is complete.

Exact identity: base `4872cd3c490baf59f2591c527f8fbb6833416e0f`, candidate `7c055b398dd8e31e6abe6118d771780c8b8cf173`, tree `2940753a5a1d18f961cbc9d3c9d01af888ab7501`.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -349,6 +349,12 @@ and transport use one trust policy; there are no separate REST YAML controls.
 
 ### `scan_manager.object_store` — S3 credentials & endpoint (`io/object_store_config.hpp`)
 
+The endpoint, region, access key, and secret key form one activation unit. An
+empty quartet leaves the REST object-store backend disabled. If any member is
+configured, all four are required; partial configurations are rejected while
+loading the file instead of disabling REST later. The session token and TLS,
+transport, or signing choices do not activate the backend by themselves.
+
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | `endpoint` | string | "" | S3 endpoint URL. |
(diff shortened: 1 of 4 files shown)


infblueocean:
Closing this draft back into the local validated patch queue while #1340 replaces the S3/REST configuration and object-store activation boundary it targets. The exact head and evidence remain retained; after #1340 lands we will re-audit the new types and reopen only the surviving behavior, rather than leave a knowingly stale review item open.

PR #1510, opened 2026-08-13
fix(config): reject incomplete object-store credentials

infblueocean:
## Summary

- treat endpoint, region, access_key, and secret_key as one object-store activation unit
- reject partial activation during configuration loading instead of silently disabling REST later
- keep the all-empty quartet disabled, including when dormant session/TLS policy fields are present
- document the activation contract

## Why

The runtime object-store authorizer already returns null when any member of this quartet is empty. That turns a typo or omitted credential into a warning and a disabled REST backend. The built-in factory does not recover the missing value from environment variables, AWS profiles, or IMDS, so the configuration cannot become usable later. Rejecting the partial file at load time gives the operator a specific missing-field error.

## Validation

- xcrun clang-format 21 on changed C++ files
- git diff --check
- python3 scripts/check_orphan_tests.py
- focused configuration fixtures cover every single missing member plus the disabled all-empty case
- full build and runtime tests delegated to upstream CI because this macOS host has no Sirius CUDA toolchain

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -360,6 +360,12 @@ and transport use one trust policy; there are no separate REST YAML controls.
 | `ca_bundle_path` | string | "" | Sole YAML source for the REST endpoint's PEM CA bundle. |
 | `tls_verify` | bool | true | Sole YAML source for REST endpoint certificate verification. |
 
+The endpoint, region, access key, and secret key form one activation unit. An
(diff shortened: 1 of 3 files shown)


infblueocean:
Closing as a duplicate of #1509 to keep one upstream review unit. This pass identified one activation-boundary refinement that will be folded into #1509: only endpoint, region, access_key, or secret_key activate quartet completeness validation; session_token, signing, transport, and TLS policy fields alone remain dormant and must not activate the backend.

PR #1512, opened 2026-08-13
fix(config): reject zero fused-chunk limits

infblueocean:
## Summary

- reject `scan_manager.rest.max_n_chunks: 0` at YAML load instead of silently clamping it to one buffer per ranged GET
- reject `scan_manager.local.max_n_chunks: 0` instead of silently disabling contiguous-segment fusion
- preserve omitted/null defaults, every positive value, and destination state after refusal

## Why

Both fields describe a maximum number of chunks. Zero currently has no literal implementation: the REST path applies `std::max(max_n_chunks, 1)`, while the local path simply cannot extend a group and behaves like one. Accepting zero therefore gives users a configuration Sirius does not actually honor. Rejecting it at the source surface makes the behavior explicit without selecting a new default or tuning policy.

## Validation

- exact base: `0d7f98d9a662e9e3b47d6f7d6e2c8797d1e6572b`
- candidate: `094d41d71b1695e8154e096d0d6acdfa0c4b495b`
- candidate tree: `182b009df7a9bef2473b02f7897843a380f2f6b7`
- retained patch SHA-256: `3e286986b780f3b05f33f92b6703afb90d0ec5f0de160048ba96c74a9c6b7c92`
- regression covers both surfaces; omitted/null preserve defaults, positive values round-trip, and zero, negative, and signed-overflow values are refused without mutation
- rebased over live `dev`; first expanded-test lint exposed only `clang-format`, repaired exactly in the candidate's second commit
- exact-current AWS L4 focused selector: pass, 60 assertions / 1 test case; CI artifact SHA-256 `1b06f075a134d15a936de87b72df007ebd19b03bbfb1b26b88e7fafa4e80f0f9`; test binary SHA-256 `43d211718ec41a04aa52b7cb5989a127acdfa6234d90dd0954ccaaf595c1ed6f`
- exact-head hosted lint, GCC release, Clang debug, CUDA 13 build, full runtime, and terminal aggregate checks pass
- hosted workflow [31694165577](https://github.com/sirius-db/sirius/actions/runs/31694165577): runtime job `94428753529` passed in 20m02s; aggregate job `94451076969` passed

The signed temporary parse closes unsigned-wraparound ambiguity before assigning the existing `size_t` destinations. The patch changes no default and makes no performance recommendation.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -311,7 +311,7 @@ single-GPU configurations only (logs a warning and disables itself otherwise).
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | `use_odirect` | bool | true | Use `O_DIRECT` for local-disk reads. |
-| `max_n_chunks` | int | 1 | Max contiguous file segments fused into one vectored read. |
+| `max_n_chunks` | int (**> 0**) | 1 | Max contiguous file segments fused into one vectored read; zero is rejected because every request contains at least one segment. |
 
 ### `scan_manager.rest` — REST / S3 backend (`io/rest/config.hpp`)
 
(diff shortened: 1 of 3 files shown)


infblueocean:
Superseded by #1514, which retains this exact REST/local chunk-limit validation alongside the adjacent IO zero-value contracts. This predecessor remains an exact passing receipt at `094d41d71b1695e8154e096d0d6acdfa0c4b495b`: workflow `31694165577`, CUDA 13 runtime job `94428753529`, terminal aggregate job `94451076969`. The consolidated exact head is `f33260e3fbc615d6b1500a03e172baec2def3bed`.

PR #1513, opened 2026-08-13
docs(config): define footer probe opt-out

infblueocean:
## Summary

- document `scan_manager.rest.footer_probe_bytes: 0` as the supported suffix-probe opt-out
- lock the existing no-suffix-GET / HEAD fallback behavior end to end
- verify negative bare and suffixed inputs fail without mutating the selected zero policy

## Why

Zero already has a coherent implementation: `fetch_footer_suffix` returns before constructing a `Range: bytes=-N` request; the caller performs HEAD and creates a stashless object; later reads use ordinary GETs. This is useful deployment policy for many tiny files or constrained links, but it was undocumented. The smallest configuration improvement is to name and test the existing escape hatch, not reject it.

## Validation

- config selector: 12 assertions, pass on exact candidate
- REST footer integration selector: 61 assertions, pass on exact candidate
- config test covers positive seed, zero readback, `-1`/`-1KiB` field-scoped refusal, and zero preservation
- REST integration proves zero emits one HEAD and no GET at open; a trailer read is oracle-correct and emits the first GET
- hosted lint, GCC release, and Clang debug checks: pass
- `git diff --check`: pass
- exact base: `0d7f98d9a662e9e3b47d6f7d6e2c8797d1e6572b`
- candidate: `3b0fa1c1ab5c106bf1818c7be2a78c9c11a1270d`
- tree: `ea1128def00a55fd1fbeae50a1d35d735a025521`
- zero-context patch SHA-256: `a045d2a11ff66afc9abff0739e7e01225c3e77a7d1f938d7e32e79fb02443642`

Independent review is clear. The initial lint run found one clang-format spacing edit; the current candidate applies that exact formatter repair.

- exact-head hosted receipt for `3b0fa1c1ab5c106bf1818c7be2a78c9c11a1270d`: CUDA 13 runtime and terminal aggregate checks passed
- hosted workflow 31695219911: runtime job 94432177546 passed from 2026-08-13T12:23:09Z through 2026-08-13T12:42:37Z; aggregate job 94450935549 passed at 2026-08-13T12:42:47Z

PR #1514, opened 2026-08-13
fix(config): define IO configuration contracts

infblueocean:
## Review question

Should `request_timeout_s: 0` explicitly disable libcurl's whole-transfer deadline while presigned URLs remain bounded (300-second fallback at zero, 60-second headroom for positive values, seven-day S3 maximum)? The previous implementation documented `0 = no limit` but silently retained a generic 30-second curl timeout.

This is judgment-bearing because removing that hidden deadline can permit an intentionally unlimited request to remain hung. Please push back if Sirius wants a bounded policy instead; negative values should still be rejected rather than acting as an undocumented alias.

## Summary

- reject zero REST and local `max_n_chunks` values that Sirius cannot honor
- define `footer_probe_bytes: 0` as the suffix-probe opt-out using HEAD plus ordinary reads
- define `use_odirect: false` as forced buffered reads without waiving the backend's current O_DIRECT-open requirement
- reject negative `request_timeout_s`
- make documented `request_timeout_s: 0` real by clearing libcurl's hidden 30-second transfer timeout
- add 60 seconds of signing headroom without overflow and cap SigV4 TTL at S3's seven-day maximum
- preserve every shipped default and positive curl deadline

## Why one PR

These settings define boundary values for the same REST and local IO surfaces. Zero selects an existing path for the footer probe and request deadline, but is incoherent for either chunk-count maximum. A negative request timeout previously followed the non-positive path silently. The hidden generic curl timeout also made the documented zero contract false.

This PR makes those existing contracts executable without choosing a new IO architecture or tuning value. It supersedes #1512 and retains the earlier #1513/#1514 IO-contract union.

## Contract

- omitted/null and positive chunk limits preserve current behavior
- zero, negative, and signed-overflow chunk limits fail without mutation
- zero footer probe emits no suffix GET and falls back to HEAD plus ordinary reads
- `use_odirect: false` selects buffered reads under the backend's current open requirements
- request timeout: omitted keeps 30 seconds, positive values remain curl deadlines, `0` clears the curl deadline, negatives fail without mutation
- signing TTL: zero maps to 300 seconds; positive values receive 60 seconds of headroom and saturate at 604800 seconds

## Current identity

- exact base: `3e4d98cfb760c716540b8d4955a9de37d5acfa53`
- exact head: `7ca3768e5a4ce206d7efd60954e2da12afd8edfb`
- base tree: `32f681ebd0ad17bbd942988dfc57e010c46412a2`
- candidate tree: `afa777945c89e95bb68290d444fe770ceea4332a`
- zero-context source-diff SHA-256: `c8f2f479237e5990d65a8a527b9fa5db26de9cd2afca9a5722eba7b0271c525d`
- cumulative diff: 13 paths, 342 insertions, 27 deletions
- clean two-commit rebase onto current `dev`

## Validation

- focused YAML regression: zero accepted/read back, negative rejected without mutating the 30-second default, positive 11 preserved
- focused fake-server regression: a 1-second deadline aborts a 1.5-second response; zero permits a slower response to complete
- focused signing regression: `0 -> 300`, `1 -> 61`, and `LONG_MAX -> 604800`
- retained IO boundary tests cover null/omitted/positive/zero/negative/overflow chunk counts, footer opt-out, buffered reads, and path selection
- `git diff --check` passes at the exact head
- exact-head hosted validation is terminal green: Check workflow `31923497755`, Test workflow `31923497904`, and Distribution workflow `31923498178` passed their applicable gates

Earlier hosted receipts belong only to predecessor heads and are not execution evidence for this candidate. Internal review superseded `81659f1c` after finding the hidden 30-second curl deadline, superseded `cbbf01a2` after finding the SigV4 seven-day boundary, and superseded `b272a5ad` after hosted lint supplied the repo-pinned clang-format diff.

## Bounded follow-up

The shared `sigv4::presign_url` primitive still accepts caller-provided TTLs above seven days. This PR keeps every REST-reactor call valid; hardening the shared primitive is deliberately left as a separate mechanical patch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -632,6 +632,7 @@ set(TEST_SOURCES
     test/cpp/compression/test_compression.cpp
     test/cpp/config/test_config.cpp
     test/cpp/config/test_context.cpp
+    test/cpp/config/test_local_io_config.cpp
     test/cpp/config/test_object_store_config.cpp
     test/cpp/config/test_topology_config.cpp
     test/cpp/data/test_convertible_data_batch.cpp
--- a/docs/README.md
+++ b/docs/README.md
@@ -30,7 +30,7 @@ Running TPC-H on 1TB data, Sirius accelerates DuckDB by 5x on DGX Station (GB300
 - NVIDIA Turing or newer, with compute capability 7.5+.
 - CUDA 13.x (driver 580.65.06 or newer) or CUDA 12.x (driver 525.60.13 or newer).
 - `io_uring` enabled at runtime (`CONFIG_IO_URING`, `kernel.io_uring_disabled=0` recommended). Containers must allow `io_uring_setup`, `io_uring_enter`, and `io_uring_register`.
-- Local Parquet files should be on a filesystem/block device that supports direct I/O (`O_DIRECT`).
+- Local Parquet files should be on a filesystem/block device that supports direct I/O (`O_DIRECT`). This remains a requirement when `scan_manager.local.use_odirect` is `false`, because the local backend still opens an `O_DIRECT` handle even though requests use the buffered handle.
 
 ### Build Requirements
 
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -328,8 +328,8 @@ single-GPU configurations only (logs a warning and disables itself otherwise).
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
-| `use_odirect` | bool | true | Use `O_DIRECT` for local-disk reads. |
-| `max_n_chunks` | int | 1 | Max contiguous file segments fused into one vectored read. |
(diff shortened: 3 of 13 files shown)


infblueocean:
Ready for maintainer review: exact-head hosted lint/build/runtime are green, and the AWS L4 focused contract passed 6/6 assertions. This is docs/tests only: `use_odirect: false` selects buffered request reads but does not remove Sirius runtime filesystem O_DIRECT support requirements.

infblueocean:
Rebase receipt: exact current `dev` base `e885ffea7d0541f1e3493566e2db0ed6e431106f`; exact head `86e793fae02a7195121163a779123e834cc56efe`; tree `3dae57aeb0b3cc2f9225cdb7e09815c8b6c45ebd`; cumulative diff SHA-256 `12b03e479589b6da2ed12b92f6bd87b6a2a090e8cf0a42e1ce91001b401b8e65`. The rebase was conflict-free and patch-equivalent, preserving the 11-path `208+/12-` patch. Hosted workflow 31758752661 is terminal green; runtime job 94641013061 passed in 20m06s and terminal aggregate 94644318207 passed. Reviewer routing is unchanged.

PR #1508, opened 2026-08-13
fix(config): reject negative downgrade monitor periods

infblueocean:
## Summary
- reject negative `sirius.executor.downgrade.monitor_period` values before target-resolution conversion
- preserve `0` as the intentional monitor-loop disable value and positive durations unchanged
- document the non-negative contract

## Why
`downgrade_executor::start()` starts the pressure monitor only when `monitor_period > 0`. Previously the YAML path accepted negative durations, so a typo such as `-1ms` silently disabled automatic downgrade recovery exactly like the intentional `0` setting. Validation must happen before conversion to milliseconds: supported values such as `-1us` and `-0.5ms` otherwise truncate to `0ms` and evade a post-conversion check.

## Tests
- config-path fixtures reject bare `-1`, `-1ms`, `-1us`, and `-0.5ms`
- config-path fixtures accept and read back `0` and `25ms`
- direct reader regression proves sub-millisecond refusal happens before truncation and leaves the destination unchanged
- `python3 scripts/check_orphan_tests.py`
- YAML fixture parse + `git diff --check`
- hosted exact-head lint, GCC release, Clang debug, CUDA 13 release build, full C++ runtime suite, dataset preparation, TPC-H snapshot, artifact upload, and aggregate check passed
- CUDA runtime job: 20m12s; C++ unit-test step: 18m53s

Exact base: `54c7d02864efc42c28987bcdb0896626a7bb371f`
Candidate: `c4b48153af995b38db002d3bc137b45991ce001a`
Tree: `2bc7e9111fd5d1dab7dd0d1bbb23fe8b78be57f4`

- exact-head hosted receipt for `c4b48153af995b38db002d3bc137b45991ce001a`: CUDA 13 runtime and terminal aggregate checks passed
- hosted workflow 31689001799: runtime job 94412744234 passed from 2026-08-13T10:19:25Z through 2026-08-13T10:39:37Z; aggregate job 94421389286 passed at 2026-08-13T10:39:47Z

PR #1515, opened 2026-08-13
fix(config): reject negative REST durations

infblueocean:
## Summary

- reject negative `upkeep_interval_ms`, `conn_max_age_s`, `retry_backoff_base_ms`, and `retry_jitter_ms` values before duration-resolution conversion
- preserve each field's intentional zero behavior
- document bare-number units and supported duration suffixes
- cover omitted/null defaults, zero, bare and suffixed positives, and negative refusal without mutation

## Why

All four values currently accept negative bare and suffixed YAML values. Those values silently collapse into zero-like runtime behavior: upkeep and connection-age controls are skipped by `> 0` guards, negative jitter disables jitter, and negative backoff makes retries immediately due rather than backing off.

This branch was intentionally stacked on predecessor #1508, whose `yaml::non_negative_duration` wrapper is now consolidated in #1529. That wrapper checks the sign before converting to the target resolution, so a value such as `-1us` cannot truncate to `0ms` and evade validation. The merge diff after #1508 is limited to the four REST consumers, documentation, and focused tests.

## Semantics retained

- upkeep `0`: disable upkeep
- connection max age `0`: use curl's default
- retry base `0`: no fixed backoff
- retry jitter `0`: no random jitter

## Validation

- focused selector: 108 assertions, pass on exact candidate
- historical dependency/base: #1508 head `c4b48153af995b38db002d3bc137b45991ce001a`; current dependency owner: #1529
- candidate: `66d965f2d6f129e5a13f5ed1ed301ec56f9d80ba`
- tree: `dc9afdf895b9568e251c747ed8effdce0d7d4bf8`
- clang-format 20.1.4, codespell, whitespace, EOF, orphan-test, and `git diff --check`: pass
- fresh exact-head hosted lint, GCC release, Clang debug, CUDA 13 build, distribution, check, PR-title, full runtime, and terminal aggregate checks pass on the exact stacked head
- test workflow `31696230243` passed; runtime job `94435359145` ran from `2026-08-13T12:43:15Z` through `2026-08-13T13:03:15Z` (20m00s), including C++ unit tests and the TPC-H snapshot; terminal aggregate job `94456486837` passed

The exact stacked candidate is execution-complete. It remains dependent on that duration wrapper and must rebase onto #1529 or its merge after upstream #1340 lands; this receipt does not imply standalone merge readiness against `dev`.

## Residual

This patch validates the lower bound only. It does not claim to fix the generic duration parser's out-of-range positive suffixed conversion or bound exponential backoff/jitter arithmetic; those require a separate upper-bound contract.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -325,10 +325,10 @@ and transport use one trust policy; there are no separate REST YAML controls.
 | `chunk_size` | bytes | 8Mi | Target bytes per ranged GET (scatter/device-staging paths). |
 | `max_n_chunks` | int | 16 | Max file-adjacent segments fused into one scatter GET. |
 | `max_read_split` | int | 16 | Max parallel ranged GETs for one contiguous host read (reads < 2 MiB stay a single GET). |
-| `upkeep_interval_ms` | int (ms) | 15000 | Idle-connection keepalive interval (`curl_easy_upkeep`; 0 disables). |
-| `conn_max_age_s` | int (seconds) | 20 | Max age curl may reuse a pooled connection (`CURLOPT_MAXAGE_CONN`; 0 = curl default). |
-| `retry_backoff_base_ms` | int (ms) | 50 | Base backoff between retries. |
-| `retry_jitter_ms` | int (ms) | 50 | Random jitter added to retry backoff. |
+| `upkeep_interval_ms` | duration (**>= 0**) | 15s | Idle-connection keepalive interval (`curl_easy_upkeep`; 0 disables). Bare numbers are milliseconds; duration suffixes are accepted. |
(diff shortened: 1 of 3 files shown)


infblueocean:
Ready for maintainer review: source/build checks are green, the exact AWS L4 focused regression passed 108/108 assertions, and the hosted CUDA runtime is now terminal green. The stacked base is #1508; this patch preserves literal zero meanings and rejects only negative REST duration aliases.

infblueocean:
Closing this draft back into the local validated patch queue while #1340 replaces the S3/REST config types and parser boundary it targets. The exact head and evidence remain retained; after #1340 lands we will re-audit the new types and reopen only the surviving behavior, rather than leave a knowingly stale review item open.

PR #1494, opened 2026-08-13
fix(config): validate and derive GPU topology

infblueocean:
## Summary

- reject empty, negative, or duplicate explicit `gpu_ids`
- reject negative `num_gpus` before unsigned conversion
- make `gpu_ids` and `num_gpus` strictly mutually exclusive by key presence
- remove `executor.pipeline.cpu_affinity` from normal YAML
- retain per-GPU topology-derived pipeline affinity and other expert affinity
  controls

## Why one PR

These changes establish one GPU-topology source of truth. Users select valid
devices through `gpu_ids` or `num_gpus`; production Sirius then derives each GPU
pipeline executor's CPU affinity from discovered topology. The removed YAML
affinity could never be honored because runtime always replaced it. This review
unit validates the input and proves the downstream derivation without changing
automatic all-visible-GPU behavior or `gpus_per_query`.

This supersedes #1503. Its stale-key rejection and topology-overrides-sentinel
regression are retained alongside #1494's complete selector fixture matrix.

## Contract

- explicit `gpu_ids` is non-empty, unique, and non-negative
- `gpu_ids` and `num_gpus` cannot both be present
- negative `num_gpus` fails before unsigned conversion; zero keeps existing
  automatic all-visible-GPU behavior
- pipeline affinity is derived from the selected GPU's topology
- task-creator, downgrade, and scan-manager affinity controls remain unchanged

## Current identity

- exact upstream `dev` base: `e885ffea7d0541f1e3493566e2db0ed6e431106f`
- exact rebased head: `45392102685f3960a6726b908d600bf9c7ed3c9e`
- candidate tree: `cd3e46be9636883a97d3da4baf037930e427815b`
- zero-context source-diff SHA-256: `9efc6e7607f2537c7ed893cafb50c441c598972eb71ed788ccc100a058f401a3`
- pre-rebase safety branch:
  `infblueocean:foxglove/safety-pr1494-pre-rebase-20260813-b9c68dcf`

The sole rebase conflict was additive in `test/cpp/config/test_context.cpp`:
upstream's dynamic-filter validation test and this PR's GPU-topology validation
test are both retained. `git range-diff` shows no behavioral change to this PR.

## Validation

- scoped pre-commit on all 11 changed files: every applicable hook passes;
  the repository Markdown wrapper alone cannot start because local `pixi` is
  unavailable
- standalone `uvx rumdl check docs/super-sirius/configuration.md`: passes
- `git diff --check`: passes
- hosted exact-head CI is fully green:
[...]

PR #1518, opened 2026-08-13
fix(config): validate and enforce CPU affinity

infblueocean:
## Summary

- validate configured Linux CPU IDs against bounds and the caller's current allowed mask before creating workers
- apply the complete list as one shared eligibility mask inside every static and bounded pool worker, then require exact kernel readback
- stop and join every created worker before propagating an affinity startup error
- make current pool owners roll back cleanly when fail-closed startup rejects a mask
- document shared-mask and empty-inheritance semantics, with focused YAML, pool, and owner regressions

## Configuration contract

`cpu_affinity` remains an expert policy input for the task-creator, downgrade,
and scan-manager pools. Omitted or empty lists inherit the constructing
thread's allowed mask. Every listed CPU is an eligibility choice shared by all
workers—not a one-CPU-per-worker assignment—and must be inside the live
cgroup/cpuset-constrained process mask.

The GPU-pipeline YAML surface stays removed by merged #1494. Pipeline affinity
is derived from the selected GPU's CPU topology and passes through the same
fail-closed pool boundary.

## September 9 current-dev repair

Current `dev` added query identity to the task-creator test fixture at the same
constructor changed by this PR. The replacement keeps affinity as the fifth
parameter and query ID as the sixth, preserves query-scoped client-context
registration, and retains the original CPU-affinity behavior. Hosted lint then
identified a formatter-only layout delta; the exact generated formatting was
applied and independently re-reviewed with no semantic change.

## Exact reviewed identity

- tier: `judgment`
- exact base: `285737bc33c84d5c175c59862162b518b89fcae0`
- base tree: `8a3ac4a8c9809355bd9303ca1dc274fd6211752a`
- exact head: `88c733feee525922069b093102405c94af0d8620`
- candidate tree: `49a77d61a6a4987cae4f3e39e5755b0afb61eae3`
- cumulative binary-diff SHA-256: `3bc56020413be8a62594995c9bf0ee2f1f831bc21b0623a5763c348f3b2cc0e0`
- cumulative diff: 18 paths, 849 insertions, 128 deletions

## Author-side validation

- independent exact-head full-diff and formatter-delta review: no critical, important, or minor findings
- clean worktree, `git diff --check`, and `git show --check`: pass
- affinity, readback, partial-start rollback, retry, owner-lifecycle, and topology-derived placement regressions retained
- hosted Check `34395692616`: lint, GCC release, Clang debug, and aggregate pass
- hosted Test `34395692800`: CUDA 13 build, C++ runtime suite, TPC-H snapshot, and aggregate pass
- hosted Distribution `34395693943`: aggregate pass; CUDA distribution leaves correctly skip because distribution inputs did not change
- hosted Validate `34395690336`: title and aggregate validation pass

The replacement head is author-ready and terminal-green. Independent
maintainer approval of this exact head remains the merge gate.

### Retained predecessor evidence

The earlier Linux C++20 ASan/UBSan affinity/startup harness and previous hosted
heads remain supporting evidence for unchanged behavior; they do not replace
the current exact-head gates above.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -272,6 +272,13 @@ hardware-derived exception described below:
 | `num_threads` | int (**> 0**) | per pool (below) | Worker threads in the pool. |
 | `cpu_affinity` | list of int | — | Cores to pin task-creator and downgrade threads. GPU pipeline affinity is derived from the selected GPU's CPU topology. |
 
+For the task-creator, downgrade, and scan-manager pools, `cpu_affinity` is one shared eligibility
+mask applied to every worker; entries are Linux logical CPU IDs, not a one-core-per-worker
+assignment. The list must be a subset of the process's currently allowed CPUs, including
+cgroup/cpuset restrictions. An omitted or empty list inherits the process mask. Sirius rejects
+invalid, unavailable, or unapplied masks rather than silently running with a different affinity.
+GPU pipeline affinity is replaced at execution with the selected GPU's topology-derived CPU mask.
+
 ### `sirius.executor.task_creator`
 
 Thread pool (default `num_threads: 1`). Task creation
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -293,17 +293,31 @@ void task_creator::start_thread_pool()
   // a paired reactivate() here, subsequent schedule() pushes silently no-op
   // and the next query's manager_loop sees an empty/inactive queue forever.
   _task_creation_queue.reactivate();
-  _bounded_pool =
-    std::make_unique<exec::bounded_thread_pool>(_config.thread_pool.num_threads,
-                                                _config.thread_pool.thread_name_prefix,
-                                                _config.thread_pool.cpu_affinity_list);
-  _manager_thread = std::thread(&task_creator::manager_loop, this);
+  try {
+    _bounded_pool =
+      std::make_unique<exec::bounded_thread_pool>(_config.thread_pool.num_threads,
+                                                  _config.thread_pool.thread_name_prefix,
+                                                  _config.thread_pool.cpu_affinity_list);
+    _manager_thread = std::thread(&task_creator::manager_loop, this);
+  } catch (...) {
+    _running = false;
+    if (_bounded_pool) { _bounded_pool->interrupt(); }
+    _task_creation_queue.interrupt();
+    if (_manager_thread.joinable()) { _manager_thread.join(); }
+    if (_bounded_pool) {
+      _bounded_pool->wait_all();
+      _bounded_pool->stop();
+      _bounded_pool.reset();
+    }
+    throw;
+  }
 }
 
 void task_creator::do_stop_thread_pool()
 {
   bool expected = true;
   if (!_running.compare_exchange_strong(expected, false)) { return; }
+  if (!_bounded_pool) { return; }
   _bounded_pool->interrupt();
   _task_creation_queue.interrupt();
   if (_manager_thread.joinable()) { _manager_thread.join(); }
--- a/src/creator/task_creator.cpp
+++ b/src/creator/task_creator.cpp
@@ -364,17 +364,31 @@ void task_creator::start_thread_pool()
   // a paired reactivate() here, subsequent schedule() pushes silently no-op
   // and the next query's manager_loop sees an empty/inactive queue forever.
   _task_creation_queue.reactivate();
-  _bounded_pool =
-    std::make_unique<exec::bounded_thread_pool>(_config.thread_pool.num_threads,
-                                                _config.thread_pool.thread_name_prefix,
-                                                _config.thread_pool.cpu_affinity_list);
-  _manager_thread = std::thread(&task_creator::manager_loop, this);
+  try {
+    _bounded_pool =
+      std::make_unique<exec::bounded_thread_pool>(_config.thread_pool.num_threads,
+                                                  _config.thread_pool.thread_name_prefix,
+                                                  _config.thread_pool.cpu_affinity_list);
+    _manager_thread = std::thread(&task_creator::manager_loop, this);
+  } catch (...) {
+    _running = false;
+    if (_bounded_pool) { _bounded_pool->interrupt(); }
+    _task_creation_queue.interrupt();
+    if (_manager_thread.joinable()) { _manager_thread.join(); }
+    if (_bounded_pool) {
+      _bounded_pool->wait_all();
+      _bounded_pool->stop();
+      _bounded_pool.reset();
+    }
+    throw;
+  }
 }
 
 void task_creator::do_stop_thread_pool()
 {
   bool expected = true;
   if (!_running.compare_exchange_strong(expected, false)) { return; }
+  if (!_bounded_pool) { return; }
   _bounded_pool->interrupt();
   _task_creation_queue.interrupt();
   if (_manager_thread.joinable()) { _manager_thread.join(); }
--- a/src/downgrade/downgrade_executor.cpp
+++ b/src/downgrade/downgrade_executor.cpp
@@ -64,46 +64,62 @@ void downgrade_executor::start()
   bool expected = false;
   if (!_running.compare_exchange_strong(expected, true)) { return; }
 
-  // HOST/DISK tier memory_spaces return device_id == -1; passing that to
-  // rmm::cuda_device_id or cudaSetDevice fails with cudaErrorInvalidDevice.
-  // Default the stream pool to GPU 0 for non-GPU tiers and skip per-thread
-  // CUDA binding entirely (the stream is ordering metadata; host/disk work
-  // is CPU-side).
-  {
-    int device_id = 0;
-    if (_space_id.tier == cucascade::memory::Tier::GPU && _memory_space) {
-      device_id = _memory_space->get_device_id();
-    }
-    _stream_pool = std::make_unique<cucascade::memory::exclusive_stream_pool>(
-      rmm::cuda_device_id{device_id}, _config.thread_pool.num_threads);
-  }
-
-  _request_queue.reactivate();
-
-  absl::AnyInvocable<void() noexcept> per_thread_init = nullptr;
-  if (_memory_space && _space_id.tier == cucascade::memory::Tier::GPU) {
-    auto device_id  = _memory_space->get_device_id();
-    per_thread_init = [device_id]() noexcept {
-      // Pin each worker to its GPU; silent failure leaks downgrade memcpys
-      // across contexts. Lambda is noexcept, so check inline.
-      cudaError_t err = cudaSetDevice(device_id);
-      if (err != cudaSuccess) {
-        SIRIUS_LOG_ERROR("downgrade_executor per-thread init: cudaSetDevice({}) failed: {}",
-                         device_id,
-                         cudaGetErrorString(err));
+  try {
+    // HOST/DISK tier memory_spaces return device_id == -1; passing that to
+    // rmm::cuda_device_id or cudaSetDevice fails with cudaErrorInvalidDevice.
+    // Default the stream pool to GPU 0 for non-GPU tiers and skip per-thread
+    // CUDA binding entirely (the stream is ordering metadata; host/disk work
+    // is CPU-side).
+    {
+      int device_id = 0;
+      if (_space_id.tier == cucascade::memory::Tier::GPU && _memory_space) {
+        device_id = _memory_space->get_device_id();
       }
-    };
-  }
+      _stream_pool = std::make_unique<cucascade::memory::exclusive_stream_pool>(
+        rmm::cuda_device_id{device_id}, _config.thread_pool.num_threads);
+    }
 
-  _pool = std::make_unique<exec::bounded_thread_pool>(_config.thread_pool.num_threads,
-                                                      _config.thread_pool.thread_name_prefix,
-                                                      _config.thread_pool.cpu_affinity_list,
-                                                      std::move(per_thread_init));
+    _request_queue.reactivate();
 
-  _processing_thread = std::thread(&downgrade_executor::processing_loop, this);
+    absl::AnyInvocable<void() noexcept> per_thread_init = nullptr;
+    if (_memory_space && _space_id.tier == cucascade::memory::Tier::GPU) {
+      auto device_id  = _memory_space->get_device_id();
+      per_thread_init = [device_id]() noexcept {
+        // Pin each worker to its GPU; silent failure leaks downgrade memcpys
+        // across contexts. Lambda is noexcept, so check inline.
+        cudaError_t err = cudaSetDevice(device_id);
+        if (err != cudaSuccess) {
+          SIRIUS_LOG_ERROR("downgrade_executor per-thread init: cudaSetDevice({}) failed: {}",
+                           device_id,
+                           cudaGetErrorString(err));
+        }
+      };
+    }
 
-  if (_memory_space && _config.monitor_period > std::chrono::milliseconds::zero()) {
-    _monitor_thread = std::thread(&downgrade_executor::monitor_loop, this);
+    _pool = std::make_unique<exec::bounded_thread_pool>(_config.thread_pool.num_threads,
+                                                        _config.thread_pool.thread_name_prefix,
+                                                        _config.thread_pool.cpu_affinity_list,
+                                                        std::move(per_thread_init));
+
+    _processing_thread = std::thread(&downgrade_executor::processing_loop, this);
+
+    if (_memory_space && _config.monitor_period > std::chrono::milliseconds::zero()) {
+      _monitor_thread = std::thread(&downgrade_executor::monitor_loop, this);
+    }
+  } catch (...) {
+    _running = false;
+    if (_pool) { _pool->interrupt(); }
+    _request_queue.interrupt();
+    _monitor_cv.notify_one();
+    if (_monitor_thread.joinable()) { _monitor_thread.join(); }
+    if (_processing_thread.joinable()) { _processing_thread.join(); }
+    if (_pool) {
+      _pool->wait_all();
+      _pool->stop();
+      _pool.reset();
+    }
+    _stream_pool.reset();
+    throw;
   }
 }
 
@@ -112,6 +128,10 @@ void downgrade_executor::stop()
   bool expected = true;
   if (!_running.compare_exchange_strong(expected, false)) { return; }
 
+  if (!_pool) {
+    _stream_pool.reset();
+    return;
+  }
   _pool->interrupt();
   _request_queue.interrupt();
   _monitor_cv.notify_one();
--- a/src/include/exec/bounded_thread_pool.hpp
+++ b/src/include/exec/bounded_thread_pool.hpp
@@ -16,12 +16,14 @@
 
 #pragma once
 
+#include "exec/cpu_affinity.hpp"
+#include "exec/thread_pool_startup.hpp"
 #include "log/logging.hpp"
 
 #include <absl/functional/any_invocable.h>
 
 #include <condition_variable>
-#include <latch>
+#include <exception>
 #include <mutex>
 #include <queue>
 #include <string>
@@ -96,37 +98,20 @@ class bounded_thread_pool {
                                absl::AnyInvocable<void() noexcept> per_thread_init = nullptr)
     : capacity_(capacity)
   {
+    validated_cpu_affinity const affinity(cpu_ids);
     threads_.reserve(capacity);
 
-    cpu_set_t cpuset;
-    CPU_ZERO(&cpuset);
-    for (int id : cpu_ids) {
-      CPU_SET(id, &cpuset);
-    }
-
-    std::unique_ptr<std::latch> init_latch;
-    if (per_thread_init) { init_latch = std::make_unique<std::latch>(capacity); }
-
     auto* init_fn_ptr = per_thread_init ? &per_thread_init : nullptr;
-    auto* latch_ptr   = init_latch.get();
-
-    for (int i = 0; i < capacity; ++i) {
-      auto& t = threads_.emplace_back([this, init_fn_ptr, latch_ptr]() {
-        if (init_fn_ptr) {
-          (*init_fn_ptr)();
-          latch_ptr->count_down();
-        }
-        work_loop();
-      });
-      if (!name.empty()) {
-        pthread_setname_np(t.native_handle(), (name + "_" + std::to_string(i)).c_str());
-      }
-      if (!cpu_ids.empty()) {
-        pthread_setaffinity_np(t.native_handle(), sizeof(cpu_set_t), &cpuset);
-      }
-    }
-
-    if (init_latch) { init_latch->wait(); }
+    detail::start_thread_pool_workers(
+      threads_,
+      capacity,
+      name,
+      [affinity, init_fn_ptr] {
+        affinity.apply_to_current_thread();
+        if (init_fn_ptr) { (*init_fn_ptr)(); }
+      },
+      [this] { work_loop(); },
+      [this] { request_stop(); });
   }
 
   ~bounded_thread_pool() { stop(); }
@@ -183,14 +168,7 @@ class bounded_thread_pool {
    */
   void stop() noexcept
   {
-    {
-      std::lock_guard lock(mu_);
-      if (stop_requested_) { return; }
-      stop_requested_ = true;
-      interrupted_    = true;
-      cv_capacity_.notify_all();
-      cv_work_.notify_all();
-    }
+    request_stop();
     for (auto& t : threads_) {
       if (t.joinable()) { t.join(); }
     }
@@ -225,6 +203,16 @@ class bounded_thread_pool {
   }
 
  private:
+  void request_stop() noexcept
+  {
+    std::lock_guard lock(mu_);
+    if (stop_requested_) { return; }
+    stop_requested_ = true;
+    interrupted_    = true;
+    cv_capacity_.notify_all();
+    cv_work_.notify_all();
+  }
+
   // Called exclusively by the slot destructor — covers both the drop-without-dispatch
   // and post-task-completion cases.
   void release_slot()
--- a/src/include/exec/cpu_affinity.hpp
+++ b/src/include/exec/cpu_affinity.hpp
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2025, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <pthread.h>
+#include <sched.h>
+
+#include <cerrno>
+#include <stdexcept>
+#include <string>
+#include <system_error>
+#include <vector>
+
+namespace sirius::exec {
+
+/**
+ * A validated CPU eligibility mask shared by every worker in a thread pool.
+ *
+ * Validation is intentionally performed on the constructing thread before any
+ * workers are created. Linux workers inherit that thread's allowed mask, so a
+ * requested CPU outside it could otherwise be silently discarded by the
+ * kernel (for example under a cgroup/cpuset constraint).
+ */
+class validated_cpu_affinity {
+ public:
+  explicit validated_cpu_affinity(const std::vector<int>& cpu_ids) : enabled_(!cpu_ids.empty())
+  {
+    CPU_ZERO(&requested_);
+    if (!enabled_) { return; }
+
+    cpu_set_t allowed;
+    CPU_ZERO(&allowed);
+    if (sched_getaffinity(0, sizeof(cpu_set_t), &allowed) != 0) {
+      throw std::system_error(errno, std::generic_category(), "sched_getaffinity failed");
+    }
+
+    for (int cpu_id : cpu_ids) {
+      if (cpu_id < 0) {
+        throw std::invalid_argument("CPU affinity ID " + std::to_string(cpu_id) +
+                                    " must be non-negative");
+      }
+      if (cpu_id >= CPU_SETSIZE) {
+        throw std::invalid_argument("CPU affinity ID " + std::to_string(cpu_id) +
+                                    " must be less than CPU_SETSIZE (" +
+                                    std::to_string(CPU_SETSIZE) + ")");
+      }
+      if (!CPU_ISSET(cpu_id, &allowed)) {
+        throw std::invalid_argument("CPU affinity ID " + std::to_string(cpu_id) +
+                                    " is not in the current process allowed CPU mask");
+      }
+      CPU_SET(cpu_id, &requested_);
+    }
+  }
+
+  [[nodiscard]] bool enabled() const noexcept { return enabled_; }
+
+  /** Apply the validated mask to the calling worker and verify the kernel readback. */
+  void apply_to_current_thread() const
+  {
+    if (!enabled_) { return; }
+
+    int const set_result = pthread_setaffinity_np(pthread_self(), sizeof(cpu_set_t), &requested_);
+    if (set_result != 0) {
+      throw std::system_error(set_result, std::generic_category(), "pthread_setaffinity_np failed");
+    }
+
+    cpu_set_t actual;
+    CPU_ZERO(&actual);
+    int const get_result = pthread_getaffinity_np(pthread_self(), sizeof(cpu_set_t), &actual);
+    if (get_result != 0) {
+      throw std::system_error(get_result, std::generic_category(), "pthread_getaffinity_np failed");
+    }
+
+    for (int cpu_id = 0; cpu_id < CPU_SETSIZE; ++cpu_id) {
+      if (CPU_ISSET(cpu_id, &requested_) != CPU_ISSET(cpu_id, &actual)) {
+        throw std::runtime_error("CPU affinity readback does not match the requested mask");
+      }
+    }
+  }
+
+ private:
+  cpu_set_t requested_{};
+  bool enabled_{false};
+};
+
+/** YAML-reader validator backed by the same checks used by the pool constructors. */
+struct valid_cpu_affinity {
+  bool operator()(const std::vector<int>& cpu_ids) const
+  {
+    (void)validated_cpu_affinity(cpu_ids);
+    return true;
+  }
+};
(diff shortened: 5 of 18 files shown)


infblueocean:
Ready for maintainer review: repaired exact head `1de5a9f9174817208e111bf5622fd062def8a088` is terminal green across lint, GCC, Clang, CUDA build, and full hosted runtime. The original runtime exposed a startup-handshake UAF; the repair snapshots each worker result before releasing the latch, passed a 10,000-iteration Linux ASan/UBSan handoff stress test, and now completes the exact hosted suite.

infblueocean:
Rebase receipt: exact current `dev` base `e885ffea7d0541f1e3493566e2db0ed6e431106f`; exact head `1e132f6cb48c04f333015e8326d28e9aa24e1682`; tree `4058d1c4b4f41b458d75b650504d74b87d3b6186`; cumulative diff SHA-256 `f2f1fe325e26750f685714e8ed9246a3b5679e957430b1ad4b3d5dd667c999b1`. The rebase was conflict-free and patch-equivalent, preserving the 19-path `843+/130-` patch. Hosted workflow 31758812834 is terminal green; runtime job 94641097989 passed in 20m07s and terminal aggregate 94646925160 passed. Reviewer routing is unchanged.

infblueocean:
Temporary exact-head repair draft. Mutation actor/repair owner: infblueocean. Published replacement head: 0ddfe2b19b7977eb79273db764f9784ba34c7ce0, rebased onto dev 285737bc33c84d5c175c59862162b518b89fcae0. Independent full-diff review found no critical, important, or minor issues; the only conflict resolution combines current-dev query identity with the existing CPU-affinity fixture while preserving affinity as the fifth parameter and query_id as the sixth. Named external dependency: fresh hosted Check, Test, Distribution, and Validate results for this exact head. Ready condition: all required non-skipped jobs terminal-green, mergeable state clean, no unresolved review threads, and the public exact-head receipt refreshed. Prior-head CI does not satisfy this condition.

infblueocean:
Repair receipt update: hosted Check run 34394581731 identified one deterministic clang-format-only delta in the merged test fixture. The exact CI-generated formatting was applied and independently re-reviewed with no semantic delta. Current published head: 88c733feee525922069b093102405c94af0d8620; mutation actor/repair owner: infblueocean. This supersedes the prior 0ddfe2b repair head and restarts the named exact-head Check, Test, Distribution, and Validate dependency. The PR remains draft until the same terminal-green, mergeable, thread-clean ready condition is satisfied.

infblueocean:
Ready-state receipt: infblueocean restored this PR to ready at exact head 88c733feee525922069b093102405c94af0d8620. Basis: independent full-diff and formatter-delta review with no findings; Check 34395692616, Test 34395692800, Distribution 34395693943, source-event Validate 34395690336, and post-body Validate 34398615725 all pass; mergeable; zero unresolved review threads. Independent maintainer approval of this exact head remains required before merge.

felipeblazing (commented):
**`src/include/exec/cpu_affinity.hpp`**

**Reviewer: Concurrency, lifetime & hangs**

**minor** Validation checks the calling thread's mask, not the process mask

`sched_getaffinity(0, ...)` returns the mask of the thread that constructs the pool. That differs between the config-load thread, the `SiriusContext` init thread, and the query thread that restarts task_creator on every query. So the same YAML can pass at load and then fail (or pass) depending on which thread happens to restart the pool. The docs and the error message both say "process allowed CPU mask". Either capture the mask once at init and validate against that snapshot, or reword the contract to "constructing thread's mask" and make sure every restart runs on an unpinned thread.

---

**`src/include/exec/cpu_affinity.hpp`**

**Reviewer: Comments, docs & AI residue**

**minor** "process allowed CPU mask" is actually the calling thread's mask

`sched_getaffinity(0, ...)` returns the affinity of the calling thread, not of the process. So the error text, the docs ("process's currently allowed CPUs", "inherits the process mask") and `config.hpp` ("constructing thread's") describe three different things. YAML validation runs on the config-loading thread, but pools are built on whatever thread calls `start()`. Say "calling thread's allowed CPU mask" in the error and the docs so users can tell why a mask accepted at load time can still be rejected at start.

---

**`src/include/exec/thread_pool_startup.hpp`**

**Reviewer: Error handling & fail-closed gating**

**minor** Affinity errors don't identify the pool or device that failed

A message such as 'CPU affinity ID 64 is not in the current process allowed CPU mask' reaches the user with no pool name or GPU. For topology-derived masks the ID doesn't appear in any config either, so the error is hard to act on. In `start_thread_pool_workers` or the pool constructors, rethrow with the `name` prefix, or log it with `SIRIUS_LOG_ERROR` before propagating. The owners roll back silently, so no log records which pool failed.

---

**`src/pipeline/task_scheduler.cpp`**

**Reviewer: Configuration surface & defaults**

**major** Topology-derived GPU pipeline affinity now fails closed under any cpuset/taskset restriction

`task_scheduler` sets `config.cpu_affinity_list = it->cpu_cores`, and cucascade fills that from sysfs `local_cpulist`, which ignores the process cpuset. Now that every pool throws when a CPU is outside `sched_getaffinity`, running under `taskset -c 0-7`, `docker --cpuset-cpus`, or a Slurm cgroup makes `task_scheduler::start()` throw on every GPU whose NUMA-local list extends past the allowed set. Before this PR the kernel silently intersected the mask and the query still ran. The user can't configure this value (YAML key removed in #1494), so intersect the derived list with the allowed mask, inheriting if the result is empty, and keep strict validation for user-supplied `cpu_affinity` only.

docs/super-sirius/configuration.md
```diff
@@ -272,6 +272,13 @@ hardware-derived exception described below:
 | `num_threads` | int (**> 0**) | per pool (below) | Worker threads in the pool. |
 | `cpu_affinity` | list of int | — | Cores to pin task-creator and downgrade threads. GPU pipeline affinity is derived from the selected GPU's CPU topology. |
 
+For the task-creator, downgrade, and scan-manager pools, `cpu_affinity` is one shared eligibility
```
felipeblazing, line 275: **Reviewer: Comments, docs & AI residue**

**minor** Config reference contradicts itself on what `cpu_affinity` means

The new paragraph says `cpu_affinity` is a shared eligibility mask and not a per-core assignment, and that it covers scan-manager. Two unchanged table rows still say otherwise: the common-keys row just above (line 273) reads "Cores to pin task-creator and downgrade threads", and the scan-manager row (line 312) reads "Cores to pin scan-manager threads to". Reword both rows to "shared CPU eligibility mask (see below)" and list all three pools, so the table and the paragraph agree.

docs/super-sirius/configuration.md
```diff
@@ -272,6 +272,13 @@ hardware-derived exception described below:
...
+mask applied to every worker; entries are Linux logical CPU IDs, not a one-core-per-worker
+assignment. The list must be a subset of the process's currently allowed CPUs, including
+cgroup/cpuset restrictions. An omitted or empty list inherits the process mask. Sirius rejects
```
felipeblazing, line 278: **Reviewer: Configuration surface & defaults**

**nit** "Inherits the process mask" should say the constructing thread's mask

The docs say an empty list inherits the process mask. The code, and the new comment in `exec/config.hpp`, say it inherits the constructing thread's mask, and that thread's mask is also what the check is made against. These differ whenever the constructing thread has been pinned. Use the same wording as the comment in `config.hpp`.

docs/super-sirius/configuration.md
```diff
@@ -272,6 +272,13 @@ hardware-derived exception described below:
...
+invalid, unavailable, or unapplied masks rather than silently running with a different affinity.
+GPU pipeline affinity is replaced at execution with the selected GPU's topology-derived CPU mask.
```
felipeblazing, line 280: **Reviewer: Comments, docs & AI residue**

**minor** Docs don't say that topology-derived pipeline affinity now fails closed too

Pipeline affinity comes from `gpu.cpu_cores` and now goes through the same validation. A container whose cpuset excludes the GPU's NUMA-local cores will therefore fail `task_scheduler::start()`, and there is no YAML key to override it. This paragraph only says the mask is "replaced at execution". State that the derived mask must also fall inside the allowed CPUs, or startup fails.

src/creator/task_creator.cpp
```diff
@@ -364,17 +364,31 @@ void task_creator::start_thread_pool()
...
-                                                _config.thread_pool.thread_name_prefix,
-                                                _config.thread_pool.cpu_affinity_list);
-  _manager_thread = std::thread(&task_creator::manager_loop, this);
+  try {
+    _bounded_pool =
+      std::make_unique<exec::bounded_thread_pool>(_config.thread_pool.num_threads,
+                                                  _config.thread_pool.thread_name_prefix,
+                                                  _config.thread_pool.cpu_affinity_list);
+    _manager_thread = std::thread(&task_creator::manager_loop, this);
+  } catch (...) {
```
felipeblazing, line 373: **Reviewer: API design, layering & reuse**

**major** Rollback catch block re-implements do_stop_thread_pool() line for line

The catch body is the same teardown as `do_stop_thread_pool()` (interrupt, join the manager, wait_all, stop, reset), and `do_stop_thread_pool()` already handles a null pool and a non-joinable thread. `_running` is still true here, so the owner's own stop path can do the rollback. That way startup and shutdown can't drift apart the next time a teardown step is added.
```suggestion
  } catch (...) {
    do_stop_thread_pool();
    throw;
  }
```

src/creator/task_creator.cpp
```diff
@@ -364,17 +364,31 @@ void task_creator::start_thread_pool()
...
+    _running = false;
+    if (_bounded_pool) { _bounded_pool->interrupt(); }
+    _task_creation_queue.interrupt();
+    if (_manager_thread.joinable()) { _manager_thread.join(); }
+    if (_bounded_pool) {
+      _bounded_pool->wait_all();
+      _bounded_pool->stop();
+      _bounded_pool.reset();
+    }
+    throw;
```
felipeblazing, line 383: **Reviewer: Concurrency, lifetime & hangs**

**major** A startup throw during the per-query task_creator restart leaves the engine permanently wedged

`start_thread_pool()` isn't only called at init. `wait_for_completion()` and `drain_after_error()` call it after every query (task_scheduler.cpp:279/314/320), on the query thread. It can now throw on those paths too: the validator rejects an ID, `pthread_setaffinity_np` fails after a cpuset change, or thread creation hits EAGAIN. When it does, this rollback leaves `_task_creation_queue` interrupted and `_running=false`. The call in `drain_after_error()` runs inside the engine's catch handler, so nothing restarts the creator, and the next query's `schedule()` pushes are silently dropped, so its `future.get()` never returns. Either validate and apply the mask only once at `SiriusContext` init and reuse it for the restarts, or make the query-boundary restart surface a fatal error instead of leaving the context looking usable.

src/downgrade/downgrade_executor.cpp
```diff
@@ -64,46 +64,62 @@ void downgrade_executor::start()
...
+
+    absl::AnyInvocable<void() noexcept> per_thread_init = nullptr;
+    if (_memory_space && _space_id.tier == cucascade::memory::Tier::GPU) {
+      auto device_id  = _memory_space->get_device_id();
+      per_thread_init = [device_id]() noexcept {
+        // Pin each worker to its GPU; silent failure leaks downgrade memcpys
+        // across contexts. Lambda is noexcept, so check inline.
+        cudaError_t err = cudaSetDevice(device_id);
+        if (err != cudaSuccess) {
+          SIRIUS_LOG_ERROR("downgrade_executor per-thread init: cudaSetDevice({}) failed: {}",
```
felipeblazing, line 92: **Reviewer: Error handling & fail-closed gating**

**minor** Downgrade worker cudaSetDevice failure is still only logged even though startup errors can now propagate

This PR adds a startup hook that can fail construction. However, `per_thread_init` is still `noexcept`, and a failed `cudaSetDevice` only logs, so the worker keeps running on device 0. The comment itself says this leaks downgrade memcpys across contexts. Put the `cudaSetDevice` check in the throwing startup hook, or have `per_thread_init` report failure, so the pool fails closed in the same way as an affinity error. `gpu_pipeline_executor::get_per_thread_init` has the same pattern.

src/downgrade/downgrade_executor.cpp
```diff
@@ -64,46 +64,62 @@ void downgrade_executor::start()
...
 
-  _processing_thread = std::thread(&downgrade_executor::processing_loop, this);
+    _processing_thread = std::thread(&downgrade_executor::processing_loop, this);
 
-  if (_memory_space && _config.monitor_period > std::chrono::milliseconds::zero()) {
-    _monitor_thread = std::thread(&downgrade_executor::monitor_loop, this);
+    if (_memory_space && _config.monitor_period > std::chrono::milliseconds::zero()) {
+      _monitor_thread = std::thread(&downgrade_executor::monitor_loop, this);
+    }
+  } catch (...) {
```
felipeblazing, line 109: **Reviewer: API design, layering & reuse**

**major** downgrade_executor::start rollback duplicates stop() and skips cancel_pending_requests

This catch copies `stop()` but drops `cancel_pending_requests()`, so the two teardown paths have already drifted. Have the catch call `stop()`. `_running` is still true, and `stop()` already handles a null `_pool`, which leaves one ordered teardown sequence for this owner.

src/downgrade/downgrade_executor.cpp
```diff
@@ -64,46 +64,62 @@ void downgrade_executor::start()
...
+    _request_queue.interrupt();
+    _monitor_cv.notify_one();
+    if (_monitor_thread.joinable()) { _monitor_thread.join(); }
+    if (_processing_thread.joinable()) { _processing_thread.join(); }
+    if (_pool) {
+      _pool->wait_all();
+      _pool->stop();
+      _pool.reset();
+    }
+    _stream_pool.reset();
```
felipeblazing, line 121: **Reviewer: Concurrency, lifetime & hangs**

**minor** Downgrade startup rollback doesn't fail requests queued after reactivate()

`_request_queue.reactivate()` runs before the pool is built, so a `request_free_memory*` / `request_downgrade` call from another thread in that window pushes a request successfully. The catch only `interrupt()`s the queue. Unlike `stop()`, it never calls `cancel_pending_requests()`, so the caller's future stays pending until the executor is destroyed, or until a later retry processes a stale request. Call `cancel_pending_requests()` in the catch, or reactivate the queue only after the pool and processing thread are up.

src/downgrade/downgrade_executor.cpp
```diff
@@ -112,6 +128,10 @@ void downgrade_executor::stop()
   bool expected = true;
   if (!_running.compare_exchange_strong(expected, false)) { return; }
 
+  if (!_pool) {
```
felipeblazing, line 131: **Reviewer: CUDA streams & device affinity**

**minor** GPU worker startup is fail-closed for affinity but still only logs when cudaSetDevice fails

The new startup hook rethrows affinity failures before the pool returns. `per_thread_init` runs in that same hook, but it is typed `noexcept`, so a failed `cudaSetDevice(device_id)` is only logged. The worker then runs its CUDA calls and stream acquisition against the wrong device, usually GPU 0. Since `start_thread_pool_workers` already propagates hook exceptions, have the device binding throw (for example, check it in the startup lambda) so a worker bound to the wrong GPU fails startup the same way a wrong CPU mask does.

src/include/exec/bounded_thread_pool.hpp
```diff
@@ -96,37 +98,20 @@ class bounded_thread_pool {
                                absl::AnyInvocable<void() noexcept> per_thread_init = nullptr)
     : capacity_(capacity)
   {
+    validated_cpu_affinity const affinity(cpu_ids);
```
felipeblazing, line 101: **Reviewer: Comments, docs & AI residue**

**minor** Pool constructors can now throw and block, with no doc comment

`bounded_thread_pool` and `static_thread_pool` now throw `std::invalid_argument` or `std::system_error` for a bad or unapplied `cpu_ids`, and they block until every worker's startup (affinity plus `per_thread_init`) has finished. Neither constructor has a doc comment or `@param` for `cpu_ids` or `per_thread_init`. Add a short contract comment: `cpu_ids` is one shared mask, empty means inherit, what it throws, and that the constructor blocks until all workers are initialized.

src/include/exec/cpu_affinity.hpp
```diff
@@ -0,0 +1,109 @@
...
+#include <string>
+#include <system_error>
+#include <vector>
+
+namespace sirius::exec {
+
+/**
+ * A validated CPU eligibility mask shared by every worker in a thread pool.
+ *
+ * Validation is intentionally performed on the constructing thread before any
```
felipeblazing, line 33: **Reviewer: Comments, docs & AI residue**

**minor** Class comment misstates why the constructing-thread mask is checked

`pthread_setaffinity_np` can widen a thread's mask beyond the one it inherited. The kernel silently drops only the CPUs outside the cpuset/cgroup, and the readback in `apply_to_current_thread()` would catch that case anyway. So the claim that a CPU outside the inherited mask "could otherwise be silently discarded" is inaccurate. Say instead that the constructing thread's mask is used as a conservative proxy for the cpuset, so bad IDs are rejected before any worker is spawned.

src/include/exec/cpu_affinity.hpp
```diff
@@ -0,0 +1,109 @@
...
+                                    " must be non-negative");
+      }
+      if (cpu_id >= CPU_SETSIZE) {
+        throw std::invalid_argument("CPU affinity ID " + std::to_string(cpu_id) +
+                                    " must be less than CPU_SETSIZE (" +
+                                    std::to_string(CPU_SETSIZE) + ")");
+      }
+      if (!CPU_ISSET(cpu_id, &allowed)) {
+        throw std::invalid_argument("CPU affinity ID " + std::to_string(cpu_id) +
+                                    " is not in the current process allowed CPU mask");
```
felipeblazing, line 63: **Reviewer: API design, layering & reuse**

**minor** Contract says both "process mask" and "constructing thread's mask"

`sched_getaffinity(0, ...)` returns the calling thread's mask. config.hpp says "constructing thread's allowed mask", but configuration.md and this error message say "process". A thread pinned narrower than the process will reject CPUs the process allows, and the message will point the user at the wrong mask. Pick one semantic, and name it the same way in the docs, the comment and the error.

src/include/exec/cpu_affinity.hpp
```diff
@@ -0,0 +1,109 @@
...
+      }
+      CPU_SET(cpu_id, &requested_);
+    }
+  }
+
+  [[nodiscard]] bool enabled() const noexcept { return enabled_; }
```
felipeblazing, line 69: **Reviewer: Error handling & fail-closed gating**

**minor** Error says 'process allowed CPU mask' but the check reads the calling thread's mask

`sched_getaffinity(0, ...)` returns the calling thread's mask. The docs and this message both promise the process mask. If a pool is started from a thread that is already pinned, a valid process CPU is rejected with a misleading message. Either read the process mask (`sched_getaffinity(getpid(), ...)` returns the main thread's mask) or change the wording and docs to say 'constructing thread'.

felipeblazing, line 69: **Reviewer: API design, layering & reuse**

**nit** enabled_ flag plus a zeroed mask is a sentinel; enabled() is unused

"Empty list means inherit" is stored as a bool next to a mask that is meaningless when the bool is false. `std::optional<cpu_set_t>` states that directly. The public `enabled()` accessor has no callers, so drop it until one needs it.

src/include/exec/cpu_affinity.hpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    cpu_set_t actual;
+    CPU_ZERO(&actual);
+    int const get_result = pthread_getaffinity_np(pthread_self(), sizeof(cpu_set_t), &actual);
+    if (get_result != 0) {
+      throw std::system_error(get_result, std::generic_category(), "pthread_getaffinity_np failed");
+    }
+
+    for (int cpu_id = 0; cpu_id < CPU_SETSIZE; ++cpu_id) {
```
felipeblazing, line 88: **Reviewer: API design, layering & reuse**

**minor** Hand-rolled mask comparison instead of CPU_EQUAL

glibc already provides `CPU_EQUAL` for exactly this 1024-iteration loop. The test file also redefines it as `masks_equal`.
```suggestion
    if (!CPU_EQUAL(&requested_, &actual)) {
      throw std::runtime_error("CPU affinity readback does not match the requested mask");
    }
```
(Replace the whole `for` loop with that.)

src/include/exec/cpu_affinity.hpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+ private:
+  cpu_set_t requested_{};
+  bool enabled_{false};
+};
+
+/** YAML-reader validator backed by the same checks used by the pool constructors. */
+struct valid_cpu_affinity {
```
felipeblazing, line 101: **Reviewer: API design, layering & reuse**

**minor** valid_cpu_affinity misuses the yaml validator contract

Every other `yaml::` validator is a `noexcept` bool predicate, and `reader::optional` turns `false` into "value out of range". This one never returns false and reports by throwing out of the predicate instead. If `reader::optional` needs to carry a reason, extend it (for example, a validator returning `std::optional<std::string>` error text) rather than smuggling the reason through an exception.

src/include/exec/cpu_affinity.hpp
```diff
@@ -0,0 +1,109 @@
...
+  bool operator()(const std::vector<int>& cpu_ids) const
```
felipeblazing, line 102: **Reviewer: API design, layering & reuse**

**major** Config parsing checks the live thread mask, and the pools repeat the check

`valid_cpu_affinity` calls `sched_getaffinity` on whatever thread loads the YAML. That thread is not the one that builds the pools, which re-validate against their own thread's mask anyway. The result is environment-dependent config loading, since the same YAML parses differently depending on the loading thread. It also runs the same check twice, at two layers that can disagree. Keep only the static checks (`>= 0`, `< CPU_SETSIZE`) in the config validator and leave the allowed-mask check to pool construction.

src/include/exec/thread_pool.hpp
```diff
@@ -40,48 +41,26 @@ class static_thread_pool {
...
   static_thread_pool& operator=(const static_thread_pool&) = delete;
 
-  ~static_thread_pool()
-  {
-    stop();
-    for (auto& t : threads_) {
-      if (t.joinable()) { t.join(); }
-    }
-  }
+  ~static_thread_pool() { stop_and_join(); }
```
felipeblazing, line 63: **Reviewer: API design, layering & reuse**

**nit** stop_and_join() has a single caller and adds nothing

The helper just moves the destructor body into a private function that only the destructor calls. The constructor's failure path uses `stop()` plus the helper's own join. Keep the original destructor body.

src/include/exec/thread_pool_startup.hpp
```diff
@@ -0,0 +1,95 @@
...
+#include <string>
+#include <thread>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace sirius::exec::detail {
+
+/**
+ * Start a pool's workers and wait until every worker has completed its startup hook.
```
felipeblazing, line 33: **Reviewer: Comments, docs & AI residue**

**minor** `start_thread_pool_workers` doesn't document what callers must provide

The doc covers the failure behaviour but not the requirements on the arguments. `threads` must start empty, because `join_workers` joins every thread in the vector. `request_stop` must be noexcept, safe to call more than once, and must make `work` return, or the join hangs. The call blocks until every startup hook has finished. Add `@param` lines covering these, since both pool constructors rely on them.

src/include/exec/thread_pool_startup.hpp
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * If thread creation or any startup hook fails, all workers are stopped and joined
+ * before the exception is propagated. The startup hook is therefore the safe place
+ * to apply and verify CPU affinity before the pool constructor returns.
+ */
+template <typename StartupFn, typename WorkFn, typename RequestStopFn>
```
felipeblazing, line 39: **Reviewer: API design, layering & reuse**

**minor** Unconstrained callable template parameters on start_thread_pool_workers

The pools' own APIs already constrain callables with `std::invocable` (for example `schedule(std::invocable auto&&)`). A bad hook here fails deep inside the thread lambda. Constrain the three parameters (for example `std::invocable StartupFn`, and require `std::is_nothrow_invocable_v<RequestStopFn>`, since it runs inside a `noexcept` lambda).

src/parallel/task_executor.cpp
```diff
@@ -67,19 +67,33 @@ void itask_executor::start()
...
-  on_start();
+  try {
+    _bounded_pool = std::make_unique<exec::bounded_thread_pool>(_config.num_threads,
+                                                                _config.thread_name_prefix,
+                                                                _config.cpu_affinity_list,
+                                                                get_per_thread_init());
+    _task_queue.reactivate();
+    _manager_thread = std::thread([this] { manager_loop(); });
+    on_start();
+  } catch (...) {
```
felipeblazing, line 78: **Reviewer: API design, layering & reuse**

**minor** itask_executor::start rollback is a third copy of stop()

Same pattern as task_creator and downgrade_executor: the catch repeats `stop()` without the `on_stop()`/`on_stopped()` hooks. If skipping the hooks is intended (`on_start()` may not have run), factor the teardown into a private `teardown_pool()` that both `stop()` and the catch call, rather than keeping two hand-maintained sequences.

src/parallel/task_executor.cpp
```diff
@@ -67,19 +67,33 @@ void itask_executor::start()
...
+    }
+    throw;
+  }
 }
 
 void itask_executor::stop()
 {
   bool expected = true;
   if (!_running.compare_exchange_strong(expected, false)) { return; }
+  if (!_bounded_pool) { return; }
```
felipeblazing, line 96: **Reviewer: Concurrency, lifetime & hangs**

**nit** The null-pool guard in stop() can only fire when start()/stop() race, and then it strands the pool

Because every rollback now resets `_running`, `_running==true && !_bounded_pool` can only happen while another thread is inside `start()`. There is no mutex here, so this is an unsynchronized read of `_bounded_pool`. In that window, `stop()` flips `_running` to false and returns. `start()` then finishes, leaving the pool and manager thread running, and the destructor's `stop()` becomes a no-op, so the joinable `_manager_thread` is destroyed and `std::terminate` runs. Remove the guard, or serialize start/stop with a mutex as task_creator does. The same applies to `downgrade_executor::stop()`.

src/pipeline/task_scheduler.cpp
```diff
@@ -119,12 +119,30 @@ void task_scheduler::schedule(std::unique_ptr<sirius::parallel::itask> task)
 
 void task_scheduler::start()
 {
+  std::vector<std::pair<int, gpu_pipeline_executor*>> executors;
```
felipeblazing, line 122: **Reviewer: API design, layering & reuse**

**minor** Sort and started-list exist only to make one test deterministic

Building and sorting a pair vector on every `start()` (even before the CAS no-op return) just fixes the start order for the 2-GPU test. `itask_executor::stop()` is already a no-op on an executor that isn't running, so the catch can simply stop every entry of `_gpu_executors` without tracking `started`. If a deterministic order really matters, make `_gpu_executors` an ordered `std::map` instead of re-sorting on each call.

src/pipeline/task_scheduler.cpp
```diff
@@ -119,12 +119,30 @@ void task_scheduler::schedule(std::unique_ptr<sirius::parallel::itask> task)
...
+  executors.reserve(_gpu_executors.size());
+  for (auto& [device_id, gpu_exec] : _gpu_executors) {
+    executors.emplace_back(device_id, gpu_exec.get());
+  }
+  std::ranges::sort(executors, {}, &std::pair<int, gpu_pipeline_executor*>::first);
```
felipeblazing, line 127: **Reviewer: Comments, docs & AI residue**

**nit** The device-ID sort in `task_scheduler::start()` is unexplained

The only explanation for sorting executors by device ID is in the test comment. A reader of `start()` sees a copy-and-sort before iterating an unordered map and has no reason to keep it. Add a one-line comment here and trim the test comment to what the test sets up.

```suggestion
  // Start in device-ID order so startup and partial-start rollback are deterministic.
  std::ranges::sort(executors, {}, &std::pair<int, gpu_pipeline_executor*>::first);
```

src/pipeline/task_scheduler.cpp
```diff
@@ -119,12 +119,30 @@ void task_scheduler::schedule(std::unique_ptr<sirius::parallel::itask> task)
...
+
+  std::vector<gpu_pipeline_executor*> started;
+  started.reserve(executors.size());
   bool expected = false;
   if (!_running.compare_exchange_strong(expected, true)) { return; }
-  for (auto& [device_id, gpu_exec] : _gpu_executors) {
-    gpu_exec->start();
+  try {
+    for (auto const& executor : executors) {
+      executor.second->start();
```
felipeblazing, line 135: **Reviewer: CUDA streams & device affinity**

**major** Fail-closed validation now rejects topology-derived GPU pipeline masks under any restricted cpuset

`gpu_pipeline_executor` gets its `cpu_affinity_list` from `gpu.cpu_cores`, and cucascade reads that from sysfs `local_cpulist`, so the cgroup/cpuset is not applied. Before this PR the kernel quietly intersected the mask. Now `validated_cpu_affinity` throws when any listed core is outside `sched_getaffinity`. Example: a 2-socket box running Sirius under `taskset -c 0-15`, Docker `--cpuset-cpus`, or a Slurm allocation where the GPU's NUMA node lists 0-31. `task_scheduler::start()` then throws and no GPU work can run. Intersect the topology list with the process's allowed mask in `task_scheduler`, and fall back to inheriting the mask if the intersection is empty. Keep the strict check for user-supplied YAML only.

felipeblazing, line 135: **Reviewer: Correctness**

**blocking** Topology-derived GPU affinity isn't intersected with the process mask, so Sirius fails to start under cpuset/taskset

`config.cpu_affinity_list = it->cpu_cores` (task_scheduler.cpp:78) is the raw sysfs `local_cpulist`, i.e. every NUMA-local CPU. It is not filtered by `sched_getaffinity`. `validated_cpu_affinity` now throws for any CPU outside the caller's mask. For example, `docker run --cpuset-cpus=0-15` or `taskset -c 0-15` on a host where GPU0 is local to CPUs 0-31 makes `task_scheduler_->start()` throw, and `SiriusContext::initialize` fails. Before this change the kernel intersected the mask and startup succeeded. Intersect the topology cores with the current allowed mask when building the per-GPU config (and inherit if the intersection is empty). Keep fail-closed only for explicit user-supplied lists, and add a test with a topology core outside the mask.

felipeblazing, line 135: **Reviewer: Error handling & fail-closed gating**

**major** Topology-derived GPU masks now fail Sirius init under any cgroup/cpuset restriction

The GPU executor's `cpu_affinity_list` is copied straight from cuCascade's `cpu_cores`, which is the sysfs `local_cpulist` for the whole NUMA node and is never intersected with the process's allowed CPUs. Before this PR the kernel silently narrowed the mask. Now `validated_cpu_affinity` throws, `task_scheduler::start()` throws, and `SiriusContext::initialize` fails. That happens under `docker --cpuset-cpus`, SLURM, k8s CPU pinning or `taskset`, and the user has no config key to fix it. Intersect the derived list with `sched_getaffinity` in the task_scheduler constructor, and throw only if the intersection is empty. Fail-closed validation should stay for user-supplied lists.

felipeblazing, line 135: **Reviewer: API design, layering & reuse**

**major** Topology-derived GPU pipeline affinity now fails closed instead of being clamped to the allowed mask

`task_scheduler` copies `gpu_topology_info::cpu_cores` into `cpu_affinity_list` (task_scheduler.cpp:78). Those cores now go through the same strict "must be inside the caller's mask" check. In a container or cpuset that exposes only some of a GPU's NUMA-local cores, `task_scheduler::start()` would throw, even though the user never set any affinity. The code that derives the mask should intersect it with the allowed set, falling back to inheritance when the result is empty. Strict rejection should apply only to user-supplied lists.

src/pipeline/task_scheduler.cpp
```diff
@@ -119,12 +119,30 @@ void task_scheduler::schedule(std::unique_ptr<sirius::parallel::itask> task)
...
+      started.push_back(executor.second);
+    }
+    _management_thread = std::thread(&task_scheduler::management_eventloop, this);
+  } catch (...) {
+    _running = false;
+    for (auto* gpu_exec : started) {
+      gpu_exec->stop();
```
felipeblazing, line 142: **Reviewer: Concurrency, lifetime & hangs**

**minor** Partial-start rollback leaves stale device_ready events in the scheduler channel

Device 0's manager_loop posts `device_ready` to `_task_request_channel` as soon as it starts. Rolling it back with `stop()` neither closes nor drains the channel. `_ready_devices` is a vector, so if a later `start()` succeeds (for example when the first failure was transient: EAGAIN, or a cpuset change), device 0 is listed twice. The management loop then dispatches two tasks into its executor queue, where the downgrade path can't see them, which is the invariant the pull model exists for. Drain the channel's pending events in the rollback, or dedupe `_ready_devices`.

test/cpp/config/test_cpu_affinity_config.cpp
```diff
@@ -0,0 +1,174 @@
...
+    std::filesystem::remove(path, ec);
+  }
+
+  scoped_yaml(scoped_yaml const&)            = delete;
+  scoped_yaml& operator=(scoped_yaml const&) = delete;
+
+  std::filesystem::path path;
+};
+
+int first_allowed_cpu()
```
felipeblazing, line 50: **Reviewer: API design, layering & reuse**

**nit** cpu_set_t test helpers copy-pasted across three test files

`first_allowed_cpu`, `first_disallowed_cpu`, `current_cpu_mask`, `cpu_ids_in` and `masks_equal` are each defined locally, and test_task_scheduler.cpp inlines the same loop again. Move them into a shared header under `test/cpp/utils/` (next to `utils.hpp`) and reuse them.

test/cpp/pipeline/test_task_scheduler.cpp
```diff
@@ -135,6 +137,53 @@ TEST_CASE("Task scheduler derives GPU executor affinity from topology", "[task_s
   REQUIRE(visited == 1);
 }
 
+TEST_CASE("Task scheduler rolls back topology-affinity startup failure",
+          "[task_scheduler][cpu_affinity]")
+{
+  // This regression needs two real executors: device 0 must start before the
+  // invalid topology mask fails device 1, so task_scheduler has something to
+  // roll back. start() orders lifecycle startup by device ID without changing
+  // the scheduler's unordered ready-device matcher. Do not construct a
```
felipeblazing, line 146: **Reviewer: Comments, docs & AI residue**

**nit** Test comment describes the implementation, not the test

"start() orders lifecycle startup by device ID without changing the scheduler's unordered ready-device matcher" describes how the production code was built, which the test reader doesn't need. Keep the setup rationale (device 0 must start before device 1 fails) and the 1-GPU caveat, and move the ordering explanation to `task_scheduler::start()`.

PR #1517, opened 2026-08-13
fix(config): reject negative log flush intervals

infblueocean:
## Summary

- reject negative `sirius_log_flush_seconds` before mutating process-global state or rebuilding the log sink
- preserve and document `0` as the explicit periodic-flush opt-out
- prove failed input preserves both DuckDB-visible and effective process-global state

## Why

The spdlog construction path treats every value at or below zero as disabled. Clean Sirius therefore accepts `-1` as a distinct visible setting while applying the documented-zero behavior. Rejecting negatives makes the setting literal without selecting a different default.

## Validation

- focused isolated-context regression covers positive setup, negative refusal/nonmutation, and zero acceptance
- `python3 scripts/check_orphan_tests.py`
- `git diff --check`
- live duplicate search: no open issue or PR for this setting
- independent merge-tree with current open PR #1516 is clean (`51b23b0c44cfea4349eb095d1f1f41697b555b13`)
- exact-head lint, GCC release, Clang debug, CUDA 13 build, full runtime, and terminal aggregate checks pass
- hosted workflow [31701205855](https://github.com/sirius-db/sirius/actions/runs/31701205855): runtime job `94451496486` passed in 19m24s; aggregate job `94456762457` passed

Exact base: `0d7f98d9a662e9e3b47d6f7d6e2c8797d1e6572b`
Candidate: `eff5220986b014cc117116d3a97aa0b058bfd369`
Candidate tree: `7ca08297201f34257df291cbf66e35fc8798ab7b`

PR #1516, opened 2026-08-13
fix(config): validate logging controls

infblueocean:
## Summary

- reject unknown `sirius_log_level` and `SIRIUS_LOG_LEVEL` values before mutation
- document and accept the full supported level set, including `critical` and `off`
- reject negative `sirius_log_flush_seconds` while preserving `0` as the explicit periodic-flush opt-out
- consolidate the adjacent #1517 flush validation into one logging-controls review unit

## Contract

Runtime and startup log-level validation accept only `trace`, `debug`, `info`, `warn`, `error`, `critical`, or `off`. Rejected values leave the visible setting, process-global value, and active sink behavior unchanged. Flush intervals must be non-negative; `0` disables periodic flushing.

## Identity

- base: `e64b8ad79ea3cee50149d7e3039dbbb92ed291a9`
- head: `a3dfe9e218c816b834de81e75579d08e629a6eea`
- tree: `bc31566b7cd7dc89e5299c31898fdc48ef1db7cd`
- zero-context source-diff SHA-256: `01d7ead8d4a9592cd6646ff7f8cd10e97f244e88a20dbe9f1e23ad0180356665`

## Validation

- exact current-base four-file diff; no unrelated source changes
- `python3 scripts/check_orphan_tests.py` passes
- `git diff --check` passes
- focused runtime/startup and nonmutation regressions included
- hosted lint, GCC, Clang, CUDA 13 build/runtime are the compiled acceptance path

Supersedes #1517.

- exact-head hosted receipt for `a3dfe9e218c816b834de81e75579d08e629a6eea`: CUDA 13 runtime and terminal aggregate checks passed
- hosted workflow 31726952876: runtime job 94540138947 passed from 2026-08-13T18:10:42Z through 2026-08-13T18:31:05Z; aggregate job 94551660464 passed at 2026-08-13T18:31:12Z

PR #1519, opened 2026-08-13
fix(config): validate and preserve logging state

infblueocean:
## What changes for users

When Sirius rebuilds an spdlog sink after a logging-setting change, a failed
rebuild must not leave the reported configuration describing a sink that never
became active. This PR now restores the prior flush interval when that rebuild
fails, matching its existing rollback behavior for backend and directory
changes.

Users still choose the supported logging backend, directory, level, and flush
interval. Sirius owns validation and the transactional sink replacement. A
focused regression forces the deterministic invalid-directory failure and
checks that both the previous current_setting value and the exact active sink
remain unchanged. The evidence boundary is that deterministic construction
failure; this does not claim recovery from arbitrary external filesystem
failures after a sink is installed.

## Summary

- reject unsupported startup and runtime logging backend/level values before mutation
- preserve the active backend, directory, sink, and startup logging tuple when replacement sink construction fails
- restore LOG_FLUSH_SECONDS when spdlog sink reconstruction fails
- reject negative sirius_log_flush_seconds while preserving 0 as the explicit periodic-flush opt-out
- consolidate the adjacent logging validation formerly carried by #1516 into the rollback transaction already reviewed here

## Contract

Runtime backend changes accept only duckdb, spdlog, or noop; runtime level
changes accept only trace, debug, info, warn, error, critical, or off. Rejected
names leave visible and effective logging state unchanged. Runtime directory,
backend, and flush-interval transitions restore their prior process-global
value when sink construction fails, and startup restores the complete
backend/directory/level tuple. Flush intervals must be non-negative; 0 disables
periodic flushing.

## Why one PR

Validation and rollback touch the same startup tuple, setters, and
isolated-context regressions. Keeping them together makes the invariant
reviewable in one place: validate before mutation, then restore the prior
coupled state if the external sink transition fails. This supersedes #1516;
#1517 and #1520 were already consolidated into those two predecessor heads.

## Current public identity

- prior reviewed public head: 6afa3b389ba4309fb30f5dd281fc65b0ff9e2783
- replacement head: 2373ed62af7324d7744682b20ebbe481f9715ec2
- candidate tree: c81c11e0851bbc75a335edfa0a986c005b954b9a
- three-path review-repair diff SHA-256:
  67a00f054e528488863a4adf60e469336f268d8d1feaeca9e4dc1ba7d5cda1a8
- current Sirius dev observed during repair:
  de71900686ddcfab66812c3f29c12f848f7c8f2e; GitHub reports the replacement
  head mergeable

## Review repair

- restores LOG_FLUSH_SECONDS on any install_configured_log_sink throw and rethrows
- adds a focused invalid-directory regression that verifies the prior
  DuckDB-visible value, process-global value, and exact sink identity
- changes the three environment pointers to auto const*
- uses non-throwing std::filesystem::remove_all overloads in the affected
  failure-test teardown paths

## Validation

- complete replacement diff independently cold-reviewed; no submodule or unrelated-source drift
- targeted repository pre-commit hooks pass, including clang-format, codespell,
  and orphan-test validation
- explicit scripts/check_orphan_tests.py, git diff --check, and git show --check pass
- native C++ execution was unavailable on the Apple-arm64 author host without
  the Sirius Pixi/CUDA environment
- replacement exact-head hosted CI passes: Check 32994600965, Test
  32994601203, Distribution 32994602370, and Validate 32994766298

### Retained predecessor hosted receipts

- rollback head 19e58e4e6d57b93d1ec2e7a2ab0f75f180052c2c: workflow
  31726205775; CUDA 13 runtime job 94536097118 and terminal aggregate job
  94545844935 passed
- validation head a3dfe9e218c816b834de81e75579d08e629a6eea: workflow
  31726952876; CUDA 13 runtime job 94540138947 and terminal aggregate job
  94551660464 passed

### Retained failed-head repair receipt

- first consolidated head 8429111bfa225f5d9ae17b088b4b9ee35d5f6c15 omitted
  the validation predecessor's utils/log_test_utils.hpp include while retaining
  its scoped_recording_log_sink regression
- exact-head workflows 31732638303 and 31732638391 failed compilation at
  test/cpp/config/test_context.cpp:350
- repaired head deb2902807c19c8bbabde9f8b87c4073819116a8: workflow
  31733223327; runtime job 94559525138 and aggregate job 94566061275 passed
- the failed predecessor remains separately retained and receives no terminal credit

### Retained pre-review current-dev receipt

- head 6afa3b389ba4309fb30f5dd281fc65b0ff9e2783: hosted lint, GCC, Clang,
  CUDA 13 build/runtime, TPC-H snapshot, and terminal aggregate passed
- Check 32350078769, Test 32350078679, Distribution 32350079434

### DIFF
--- a/src/sirius_context.cpp
+++ b/src/sirius_context.cpp
@@ -1586,13 +1586,38 @@ void install_configured_log_sink(DatabaseInstance* db)
 
 SiriusContextExtensionCallback::SiriusContextExtensionCallback()
 {
-  if (auto* env = std::getenv("SIRIUS_LOG_BACKEND")) { Config::LOG_BACKEND = env; }
-  if (auto* env = std::getenv("SIRIUS_LOG_DIR")) { Config::LOG_DIR = env; }
-  if (auto* env = std::getenv("SIRIUS_LOG_LEVEL")) { Config::LOG_LEVEL = env; }
+  auto const previous_log_backend = Config::LOG_BACKEND;
+  auto const previous_log_dir     = Config::LOG_DIR;
+  auto const previous_log_level   = Config::LOG_LEVEL;
+  auto* backend_env               = std::getenv("SIRIUS_LOG_BACKEND");
+  auto* log_dir_env               = std::getenv("SIRIUS_LOG_DIR");
+  auto* level_env                 = std::getenv("SIRIUS_LOG_LEVEL");
+  if (backend_env) {
+    std::string_view const backend{backend_env};
+    if (backend != "duckdb" && backend != "spdlog" && backend != "noop") {
+      throw InvalidInputException(
+        "SIRIUS_LOG_BACKEND must be one of: duckdb, spdlog, noop; got '%s'", backend_env);
+    }
+  }
+  if (level_env && !sirius::log::string_to_enum(level_env)) {
+    throw InvalidInputException(
+      "SIRIUS_LOG_LEVEL must be one of: trace, debug, info, warn, error, critical, off; got '%s'",
+      level_env);
+  }
+  if (backend_env) { Config::LOG_BACKEND = backend_env; }
+  if (log_dir_env) { Config::LOG_DIR = log_dir_env; }
+  if (level_env) { Config::LOG_LEVEL = level_env; }
   // Install now (no db yet) so spdlog/noop capture the logs emitted by
   // read_config_file_if_exists() below; the duckdb backend needs a db (installed
   // later).
-  install_configured_log_sink(nullptr);
+  try {
+    install_configured_log_sink(nullptr);
+  } catch (...) {
+    Config::LOG_BACKEND = previous_log_backend;
+    Config::LOG_DIR     = previous_log_dir;
+    Config::LOG_LEVEL   = previous_log_level;
+    throw;
+  }
   read_config_file_if_exists();
 }
 
--- a/src/sirius_context.cpp
+++ b/src/sirius_context.cpp
@@ -1589,9 +1589,9 @@ SiriusContextExtensionCallback::SiriusContextExtensionCallback()
   auto const previous_log_backend = Config::LOG_BACKEND;
   auto const previous_log_dir     = Config::LOG_DIR;
   auto const previous_log_level   = Config::LOG_LEVEL;
-  auto* backend_env               = std::getenv("SIRIUS_LOG_BACKEND");
-  auto* log_dir_env               = std::getenv("SIRIUS_LOG_DIR");
-  auto* level_env                 = std::getenv("SIRIUS_LOG_LEVEL");
+  auto const* backend_env         = std::getenv("SIRIUS_LOG_BACKEND");
+  auto const* log_dir_env         = std::getenv("SIRIUS_LOG_DIR");
+  auto const* level_env           = std::getenv("SIRIUS_LOG_LEVEL");
   if (backend_env) {
     std::string_view const backend{backend_env};
     if (backend != "duckdb" && backend != "spdlog" && backend != "noop") {
--- a/src/sirius_extension.cpp
+++ b/src/sirius_extension.cpp
@@ -1955,37 +1955,58 @@ static void SetLogBackend(ClientContext& context, SetScope scope, Value& paramet
     throw InvalidInputException("Unknown sirius_log_backend '%s' (expected: duckdb, spdlog, noop)",
                                 backend);
   }
-  Config::LOG_BACKEND = std::move(backend);
-  install_configured_log_sink(context.db.get());
+  auto const previous_backend = Config::LOG_BACKEND;
+  Config::LOG_BACKEND         = std::move(backend);
+  try {
+    install_configured_log_sink(context.db.get());
+  } catch (...) {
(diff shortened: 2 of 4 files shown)


infblueocean:
Ready for maintainer review: exact head `f5d439f5fdc30bbb940b6bab701705d95a908e6e` is terminal green across lint, GCC, Clang, CUDA build, and full hosted runtime. The regression proves the deterministic invalid-directory construction failure preserves DuckDB-visible `sirius_log_dir`, process-global config, and the exact active Sirius sink object.

infblueocean:
Repair pushed at deb2902807c19c8bbabde9f8b87c4073819116a8. The failed consolidated head 8429111bfa225f5d9ae17b088b4b9ee35d5f6c15 retained #1516’s recording-sink regression but omitted its `utils/log_test_utils.hpp` include during conflict resolution; workflows 31732638303 and 31732638391 correctly failed compilation at test_context.cpp:350. The current head restores that exact predecessor include. Full pre-commit, standalone rumdl, and diff-check pass; new exact-head hosted CI is live. No production semantics changed in the repair.

infblueocean:
Current-dev rebased-head validation is terminal green on `c8c68647abb1e5cb88bc4e21b9b130d9fabe91dc`: hosted lint, GCC, Clang, CUDA 13 build/runtime, and TPC-H snapshot all passed (workflow `31758550110`; runtime job `94640436327`, 20m04s; aggregate `94643728815`). The branch is current with `dev`; maintainer review is the remaining gate.

src/sirius_extension.cpp
```diff
@@ -1955,37 +1955,58 @@ static void SetLogBackend(ClientContext& context, SetScope scope, Value& paramet
...
+  auto const seconds = IntegerValue::Get(parameter);
+  if (seconds < 0) {
+    throw InvalidInputException(
+      "sirius_log_flush_seconds must be non-negative; zero disables periodic flushing; got %d",
+      seconds);
+  }
+  Config::LOG_FLUSH_SECONDS = seconds;
   // The flush interval is fixed at spdlog-sink construction, so rebuild it (only
   // the spdlog backend uses it).
   if (Config::LOG_BACKEND == "spdlog") { install_configured_log_sink(context.db.get()); }
```
johanpel: Shouldn't we also roll back here if `install_configured_log_sink` throws?

src/sirius_context.cpp
```diff
@@ -1586,13 +1586,38 @@ void install_configured_log_sink(DatabaseInstance* db)
...
 {
-  if (auto* env = std::getenv("SIRIUS_LOG_BACKEND")) { Config::LOG_BACKEND = env; }
-  if (auto* env = std::getenv("SIRIUS_LOG_DIR")) { Config::LOG_DIR = env; }
-  if (auto* env = std::getenv("SIRIUS_LOG_LEVEL")) { Config::LOG_LEVEL = env; }
+  auto const previous_log_backend = Config::LOG_BACKEND;
+  auto const previous_log_dir     = Config::LOG_DIR;
+  auto const previous_log_level   = Config::LOG_LEVEL;
+  auto* backend_env               = std::getenv("SIRIUS_LOG_BACKEND");
+  auto* log_dir_env               = std::getenv("SIRIUS_LOG_DIR");
+  auto* level_env                 = std::getenv("SIRIUS_LOG_LEVEL");
```
johanpel: Nit: make these `auto const*`

test/cpp/config/test_context.cpp
```diff
@@ -1234,6 +1423,168 @@ TEST_CASE("YAML-backed operator and compression settings are DuckDB defaults",
...
+
+  auto const previous_backend = duckdb::Config::LOG_BACKEND;
+  auto const previous_log_dir = duckdb::Config::LOG_DIR;
+  auto const previous_sink    = sirius::log::get_sink();
+  auto const test_root        = fs::temp_directory_path() / "sirius-log-dir-rollback-test";
+  finally restore_logging{[&]() {
+    duckdb::Config::LOG_BACKEND = previous_backend;
+    duckdb::Config::LOG_DIR     = previous_log_dir;
+    sirius::log::set_sink(previous_sink);
+    fs::remove_all(test_root);
```
johanpel: Nit: we could use the non-throwing overloads here

johanpel (commented):
LGTM overall, I think this needs one fix:

src/sirius_extension.cpp
infblueocean: Good catch. I now restore `LOG_FLUSH_SECONDS` if rebuilding the spdlog sink throws. The focused deterministic invalid-directory regression also verifies that both the previous reported interval and the exact active sink are preserved.

src/sirius_context.cpp
infblueocean: Done — the three environment pointers are now `auto const*`.

test/cpp/config/test_context.cpp
infblueocean: Done — the affected failure-test teardown paths now use the non-throwing `remove_all` overload so teardown cannot throw.

johanpel (approved):

mbrobbel (approved):

PR #1520, opened 2026-08-13
fix(config): preserve log backend on sink errors

infblueocean:
## Summary

- restore the prior process-global `sirius_log_backend` when installing its replacement sink fails
- preserve DuckDB-visible and process-global state on the deterministic invalid-directory construction failure
- prove the failed transition remains recoverable after correcting the staged log directory
- reject unsupported startup `SIRIUS_LOG_BACKEND` values before process-global mutation

## Why

While the `noop` backend is active, `sirius_log_dir` may legitimately stage a path without constructing a file sink. Switching to `spdlog` assigns `Config::LOG_BACKEND` before trying that path. If construction rejects it, DuckDB reports the backend `SET` as failed and keeps the visible value at `noop`, but Sirius currently retains `spdlog` internally. At startup, an unknown environment value is also assigned and only rejected later when a DatabaseInstance exists, poisoning process-global state across a retry.

## Validation

- focused isolated-context regression stages an invalid directory under `noop`, requires the `spdlog` transition to fail, and proves visible/effective backend nonmutation
- the same regression corrects the directory and proves the subsequent `spdlog` transition succeeds
- startup constructor regression proves exact unknown-environment failure/nonmutation and valid `duckdb` application
- `python3 scripts/check_orphan_tests.py`
- `git diff --check`
- clean independent merge trees with #1517 (`01dff114f8c0eb4c015f32baf871116d19de1c99`), #1519 (`9d740aebcb0d36d41d492a9d811e42202d66936d`), and #1522 (`4bf74ed20194fcfad4cda6a0774f5b49121e457b`)
- #1516 now validates the neighboring startup log-level environment value in the same constructor and adds independent tests in the same file; the combined resolution is the union of the two validation blocks and test cases
- fresh exact-head hosted lint, GCC release, Clang debug, CUDA 13 build, distribution, check, PR-title, full runtime, and terminal aggregate checks pass
- test workflow `31704934195` passed at the exact candidate; runtime job `94464053089` ran from `2026-08-13T14:56:41Z` through `2026-08-13T15:16:04Z` (19m23s), including C++ unit tests and the TPC-H snapshot; terminal aggregate job `94495565907` passed

This claim is scoped to the concrete invalid-directory sink-construction path; it does not claim every hypothetical exception in spdlog's global setup is strongly transactional.

Exact base: `0d7f98d9a662e9e3b47d6f7d6e2c8797d1e6572b`
Candidate: `c9ed92d4c6427ba817741da0394213091de13c9c`
Candidate tree: `4bae1faf666f0c129477f8973669883677c94a4f`
Zero-context source patch SHA-256: `cdbba82b50612532923d6acf61e7fd1ec783ba2540fcedb6f804d58249972066`

### DIFF
--- a/src/sirius_context.cpp
+++ b/src/sirius_context.cpp
@@ -1586,7 +1586,14 @@ void install_configured_log_sink(DatabaseInstance* db)
 
 SiriusContextExtensionCallback::SiriusContextExtensionCallback()
 {
-  if (auto* env = std::getenv("SIRIUS_LOG_BACKEND")) { Config::LOG_BACKEND = env; }
+  if (auto* env = std::getenv("SIRIUS_LOG_BACKEND")) {
+    std::string_view const backend{env};
+    if (backend != "duckdb" && backend != "spdlog" && backend != "noop") {
+      throw InvalidInputException(
+        "SIRIUS_LOG_BACKEND must be one of: duckdb, spdlog, noop; got '%s'", env);
+    }
+    Config::LOG_BACKEND = env;
+  }
   if (auto* env = std::getenv("SIRIUS_LOG_DIR")) { Config::LOG_DIR = env; }
   if (auto* env = std::getenv("SIRIUS_LOG_LEVEL")) { Config::LOG_LEVEL = env; }
   // Install now (no db yet) so spdlog/noop capture the logs emitted by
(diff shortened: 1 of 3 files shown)


infblueocean:
Superseded by #1519, which now carries this backend contract together with the same logging transaction’s directory and startup rollback, rebuilt on exact current dev. The composed tree is cca55476012b360d8f29664d22b57ba25d02ef3e.

PR #1523, opened 2026-08-13
refactor(config): keep automatic choices internal

infblueocean:
## Review focus

**Tier:** judgment — this removes ordinary user control over three internal
execution alternatives while preserving exact test and advanced benchmark seams.

**Maintainer question:** Should merge fusion, runtime distinct-build probing,
and derived CONCAT sizing remain engine policy, with ordinary sessions unable
to discover or toggle their internal reference paths?

**Main risk:** internalization can remove a useful operational escape hatch.
The exact test-option gate and explicit advanced CONCAT YAML override remain so
the implementations stay directly testable without becoming ordinary choices.

## Summary

- keep merge-pipeline fusion enabled as engine-owned production policy
- keep runtime distinct-build probing as engine-owned join policy, currently disabled pending #1600
- keep the CONCAT batch target derived from effective GPU capacity
- remove all three from the ordinary DuckDB session-settings surface
- expose the three reference paths only when `SIRIUS_ENABLE_TEST_OPTIONS=1`
- retain the explicit CONCAT YAML override for advanced benchmark envelopes
- reject the removed YAML distinct-build key with deletion guidance

## Why one PR

All three settings expose internal execution alternatives without an ordinary user-facing selection rule. Their normal paths are already engine policy, and the engine has the plan or hardware information needed to choose them. They share the same registration boundary and exact test-option guard. The union proves once that normal sessions cannot discover or mutate these choices while all three reference paths remain directly testable.

Runtime distinct-build probing is temporarily off because the cuCollections defect tracked in #1600 can affect some key distributions. That dependency changes the current internal policy value, not the configuration-surface decision: users should not need to discover and toggle an unsafe implementation path.

This supersedes #1526 and #1530. Their YAML field-path, planner propagation, session visibility, `SET`/`RESET`, and exact-opt-in checks are retained.

## Contract

- normal processes, including `SIRIUS_ENABLE_TEST_OPTIONS=true`, cannot discover or set any of the three DuckDB options
- exact `SIRIUS_ENABLE_TEST_OPTIONS=1` exposes all three options for tests
- the removed YAML distinct-build key fails with full-path deletion guidance
- runtime distinct-build probing retains current-dev's default-off policy while #1600 is unresolved
- advanced benchmark fixtures may still override `concat_batch_bytes` in YAML
- merge fusion remains automatic-on; automatic defaults and all inactive reference implementations remain intact

## Current identity

- exact base: `cbd9516367b2d3160492c8cb0575208f1dce047c`
- exact head: `8acbb29bdbf075590bd04aac9e00967f613bbd83`
- candidate tree: `01d3b8382b0d48851ff40c18513c13cc36c9cc0c`
- full-index binary diff SHA-256: `bbe7940158b4b3d1c0c1b9360292ca80a8d3c2b3f5b724733acad49a0fee985c`
- one-commit rebase onto current `dev`; conflicts were limited to current configuration documentation, ordinary-setting registration, and configuration-test expectations

## Validation

- conflict resolution preserves current-dev GPU-admission and dynamic-filter coverage while removing the three ordinary choices
- current-dev's #1600 default-off change is retained in source, documentation, YAML-refusal behavior, and test expectations
- exact test-option coverage now toggles runtime distinct-build probing to `true`, exercising the inactive reference path rather than restating its false default
- full resolved-diff review caught and repaired one stale pre-#1600 `RESET` expectation; the test now requires the current default to remain `false`
- `git diff --check`: pass
- local conflict-marker scan across all changed source, docs, and tests: pass

- exact-head hosted Test run `32193799797`: CUDA 13 build passed; runtime job
  `95894462143` passed in 19m56s, including C++ unit tests and the TPC-H
  snapshot; terminal aggregate job `95898848048` passed
- exact-head hosted Check run `32193799872`: lint, GCC release, Clang debug,
  and terminal aggregate passed
- exact-head hosted Distribution run `32193800262`: terminal aggregate passed

Final author-readiness review at exact head `7bb37ffa2bb7` found the one-commit
unit coherent, the #1600 conflict adaptation consistent across source, docs,
and tests, the #1595 dependency explicit, and no remaining author-side repair.
The engine-policy question above is ready for maintainer review rather than an
unfinished-author-work draft condition.

## Neighboring work

- #1588 adds a separate twin-scan fusion mechanism and user-facing `fuse_twin_scans` choice. It is not silently folded into this PR: that is a separate product judgment. The branches currently conflict in shared planner/config files, so whichever lands second must rebase and preserve both decisions explicitly.
- #1595 changes BUILD_PROBE admission and currently assumes `enable_runtime_distinct_build_probe` remains an ordinary setting. The branches also conflict. This PR preserves the internalization contract; integration must adapt #1595's tests/docs to the internal field or guarded test seam rather than re-expose the user knob.

## August 19 repair receipt

Maintainer feedback on predecessor head `7bb37ffa2bb7` is addressed in rebased exact head `8acbb29bdbf075590bd04aac9e00967f613bbd83`. Internal option registration is centralized at the sole consumer; scan and CONCAT retain only their named YAML benchmark envelopes; obsolete runtime-distinct YAML fails noisily; and stale post-throw and setting-readback tests are removed. Pinned pre-commit, rumdl, the exact DuckDB AddExtensionOption compile probe, diff-check, formatting, YAML, and orphan-test validation pass. Hosted exact-head Test run `32295969393` passed, including GPU runtime job `96208613358`; Check, Distribution, lint, GCC, Clang, CUDA build, title validation, and every other non-skipped job also passed. Foxglove reviewed the complete rebased diff and resolved all maintainer threads before returning the PR to ready.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -407,7 +407,6 @@ individually.
 | `max_broadcast_join_size` | 256 MiB | Max build-side size eligible for a broadcast join. A build below this size is replicated to every GPU (instead of hash-partitioned) when it is tiny, or when the DuckDB-estimated probe-to-build row ratio is at least `num_gpus * 1.25`. |
 | `max_sort_partition_memory_fraction` | 0.33 | Fraction of GPU memory per sort partition when `max_sort_partition_bytes` is 0 |
 | `mark_join_build_switch_ratio` | 8.0 | For STANDARD MARK joins, build on the smaller (left) side when `right_rows >= ratio * left_rows` (0 disables) |
-| `enable_runtime_distinct_build_probe` | false | For `BUILD_PROBE` INNER/LEFT equality joins whose build-key uniqueness the planner could not prove, test distinctness at runtime (one `cudf::distinct_count` pass over the cached build, dimension-scale builds only) and take the single-pass `cudf::distinct_hash_join` instead of the general two-pass join when the keys are distinct. Temporarily off by default until a cuCollections bug fix ships in libcudf (issue #1600). |
 | `enable_dynamic_filter_pushdown` | true | Master switch for dynamic table-filter pushdown. An eligible `BUILD_PROBE` hash-join build selects a raw exact IN-list for 1–12 supported build rows, otherwise a hash IN-list if it fits the smallest probe-GPU L2 or a Bloom, for post-decode application by the probe scan. |
 | `enable_dynamic_zone_map_filter` | false | Additionally publish build-key min/max bounds. Parquet scans use them for read-time row-group pruning; duckdb-native scans apply them row-wise post-decode. Requires `enable_dynamic_filter_pushdown`; intended for clustered-keyset workloads. |
 | `dynamic_filter_domain_coverage_threshold` | 0.9 | Positive finite threshold for skipping publication when the build covers at least this fraction of the key's domain; ≥ 1.0 effectively disables the gate. |
@@ -420,6 +419,11 @@ individually.
 **Note:** `admission_bytes_per_gpu` is a parallelism dial, not a memory budget. Peak GPU residency is bounded by partition sizing (`hash_partition_bytes` and the batch settings), not by the admitted GPU count — a query on fewer GPUs processes more partitions sequentially at roughly unchanged peak memory, trading wall-clock for freed devices. Tune it against how much of the fleet a query should occupy, not against VRAM.
 
 **Note:** `max_build_hash_table_bytes` can be larger than `concat_batch_bytes`. When it is, the partition operator configures CONCAT to concatenate all batches, enabling the more efficient BUILD_PROBE join mode for larger build sides. Other joins (STANDARD, MIXED) still use `concat_batch_bytes` as the batch size threshold.
+
+Runtime distinct-build probing is an internal join policy, not a user configuration choice. It is
+temporarily disabled while the cuCollections defect tracked in #1600 remains unresolved. The
+engine retains the guarded single-pass `cudf::distinct_hash_join` path for policy-controlled use
+after that dependency is fixed.
 
 ## Telemetry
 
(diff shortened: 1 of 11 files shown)


infblueocean:
Rebased-head validation is terminal green on `82197fe713fa1a461f9209161b15b95154f96c43`: hosted lint, GCC, Clang, CUDA 13 build/runtime, and TPC-H snapshot all passed (workflow `31755962317`; runtime job `94632324275`, 19m32s). The branch is current with `dev`; maintainer review is the remaining gate.

docs/super-sirius/configuration.md
9prady9, line 1: When you say internal ? are these spread across different files ? If yes, that is perhaps going to cause a different kind of problem where it is hard to see how these are related and cross related.

If we can group them appropriately in a single internal header, perhaps, wouldn't that be better ?

test/cpp/config/test_context.cpp
```diff
@@ -281,9 +298,34 @@ TEST_CASE("Test-only settings require explicit process opt-in",
...
+{
+  std::source_location loc = std::source_location::current();
+  auto const data_dir      = fs::path(loc.file_name()).parent_path() / "data";
+
+  sirius::sirius_config config;
+  REQUIRE_THROWS_WITH(
+    config.load_from_file(data_dir / "invalid_runtime_distinct_build_probe.yaml"),
+    Catch::Contains("sirius.operator_params.enable_runtime_distinct_build_probe") &&
+      Catch::Contains("removed") && Catch::Contains("remove this key"));
+  CHECK_FALSE(config.get_operator_params().enable_runtime_distinct_build_probe);
```
9prady9: CHECK_FALSE will not even run right ?

test/cpp/config/test_context.cpp
9prady9, line 1: there are more locations in this test where `enable_runtime_distinct_build_probe` is still being tested/used, most of these can be removed I think right ?

docs/super-sirius/configuration.md
9prady9: Below is pseudo code. Was wondering if we do something like
```c++
enum class option_visibility { user, internal };                                                                                                                                                                                                                                                                                                                                                                                
void add_sirius_option(DBConfig& config, option_visibility vis, const char* name,                                                                                                                                        
                       string description, LogicalType type, Value def,                                                                                                                                           
                       set_option_callback_t cb = nullptr) {                                                                                                                                                     
    if (vis == option_visibility::internal && !test_options_enabled()) { return; }                                                                                                                                   
    if (vis == option_visibility::internal) { description = "TEST ONLY: " + description; }                            
    config.AddExtensionOption(name, std::move(description), std::move(type), def, cb);                    
}
```
Or have a allowlist and avoid anything that is not in the allowlist, I think DuckDB does something like this

9prady9: Why did we keep some options yaml overrides and throw for one of them ? Almost all of these are internal options right ?

abelatnvidia: Agreed. I centralized visibility and registration in src/sirius_extension.cpp, the sole registration consumer, using option_visibility and add_sirius_option, and grouped the internal options together. I kept the helper local instead of adding a cross-module header/API.

abelatnvidia: Implemented this pattern. Internal options are absent unless SIRIUS_ENABLE_TEST_OPTIONS=1, and the helper prefixes their descriptions automatically. I used the same helper for representative user options while leaving unrelated existing registrations untouched to avoid whole-file mechanical churn.

abelatnvidia: These are engine-owned for ordinary users. Scan and CONCAT retain YAML-only overrides because benchmark/test envelopes require explicit batch targets; they are no longer normal session settings. Merge fusion has no YAML override. Runtime distinct-build is correctness-sensitive and disabled pending #1600, so stale YAML now fails noisily instead of being silently accepted or ignored.

test/cpp/config/test_context.cpp
abelatnvidia: Correct. I removed the redundant assertion after REQUIRE_THROWS_WITH; the test now verifies only the intended fail-noisy contract.

test/cpp/config/test_context.cpp
abelatnvidia: Removed the stale YAML-default readback, SET/RESET, and state assertions. The remaining references cover only the intended boundaries: test-only registration and rejection of obsolete YAML.

9prady9 (approved):
looks great, thank you!

PR #1526, opened 2026-08-13
refactor(config): keep distinct-build probing internal

infblueocean:
## Summary

- keep runtime distinct-build probing enabled as an engine-owned join policy
- reject the removed YAML key with its full path and deletion guidance
- expose the general two-pass reference path only when `SIRIUS_ENABLE_TEST_OPTIONS=1`
- preserve the internal field, planner propagation, and runtime false path for tests

## Why

`enable_runtime_distinct_build_probe` selects between internal BUILD_PROBE implementations and already defaults on. Sirius can prove distinctness from the actual build keys at runtime; users have no better workload rule for disabling that proof. Keeping the switch public asks them to select an implementation path the engine can own.

## Validation

- repository pre-commit hooks on every changed file: PASS
- `git diff --check`: PASS
- explicit YAML key rejects with full-path removal guidance and leaves the internal default enabled
- normal and non-exact-opt-in processes cannot discover or set the DuckDB option; exact test opt-in can set/reset it
- exact zero-context replay reproduces candidate tree and reverses to the base tree
- duplicate search: no open issue or PR for this setting
- exact composition with #1523 resolved and pre-commit-clean at tree `c9bb796d1cc4ec29fab1653388a3d4072b9c0b7b`

Hosted CUDA-linked tests will execute the config regression and retain the existing internal join-path coverage.

## Identity

- base: `e64b8ad79ea3cee50149d7e3039dbbb92ed291a9`
- head: `6e2ee85ea15b6a7304e90071169f3d5f04929b99`
- tree: `1ed675dc395ec1bd37c2208d68e72888468e03e8`

- exact-head hosted receipt for `6e2ee85ea15b6a7304e90071169f3d5f04929b99`: CUDA 13 runtime and terminal aggregate checks passed
- hosted workflow 31707971632: runtime job 94474358746 passed from 2026-08-13T16:31:32Z through 2026-08-13T16:50:54Z; aggregate job 94523470997 passed at 2026-08-13T16:51:04Z

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -387,7 +387,6 @@ individually.
 | `max_broadcast_join_size` | 256 MiB | Max build-side size eligible for a broadcast join. A build below this size is replicated to every GPU (instead of hash-partitioned) when it is tiny, or when the DuckDB-estimated probe-to-build row ratio is at least `num_gpus * 1.25`. |
 | `max_sort_partition_memory_fraction` | 0.33 | Fraction of GPU memory per sort partition when `max_sort_partition_bytes` is 0 |
 | `mark_join_build_switch_ratio` | 8.0 | For STANDARD MARK joins, build on the smaller (left) side when `right_rows >= ratio * left_rows` (0 disables) |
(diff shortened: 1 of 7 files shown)


infblueocean:
Superseded by #1523, which retains this exact distinct-build policy change, YAML regression, planner propagation, and guarded test path alongside merge-fusion internalization. This predecessor remains an exact passing receipt at `6e2ee85ea15b6a7304e90071169f3d5f04929b99`: workflow `31707971632`, CUDA 13 runtime job `94474358746`, terminal aggregate job `94523470997`. The consolidated exact head is `04eb2ea4683cf371961b4c5455a0da26d4e704f7`.

PR #1527, opened 2026-08-13
fix(config): validate and simplify prefetch controls

### LINKED ISSUE #1312: Prefetch cache returns silently wrong results on multi-GPU
## Summary

Setting `sirius.executor.scan_manager.enable_prefetch_cache: true` together with
`sirius.topology.num_gpus > 1` makes TPC-H queries return intermittent silent results. Example, TPC-H q7:

```
expected:  ('EGYPT', 'UNITED STATES', 1995, Decimal('15896233595.8341'))
actual:    ('EGYPT', 'UNITED STATES', 1995, Decimal('15877063427.5913'))   # ~0.1% low
```

Affected queries across runs: q5, q6, q7, q14, q15, q20 — all lineitem-heavy scans. Which
queries fail varies run to run.

With `num_gpus: 1` the prefetch cache is correct. With the cache disabled, 1/2/4 GPUs are
all correct.

Observed on `dev` @ `50352d1b`, 4x GB200, TPC-H SF300 parquet.

## Reproducer

Needs a TPC-H SF300 parquet dataset and >= 2 GPUs. Substitute your own path for `$DATA`.

**1. Config** — save as `prefetch_2gpu.yaml`:

```yaml
sirius:
  topology:
    num_gpus: 2                      # >= 2 is required; num_gpus: 1 does not reproduce
  memory:
    gpu:
      usage_limit_fraction: 0.9
      reservation_limit_fraction: 1.0
    host:
      capacity_bytes: 700Gi
      initial_number_pools: 20
      pool_size: 8192
      block_size: 1Mi
    disk:
      disk_id: 0
      capacity_bytes: 1Ti
      downgrade_root_dirs: /tmp/sirius_spill_repro
  executor:
    scan_manager:
      num_threads: 16
      uring_n_reactors: 8
      enable_prefetch_cache: true    # the trigger
      cache:
        dispose_after_use: false     # keep the cache warm -> exercises the hit path
        inflight_io_chunk_budget: 4096
        eviction_threshold_fraction: 0.99
    pipeline:
      num_threads: 8
    task_creator:
      num_threads: 12
```

**2. Reference run** (prefetch off, 1 GPU) — this is ground truth:

```bash
sed -e 's/num_gpus: 2/num_gpus: 1/' -e 's/enable_prefetch_cache: true/enable_prefetch_cache: false/' \
    prefetch_2gpu.yaml > reference.yaml

SIRIUS_CONFIG_FILE=$PWD/reference.yaml \
  pixi run python test/tpch_performance/performance_test.py \
    --input $DATA --engine gpu --iterations 2 --queries 6,7,14 \
    --output /tmp/repro --name reference
```

**3. Test run** (prefetch on, 2 GPU) — repeat a few times, it is intermittent:

```bash
for r in 1 2 3 4; do
  SIRIUS_CONFIG_FILE=$PWD/prefetch_2gpu.yaml \
    pixi run python test/tpch_performance/performance_test.py \
      --input $DATA --engine gpu --iterations 2 --queries 6,7,14 \
      --output /tmp/repro --name "test$r"

  for q in 6 7 14; do
    cmp -s /tmp/repro/*_reference/sirius/q$q/result.txt \
           /tmp/repro/*_test$r/sirius/q$q/result.txt \
      || echo "run$r: q$q MISMATCH"
  done
done
```

Expected: `MISMATCH` on roughly 2-3 of 4 runs.

## Where the bug is

The read path in `prefetching_cache::device_read_async`
(`src/io/cache/prefetching_cache.cpp`) classifies each chunk-aligned position of a request
into three cases:

1. already cached — Sirius copies host->device itself
2. allocated but not loaded — read disk into the chunk's own buffer, then H2D
3. missing / busy / partial-boundary — read a block-aligned span through one of the
   **reactor's internal bounce slots**, leaving the cache untouched

**The defect is in case (3).** Patching `device_read_async` to bail out of the cache
entirely (falling back to a plain direct device read) whenever any position needs case (3):

| case (3) | runs | wrong results |
|---|---|---|
| disabled | 4 | **0** |
| enabled (control, same binary) | 3 | 2 |

A narrower variant that only refuses to *mix* case (1) with case (2)/(3) in one request was
clean at 2 GPU but still failed at 4 GPU — consistent with case (3) being the culprit, since
that variant leaves pure case-3 requests exposed.

Suspicion, not yet demonstrated: the slot-lifetime/event bookkeeping in
`uring_reactor::worker_loop` — `poll_copy_completions` resolves the completion event through
`slots[si].event`, a pointer that `io_slot::on_request` overwrites when the slot is reused.

infblueocean:
## Current repair identity

- exact base: `cbd9516367b2d3160492c8cb0575208f1dce047c`
- exact head: `c5ae05c03d488341ed32df589c1a0840500172a3`
- candidate tree: `984d9fa0333d3e09a51bc6f2da53153274bba4ef`
- full-index binary diff SHA-256: `7dfac65d012b5a14e84997a2494e87abcca2e7d4ee5295ac443fcbebe65a1ad6`
- current-dev conflict resolution preserves merged #1532 scan internalization and the complete prefetch validation unit
- complete diff and independent cold review: clear
- diff, clang-format, rumdl, YAML-fixture, test-registration, and single-GPU documentation consistency checks: pass
- hosted exact-head Test run 32296818402 passed, including GPU runtime job 96211865477; Check, Distribution, lint, GCC, Clang, CUDA build, title validation, and every other non-skipped job passed

Foxglove has completed author-side review of this exact head and returned it to ready. The receipt below is preserved explicitly as predecessor evidence.

## Review focus

**Tier:** judgment — this changes the accepted configuration envelope by
refusing a known-incorrect multi-GPU prefetch-cache topology.

**Maintainer question:** Is load-time refusal the right containment for the
known multi-GPU wrong-result path while #1312 remains open, or is there a
topology-safe cache implementation constraint this guard should encode instead?

**Main risk:** the guard intentionally rejects a configuration previously
accepted by the loader; that configuration is retained only when the cache is
disabled because the enabled path is known to return wrong results.

## Summary

- reject non-null cache controls while `enable_prefetch_cache` is false or omitted
- reject a cache reservation floor above its eviction threshold, retaining equality
- remove memory-prefetcher polling and drain timing from YAML; keep the existing internal policy
- reject memory-prefetcher overrides unless that prefetcher is enabled
- reject `enable_prefetch_cache: true` when more than one GPU is configured
- preserve null/empty cache blocks and every supported enabled-path readback

## Why one PR

The cache and memory prefetcher are adjacent scan-manager prefetch components. This PR gives them one load-time contract: dependent settings require their named component, contradictory budgets fail, worker timing stays internal, and the prefetch cache is restricted to its known-correct single-GPU path.

The topology guard mitigates the silent wrong-result configuration reported in #1312. It deliberately does **not** close #1312: the underlying multi-GPU cache defect remains. A load-time refusal is safer than silently disabling the requested cache or allowing a configuration known to produce incorrect results.

This supersedes #1531 and retains the complete cache activation contract formerly carried by #1488.

## Contract

- implicit or explicit disablement rejects non-null dependent settings
- null and empty cache blocks remain equivalent to omission
- enabled components accept and read back their remaining controls
- cache budget floor must be less than or equal to its eviction threshold
- removed memory-prefetcher timing keys fail with deletion guidance
- omitted timing retains the existing 2 ms polling and 100 ms drain policy
- one configured GPU may enable the prefetch cache
- multiple configured GPUs may run with the cache disabled, but enabling it fails with corrective guidance

## August 18 predecessor identity

- exact base: `34311516f1233c1f4aca4d16b0a7bf190fa197f9`
- exact head: `d5765f1820e1ac125bf8180eadef44d35381fcf7`
- candidate tree: `0d2fedb6542d018244b19fec5b1b4e27bd736cca`
- full-index binary diff SHA-256: `3b1023c0af8d14ee914f49251b796cc92d8449203fd1f4787b14daafda0dc87a`
- clean two-commit rebase onto current `dev`; the only conflict was additive test placement in `test_topology_config.cpp`

## Validation

- conflict resolution retains current-dev topology validation and the complete one-GPU/multi-GPU prefetch-cache regression
- range-diff preserves the production/source patch; only the topology-test insertion context moved
- live comparison with #1588 and #1595 found shared aggregate config files but no prefetch/cache semantic delta to consolidate or resolve
- `git diff --check`: pass
- local conflict-marker scan across the changed source, docs, and config tests: pass
- new hardware-independent topology regression covers: one GPU + cache enabled; two GPUs + cache disabled; two GPUs + cache enabled and rejected
- adjacent config selectors cover topology, scan-manager activation, object-store cache use, and full Sirius config loading
- exact-head hosted Test run `32193359123`: CUDA 13 build passed; runtime job `95893704932` passed in 20m27s, including C++ unit tests and the TPC-H snapshot; terminal aggregate job `95898222615` passed
- exact-head hosted Check run `32193359054`: lint, GCC release, Clang debug, and terminal aggregate passed
- exact-head hosted Distribution run `32193359694`: terminal aggregate passed

Final author-readiness review at exact head `d5765f1820e` found the two-commit
unit coherent, the conflict resolution faithful, the dependency current, and
no remaining author-side repair. The judgment question above is ready for
maintainer review rather than an unfinished-author-work draft condition.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -323,8 +323,13 @@ single-GPU configurations only (logs a warning and disables itself otherwise).
 | `enable` | bool | false | Master switch for the prefetcher. |
 | `num_threads` | int (**> 0**) | 2 | Prefetch worker threads; each drives one in-flight batch conversion on its own stream. |
 | `min_free_fraction` | double [0,1] | 0.4 | Keep at least this fraction of the GPU space free after each prefetch; conversions (and their reservations) are only attempted above this floor. |
-| `poll_interval_ms` | int (**> 0**) | 2 | Worker sweep interval while waiting for headroom / new splits. |
-| `drain_quiet_ms` | int (ms) | 100 | A connector counts as actively draining (and is skipped) until this long passes since its last pop. Must exceed the scan's inter-pop interval. |
+
+`num_threads` and `min_free_fraction` are active only when `enable: true`; configuration loading
+rejects non-null overrides while the prefetcher is disabled.
+
+Worker polling and connector-drain timing are internal scheduling policy. The
+former `poll_interval_ms` and `drain_quiet_ms` keys have been removed;
+configurations that still contain either key must delete it.
 
 ### `scan_manager.local` — io_uring backend (`io/uring/config.hpp`)
 
@@ -363,8 +368,8 @@ and transport use one trust policy; there are no separate REST YAML controls.
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | `inflight_io_chunk_budget` | int (**> 0**) | 2048 | Max in-flight IO chunks (enforced by admission control). |
-| `eviction_threshold_fraction` | double [0,1] | 0.6 | Start evicting when the pool fills to this fraction. |
-| `min_prefetching_budget_fraction` | double [0,1] | 0.05 | Floor of the budget reserved for prefetching. |
+| `eviction_threshold_fraction` | double [0,1] | 0.6 | Start evicting when the pool fills to this fraction. Must be at least `min_prefetching_budget_fraction`. |
+| `min_prefetching_budget_fraction` | double [0,1] | 0.05 | Floor of the budget reserved for prefetching. Must not exceed `eviction_threshold_fraction`. |
 | `dispose_after_use` | bool | false | Discard chunks immediately after use. |
 
 ### `scan_manager.object_store` — S3 credentials & endpoint (`io/object_store_config.hpp`)
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -306,7 +306,7 @@ The `sirius.executor.scan_manager` block configures the scan-metadata thread poo
 | `use_sirius_datasource` | bool | true | Route reads through the Sirius `io_uring` datasource. When false, the kvikio fallback is used (single-GPU only; multi-GPU requires the Sirius datasource). |
 | `uring_n_reactors` | int (**> 0**) | 1 | Number of io_uring reactor threads for local-disk reads. |
 | `rest_n_reactors` | int (**> 0**) | 2 | Number of REST reactor threads for object-store (`s3://`) reads. |
-| `enable_prefetch_cache` | bool | false | Attach the pinned-memory prefetching cache in front of the backend. |
+| `enable_prefetch_cache` | bool | false | Attach the pinned-memory prefetching cache in front of the backend. Supported only with one configured GPU; configuration loading rejects it for multiple GPUs because that path can return incorrect results. |
 
 Five optional nested sub-configs tune the individual backends, caches, and the memory prefetcher:
 
--- a/docs/super-sirius/optimizations.md
+++ b/docs/super-sirius/optimizations.md
@@ -428,7 +428,9 @@ If translation fails, filtering falls back to `expression_evaluator` on the deco
 - `src/io/io_context.cpp` — `sirius_ioctx` cache integration and coverage policy
 - `src/include/exec/semi_future.hpp` — async I/O completion primitive
 
-**Config:** `enable_prefetch_cache` and the `cache` sub-config under `executor.scan_manager`
+**Config:** `enable_prefetch_cache` and the `cache` sub-config under `executor.scan_manager`.
+The prefetch cache is currently supported only with exactly one configured GPU;
+configuration loading rejects an enabled cache for multi-GPU execution.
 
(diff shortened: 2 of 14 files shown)


infblueocean:
Current-dev rebased-head validation is terminal green on `ab5a407df131ba1cbba87447b7faf666a79896b9`: hosted lint, GCC, Clang, CUDA 13 build/runtime, TPC-H snapshot, and terminal aggregate all passed (workflow `31758632101`; runtime job `94640671106`, 20m30s; aggregate `94644222416`). The branch is current with `dev`; maintainer review is the remaining gate.

PR #1528, opened 2026-08-13
fix(config): validate SIRIUS_DISABLE values

infblueocean:
## Summary

- accept only explicit `SIRIUS_DISABLE=0` or `SIRIUS_DISABLE=1`
- centralize the environment interpretation used by runtime initialization and optimizer-mask publication
- reject misspellings instead of silently treating every non-`0` string as disabled

## Configuration contract

`SIRIUS_DISABLE` is a deployment policy switch, not a tuning knob. Unset and
`0` leave Sirius enabled; `1` disables it. Empty, textual booleans, other
numbers, and misspellings fail with the setting name and accepted values before
runtime initialization or optimizer-mask publication.

## September 9 current-dev repair

The prior head conflicted after merged #1519 inserted its logging rollback
regressions at the same test anchor. This replacement retains every #1519
logging test and include, then appends the unchanged `SIRIUS_DISABLE`
regression. The original five-file changed-line payload is byte-identical after
the rebase.

## Exact reviewed identity

- tier: `judgment`
- exact base: `285737bc33c84d5c175c59862162b518b89fcae0`
- base tree: `8a3ac4a8c9809355bd9303ca1dc274fd6211752a`
- exact head: `4add07a8e3865c66abe868573e9b46d1430b53ee`
- candidate tree: `3dfee4a90f2b5c6e115c0ea90089858f25dc12e6`
- cumulative binary-diff SHA-256: `39950b05cb18dde14a703d30df7adc378271fb654ec2e362719f450543262b47`
- cumulative diff: 5 paths, 48 insertions, 6 deletions

## Author-side validation

- independent exact-head full-diff review: no critical, important, or minor findings
- clean worktree, `git diff --check`, and `git show --check`: pass
- focused regression covers unset, `0`, `1`, and invalid empty, `false`, `true`, `2`, and `yes` values with exact environment restoration
- both production call sites use the same helper before their visible mutation boundaries
- hosted Check `34394470477`: lint, GCC release, Clang debug, and aggregate pass
- hosted Test `34394470184`: CUDA 13 build, runtime suite, and aggregate pass
- hosted Distribution `34394471718`: aggregate pass; CUDA distribution leaves correctly skip because distribution inputs did not change
- hosted Validate `34394467550`: title and aggregate validation pass

The replacement head is author-ready and terminal-green. Independent
maintainer approval of this exact head remains the merge gate.

### Retained predecessor evidence

Earlier exact heads also passed the hosted runtime suite. Those receipts support
the unchanged behavior but do not replace the current exact-head gates above.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -26,7 +26,12 @@ If no config file is found, Sirius initializes with built-in defaults (95% GPU m
 
 ### `SIRIUS_DISABLE`
 
-Set `SIRIUS_DISABLE=1` to prevent Super Sirius from initializing. This is **required** when using the legacy code path (`gpu_buffer_init`/`gpu_processing`), because Super Sirius claims most GPU and pinned host memory on startup, leaving insufficient memory for the legacy buffer manager. It is also useful for CPU-only benchmarks.
+Set `SIRIUS_DISABLE=1` to prevent Super Sirius from initializing. `0` explicitly
+enables initialization; other values are rejected so a misspelling cannot
+silently disable Sirius. This is **required** when using the legacy code path
+(`gpu_buffer_init`/`gpu_processing`), because Super Sirius claims most GPU and
+pinned host memory on startup, leaving insufficient memory for the legacy
+buffer manager. It is also useful for CPU-only benchmarks.
 
 ```bash
 export SIRIUS_DISABLE=1
--- a/src/include/sirius_context.hpp
+++ b/src/include/sirius_context.hpp
@@ -703,6 +703,11 @@ class SiriusContext : public ClientContextState {
 /// null (best-effort) path never throws.
 void install_configured_log_sink(DatabaseInstance* db);
 
+/// True only when `SIRIUS_DISABLE` is exactly `1`; unset and exactly `0` enable
+/// Sirius, while every other set value is rejected before runtime
+/// initialization and optimizer-mask publication.
+bool sirius_disabled_from_environment();
+
 /// todo(amin): when duckdb is updated, we need to enable OnExtensionLoaded to support sirius
 /// extensions
 class SiriusContextExtensionCallback : public ExtensionCallback {
--- a/src/include/sirius_context.hpp
+++ b/src/include/sirius_context.hpp
@@ -747,6 +747,11 @@ class SiriusContext : public ClientContextState {
 /// null (best-effort) path never throws.
 void install_configured_log_sink(DatabaseInstance* db);
 
+/// True only when `SIRIUS_DISABLE` is exactly `1`; unset and exactly `0` enable
+/// Sirius, while every other set value is rejected before runtime
+/// initialization and optimizer-mask publication.
+bool sirius_disabled_from_environment();
+
 /// todo(amin): when duckdb is updated, we need to enable OnExtensionLoaded to support sirius
 /// extensions
 class SiriusContextExtensionCallback : public ExtensionCallback {
--- a/src/sirius_context.cpp
+++ b/src/sirius_context.cpp
@@ -1584,6 +1584,14 @@ void install_configured_log_sink(DatabaseInstance* db)
   }
 }
 
+bool sirius_disabled_from_environment()
+{
+  auto const* value = std::getenv("SIRIUS_DISABLE");
+  if (value == nullptr || std::string_view{value} == "0") { return false; }
+  if (std::string_view{value} == "1") { return true; }
+  throw InvalidInputException("Invalid SIRIUS_DISABLE '%s' (expected: 0 or 1)", value);
+}
+
 SiriusContextExtensionCallback::SiriusContextExtensionCallback()
 {
   if (auto* env = std::getenv("SIRIUS_LOG_BACKEND")) { Config::LOG_BACKEND = env; }
@@ -1636,7 +1644,7 @@ void SiriusContextExtensionCallback::OnExtensionLoadFail(DatabaseInstance& db,
 void SiriusContextExtensionCallback::read_config_file_if_exists()
 {
   // Check for explicit disable (used by benchmarks/tests that need pure CPU execution)
-  if (auto* val = std::getenv("SIRIUS_DISABLE"); val != nullptr && std::string(val) != "0") {
+  if (sirius_disabled_from_environment()) {
     SIRIUS_LOG_INFO("Sirius disabled via SIRIUS_DISABLE environment variable.");
     return;
   }
--- a/src/sirius_context.cpp
+++ b/src/sirius_context.cpp
@@ -1653,6 +1653,14 @@ void install_configured_log_sink(DatabaseInstance* db)
   }
 }
 
+bool sirius_disabled_from_environment()
+{
+  auto const* value = std::getenv("SIRIUS_DISABLE");
+  if (value == nullptr || std::string_view{value} == "0") { return false; }
+  if (std::string_view{value} == "1") { return true; }
+  throw InvalidInputException("Invalid SIRIUS_DISABLE '%s' (expected: 0 or 1)", value);
+}
+
 SiriusContextExtensionCallback::SiriusContextExtensionCallback()
 {
   auto const previous_log_backend = Config::LOG_BACKEND;
@@ -1736,7 +1744,7 @@ void SiriusContextExtensionCallback::OnExtensionLoadFail(DatabaseInstance& db,
 void SiriusContextExtensionCallback::read_config_file_if_exists()
 {
   // Check for explicit disable (used by benchmarks/tests that need pure CPU execution)
-  if (auto* val = std::getenv("SIRIUS_DISABLE"); val != nullptr && std::string(val) != "0") {
+  if (sirius_disabled_from_environment()) {
     SIRIUS_LOG_INFO("Sirius disabled via SIRIUS_DISABLE environment variable.");
     return;
   }
--- a/src/sirius_extension.cpp
+++ b/src/sirius_extension.cpp
@@ -2627,10 +2627,7 @@ static void LoadInternal(ExtensionLoader& loader)
   // above succeeded, so a failed load leaves no mask behind. SIRIUS_DISABLE
   // means no Sirius runtime initialization and no mask publication (the
   // extension binary itself may still be loaded).
-  bool const sirius_disabled = [] {
-    auto* val = std::getenv("SIRIUS_DISABLE");
-    return val != nullptr && std::string(val) != "0";
-  }();
+  bool const sirius_disabled = sirius_disabled_from_environment();
   if (!sirius_disabled) { publish_transparent_optimizer_mask(config); }
 }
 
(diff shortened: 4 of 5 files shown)


infblueocean:
Current-dev rebased-head validation is terminal green on `cbf3a91b51018ae16e2964c2cf74cf6323971d49`: hosted lint, GCC, Clang, CUDA 13 build/runtime, TPC-H snapshot, and terminal aggregate all passed (workflow `31758653196`; runtime job `94640745217`, 20m18s; aggregate `94644189032`). The branch is current with `dev`; maintainer review is the remaining gate.

infblueocean:
Temporary exact-head repair draft. Mutation actor/repair owner: infblueocean. Published replacement head: 4add07a8e3865c66abe868573e9b46d1430b53ee, rebased onto dev 285737bc33c84d5c175c59862162b518b89fcae0. Independent full-diff review found no critical, important, or minor issues; the only conflict resolution retains merged #1519 logging tests and appends the unchanged SIRIUS_DISABLE regression. Named external dependency: fresh hosted Check, Test, Distribution, and Validate results for this exact head. Ready condition: all required non-skipped jobs terminal-green, mergeable state clean, no unresolved review threads, and the public exact-head receipt refreshed. Prior-head CI does not satisfy this condition.

infblueocean:
Ready-state receipt: infblueocean restored this PR to ready at exact head 4add07a8e3865c66abe868573e9b46d1430b53ee. Basis: independent full-diff review with no findings; Check 34394470477, Test 34394470184, Distribution 34394471718, source-event Validate 34394467550, and post-body Validate 34397556017 all pass; mergeable; zero unresolved review threads. Independent maintainer approval of this exact head remains required before merge.

docs/super-sirius/configuration.md
```diff
@@ -26,7 +26,12 @@ If no config file is found, Sirius initializes with built-in defaults (95% GPU m
 
 ### `SIRIUS_DISABLE`
 
-Set `SIRIUS_DISABLE=1` to prevent Super Sirius from initializing. This is **required** when using the legacy code path (`gpu_buffer_init`/`gpu_processing`), because Super Sirius claims most GPU and pinned host memory on startup, leaving insufficient memory for the legacy buffer manager. It is also useful for CPU-only benchmarks.
+Set `SIRIUS_DISABLE=1` to prevent Super Sirius from initializing. `0` explicitly
```
felipeblazing, line 29: **Reviewer: Comments, docs & AI residue**

**nit** Paragraph hard-wrapped, unlike the rest of the file

The rest of `configuration.md` puts each paragraph on one line (e.g. lines 3 and 25). Hard-wrapping this paragraph at 80 columns makes the diff show every line as changed when only one sentence was added. Keep the original single line and insert the new sentence into it.

docs/super-sirius/configuration.md
```diff
@@ -26,7 +26,12 @@ If no config file is found, Sirius initializes with built-in defaults (95% GPU m
...
+enables initialization; other values are rejected so a misspelling cannot
```
felipeblazing, line 30: **Reviewer: Comments, docs & AI residue**

**minor** User doc should say a rejected value fails `LOAD` and call out the behaviour change

"Rejected" doesn't tell users what happens. In practice the extension load fails with an error. Before this PR any non-`0` value disabled Sirius, so scripts using `SIRIUS_DISABLE=true`, `yes`, or an empty `SIRIUS_DISABLE=` (a common way to "clear" a variable) will now fail to load instead. State both explicitly so people hitting the error can find the cause.

src/include/sirius_context.hpp
```diff
@@ -747,6 +747,11 @@ class SiriusContext : public ClientContextState {
 /// null (best-effort) path never throws.
 void install_configured_log_sink(DatabaseInstance* db);
 
+/// True only when `SIRIUS_DISABLE` is exactly `1`; unset and exactly `0` enable
```
felipeblazing, line 750: **Reviewer: Comments, docs & AI residue**

**minor** Header doc should state the throw, not what the callers do

The function itself doesn't know about "runtime initialization and optimizer-mask publication"; that describes its two call sites, and the doc will go stale if a third caller is added. What a caller needs to know is missing: it throws `InvalidInputException` for any other value.
```suggestion
/// Reads `SIRIUS_DISABLE`: returns true for `1`, false when unset or `0`, and throws
/// `InvalidInputException` for any other value (including empty).
```

src/sirius_context.cpp
```diff
@@ -1653,6 +1653,14 @@ void install_configured_log_sink(DatabaseInstance* db)
   }
 }
 
+bool sirius_disabled_from_environment()
+{
+  auto const* value = std::getenv("SIRIUS_DISABLE");
+  if (value == nullptr || std::string_view{value} == "0") { return false; }
+  if (std::string_view{value} == "1") { return true; }
+  throw InvalidInputException("Invalid SIRIUS_DISABLE '%s' (expected: 0 or 1)", value);
```
felipeblazing, line 1661: **Reviewer: Configuration surface & defaults**

**nit** Error message format differs from the adjacent env-var validation

The neighbouring startup checks say `SIRIUS_LOG_BACKEND must be one of: ...; got '%s'`. Using the same shape here keeps the startup errors consistent.

```suggestion
  throw InvalidInputException("SIRIUS_DISABLE must be one of: 0, 1; got '%s'", value);
```

(The test's `Catch::Contains("0 or 1")` would need to change to `"0, 1"`.)

src/sirius_context.cpp
```diff
@@ -1736,7 +1744,7 @@ void SiriusContextExtensionCallback::OnExtensionLoadFail(DatabaseInstance& db,
 void SiriusContextExtensionCallback::read_config_file_if_exists()
 {
   // Check for explicit disable (used by benchmarks/tests that need pure CPU execution)
-  if (auto* val = std::getenv("SIRIUS_DISABLE"); val != nullptr && std::string(val) != "0") {
+  if (sirius_disabled_from_environment()) {
```
felipeblazing, line 1747: **Reviewer: Error handling & fail-closed gating**

**minor** Invalid SIRIUS_DISABLE is rejected only after the ctor has swapped the global log sink

The ctor applies `SIRIUS_LOG_*` to `Config::` and calls `install_configured_log_sink(nullptr)` before `read_config_file_if_exists()` reaches this validation. When the validation throws, the ctor does not roll that mutation back, so a failed load leaves the process with a different backend, directory, level and sink. That breaks the #1519 contract that config is validated before mutation. Validate `SIRIUS_DISABLE` at the top of the ctor next to the backend/level checks (or cache the result there), and add a test that a bad value leaves `Config::LOG_*` and the sink unchanged.

src/sirius_extension.cpp
```diff
@@ -3536,10 +3536,7 @@ static void LoadInternal(ExtensionLoader& loader)
   // above succeeded, so a failed load leaves no mask behind. SIRIUS_DISABLE
   // means no Sirius runtime initialization and no mask publication (the
   // extension binary itself may still be loaded).
-  bool const sirius_disabled = [] {
-    auto* val = std::getenv("SIRIUS_DISABLE");
-    return val != nullptr && std::string(val) != "0";
-  }();
+  bool const sirius_disabled = sirius_disabled_from_environment();
```
felipeblazing, line 3539: **Reviewer: Error handling & fail-closed gating**

**major** Mask decision re-reads the env instead of reusing the ctor's decision; the new throw lands after partial registration

The callback ctor decides init-or-not from `SIRIUS_DISABLE`. By the time line 3539 reads it again, the callback, GPU functions, the s3 filesystem and per-connection state have all been registered. If the env changes in between (the test suite does setenv/unsetenv in-process), the two decisions disagree: either the mask is published with no `context_`, or the context is initialized with no mask. And if the value is now invalid, the throw leaves the extension half-loaded. Record the decision once in the ctor (for example `[[nodiscard]] bool is_disabled() const noexcept` on `SiriusContextExtensionCallback`) and use `callback_ptr->is_disabled()` here.

felipeblazing, line 3539: **Reviewer: Test coverage & rigor**

**minor** No test that loading with an invalid value fails cleanly and publishes no optimizer mask

The PR says an invalid value fails 'before runtime initialization or optimizer-mask publication', but no test loads the extension with a bad value. A `duckdb::DuckDB db(nullptr)` under `SIRIUS_DISABLE=yes` should show whether the error surfaces as a load failure or something worse, and whether any mask remains. Note that the helper call at this line can never throw for a bad value: the callback constructor has already thrown earlier in `LoadInternal`.

test/cpp/config/test_context.cpp
```diff
@@ -611,6 +611,33 @@ TEST_CASE("Sirius startup rejects unknown environment log levels before mutation
   REQUIRE_FALSE(sirius::log::get_sink()->should_log(sirius::log::level::critical));
 }
 
+TEST_CASE("SIRIUS_DISABLE accepts only explicit boolean values", "[sirius][config]")
```
felipeblazing, line 614: **Reviewer: Test coverage & rigor**

**major** Test would still pass if both call-site changes were reverted

The test only calls `sirius_disabled_from_environment()` directly. If `read_config_file_if_exists` or `LoadInternal` went back to the old inline `getenv(...) != "0"` check, the test would still pass, and those call sites are where the user-visible behaviour lives. Add a case that mirrors the log-level test at line 599: set `SIRIUS_DISABLE=yes` and `REQUIRE_THROWS_WITH(duckdb::SiriusContextExtensionCallback{}, Catch::Contains("SIRIUS_DISABLE"))`.

test/cpp/config/test_context.cpp
```diff
@@ -611,6 +611,33 @@ TEST_CASE("Sirius startup rejects unknown environment log levels before mutation
...
+{
+  std::optional<std::string> previous;
```
felipeblazing, line 616: **Reviewer: Test coverage & rigor**

**nit** Hand-rolled env restore duplicates `scoped_env_assignment`

Lines 616-624 repeat what the `scoped_env_assignment` helper at line 132 already does. Use `scoped_env_assignment restore_env{"SIRIUS_DISABLE", "1"};` followed by the `unsetenv` at line 626 instead.

test/cpp/config/test_context.cpp
```diff
@@ -611,6 +611,33 @@ TEST_CASE("Sirius startup rejects unknown environment log levels before mutation
...
+  }};
+
+  unsetenv("SIRIUS_DISABLE");
+  REQUIRE_FALSE(duckdb::sirius_disabled_from_environment());
+  setenv("SIRIUS_DISABLE", "0", 1);
+  REQUIRE_FALSE(duckdb::sirius_disabled_from_environment());
+  setenv("SIRIUS_DISABLE", "1", 1);
+  REQUIRE(duckdb::sirius_disabled_from_environment());
+
+  for (auto const* invalid : {"", "false", "true", "2", "yes"}) {
```
felipeblazing, line 633: **Reviewer: Test coverage & rigor**

**nit** Invalid-value list misses the most likely near-misses

The list has no whitespace or formatting variants, which are the realistic ways a value like `1` gets mangled in env files and YAML. Add them so an accidental trim or `stoi` in a later rewrite gets caught.
```suggestion
  for (auto const* invalid : {"", "false", "true", "2", "yes", " 1", "1 ", "01", "-1"}) {
```

PR #1522, opened 2026-08-13
fix(config): reject zero scan task batches

infblueocean:
## Summary

- reject zero `scan_task_batch_size` on YAML and DuckDB runtime surfaces
- validate runtime input before Sirius-context lookup so disabled-context sessions refuse the same value
- preserve prior state on refusal and document the positive-byte contract

## Why

Zero currently has incompatible meanings across the two scan paths. The DuckDB-native coalescer explicitly treats `0` as no byte cap, producing one unbounded batch. The Parquet insert path compares every subsequent row group against the zero budget, producing one row group per batch. One accepted setting therefore selects opposite batching extremes by storage format. The derived/default value is positive and the user documentation describes a target batch size, not a format-dependent sentinel.

## Validation

- YAML regression rejects `scan_task_batch_size: 0` with the exact setting path
- isolated DuckDB regression rejects runtime zero without a Sirius context and proves the visible default is unchanged
- `python3 scripts/check_orphan_tests.py`
- `git diff --check`
- live duplicate search found no open zero-validation work for this setting
- clean independent merge trees with #1516 (`5a7797d0cab45630ce4a9669ca009ac2571e6c5e`), #1517 (`7918e91391d15a6a1842769f01826c4052ae5207`), #1519 (`29026214cd3a0943695106154384750d447c3bf5`), and #1520 (`d7c8a3bee12c5e2843f3e99e9ebca0cf9f96356a`)

Exact base: `0d7f98d9a662e9e3b47d6f7d6e2c8797d1e6572b`
Candidate: `e2d0cab113fc9161caaad65847807d9a5f3793f1`
Candidate tree: `c3c7fc2f48a1e2d1a504ce47397e1c35dfb7e428`
Zero-context source patch SHA-256: `17a0cf1cba642546365b5cacd6082338269c305eebc159de79d38d9ec6cfeb79`

Exact-head hosted CI is terminal green: lint, GCC release, Clang debug, CUDA 13 build, full C++ runtime (job 94464384943, 20m11s), and aggregate checks passed.

PR #1532, opened 2026-08-13
refactor(config): keep scan sizing automatic

infblueocean:
## Summary

- keep `scan_task_batch_size` automatic for ordinary DuckDB sessions
- retain the explicit YAML override for advanced benchmark envelopes and `SET`/`RESET` only behind exact `SIRIUS_ENABLE_TEST_OPTIONS=1`
- reject zero on both retained YAML and test-only session surfaces
- preserve the effective-GPU-capacity-derived default introduced by #1447

## Why one PR

This is one setting contract: ordinary users do not choose the scan batch target; Sirius derives it from effective GPU capacity. Advanced benchmark/test overrides remain possible but must be positive because clean Sirius interprets zero incompatibly across scan paths—unbounded DuckDB-native coalescing versus one Parquet row group per batch.

The exact union supersedes #1522 while retaining its YAML and disabled-context session regressions. The session validation executes before Sirius-context lookup, so the exact test surface refuses zero consistently even without a live Sirius context.

Foxglove's capped-pressure screen found that cap-relative values retained oracle-correct feasibility at 1 GiB, 512 MiB, and 256 MiB effective GPU caps (9/9 default cells). That automatic behavior replaces ordinary user tuning; it is not evidence that zero is a useful sentinel.

## Rebased identity

- base: `e885ffea7d0541f1e3493566e2db0ed6e431106f`
- head: `f4a7aa1ed971b8383da237629d6c469d3da52a1f`
- tree: `ec6956346841ed1335a36c811333c5f57cc75132`
- full-index binary diff SHA-256: `3d1d885f986bd8f247e14934dd66ad3f104efdf5d7feef817ac53ddbb92869e5`

## Validation

- rebased onto current `dev`; conflicts only shared the test-option gate and config tests with merged #1536, resolved additively so both feature families remain
- independent source/range comparison across the same five paths: semantics preserved; no legacy edits
- current-dev ancestor and recursive submodule identity/cleanliness: pass
- `git diff --check`: pass
- normal and non-exact sessions omit the setting; exact test opt-in retains `SET`/`RESET`; YAML and test-only runtime zero rejection remain covered
- pre-rebase exact-head hosted lint, GCC, Clang, CUDA 13 build/runtime: pass (workflow 31733921150)
- rebased-head hosted lint, GCC release, Clang debug, CUDA 13 build/runtime,
  TPC-H snapshot, and terminal aggregate all passed on
  `f4a7aa1ed971b8383da237629d6c469d3da52a1f`
- hosted workflow 31754904530: build job 94628495369 passed; runtime job
  94629221380 passed in 19m45s; terminal aggregate job 94632923981 passed

Supersedes #1522.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -377,7 +377,7 @@ individually.
 
 | Parameter | Default | Description |
 |-----------|---------|-------------|
-| `scan_task_batch_size` | Shared physical/effective GPU batch default described above | Target batch size for DuckDB scan tasks |
+| `scan_task_batch_size` | Shared physical/effective GPU batch default described above | Target batch size for DuckDB scan tasks; must be greater than zero |
 | `enable_compressed_materialization` | true | Store eligible integer and fixed-point DECIMAL values in value-preserving narrower carriers when exact pin-time bounds permit it; restore native carriers at type-sensitive boundaries. |
 | `max_sort_partition_bytes` | 0 (auto) | Max bytes per sort partition. Auto = 33% of GPU memory. |
 | `hash_partition_bytes` | Shared physical/effective GPU batch default described above | Target partition size for hash joins and group-bys; must be greater than zero |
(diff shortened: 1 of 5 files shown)


infblueocean:
Rebased onto current dev after #1536 and revalidated the exact head. Conflicts were additive overlap in the test-option gate/config tests; both merged dynamic-filter coverage and this PR's scan-setting coverage are retained. Hosted lint, GCC, Clang, CUDA 13 build/runtime, TPC-H snapshot, and aggregate are all green on f4a7aa1ed971b8383da237629d6c469d3da52a1f. Ready for review.

9prady9 (approved):

PR #1154, opened 2026-07-11
perf: zone-maps on pinned table chunks

kevkrist:
This PR introduces zone-map statistics for pinned batches. The statistics are generated at pin time via min/max reductions. The cache checks filters against the batch zone maps before serving them. 

When data is clustered on common ordered filter columns (e.g., for TPC-H, `l_shipdate` for `lineitem` and `o_orderdate` for `orders`), the cache can significantly prune served data batches, resulting in data processing reduction and performance improvement. See the comments below for a performance analysis on gb200.

TLDR: 1.22x speedup in single gpu, 1.06x speedup in multi-gpu.

NOTE: to run performance evaluation (`performance_test.py`), use the following command to generate the clustered duckdb data:
`pixi run bash test/tpch_performance/generate_tpch_data.sh <desired sf> --format duckdb --cluster`

PR #1535, opened 2026-08-13
refactor(config): internalize pinned zone-map pruning

infblueocean:
## Summary

- keep pinned-table zone-map capture/pruning automatic and enabled by default
- keep `sirius.operator_params.enable_pinned_zone_map_pruning` as the advanced YAML benchmark/diagnosis envelope
- register the direct DuckDB session override only when `SIRIUS_ENABLE_TEST_OPTIONS=1`
- prove the option is absent normally, remains absent for loose truthy opt-in values, and is available to the existing transition tests only under exact opt-in
- update the user docs to describe the smaller surface

## Why

This is an internal cache capture/serve mechanism, not a useful normal-session choice. Changing it before pinning creates a statless cached entry that stays statless until re-pinned, so exposing a mutable session knob asks users to reason about cache history.

The default and runtime behavior do not change. YAML remains available for an explicit benchmark or diagnosis envelope, and Sirius's test binary already opts into test-only settings before constructing its databases.

The merged implementation in #1154 made the feature default-on and retained oracle correctness in its end-to-end clustered-table benchmark (22/22 queries; 1.220x single-GPU and 1.058x multi-GPU suite speedups for the combined clustered-plus-pruning treatment). This PR makes no isolated performance claim; it only removes the unnecessary normal session choice.

## Validation

- pre-commit: all applicable hooks passed; the local `rumdl` wrapper alone could not start because `pixi` is not installed on the macOS host
- `rumdl==0.2.44 check` passed independently on all changed Markdown
- `git diff --check` passed
- hosted CUDA build/tests requested by this PR

- hosted workflow [31716482009](https://github.com/sirius-db/sirius/actions/runs/31716482009): runtime job 94503737159 passed in 20m12s; aggregate job 94534354523 passed

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -586,21 +586,20 @@ SET enable_dynamic_zone_map_filter = false;
 
 ### Pinned Tables
 
-This setting is accepted both as a DuckDB `SET` variable and in YAML under
-`sirius.operator_params`.
+Pinned-table zone-map capture and pruning are automatic. The advanced YAML escape hatch is under
+`sirius.operator_params` for benchmark and diagnosis envelopes; it is not a normal session choice.
 
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `enable_pinned_zone_map_pruning` | true | Capture pinned-chunk zone maps at pin time and use them to prune cached scans. |
 
-Setting this to `false` before `pin_table` avoids the extra GPU reductions and creates a
+Setting the YAML value to `false` before startup avoids the extra GPU reductions and creates a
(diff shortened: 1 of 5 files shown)


infblueocean:
Superseded by the consolidated dynamic-filter unit in #1536 at exact head 929dcf7b38dc50c80c305a6864a6b28d85185ec6 (tree f19bb8bff85a0e51c711abf4de30e6c84f529dd1). This PR's exact head 1280834c9146f4099440acf1c4fc24547864a619 and hosted run 31716482009 remain recorded as predecessor evidence; its pinned zone-map surface reduction, docs, and tests are retained in #1536.

PR #1540, opened 2026-08-13
fix(config): require disk spill mount paths

infblueocean:
## Summary

- reject low-level `sirius.space.disk[]` entries that omit or null `mount_path` during YAML loading
- reject an explicitly empty mount path before allocator construction
- document the spill path as required and non-empty

## Why

The low-level `sirius.space.*` surface is an intentional expert replacement API, so this leaves the surface intact. A disk-space entry is nevertheless unusable without a mount path: `cucascade::memory::memory_space` already throws for an empty path, but the YAML loader currently accepts the invalid state and defers the error until allocator construction. This moves the existing invariant to the user-facing configuration boundary.

## Validation

- regression fixtures cover omitted, null, and empty `mount_path`
- the existing `spaces.yaml` configuration test remains the positive path
- applicable pre-commit hooks pass (the repository rumdl hook could not launch because local `pixi` is unavailable)
- `git diff --check` passes

- hosted workflow [31722321301](https://github.com/sirius-db/sirius/actions/runs/31722321301): runtime job `94523031030` passed in 20m07s; aggregate job `94533693742` passed

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -219,7 +219,7 @@ Each tier is a **YAML sequence** of space configs. Byte fields accept suffixes; 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | `disk_id` | int | 0 | Identifier for the disk space. |
-| `mount_path` | string | "" | Spill directory. |
+| `mount_path` | string | required | Non-empty spill directory. |
 | `memory_capacity` | bytes | space default | Maximum disk space for spill files. |
 
 ```yaml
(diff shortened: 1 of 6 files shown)


infblueocean:
Superseded by the consolidated low-level memory-space validation unit in #1542 at exact head f9407cd6fe06b30b9bdb1060b425a35f3de3aba1 (tree 58b44c2d2c15422ccb6a9f984692a4b134af0284). This PR's exact head af1c30cb8515f6605ce2223c64d22d343a5b44fc and hosted run 31722321301 remain predecessor evidence; its disk mount-path contract and fixtures are retained in #1542.

PR #1542, opened 2026-08-13
fix(config): validate low-level memory spaces

infblueocean:
## Summary

- require every low-level GPU entry to provide a non-negative `device_id`
  that exists in discovered topology
- reject duplicate GPU `device_id`, host `numa_id`, and disk `disk_id` values,
  including collisions between omitted host IDs that resolve to `-1`
- accept only host NUMA IDs of `-1` or non-negative
- require every low-level disk entry to provide a non-empty `mount_path`

This consolidates the low-level memory-space validation family previously
split across #1540, #1541, and #1542. It does not change high-level memory
policy, default capacities, automatic topology selection, or compression
policy.

## Why

Any non-empty `sirius.space` list replaces the high-level memory-space set.
Missing or colliding identities make lookup ambiguous, an unknown GPU ID
cannot be executed, and a disk tier without a mount path cannot spill. These
are malformed low-level configurations, not tuning choices.

## Exact current-dev composition

- current base: `3e4d98cfb760c716540b8d4955a9de37d5acfa53`
- exact rebased head: `577f66f9b7d5beaf053943a9b24c0ee2327a25e2`
- candidate tree: `84823a77dec580b5714e06abf1382ba42bffb8c7`
- zero-context diff SHA-256 (`git diff --unified=0 BASE..HEAD`):
  `f53970bc3a65aa2f3df4c5b151fda03e7db7e669fe381773b1fb9124cacb6106`
- guarded force-push lease: previous remote head
  `77c45d974cad50cb91704376f0a9828d5c814d4a`

The one rebase conflict was an adjacent include addition in
`src/sirius_config.cpp`: merged #1498 requires `<cmath>` and this change
requires `<cstdint>`. The resolution retains both. Range-diff is 1:1; its only
delta is that additive include context. All five #1542 test blocks and all 12
fixtures are byte-identical to the prior head. The #1494 topology/affinity and
#1498 compression test blocks are also byte-identical to current `dev`.

## Validation

- source-contract inspection: required identities, bounds, per-tier uniqueness,
  disk path enforcement, and discovered-topology validation are present
- 12 focused invalid fixtures cover missing/null/negative/unknown GPU IDs,
  invalid host IDs, duplicate IDs including the default-host collision, and
  missing/null/empty disk paths
- scoped pre-commit suite: all applicable hooks pass
- standalone `rumdl`: changed documentation passes
- orphan-test check: pass
- `git diff --check`: pass
- no compiled Sirius test binary or `pixi` was available in the isolated
[...]

PR #1543, opened 2026-08-13
fix(config): validate low-level host NUMA ids

infblueocean:
## Summary

- reject low-level `sirius.space.host[].numa_id` values below the supported `-1` sentinel
- document the actual `-1` default and its unbound-host-allocation meaning
- retain omitted and explicit `-1` behavior

## Why

The expert `sirius.space.host[]` replacement API defaults `numa_id` to `-1`. cuCascade handles exactly that value as the unbound path using `cudaHostAlloc`; every other value is passed to `numa_alloc_onnode`. Values below `-1` currently survive YAML loading and reach libnuma as node IDs. This validates the existing two-state contract at the configuration boundary without removing the intentional `-1` escape hatch.

## Validation

- regression fixture rejects `numa_id: -2` with the low-level field named in the error
- existing `spaces.yaml` remains the positive explicit `-1` path
- `git diff --check` passes

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -205,7 +205,7 @@ Each tier is a **YAML sequence** of space configs. Byte fields accept suffixes; 
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
-| `numa_id` | int | 0 | NUMA node the pinned pool is bound to. |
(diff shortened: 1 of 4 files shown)


infblueocean:
Superseded by #1542. The valid host NUMA-ID domain and duplicate per-tier identity checks are one coherent low-level memory-space identity contract, now reviewed together.

PR #1541, opened 2026-08-13
fix(config): validate low-level GPU device ids

infblueocean:
## Summary

- require each expert `sirius.space.gpu[]` entry to select a CUDA device explicitly
- reject negative low-level GPU device IDs during YAML loading
- reject IDs absent from Sirius's discovered GPU topology before allocator construction
- document the complete explicit-ID contract and add omitted/null/negative/unknown regressions

## Why

The low-level `sirius.space.*` surface remains an intentional expert replacement API. Its `gpu_memory_space_config` default is `-1`, however, and the low-level path copies entries directly into allocator construction, where `memory_space` passes `device_id` to `rmm::cuda_set_device_raii`. Omitted, negative, and non-negative-but-unavailable IDs therefore survive configuration loading and fail later as CUDA-device errors. This moves the complete device-selection invariant to the user-facing configuration boundary without changing high-level topology selection.

## Validation

- regression fixtures cover omitted, null, negative, and deterministically unavailable `device_id`
- existing `spaces.yaml` remains the positive explicit discovered-device path
- applicable pre-commit hooks pass (repository rumdl hook cannot launch locally because `pixi` is unavailable)
- direct `rumdl 0.2.44` passes
- `git diff --check` passes

Base: `e64b8ad79ea3cee50149d7e3039dbbb92ed291a9`

- fresh exact-head hosted lint, GCC release, Clang debug, CUDA 13 build, full runtime, and terminal aggregate checks: pass
- hosted workflow 31723114178: runtime job 94525886838 passed in 20m55s; aggregate job 94538376690 passed

PR #1544, opened 2026-08-13
fix(config): validate low-level GPU topology

infblueocean:
## Summary

- reject expert `sirius.space.gpu[]` device IDs absent from discovered topology
- report the invalid field and ID during configuration loading
- document that explicit low-level IDs must refer to discovered GPUs

## Why

High-level `sirius.topology` selections are resolved against topology by the configurator, but the low-level replacement path copies `gpu_memory_space_config` entries directly into allocator construction. A non-negative but unavailable `device_id` therefore survives configuration loading and only fails when the GPU memory space calls RMM/CUDA. This gives the expert path the same fail-early hardware identity boundary without changing its replacement semantics.

## Validation

- regression uses `INT_MAX` as a deterministically absent GPU ID
- existing `spaces.yaml` remains the positive discovered-device path
- `git diff --check` passes


infblueocean:
Superseded by #1541. The topology-membership check belongs to the same low-level GPU device-ID contract and has been folded into that review unit.

PR #1511, opened 2026-08-13
fix(config): validate memory policy inputs

infblueocean:
## Summary

- reject zero host pool dimensions and zero high-level GPU/host capacities or fractions
- bound the initial host-pool footprint by each resolved non-zero host-space capacity with overflow-safe arithmetic
- honor absolute GPU/host reservation bytes against resolved effective space capacity and reject values that cannot fit
- remove the high-level single-disk `disk_id` choice and use internal ID 0
- reject positive high-level spill capacity without a spill directory, while retaining `capacity_bytes: 0` as the explicit spill opt-out
- preserve explicit IDs on expert low-level `sirius.space.disk[]`

## Why one PR

These are the shared allocator-safety and high-level `sirius.memory` source-of-truth rules. Capacity and reservation values must mean the bytes users state; pool preallocation must fit the resolved capacity; and the one high-level disk tier should express spill path/capacity policy rather than an internal identity. Low-level host-pool dimensions receive the same safety checks; low-level multi-space identities and topology remain separate in #1542.

This supersedes #1545 and preserves its single-disk/partial-activation regressions alongside the allocator budget contract.

## Contract

- high-level capacity and fraction fields that construct memory tiers are positive; intentional reservation/disk zero opt-outs remain accepted
- initial host-pool bytes fit each non-zero effective host capacity
- absolute reservations are applied literally after effective capacity resolution and cannot exceed it
- the single high-level disk tier always uses internal ID 0
- positive spill capacity requires a spill directory; zero capacity disables the tier
- low-level host-pool dimensions are positive and bounded; low-level space identities remain the separate #1542 contract

## Identity

- exact base: `e885ffea7d0541f1e3493566e2db0ed6e431106f`
- exact consolidated head: `dc8bbce33177a381ddaa12182126f0e44832064b`
- base tree: `df560dabe752232176201ed43bbaaa9c91be5805`
- candidate tree: `70621cf77dd11764d9764c451403bc19de8383cb`
- zero-context source-diff SHA-256: `a10d8e51374b93f517890b6a7401d54c9ede0527d3b520ace3a2a1ac99c06b75`
- prior #1511 head: `fc975c4ae07694d287400bc11512620fc1e73e7e`
- #1545 predecessor: `d3ac266b35639c6ada5994d82d25344668e2f1c0`

## Validation

- clean conflict-free rebase; `git range-diff` reports the single commit as patch-equivalent
[...]

PR #1545, opened 2026-08-13
refactor(config): simplify high-level disk policy

infblueocean:
## Summary

- remove `sirius.memory.disk.disk_id` from the normal high-level configuration path and fix the single high-level disk tier to internal ID 0
- reject an explicitly positive spill capacity when no spill directory activates it
- preserve `capacity_bytes: 0` as the explicit high-level spill-tier opt-out
- retain explicit IDs on expert `sirius.space.disk[]`, where multiple disk spaces make identity meaningful
- update the shipped config examples and SF1000 reproduction config

## Why

High-level `sirius.memory.disk` can construct at most one disk space. Its `disk_id` only renamed the internal `memory_space_id`; it did not choose a device, path, capacity, or recovery policy. Requiring users to choose that implementation key added configuration surface without expressing intent.

The same high-level block previously accepted `capacity_bytes` by itself and silently skipped constructing the disk tier because `downgrade_root_dirs` was empty. That partial configuration now fails at YAML load with an actionable message instead of looking active while doing nothing.

The low-level replacement API still supports multiple disk spaces and keeps explicit IDs.

## Validation

- configurator integration asserts the high-level disk space uses internal ID 0
- stale high-level `disk_id` fixture checks removal and low-level replacement guidance
- positive-capacity-without-directory fixture checks inactive-setting rejection
- zero-capacity fixture proves the explicit opt-out remains accepted and produces no disk space
- existing low-level `spaces.yaml` continues to cover IDs 0 and 1
- `git diff --check` passes

- hosted workflow 31723993656: runtime job 94528692203 passed in 20m20s; aggregate job 94540105396 passed

### DIFF
--- a/bench/sf1000-repro/sirius-sf1000.yaml
+++ b/bench/sf1000-repro/sirius-sf1000.yaml
@@ -25,7 +25,6 @@ sirius:
             pool_size: 8
             block_size: 67108864
         disk:
-            disk_id: 0
             capacity_bytes: 1000000000000
             downgrade_root_dirs: "/localhome/local-faramburu/sirius_disk_memory"
     executor:
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -78,7 +78,7 @@ sirius:
       downgrade_trigger_fraction: 0.8
       downgrade_stop_fraction: 0.6
     host: { capacity_bytes: 25GB, initial_number_pools: 50, pool_size: 512, block_size: 1048576 }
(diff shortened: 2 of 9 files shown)


infblueocean:
Superseded by #1511, which retains the exact high-level disk-ID removal, spill activation contract, and zero opt-out regressions alongside the high-level capacity/reservation source-of-truth rules. This predecessor remains an exact passing receipt at `d3ac266b35639c6ada5994d82d25344668e2f1c0`: workflow `31723993656`, CUDA 13 runtime job `94528692203`, terminal aggregate job `94540105396`. The consolidated exact head is `fc975c4ae07694d287400bc11512620fc1e73e7e`.

Issue #1143, opened 2026-07-10
Support INSERT/DELETE for persistent DuckDB tables on storage (MVCC-blind native scans)


bwyogatama:
## Goal

Support INSERT/DELETE on duckdb tables served by the **duckdb-native GPU scan of persistent storage**: GPU results must equal DuckDB CPU at the query's snapshot while rows are inserted and deleted — the same contract #819 gives MVCC-pinned tables. Today the scan reads the table's **last-checkpointed disk image byte-faithfully and ignores MVCC state entirely**; whenever the live table has diverged from that image, GPU results silently differ.

## Verified failure modes (dev)

1. **Persisted tombstones are served.** Sparse DELETEs that survive a `CHECKPOINT` are serialized as row-group delete info (`deletes_pointers`); the CPU scan applies them via `RowVersionManager::GetSelVector`, but the GPU walker has **no delete/version handling at all** (grep `duckdb_native_metadata.cpp` for delete/version — zero hits) and the decoder copies row-group blocks byte-faithfully, tombstoned rows included. Repro: `CREATE TABLE t AS ... range(300000); CHECKPOINT; DELETE FROM t WHERE rowid % 1000 = 7; CHECKPOINT;` → GPU scan returns the deleted rows.
2. **Uncheckpointed committed DELETEs are served.** Same mechanism without the second checkpoint — the delete lives only in the in-memory `RowVersionManager`, and auto-checkpoint (16 MB WAL default) may not fire for a long time.
3. **Committed-but-uncheckpointed INSERTs fail loudly** (not silently): the walker counts transient rows via `GetPartitionStats` but stages zero bytes for them, and decoder pre-validation throws — at execution time, which surfaces as `InternalException` with no CPU fallback (see the fallback constraint recorded in #819).

In-place UPDATEs (stale `UpdateSegment` values, formerly failure mode 3 here) are tracked separately in #1162.

## Status

#819 PR3 (#1142) closed this for **MVCC-pinned tables** via plan-time cache-or-CPU guards. The same PR briefly carried a plan-time refusal for every duckdb-native scan, then **deferred it before merge** — the exactness walk touched every row group of the scanned table per query at plan time. #1160 tracks the execution-time replacement.

## Direction

[...]

PR #1547, opened 2026-08-13
fix(scan): column-aware pinned-entry lookup for split pin layouts

felipeblazing:
## Problem

`find_pinned_entry_for_duckdb_table` returned the first identity match in
`_pinned_entries` iteration order, ignoring the requested columns. Under a split
pin layout — one table pinned as two entries with different column sets — the
plan-time MVCC guard in `sirius_plan_get.cpp` could land on an entry that cannot
serve the scan, declare it unservable, and silently fall back to CPU. The serving
side (`try_match_cached_entry`) is column-aware and would have matched the other
entry, so the two disagreed.

Results stay correct, so nothing in validation flags it. Measured at SF1000
(GB300, TPC-H, split pin over `orders`): ~10 queries per pass fell back this way,
stream time 108 s instead of 11 s.

## Fix

The lookup takes the scan's `column_ids` and `returned_types` (both optional
pointers, `nullptr` default) and prefers, among identity matches:

1. an entry that covers the requested columns **and** whose recorded pin-time
   native types still equal the ones the scan declares;
2. failing that, a covering entry whose recorded types no longer match;
3. failing that, the first identity match.

Those are the same two gates the plan-time guard itself applies, so the lookup
can no longer hand the guard an entry it will reject while a usable one exists.
It never returns `nullptr` where an identity match exists — that would flip the
guard from "decline this scan" to "table is unpinned", which routes to the
MVCC-blind disk read.

`pinned_native_types_match_columns` moves out of `sirius_plan_get.cpp`'s
anonymous namespace into `sirius::scan_manager` so the guard and the lookup call
one predicate. The serve-side `pinned_native_types_match_scan` is unchanged: its
missing-column policy differs deliberately (coverage and types are one gate at
serve time, two at plan time).

With no covering entry, or when the optional arguments are omitted, behavior is
identical to before for every non-split layout. The two other callers keep the
three-argument form.

## Tests

`test/cpp/scan_manager/test_pinned_entry_column_lookup.cpp` (`[pinned_lookup]`).

Column coverage: pins the split shape as two zero-chunk entries (`orders` =
{0..7}, `main.orders` = {0,1,8}) and asserts the covering entry is chosen in each
direction, that an uncovered request falls back to a non-null identity match, and
that null/empty/defaulted ids preserve the previous behavior.

Type matching: pins two single-chunk entries that both cover {0,1}, one recording
the INT64 the scan's BIGINT maps to and one recording INT32, and asserts the
matching entry wins — run in both assignments, so a coverage-only rule fails one
of them for any iteration order. A second case asserts a non-null result when no
covering entry matches.

Verified on this branch: full suite green; reverting just the preference loop
fails the directed cases, confirming they are directed rather than incidental.

## Note for reviewers

An entry whose recorded types no longer match is treated as unpinned by the
guard, and the branch that would otherwise protect an unpinned table is the
disabled `#if 0` block pending #1160 — so a *single* such pin still reaches the
MVCC-blind disk read. That is pre-existing on `dev` and unchanged here; this PR
only stops the lookup from preferring such an entry over a usable one.

### DIFF
--- a/src/scan_manager/sirius_scan_manager.cpp
+++ b/src/scan_manager/sirius_scan_manager.cpp
@@ -22,6 +22,7 @@
 #include "data/data_batch_utils.hpp"
 #include "exec/thread_pool.hpp"
 #include "helper/numeric_narrowing.hpp"
+#include "helper/type_conversions.hpp"
 #include "io/cache/prefetching_cache.hpp"
 #include "io/io_context.hpp"
 #include "io/parquet_helpers.hpp"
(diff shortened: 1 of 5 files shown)


src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -1576,14 +1576,21 @@ void sirius_scan_manager::visit_pinned_entries(
...
+  std::string_view schema_name,
+  std::string_view table_name,
+  duckdb::vector<duckdb::ColumnIndex> const* requested_ids) const
 {
+  pinned_entry const* identity_match = nullptr;
   for (auto const& [name, entry] : _pinned_entries) {
-    if (entry.cache_info.matches_duckdb_table(catalog_name, schema_name, table_name)) {
+    if (!entry.cache_info.matches_duckdb_table(catalog_name, schema_name, table_name)) { continue; }
+    if (identity_match == nullptr) { identity_match = &entry; }
+    if (requested_ids == nullptr || requested_ids->empty() ||
```
kevkrist: Can we include native-type eligibility here, not just column coverage? `try_match_cached_entry` also rejects type-drifted pins, so in the presence of type-drift `prepare_for_query()` will reject the match there after it's passed here, and a fresh DuckDB-native scan can be constructed which would then violate MVCC, I think.

kevkrist (changes requested):
Just 1 potential issue.

src/scan_manager/sirius_scan_manager.cpp
felipeblazing: Good catch — fixed. The lookup now applies both gates.

`pinned_native_types_match_columns` moved out of the planner's anonymous
namespace into `sirius::scan_manager` so the guard and the lookup share one
predicate, and the preference is now: covers + types still match → covers but
types no longer match → first identity match. It never returns `nullptr` where an
identity match exists, since that would flip the guard from "decline this scan"
to "table is unpinned" — the path you're pointing at. `pinned_native_types_match_scan`
is untouched; its missing-column policy differs from the plan-time one on purpose.

Two tests, both directed: the type-preference case runs each assignment of
matching/non-matching entry, so a coverage-only rule fails one of them for any
iteration order (verified by ablation). Full suite green.

Separately, and pre-existing: a *single* pin whose recorded types no longer match
still reaches the MVCC-blind disk read, because the branch that would catch it is
the disabled `#if 0` block pending #1160. Out of scope here — happy to file it
against #1143/#1160 if you think it's worth its own issue.

kevkrist (approved):

PR #1548, opened 2026-08-13
perf: build the Sirius physical plan once per query; skip the row-group metadata walk for pin-served scans

felipeblazing:
## Description

Two independent frontend fixes, one per commit, that cut redundant plan-construction work out
of the query path.

### 1. Build the Sirius physical plan once per query

The transparent path built the Sirius GPU physical plan **twice per query**: once in
`OnFinalizePrepare` purely to validate GPU support (the resulting plan was discarded), and
again in `PhysicalSiriusExecution::GetDataInternal` for the actual run. The validated plan is
now handed to `PhysicalSiriusExecution` and consumed by the first `GetData` (one-shot — the
move empties the slot). Re-executions of the same prepared operator rebuild from the logical
plan template exactly as before, so data-drift and rebind semantics are unchanged.

### 2. Defer the row-group metadata walk for pin-served scans

Every plan build constructs one duckdb-native ingestible per `seq_scan`, whose constructor ran
`prepare_duckdb_native_walk`: `GetPartitionStats` plus a per-row-group statistics pass per
prunable filter column and per projected varchar column. For a **pin-served scan its output is
never consumed** — splits come from the pinned entry's chunk plan and insert-delta bundles, not
from the walk.

`sirius_plan_get` now marks a scan whose MVCC cache-or-CPU guards all passed with
`mvcc_pin_serves_scan`; the plan generator forwards it as `defer_metadata_walk`; the ingestible
skips the walk at construction; and the new `gpu_ingestible::ensure_metadata_prepared()` hook
runs it lazily from the scan manager, on the query thread, for exactly the scans that did *not*
match a pinned entry at execution time. A scan that actually reads disk always gets its walk.

The two compound: fix 1 halves the number of plan builds, fix 2 removes the walk from the
remaining build on pinned workloads.

### Refusal semantics preserved

- The projected-type gate stays **eager in both modes** (factored out as
  `unsupported_projected_type_reason`), so unsupported types still refuse at plan time, where
  the throw is a clean CPU fallback.
- `sirius_plan_get`'s table-level overflow-string probe strictly dominates the walk's
  per-row-group overflow check. For deferred scans the per-row-group refusal moves to the lazy
  walk (a runtime scan error → CPU fallback); for unpinned scans it stays at construction.
- A lazy-walk failure throws out of `prepare_for_query` and surfaces as a runtime scan error,
  matching the walk's other runtime refusals.

No new configuration knobs; behavior is unconditional.

## Measurements

**Re-measured 2026-09-16 after rebasing onto `dev` @ `05148c78`** (RAPIDS 26.08), same GB300 box,
same duckdb-native SF1000 database, unpinned, q1–q19, best-of-3 per query, arms alternated, each
arm built in its own worktree. Only runs where the process was in the box's fast mode are
comparable (see the caveat below):

| | dev @ 05148c78 | this branch | |
|---|---|---|---|
| suite total, run A | 36.362 s | 35.907 s | |
| suite total, run B | 36.696 s | 35.922 s | |
| suite total, min | 36.362 s | 35.907 s | **−1.25%** |
| queries equal or faster | | | 19/19 |

Largest movers: q12 −4.2%, q6 −3.6%, q5 −2.2%, q3 −1.9%. No query slower outside its own run
spread. This matches the −1.05…−1.23% measured on the previous base — as expected, fix 1 is the
whole effect on an unpinned workload.

**Caveat — bimodal box, not bimodal code.** During this session the machine put roughly a third of
launched Sirius processes into a persistent "slow mode": cold first iterations identical, every
warm iteration ~2.6× slower, for the process's whole lifetime, on *either* binary. Twelve
alternating single-process probes (q6+q12, 3 iterations) split dev 2 fast / 4 slow and this branch
4 fast / 2 slow, with the two binaries equal inside each mode (q6 1.66–1.71 s vs 1.66–1.71 s fast;
4.4–4.5 s vs 4.4–4.5 s slow). Slow-mode runs are excluded from the table above. The mechanism is not
diagnosed here (best guess: host-pool placement vs page cache); it predates and is independent of
this PR.

**Also on both arms:** the pre-existing `SIGSEGV` in `cuMemcpyBatchAsync_v2` under
`decode_duckdb_native_split` now also fires intermittently on q2 (and q9), not only q20/q3. Dev
bug, reproduces on the base binary.

<details><summary>Original measurements against dev @ 9624f063</summary>

Measured on a GB300 workstation against a duckdb-native TPC-H SF1000 database, comparing this
branch against its base (`dev` @ `9624f063`), each arm built in its own worktree.

**Unpinned — the case where the walk cannot be skipped and must run anyway.** Best-of-3 per
query, 4 complete runs on the base and 3 on this branch, arms alternated:

| | base | this branch | |
|---|---|---|---|
| suite total, min of runs | 37.784 s | 37.389 s | **−1.05%** |
| suite total, median of runs | 37.882 s | 37.416 s | **−1.23%** |
| per-run spread | 37.78–38.15 s | 37.39–37.46 s | non-overlapping |
| queries faster | | | 15/19 |

The two arms' spreads do not overlap, so this is not run-to-run noise. Largest movers: q6
−4.1%, q12 −3.4%, q19 −2.3%, q4 −2.2%. Two queries moved the other way (q14 +3.2%, q13 +2.5%)
but both sit inside their own per-run spread.

This is fix 1 acting alone: with nothing pinned, `defer_metadata_walk` is never set and the
walk runs eagerly on both arms — but the base runs it *twice per scan* and this branch once.
Counting the per-operator plan-build log marker on a single-scan query confirms the mechanism:
**6 emissions on the base, 3 on this branch** (3 logical operators × 2 builds vs × 1).

**Pinned.** Verified by log markers that the feature engages as designed — one
`deferring metadata walk` and one `reusing finalize-validated Sirius plan` per query, with the
walk skipped entirely. The magnitude of the pinned-workload win is **not** re-measured here;
the numbers previously quoted on this PR were taken against a superseded base and have been
removed rather than carried forward.

**Not measured:** the deferred-walk-then-pin-missing path (the walk runs from
`prepare_for_query` instead of at plan time). Reaching it requires a pin to disappear between
plan and execute, which I could not force deterministically at scale. It runs the same function
on the same thread, just later.

**Unrelated pre-existing failure:** q20 segfaults on the unpinned duckdb-native path on *both*
arms (4/4 runs), and q3 does so intermittently — `SIGSEGV` in `cuMemcpyBatchAsync_v2` under
`decode_duckdb_native_split`. It reproduces identically on `dev`, so it is not from this PR;
it is why the table above covers q1–q19.

</details>

## Checklist
- [x] Read CONTRIBUTING.md
- [ ] Update human and agent documentation (README.md, docs/, skills, CLAUDE.md) — no
      user-facing behavior or configuration changes to document

## Tests

- New `test/cpp/scan/test_duckdb_native_deferred_walk.cpp` (`[duckdb_native_deferred_walk]`,
  CPU-side, file-backed duckdb): deferred construction skips the walk;
  `ensure_metadata_prepared()` produces a plan identical to the eager one (row groups, starts,
  counts, filter-stat pruning); unsupported types still refuse at construction in both modes;
  overflow-string refusal placement per mode; a failed walk is retried rather than latched;
  deferred split claims match eager claims.
- **AC-12 lifecycle-test adaptation** (`test/cpp/integration/test_query_lifecycle_slot.cpp`):
  the runtime-window parser required a "Creating sirius physical plan" line inside every
  execution window. With plan reuse the first execution's plan is created at finalize, so the
  parser now also accepts the plan-reuse line as plan provenance. The acceptance criterion
  itself is unchanged — the id-zero check and the cross-execution operator-id-sequence-equality
  check still verify that operator ids restart for every repeated execution.
- Local: `[duckdb_native_deferred_walk]` + `[duckdb_native_walker]` 10 cases / 101 assertions
  pass; full default suite green (2828 cases, 32.7M assertions).

## References

### DIFF
--- a/src/scan_manager/sirius_scan_manager.cpp
+++ b/src/scan_manager/sirius_scan_manager.cpp
@@ -663,6 +663,9 @@ void sirius_scan_manager::prepare_for_query(const sirius::planner::query& query,
       _scan_op_order.push_back(op);
       continue;
     }
+    // This scan reads disk, so it needs any walk the ingestible deferred. Must run here on the
+    // query thread: GetPartitionStats touches ClientContext/LocalStorage.
+    op->get_ingestible().ensure_metadata_prepared();
     auto provider = std::make_unique<split_provider>(
       op->get_ingestible(),
       [this](std::string_view file_path) -> std::shared_ptr<io::sirius_ioctx> {
--- a/src/sirius_context.cpp
+++ b/src/sirius_context.cpp
@@ -1432,13 +1432,16 @@ RebindQueryInfo SiriusContext::OnFinalizePrepare(ClientContext& context,
     } catch (NotImplementedException&) {
       plan_is_copyable = false;
     }
+    // Hand the validated plan over instead of discarding it, so the first execution can skip
+    // an identical rebuild.
+    duckdb::unique_ptr<sirius::op::sirius_physical_operator> validated_sirius_plan;
     if (plan_is_copyable) {
-      planner.create_plan(std::move(validation_plan));
+      validated_sirius_plan = planner.create_plan(std::move(validation_plan));
     } else {
       // Validate by consuming the freshly re-planned logical_plan; the
-      // PhysicalSiriusExecution operator will re-plan again at execute time
-      // using the SQL string we cached above.
-      planner.create_plan(std::move(logical_plan));
+      // PhysicalSiriusExecution operator re-plans from the SQL string we
+      // cached above when the one-shot validated plan has been consumed.
+      validated_sirius_plan = planner.create_plan(std::move(logical_plan));
       logical_plan.reset();  // signal PhysicalSiriusExecution to use the SQL replan path
     }
 
@@ -1456,14 +1459,15 @@ RebindQueryInfo SiriusContext::OnFinalizePrepare(ClientContext& context,
 
     // Create a new DuckDB PhysicalPlan containing our custom operator.
     auto new_physical_plan = make_uniq<PhysicalPlan>(Allocator::Get(context));
-    auto& sirius_op =
-      new_physical_plan->Make<sirius::transparent::PhysicalSiriusExecution>(std::move(logical_plan),
-                                                                            current_query_sql,
-                                                                            prepared.types,
-                                                                            prepared.names,
-                                                                            std::move(cpu_fallback),
-                                                                            plan_reads_s3,
-                                                                            0);
+    auto& sirius_op        = new_physical_plan->Make<sirius::transparent::PhysicalSiriusExecution>(
+      std::move(logical_plan),
+      current_query_sql,
+      prepared.types,
+      prepared.names,
+      std::move(cpu_fallback),
+      plan_reads_s3,
+      0,
+      std::move(validated_sirius_plan));
     new_physical_plan->SetRoot(sirius_op);
 
     // Replace the DuckDB CPU physical plan.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -829,6 +829,7 @@ set(TEST_SOURCES
     test/cpp/scan/test_duckdb_insert_delta.cpp
     test/cpp/scan/test_duckdb_mvcc_visibility.cpp
     test/cpp/scan/test_duckdb_native_batch_coalescer.cpp
+    test/cpp/scan/test_duckdb_native_deferred_walk.cpp
     test/cpp/scan/test_duckdb_native_empty_split_decode.cpp
     test/cpp/scan/test_duckdb_native_host_backed_decode.cpp
     test/cpp/scan/test_duckdb_native_walker.cpp
--- a/src/include/op/scan/duckdb_native_gpu_ingestible.hpp
+++ b/src/include/op/scan/duckdb_native_gpu_ingestible.hpp
@@ -36,6 +36,7 @@
 #include <atomic>
 #include <cstddef>
 #include <memory>
+#include <mutex>
 #include <span>
 #include <string>
 #include <vector>
@@ -69,6 +70,10 @@ class duckdb_native_ingestible_table_info : public op::scan::ingestible_table_in
   /// dynamic path).
   std::shared_ptr<sirius::op::sirius_dynamic_filter_set> sirius_dynamic_filters;
   std::size_t approximate_batch_size = sirius::config::DEFAULT_SCAN_TASK_BATCH_SIZE;
+
+  /// Skip the constructor's metadata walk: a pin-served scan takes its splits from the pinned
+  /// entry. ensure_metadata_prepared() runs the walk if the scan falls to disk after all.
+  bool defer_metadata_walk = false;
 
   duckdb::DataTable* storage     = nullptr;
   duckdb::ClientContext* context = nullptr;
@@ -161,6 +166,19 @@ class duckdb_native_gpu_ingestible : public op::scan::gpu_ingestible {
 
   std::unique_ptr<batch_coalescer> create_batch_coalescer() const override;
 
+  void ensure_metadata_prepared() override;
+
+  [[nodiscard]] bool metadata_walk_pending() const noexcept
+  {
+    return _walk_deferred && !_walk_ready.load(std::memory_order_acquire);
+  }
+
+  /// Call only when !metadata_walk_pending().
+  [[nodiscard]] duckdb_native_walk_plan const& walk_plan_for_testing() const noexcept
+  {
+    return _plan;
+  }
+
   [[nodiscard]] bool has_processed_all_metadata() const override;
 
   metadata_scan_task_t next_split_provider(io::ioctx_resolver resolve) override;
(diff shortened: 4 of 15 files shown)


src/sirius_context.cpp
```diff
@@ -1432,13 +1432,16 @@ RebindQueryInfo SiriusContext::OnFinalizePrepare(ClientContext& context,
...
+    // Hand the validated plan over instead of discarding it, so the first execution can skip
+    // an identical rebuild.
+    duckdb::unique_ptr<sirius::op::sirius_physical_operator> validated_sirius_plan;
     if (plan_is_copyable) {
-      planner.create_plan(std::move(validation_plan));
+      validated_sirius_plan = planner.create_plan(std::move(validation_plan));
     } else {
       // Validate by consuming the freshly re-planned logical_plan; the
-      // PhysicalSiriusExecution operator will re-plan again at execute time
-      // using the SQL string we cached above.
```
kevkrist, line 1440: I'm not totally sure what this SlotGuard above is doing, but it seems that it freezes the scan manager's pin registry while this plan is created. The plan then is executed later outside the SlotGuard. Is it possible that, in the window between this `SlotGuard plan_window` and execution, the pin registry state has changed? Thinking about concurrency. Can an unpin between windows leave stale data on the plan that then misses the cache?

kevkrist (commented):
1 potential issue

bwyogatama (approved):

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -1402,6 +1405,7 @@ void sirius_scan_manager::insert_pinned_entry(
   }
 
   _pinned_entries[name] = std::move(entry);
+  bump_pin_registry_epoch();
```
kevkrist, line 1408: If someone pins a new column that goes through the same-row-count merge path above, which doesn't reach this epoch bump, that gets narrowed between `OnFinalizePrepare()` and execution-time, the plan won't get rebuilt and won't take advantage of the narrowing (not a bug -- the column will just get upcast immediately). Probably not a problem but worth noting (this only applies to Parquet I think). If you think it's valuable, could be fixed by conservatively bumping the epoch during merging, or, if the info is there, bumping epoch only when merging narrowed columns.

src/scan_manager/sirius_scan_manager.cpp
kevkrist, line 1383: I suppose it's possible for a throw to occur here between this erasure and the construction and insertion of a new entry. In that case, the old pin is gone, but the epoch hasn't changed. Not sure if it's a good fix but could increment the epoch before erasing?

src/sirius_context.cpp
felipeblazing: Good catch — this is exactly the window the epoch mechanism (now e563ad03, originally 0666a0df) closes. `OnFinalizePrepare` stamps `pin_registry_epoch()` next to the validated plan, and `PhysicalSiriusExecution::GetDataInternal` re-reads it inside its own slot window: if any pin/unpin landed between the two windows the plan is discarded and rebuilt against what that window actually sees ([physical_sirius_execution.cpp:211](https://github.com/sirius-db/sirius/blob/94ed2a0d/src/transparent/physical_sirius_execution.cpp#L211)). Covered by `test_pin_registry_epoch.cpp` (injected registry change → `discarding finalize-validated Sirius plan`, results still correct).

src/scan_manager/sirius_scan_manager.cpp
felipeblazing: Agreed it's a missed optimisation rather than a bug, but the header claimed "bumped on every mutation" and this path broke that, so I fixed it in 94ed2a0d: the six explicit bumps are replaced by a private RAII `pin_registry_mutation_scope` declared first in every registry-mutating member ([sirius_scan_manager.hpp:798](https://github.com/sirius-db/sirius/blob/94ed2a0d/src/include/scan_manager/sirius_scan_manager.hpp#L798)), so the merge branch's early `return` bumps too. Also picks up `attach_proven_unique_columns`, which dev added after this branch forked and which the comparison-join planner reads at plan time. New unit test `[scan_manager][pin_registry_epoch]` pins the merge path.

src/scan_manager/sirius_scan_manager.cpp
felipeblazing: You're right, and this one is a real (if narrow) hole: the column-count mismatch throw in the rebuild branch lands after the erase, so the pin was gone while the epoch said nothing changed — a validated plan built while the table was pinned would then be reused against an unpinned table, skipping the plan-time visibility guards the unpinned path relies on. Fixed in 94ed2a0d with the RAII scope guard (bumps on every exit path incl. throw); the asymmetry is deliberate — a spurious bump costs one plan rebuild, a missed one runs a stale plan. Unit test `a throw after the stale entry is erased still moves the pin-registry epoch` reproduces the exact sequence (8-row pin, then a 4-row re-pin declaring two columns but carrying one).

kevkrist (approved):

PR #1549, opened 2026-08-13
perf(aggregate): hash-based label assignment for the COLLECT_SET group-key encode

felipeblazing:
# perf(aggregate): hash-based label assignment for the COLLECT_SET group-key encode

## Summary

Replaces `cudf::encode` with a hash-based drop-in (`hash_encode`) in the sorted-groupby
COUNT(DISTINCT) label-encode path of `gpu_aggregate_impl::local_grouped_aggregate`.
TPC-H q16 at SF1000: **-38% phase-stable**; full suite **~-1.2%** (q16 is essentially all
of it). Output is **bit-identical** to `cudf::encode`.

## Mechanism

The COUNT(DISTINCT) sorted-groupby path collapses multi-column group keys into one
radix-sortable INT32 label via `cudf::encode`, whose per-row index assignment is a
`thrust::lower_bound` binary search driven by a lexicographic row comparator:
O(N log G) comparator calls of random access over every key column. On TPC-H SF1000 q16
(59.4M rows x [p_brand, p_type, p_size], G~18k) that single kernel measured 118.8 ms —
41% of the query.

`hash_encode` produces exactly what `cudf::encode` produces, at hash-probe cost:

1. `cudf::distinct` — hash-based distinct set (NULL==NULL / NaN==NaN, identical row
   identity to `cudf::encode` and to cudf's sorted groupby under `null_policy::INCLUDE`);
2. sort at **distinct** cardinality only (ascending, NULLs last — the exact
   `cudf::encode` output ordering; no stability concern, distinct rows have no ties);
3. `cudf::distinct_hash_join::left_join` — one hash probe per input row returns each
   row's index into the sorted distinct table, in probe-row order (no scatter needed);
4. a min-reduce guard: an unmatched probe row would carry the negative `JoinNoMatch`
   sentinel and silently corrupt the post-aggregate key gather, so fail loudly (the
   caller's existing multi-column-key fallback path remains intact).

The existing gates are unchanged: only large inputs, only multi-column / non-radix keys,
and only when the HLL group-cardinality estimate says label-encoding wins.

## Measured (GB300, SF1000, master DB read-only, host-tier compressed pins)

- Single-query A/B, 7 reps, median of last 6: **q16 0.2883 s → 0.2155 s (-25.3%)**;
  27,840-row result byte-identical to the baseline binary.
- Full varied power-only run vs `tpch_power_20260813_102701` baseline:
  **q16 0.312 s → 0.193 s power pass (-38.1%)**, phase-stable at 0.192–0.193 s across
  clean/power/postrf2 (-105..-119 ms each).
- Suite: clean pass **-1.27%**, postrf2 **-1.14%** (q16 is ~all of it); the power pass
  was polluted by q9's known RF1-delta spill-cliff trip (+146 ms — the COLLECT_SET-gated
  path cannot touch q9).

## Correctness evidence

- Output contract is differential-tested against `cudf::encode` as the oracle
  (`test/cpp/operator/aggregate/test_hash_encode.cpp`, tag `[hash_encode]`):
  q16-shaped string keys with duplicates, NULL groups sorted last, NaN collapse,
  empty/single-row/all-identical inputs, and a 200k-row randomized large-cardinality
  differential including the all-distinct case.
- End-to-end COUNT DISTINCT run through `local_grouped_aggregate` at the 2^20-row
  label-encode gate scale, validating group-key recovery and per-group sets.
- Benchmark output byte-identical to the baseline binary at SF1000.

## Knobs

None added. The change is inside the existing label-encode path and keeps its gates
(input-size gate, non-radix multi-column key gate, HLL cardinality-ratio gate). The
runtime `JoinNoMatch` min-reduce check throws into the pre-existing non-label fallback
rather than producing wrong results.

## Files

- `src/op/aggregate/gpu_aggregate_impl.cpp` — `hash_encode` + call-site swap
- `src/include/op/aggregate/gpu_aggregate_impl.hpp` — declaration + contract docs
- `test/cpp/operator/aggregate/test_hash_encode.cpp` — new differential tests
- `CMakeLists.txt` — register the test source

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -761,6 +761,7 @@ set(TEST_SOURCES
     test/cpp/operator/test_build_probe_scheduling.cpp
     test/cpp/operator/test_cross_schedule.cpp
     test/cpp/operator/aggregate/test_gpu_merge_impl.cpp
+    test/cpp/operator/aggregate/test_hash_encode.cpp
     test/cpp/operator/aggregate/test_physical_grouped_aggregate.cpp
     test/cpp/operator/aggregate/test_physical_grouped_aggregate_count_distinct.cpp
     test/cpp/operator/aggregate/test_physical_grouped_aggregate_merge.cpp
--- a/src/include/op/aggregate/gpu_aggregate_impl.hpp
+++ b/src/include/op/aggregate/gpu_aggregate_impl.hpp
@@ -91,6 +91,33 @@ class gpu_aggregate_impl {
     rmm::cuda_stream_view stream,
     cucascade::memory::memory_space& memory_space,
     const telemetry::batch_telemetry_info& telemetry_info = {});
+
+  /**
+   * @brief Hash-based drop-in replacement for `cudf::encode`.
+   *
+   * Returns exactly what `cudf::encode(keys)` returns: the distinct rows of `keys` in
+   * ascending lexicographic order (NULLs last, per column), plus one INT32 index per input
+   * row into that distinct table (`distinct[indices[i]] == keys[i]`). Row identity follows
+   * `null_equality::EQUAL` / `nan_equality::ALL_EQUAL`, matching both `cudf::encode` and
+   * cudf's sorted groupby under `null_policy::INCLUDE`.
+   *
+   * The difference is the cost model: `cudf::encode` assigns indices with a per-row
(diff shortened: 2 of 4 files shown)


joosthooz:
This is more or less the same optimization as #1605. I'm closing this one because the other is slightly more developed and tested.

PR #1550, opened 2026-08-13
perf(join): prove DELIM_GET build-side uniqueness — skip the count pass on unique build keys

felipeblazing:
# perf(join): prove DELIM_GET build-side uniqueness — q2/q22 count-pass skip

## Summary

Teaches the planner's build-key uniqueness proof (`prove_unique_columns`) that a
`LOGICAL_DELIM_GET` build side is unique by construction, so the hottest PK-shaped
TPC-H joins take the single-pass `cudf::distinct_hash_join` fast path at plan time.
Measured at SF1000: **q22 -24%, q2 -9%**, results **byte-identical**.

The production change is 16 lines in one planner file.

## Mechanism

The distinct_hash_join fast path (#558) skips cudf `hash_join`'s cuco count pass, but
its plan-time uniqueness proof only knew PRIMARY KEY constraints and GROUP BY outputs.
The TPC-H benchmark schema declares no PKs, so the proof never fired on the SF1000
suite's hottest PK-shaped joins: q2's partsupp ⋈ delim part keys and q22's orders ⋈
delim cust keys both walk the probe twice (count + retrieve) purely to size the output.

This adds a `LOGICAL_DELIM_GET` case to `prove_unique_columns`: a delim scan replays the
enclosing delim join's duplicate-eliminated chunk, which Sirius materializes with a
grouped aggregate over ALL delim columns (globalized by PARTITION_DISTINCT →
DISTINCT_MERGE), so the full output row set is unique by construction — under EQUAL and
UNEQUAL null semantics alike (GROUP BY collapses NULL keys into a single group). Only
the complete column set is claimed; subsets stay refused.

### Relation to #1466 (runtime distinct-build probe)

Complementary, not overlapping. #1466 tests build distinctness at runtime, in
BUILD_PROBE mode only, at the cost of a `cudf::distinct_count` pass (plus a
host-synchronizing readback) per cached build, and it caps out at 128M build rows.
The plan-time DELIM_GET proof (a) also fires in STANDARD mode (q22's shape), (b) skips
the runtime distinct-count pass and its stream sync entirely, and (c) has no build-size
cap. The measured numbers below predate #1466 on the measurement branch; on dev the q2
BUILD_PROBE share may partially overlap with #1466's runtime win, while the STANDARD-mode
and count-pass-avoidance components remain additive.

## Measured (GB300, SF1000)

- **q22 -24%**, **q2 -9%**, results byte-identical to baseline.
- Evidence trail (`tpch_power_20260812_014959_sf1000_s7`): q2 Pipeline 22 join_retrieve
  35.5 ms (29% of q2); q22 count kernel 13.9 ms; both builds are delim-deduplicated key
  sets that the old proof could not claim.

## Correctness evidence

A wrong uniqueness proof would *silently drop matches* (distinct_hash_join retrieves at
most one build row per probe row), so the change ships with two test layers:

- **Plan-time detection** (`test/cpp/planner/test_distinct_hash_join_detection.cpp`,
  tags `[distinct_hash_join][isolated_context]`): the DELIM_GET shape is claimed, and a
  non-unique build alongside it stays refused — asserted per-join by discriminating on
  whether the build side contains a DELIM_SCAN, not by a whole-plan "not all unique".
- **Integration** (`test/cpp/integration/test_gpu_execution_unique_join.cpp`, tags
  `[integration][gpu_execution][join][unique_build_keys]`): GPU-vs-CPU comparison over
  the delim / PK / GROUP BY / non-unique paths, including LEFT joins, empty builds, NULL
  keys, and duplicate probe keys. A wrong proof drops rows, so the GPU result diverges
  from DuckDB's and the comparison fails.

Dropped matches are exactly what the GPU-vs-CPU comparison detects, so it carries the
correctness argument directly — no runtime instrumentation ships to make that work.

An earlier revision of this PR added a `SIRIUS_VERIFY_UNIQUE_BUILD_KEYS=1` device
recount guard in `sirius_physical_hash_join`. It was used during bring-up to validate
the full TPC-H suite (green with the guard enabled) and has been removed: being env-gated
and off by default, it contributed a `getenv` per hash-table build, two exported symbols,
and two cudf/rmm includes in a widely-included header while detecting nothing in any
build anyone actually runs. Its unit tests asserted that `cudf::distinct_count` counts
distinct values, which is cudf's contract rather than this change's.

### Test results

- `[distinct_hash_join]` — 12 cases, 80 assertions, green
- `[unique_build_keys]` — 7 cases, 449 assertions, green
- `[hash_join]` — 37 cases, 172 assertions, green
- `[gpu_execution]` — 690 cases, 689 passed, 1 skipped, 0 failed

Negative control: temporarily removing the `LOGICAL_DELIM_GET` case fails both plan-time
tests. Worth noting because the earlier revision's whole-plan `CHECK_FALSE(all_unique)`
*passed* under that same control — with no proof firing, "not every join is unique" is
trivially true — so it could not have caught the regression it was written for.

## Files

- `src/planner/sirius_plan_comparison_join.cpp` — `LOGICAL_DELIM_GET` case in
  `prove_unique_columns` (the entire production change)
- `test/cpp/integration/test_gpu_execution_unique_join.cpp` — new
- `test/cpp/planner/test_distinct_hash_join_detection.cpp` — extended
- `CMakeLists.txt` — register the new test source

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -722,6 +722,7 @@ set(TEST_SOURCES
     test/cpp/integration/test_gpu_execution_tpcds_nulls.cpp
     test/cpp/integration/test_gpu_execution_tpch.cpp
     test/cpp/integration/test_gpu_execution_tpch_mgpu_audit.cpp
+    test/cpp/integration/test_gpu_execution_unique_join.cpp
     test/cpp/integration/test_gpu_execution_vector_search.cpp
     test/cpp/integration/test_pin_table_host_streaming.cpp
     test/cpp/integration/test_pin_table_merge_columns.cpp
--- a/src/planner/sirius_plan_comparison_join.cpp
+++ b/src/planner/sirius_plan_comparison_join.cpp
@@ -23,6 +23,7 @@
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
 #include "duckdb/planner/operator/logical_aggregate.hpp"
 #include "duckdb/planner/operator/logical_comparison_join.hpp"
+#include "duckdb/planner/operator/logical_delim_get.hpp"
 #include "duckdb/planner/operator/logical_filter.hpp"
 #include "duckdb/planner/operator/logical_get.hpp"
 #include "duckdb/planner/operator/logical_order.hpp"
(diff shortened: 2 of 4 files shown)


mbrobbel (approved):
Nice!

PR #1551, opened 2026-08-13
feat(benchmark): staged refresh — pre-load update sets, CSV parsing off the timed RF path

felipeblazing:
# bench: staged refresh — take the CSV parse off the timed RF path (T1)

**Branch:** `pr/staged-refresh` → base: `dev`
**Default on:** `--staged-refresh` on `test/tpch_performance/tpch_power_throughput.py`; pass
`--no-staged-refresh` for the legacy COPY/read_csv path, which is unchanged.

## Mechanism

The timed refresh functions currently pay dbgen CSV parsing inside the measured window: RF1 is
two `COPY ... FROM '<update set>'` and RF2 embeds `read_csv('delete.n')` twice (once per DELETE),
so the delete key file is parsed twice per pair.

With `--staged-refresh` (the default), every needed update set is pre-loaded into native staging
tables during **untimed prep**, right after the pins:

- `staging_orders_u<n>` / `staging_lineitem_u<n>`: a LIMIT-0 CTAS off the base table clones the
  exact schema, then the *same* COPY statement the legacy RF1 uses fills it — parse semantics are
  identical by construction. Row counts are checked against the files at stage time.
- `staging_delete_u<n>`: the delete keys as BIGINT from one `read_csv`.

The timed refreshes become pure native DML (timing stays submission-to-commit, clause 5.3.7.3):

- RF1: two `INSERT INTO ... SELECT * FROM staging_*`
- RF2: two `DELETE ... WHERE key IN (SELECT orderkey FROM staging_delete_u<n>)` — DuckDB plans
  the IN as a hash semi-join, so the keys are parsed zero times in the timed window.

Staging tables are dropped in untimed teardown (best-effort — a run that already earned its
metrics never loses them to a teardown failure). Validation and post-RF count verification run
unchanged in either mode; the deferred CPU validation child always replays the legacy CSV form,
so final-table-state equivalence is checked on every validated run.

`--refresh-write-config` (defaults to following `--staged-refresh`, so it is also on by default)
additionally sets
`enable_optimistic_write=true` and `preserve_insertion_order=false` on the refresh cursors,
best-effort; pass it alone for the config-only A/B, or `--staged-refresh
--no-refresh-write-config` to isolate staging. Both knobs are recorded in `run_info.txt` and
`metrics.json`.

## Spec legality

TPC-H v3.0.1 **clause 2.5.3.1** explicitly permits providing the SUT with the RF1 data / RF2
delete keys *before* the benchmark, in any implementation language — only **pre-executing** the
refresh functions is banned. This change moves only the data-provisioning (CSV parse) into
untimed prep; the refresh DML itself still executes inside the timed window
(submission-to-commit per clause 5.3.7.3), and the refresh stream's pairs stay sequential
(clause 5.1.2.4).

## Measured

SF1000, GB300: **RF1 10.75 s → ~2.5 s (−75 %)** on the power run. RF2 stops parsing the delete
key CSV twice per pair.

## Testing

- Pure-DuckDB equivalence test: staged and legacy paths produce identical final table state on a
  synthetic orders/lineitem DB with dbgen-shaped update files (insert + keyed delete, staging
  row-count checks, teardown drops verified).
- `black`, `codespell`, and the full pre-commit hook set pass on the changed files.
- `--help` / argparse wiring smoke-tested.

## Scope note

This PR is staged-refresh only. The `bench/sf1000-repro/run-power.sh` `STAGED_REFRESH=1` wrapper
wiring rides with the broader official power/throughput harness PR (the wrapper script does not
exist on `dev`); the harness PR's `--update-set-offset` interplay (power set ≠ 1 under database
evolution) is likewise re-applied there.

### DIFF
--- a/test/tpch_performance/CLAUDE.md
+++ b/test/tpch_performance/CLAUDE.md
@@ -354,6 +354,26 @@ sqrt(Power · Throughput)`.
   following queries from `pinned base + insert delta − delete mask`, with no CHECKPOINT between a
   refresh and the queries that observe it. The delta is re-decoded and the mask re-applied per
   query, so the post-refresh passes measure a stable recurring cost.
+- `--staged-refresh` (default on; `--no-staged-refresh` restores the legacy path) takes the CSV
+  parse off the timed refresh path. In untimed prep (right after the pins), every needed update
+  set is loaded into native staging tables in the scratch DB — `staging_orders_u<n>` /
+  `staging_lineitem_u<n>` via a LIMIT-0 CTAS off the base table plus the *same* COPY statement the
+  legacy RF1 uses (parse semantics identical by construction), and `staging_delete_u<n>` holding
+  the delete keys as BIGINT from one read_csv. The timed RF1 becomes two
+  `INSERT INTO ... SELECT * FROM staging_*` and RF2 two
+  `DELETE ... WHERE key IN (SELECT orderkey FROM staging_delete_u<n>)`, which DuckDB plans as a
+  hash semi-join (verified with EXPLAIN on CPU) — the delete key CSV is no longer parsed twice
+  per pair. Staging tables are dropped in untimed teardown. **Spec legality**: TPC-H v3.0.1
+  clause 2.5.3.1 explicitly permits providing the SUT with the RF1 data / RF2 keys before the
+  benchmark in any implementation language — only *pre-executing* the refresh functions is
+  banned; RF timing remains submission-to-commit (clause 5.3.7.3) and the refresh stream's pairs
+  stay sequential (clause 5.1.2.4). The mode also applies `enable_optimistic_write=true` and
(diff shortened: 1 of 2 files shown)


wmalpica (approved):

test/tpch_performance/CLAUDE.md
```diff
@@ -354,6 +354,25 @@ sqrt(Power · Throughput)`.
   following queries from `pinned base + insert delta − delete mask`, with no CHECKPOINT between a
   refresh and the queries that observe it. The delta is re-decoded and the mask re-applied per
   query, so the post-refresh passes measure a stable recurring cost.
+- `--staged-refresh` (default off) takes the CSV parse off the timed refresh path. In untimed
```
bwyogatama: why don't we just enable this at all time?

felipeblazing: ok will do that

PR #1552, opened 2026-08-13
perf(spill): fix the partition-spill cliff; chunk and pipeline downgrade/offload copies

felipeblazing:
# spill: fix the partition-spill cliff and monolithic downgrade copies (T8)

**Branch:** `pr/spill-cliff-chunked-offload` → base: `dev` (`3a0ef61f`). No stacking dependency.

## Problem

The q9-class RF1 spill cliff: +0.1 % delta rows tipped two extra ~4.5 GB partition round-trips
through host (D2H 57→75 GB / H2D 75→93 GB, post-RF1 +117 ms), delivered as single 18–28 GB
blocking copy submissions on an un-instrumented thread. Two independent causes:

1. The builtin cucascade GPU→HOST fast converter submits an entire batch's D2H copies as one
   monolithic batched call plus a blocking sync.
2. Monitor-issued downgrades force-flush the whole trigger→stop band (≥ 12 GB per marginal
   crossing on the SF1000 GB300 config) and spill whole partitions in policy order, so a
   marginal working-set increase becomes a multi-GB spill.

## Change

1. **Chunked, pipelined GPU→HOST spill copies** (`src/data/spill_chunked_converters.cpp`,
   `chunked_spill_copy.hpp`): `SiriusContext::initialize` replaces the builtin converter with a
   Sirius-side transcription that flushes D2H submissions every `~copy_chunk_bytes` (default
   1 GiB; yaml `executor.downgrade.copy_chunk_bytes`, `0` keeps the builtin) while the column
   tree is still being walked — chunk k's DMA overlaps chunk k+1's prep. Produces a
   byte-identical `host_data_representation` (same self-describing column_metadata layout); the
   HOST→GPU restore direction deliberately stays on the unchanged builtin converter. Adds NVTX
   ranges to the downgrade worker's conversions (no timing instrumentation is compiled into the
   production path; measurement is via the profiler).
2. **Overflow-proportional spill sizing** (`downgrade_executor.cpp`, new `spill_policy.hpp`):
   - Monitor requests (yaml `executor.downgrade.overflow_proportional_spill`, default true) stop
     as soon as live pressure drops back below the *trigger* threshold instead of flushing down
     to the stop threshold.
   - Byte-targeted requests carry `target_bytes`: dispatch stops once planned (freed +
     in-flight) bytes cover the target, bounding overshoot to < 1 batch at any pool width, and
     the final per-repo pick is best-fit (smallest candidate covering the remaining deficit)
     instead of the next whole partition in policy order.
   - The live predicate is evaluated before every dispatch, so an already-satisfied request
     spills nothing instead of at least one batch.

Reservation-driven `request_downgrade` keeps its dispatch-until-satisfied envelope
(livelock-sensitive at 251/256 GB); it only gains the earlier predicate checks. This decouples
spilled bytes from both the trigger→stop band and whole-partition granularity **without**
retuning `hash_partition_bytes` (a measured trade, not a win).

## Measured

SF1000, GB300: **q9 post-RF1 spill cliff eliminated (+117 ms → −71 ms)**; spilled bytes become
proportional to the actual overflow instead of the trigger→stop band.

## Testing

- New unit tests: `test_chunked_spill_copy.cpp` (chunked converter produces byte-identical host
  representations across chunk-size sweeps), `test_spill_policy.cpp` (best-fit picks), extended
  `test_downgrade_executor.cpp` (predicate-stops, target_bytes bounding).
- Full `sirius_unittest` suite green on GB300: 2842 test cases, 32,778,745 assertions.
- Full pre-commit hook set passes on the changed files.

## Related follow-up (not in this PR)

The stack lineage carries `dcc4c50a` "fix(downgrade): wake the scheduler matcher after returning
extracted tasks" — a pre-existing whole-process deadlock (task returned via RAII emits no
`task_available`; deterministic at SF1000 host-pinned q18, 7/7 hangs). It applies to `dev`
independently of this PR and should be its own small PR.

### DIFF
--- a/src/data/spill_chunked_converters.cpp
+++ b/src/data/spill_chunked_converters.cpp
@@ -1,0 +1,333 @@
+/*
+ * Copyright 2025, Sirius Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "data/spill_chunked_converters.hpp"
+
+#include "data/chunked_spill_copy.hpp"
+#include "log/logging.hpp"
+
+#include <cudf/column/column_view.hpp>
+#include <cudf/null_mask.hpp>
+#include <cudf/strings/strings_column_view.hpp>
+#include <cudf/table/table_view.hpp>
+#include <cudf/utilities/traits.hpp>
+#include <cudf/utilities/type_dispatcher.hpp>
+
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/detail/error.hpp>
+
+#include <cuda_runtime.h>
+#include <nvtx3/nvtx3.hpp>
+
+#include <cucascade/cudf/gpu_data_representation.hpp>
+#include <cucascade/cudf/host_data_representation.hpp>
+#include <cucascade/cudf/host_table.hpp>
+#include <cucascade/memory/column_metadata.hpp>
+#include <cucascade/memory/fixed_size_host_memory_resource.hpp>
+#include <cucascade/memory/memory_space.hpp>
+
+#include <cassert>
+#include <cstdint>
+#include <vector>
+
+namespace sirius {
+namespace spill {
+
+namespace {
+
+// Layout planning is a faithful transcription of the cucascade fast converter's planning pass
+// (cucascade/src/cudf/representation_converter_builtins.cpp, convert_gpu_to_host_fast). Keeping
+// the algorithm identical keeps the emitted column_metadata layout byte-identical, so the
+// unchanged builtin HOST->GPU converter can restore from it.
+
+/// @p alignment must be a power of two.
+std::size_t align_up(std::size_t offset, std::size_t alignment) noexcept
+{
+  return (offset + alignment - 1u) & ~(alignment - 1u);
+}
+
+/// True for column types that store element data in a flat device buffer. Nested types carry
+/// their payload in children instead (STRING's chars buffer is handled in plan_column_copy).
+bool column_has_data_buffer(const cudf::column_view& col) noexcept
+{
+  switch (col.type().id()) {
+    case cudf::type_id::STRING:
+    case cudf::type_id::LIST:
+    case cudf::type_id::STRUCT:
+    case cudf::type_id::DICTIONARY32:
+    case cudf::type_id::EMPTY: return false;
+    default: return true;
+  }
+}
+
+/**
+ * @brief Recursively plan the buffer layout for one column, filling in column_metadata.
+ *
+ * @note Assumes the column view has offset == 0 (i.e. the table is not a slice) — same contract
+ * as the builtin converter this replaces.
+ */
+cucascade::memory::column_metadata plan_column_copy(const cudf::column_view& col,
+                                                    std::size_t& current_offset,
+                                                    rmm::cuda_stream_view stream)
+{
+  assert(col.offset() == 0 && "column_view with non-zero offset is not supported");
+
+  cucascade::memory::column_metadata meta{};
+  meta.type_id    = static_cast<int32_t>(col.type().id());
+  meta.num_rows   = col.size();
+  meta.null_count = col.null_count();
+  meta.scale      = 0;
+
+  if (col.type().id() == cudf::type_id::DECIMAL32 || col.type().id() == cudf::type_id::DECIMAL64 ||
+      col.type().id() == cudf::type_id::DECIMAL128) {
+    meta.scale = col.type().scale();
+  }
+
+  if (col.nullable()) {
+    meta.has_null_mask    = true;
+    meta.null_mask_size   = cudf::bitmask_allocation_size_bytes(col.size());
+    current_offset        = align_up(current_offset, 8u);
+    meta.null_mask_offset = current_offset;
+    current_offset += meta.null_mask_size;
+  } else {
+    meta.has_null_mask    = false;
+    meta.null_mask_offset = 0;
+    meta.null_mask_size   = 0;
+  }
+
+  if (col.type().id() == cudf::type_id::STRING) {
+    // Offsets live in child(0) and may be INT32 or INT64 (large strings); chars_size() handles
+    // both widths.
+    if (col.size() > 0 && col.num_children() > 0 && col.data<char>() != nullptr) {
+      auto const chars_bytes = cudf::strings_column_view(col).chars_size(stream);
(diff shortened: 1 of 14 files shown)


test/cpp/data/test_chunked_spill_copy.cpp
```diff
@@ -0,0 +1,187 @@
...
+{
+  std::size_t flushes = 0;
+  chunked_copy_batcher batcher(8, [&](std::span<const copy_op>) { ++flushes; });
+  batcher.add(fake_ptr(0x1000), fake_ptr(0x2000), 4);
+  batcher.flush_pending();
+  batcher.flush_pending();
+  REQUIRE(flushes == 1);
+}
+
+TEST_CASE("register_chunked_spill_converters replaces the GPU->HOST pair", "[chunked_spill_copy]")
```
kevkrist, line 157: Is there a way to test replacement here? It looks like this REQUIRE statements would still pass even with the built-in monolithic converters still installed, rgith?

src/sirius_context.cpp
```diff
@@ -820,6 +822,14 @@ void SiriusContext::initialize(const sirius::sirius_config& config)
   // Managers are created per execution window (begin_execution_window), not here; the registry
   // starts empty and only ever holds entries for in-flight queries.
 
+  // Swap the builtin GPU->HOST spill converter for the chunked one before any executor can
+  // spill. Done here rather than at converter_registry::initialize() because the chunk size
+  // comes from the loaded config, which does not exist yet at extension load.
+  if (sirius::converter_registry::is_initialized()) {
```
kevkrist, line 828: Does the registry exist yet? It looks like the registry isn't initialized until after the context is initialized (happening here) during LoadInternal() at sirius_extension.cpp:2844. Of course, if that's true, the monolithic converters aren't replaced, so it's possible I'm missing something here.

src/downgrade/downgrade_executor.cpp
```diff
@@ -200,6 +204,29 @@ void downgrade_executor::processing_loop()
       target_spaces.push_back(ds);
     }
 
+    // Workers evaluate the predicate after each conversion; evaluating it before each dispatch
+    // too means a request already satisfied by other means (the caller's reservation landing,
+    // the running query freeing memory) spills nothing at all instead of at least one batch.
+    auto predicate_satisfied = [&req]() {
+      if (req->satisfied.load(std::memory_order_acquire)) { return true; }
+      if (req->predicate && req->predicate()) {
```
kevkrist, line 212: nit: consider wrapping the predicate invocation in a try/catch to capture the exception in the std::future in case this predicate someday is throwable.

src/downgrade/downgrade_executor.cpp
```diff
@@ -461,14 +517,30 @@ void downgrade_executor::monitor_loop()
...
+        // GB on a large card) no matter how marginal the crossing was. It stays the byte target
+        // (an upper bound) in both modes.
         size_t amount = _memory_space->get_amount_to_downgrade();
         if (amount > 0) {
           auto req                = std::make_unique<downgrade_request>();
           req->is_monitor_request = true;
-          req->predicate          = [&freed = req->bytes_freed, amount]() {
-            return freed.load(std::memory_order_relaxed) >= amount;
-          };
+          req->target_bytes       = amount;
```
kevkrist, line 528: This is set to `get_amount_to_downgrade()`, which is consumed - stop threshold. If consumed is 3, stop is 1, and trigger is 2, this is 2, and the trigger overflow is 1. The best fit downgrade policy will fall back to policy order if no batch is 2, but I think the intention with this PR is that, if there is a batch of 1, that should be chosen first, right?
So, the question is should this be the trigger overflow consumed - trigger rather than consumed - stop?

src/downgrade/downgrade_executor.cpp
```diff
@@ -245,6 +291,8 @@ void downgrade_executor::processing_loop()
           // have set satisfied while we were blocked waiting for a thread slot.
           if (req->satisfied.load()) break;
 
+          planned_bytes += candidate_bytes;
```
kevkrist, line 294: This is added *before* conversion, which might fail, eg.g., if HOST memory is pressurized, in which case it looks like smaller candidates won't get planned for downgrade with the priority policy (and the same downgrade request will get retried). Maybe this is intended, but should planned_bytes get decremented by candidate_bytes when a conversion fails? Not sure that this would in and of itself cause the policy to choose a smaller batch that might successfully downgrade.

kevkrist (commented):
This is rad!

src/downgrade/downgrade_executor.cpp
```diff
@@ -461,14 +517,30 @@ void downgrade_executor::monitor_loop()
...
+          req->target_bytes       = amount;
+          if (_config.overflow_proportional_spill) {
+            // Stop as soon as live pressure drops back below the *trigger* threshold, so a
+            // marginal crossing spills roughly the overflow instead of the whole band. The
+            // check is live, so memory the running query frees concurrently counts as relief.
+            // Headroom is preserved because reservations need the trigger->limit gap, not the
+            // stop threshold, and the reservation path issues its own targeted requests.
+            req->predicate = [space = _memory_space, &freed = req->bytes_freed, amount]() {
+              return freed.load(std::memory_order_relaxed) >= amount ||
+                     !space->should_downgrade_memory();
```
wmalpica, line 537: I think that _config.overflow_proportional_spill is a good idea in general and we should just always have it on. We dont need to make it a config option.
BUT, at the same time, I think:
`!space->should_downgrade_memory()` 
should become:
`space->should_stop_downgrading_memory()`.
And then the configuration tuning becomes, how close do we want _stop_downgrading_memory_threshold to be to _start_downgrading_memory_threshold
And not a configuration tuning of do we want overflow_proportional_spill on or off

src/data/spill_chunked_converters.cpp
```diff
@@ -0,0 +1,333 @@
...
+  // also makes the call idempotent across context re-initializations.
+  registry.unregister_converter<cucascade::gpu_table_representation,
+                                cucascade::host_data_representation>();
+  registry
+    .register_converter<cucascade::gpu_table_representation, cucascade::host_data_representation>(
+      [chunk_bytes](cucascade::idata_representation& source,
+                    const cucascade::memory::memory_space* target_memory_space,
+                    rmm::cuda_stream_view stream,
+                    cucascade::memory::reservation* reservation) {
+        return convert_gpu_to_host_chunked(
```
wmalpica, line 325: out of curiosity, would this converter eventually make it into cucascade as a default? why, why  not?

wmalpica (changes requested):

src/sirius_context.cpp
bwyogatama: This will be addressed

src/downgrade/downgrade_executor.cpp
bwyogatama: Will do

src/downgrade/downgrade_executor.cpp
bwyogatama: i believe this should be consumed - trigger

src/downgrade/downgrade_executor.cpp
bwyogatama: this will be addressed

bwyogatama (approved):

wmalpica (approved):

PR #1553, opened 2026-08-13
feat(telemetry): sticky per-connection session labels for query grouping

felipeblazing:
## Problem

Every query in a process reports under one telemetry query group,
`{engine}-session-{pid}`. A benchmark's warmup, power, and throughput-stream
phases all collapse into that single bucket.

## Change

`CALL sirius_set_session_label('<label>')` sets a sticky per-connection label;
every subsequent query reports under the query group `{engine}-{label}`,
declared on first use. `''` reverts to the default session group. Unlike
`sirius_set_query_label`, the label is not consumed per query.

The label rides on `sirius_interface` (new defaulted third constructor arg) and
selects the group via `telemetry_context::query_group_id_for`. Both interface
construction sites are covered — the `gpu_execution` table function and the
transparent interception path, which is the one plain SQL takes by default.

```sql
CALL sirius_set_session_label('power');
-- ... queries land in '{engine}-power' ...
CALL sirius_set_session_label('');  -- back to the default group
```

### DIFF
--- a/src/include/sirius_context.hpp
+++ b/src/include/sirius_context.hpp
@@ -150,6 +150,20 @@ class SiriusConnectionState : public ClientContextState {
     return label;
   }
 
+  /// Sets the telemetry query-group label for subsequent queries on this connection.
+  ///
+  /// The label remains active until replaced.
+  /// An empty label restores the default session group.
+  void set_session_label(std::string label)
+  {
+    if (label.empty()) {
+      session_label_.reset();
+    } else {
+      session_label_ = std::move(label);
+    }
+  }
+  [[nodiscard]] const std::optional<std::string>& session_label() const { return session_label_; }
+
   void enter_internal_query() noexcept
   {
     internal_query_depth_.fetch_add(1, std::memory_order_relaxed);
@@ -185,6 +199,8 @@ class SiriusConnectionState : public ClientContextState {
   /// Label set by `sirius_set_query_label`, consumed by the next
   /// sirius_interface construction on this connection.
   std::optional<std::string> pending_query_label_;
+  /// Sticky label set by `sirius_set_session_label`; never consumed.
+  std::optional<std::string> session_label_;
   std::atomic<int> internal_query_depth_{0};
   std::atomic<int> cpu_fallback_depth_{0};
   std::optional<std::shared_lock<std::shared_mutex>> pinned_update_guard_;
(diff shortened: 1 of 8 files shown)


src/include/sirius_context.hpp
```diff
@@ -150,6 +150,17 @@ class SiriusConnectionState : public ClientContextState {
     return label;
   }
 
+  /// Sticky telemetry query-group label for this connection; '' clears it.
```
johanpel: Nit: to clarify what "sticky" means:
```suggestion
  /// Sets the telemetry query-group label for subsequent queries on this connection.
  ///
  /// The label remains active until replaced. 
  /// An empty label restores the default session group.
```

src/include/telemetry/telemetry_context.hpp
```diff
@@ -71,6 +71,10 @@ class telemetry_context {
   [[nodiscard]] const uuid::UUID& worker_id() const { return worker_uuid_; }
   /// The single, session-scoped query group that every query in this context is reported under.
   [[nodiscard]] const uuid::UUID& query_group_id() const { return query_group_uuid_; }
+  /// The `{engine}-{session_label}` query group, declared on first use; empty or
+  /// nullopt falls back to the default session-scoped group. Thread-safe.
+  [[nodiscard]] const uuid::UUID& query_group_id_for(
+    const std::optional<std::string>& session_label) const;
```
johanpel: Nit: UUIDs should be cheap copies, even the one backed by Rust structs here. Should we return a copy here instead so we'll not run into dangling reference problems etc. if we ever implement cleaning up the map?

src/sirius_extension.cpp
```diff
@@ -601,6 +603,8 @@ unique_ptr<FunctionData> SiriusExtension::GPUExecutionBind(ClientContext& contex
   // (connection-scoped: it lives on THIS connection's per-connection state).
   if (auto conn_state = get_sirius_connection_state(context)) {
     query_label = conn_state->take_pending_query_label();
+    // sticky, NOT consumed
+    result->session_label = conn_state->session_label();
```
johanpel: If we use `PREPARE` should we capture the label at execution time or at prepare time?

johanpel (commented):
Nice, thanks. I'd like to clear up one thing before approval (and some nits):

src/include/sirius_context.hpp
felipeblazing: Applied your suggestion verbatim.

src/include/telemetry/telemetry_context.hpp
felipeblazing: Good call — changed to return `uuid::UUID` by value, and dropped the "node-stable map" note on the member since the return no longer depends on it. Verified the type copies fine (the ctor already does `.engine_id = engine_uuid_` and the map stores by value).

src/sirius_extension.cpp
felipeblazing: moved to execution time. I think that's better

johanpel (approved):

PR #1524, opened 2026-08-13
perf: Late materialization (stacked on decompression pushdown)

joosthooz:
## Description
Stacking the work done in #1409 on top of the refactored decompression pushdown PR #1474.

## Checklist
- [x] Read CONTRIBUTING.md
- [x] Cover changes with new or existing tests
- [x] Document configuration changes in code and summarize in the description above
- [x] Update human and agent documentation (README.md, docs/, skills, CLAUDE.md)

PR #1499, opened 2026-08-13
refactor(config): keep diagnostic controls internal

infblueocean:
## Summary

- remove high-level `memory.gpu.track_per_stream_reservation` from YAML
- preserve explicit low-level per-space reservation accounting for diagnostics
- remove four executor `thread_name_prefix` choices from normal YAML
- keep stable internal labels and UUID-backed telemetry identity unchanged
- reject every stale removed path with contextual configuration errors

## Why one PR

These five settings are diagnostic controls, not workload policy. Per-stream
reservation accounting changes debugging detail and remains available through
the explicit low-level space model. Thread-label strings rename OS threads and
Quent instances without changing scheduling, capacity, affinity, or execution.
Keeping stable labels engine-owned makes diagnostics comparable while removing
ordinary choices users cannot tune productively.

This supersedes #1501. Its four field-path regressions and resolved-label checks
are retained alongside #1499's high-level/low-level reservation boundary.

## Contract

- high-level GPU memory policy fixes per-stream reservation tracking off
- explicit low-level `space.gpu[].per_stream_reservation` remains available
- task-creator, pipeline, downgrade, and scan-manager labels retain exact stable
  defaults
- pipeline affinity remains hardware-derived from selected GPU topology; the
  removed YAML `cpu_affinity` path from merged #1494 is not restored
- all five removed high-level paths fail at configuration load
- UUID-backed telemetry identity is unchanged

## Rebased identity

- exact base: `e3e37aec4b19bd3959c87d24c4e3f60bfb8b046b`
- exact head: `ca0acc7742c8704df8cb6e840b59e8f34653ea23`
- candidate tree: `9c814aa42346a6d0cbdc4a647c237189f738c3e6`
- zero-context cumulative diff SHA-256: `c628c6d69124b674fd4b36af759e8cea242cae49334f3f52bb9848a41a4e44c9`
- cumulative diff: 12 paths, 79 insertions, 25 deletions
- reservation predecessor: `caebd91dc366b597f079d863e725edabeff44bcd`
- thread-label predecessor from #1501: `62e48601658ad9304c2c6573d78bdeed3e3e537f`

## Validation

- rebased onto current `dev` after #1494 merged
- range-diff: the SF1000 migration repair remains patch-identical; the first
  commit changes only where required to preserve #1494's topology-derived
  pipeline affinity while also removing the thread-name choice
- the resolved generic pipeline parser accepts only `num_threads`; independent
  tests reject both stale `cpu_affinity` and `thread_name_prefix` keys
[...]

PR #1409, opened 2026-08-04
perf: late materialization v1-v3 — deferred columns via pin-order rowids + group-by-rowid (SF1000 6.912 -> 6.654 s, q10 -45%)

felipeblazing:
## Late materialization v1–v3 (stacked on #1391)

**Stacked on #1391** (fused scan-filter, which stacks on #1371) — this branch contains that full history; **review from `6f7f7041` onward** for this PR's own delta (3 commits: v1, v2, v3).

Transparently defers pinned payload columns past cardinality-reducing operators as a compact pin-order rowid, materializing by gather/sparse-decode only when and where consumed. All behavior is gated (`SIRIUS_EXP_LATE_MAT`, `_V2`, `_V3`) and gate-off is byte-identical.

### Results (TPC-H SF1000, GB300, on top of the 6.918 s fused stack)

| version | headline | disposition |
|---|---|---|
| v1 — stop-port deferral | q10 −48 ms; suite −33..−73 ms across 6 A/B pairs (mean −56) | banked |
| v2 — planner pass + group-by-rowid | **q10 −198 ms (−42.8%); suite 6.912 → 6.677 on one binary (−3.4%)** | banked |
| v2 — compressed width-arbitrage (arm C) | q9 +66 ms vs a −5..−35 band | killed per pre-registration; machinery knob-gated |
| v2 — count-on-deferred (q13) | pooled −2.9 ms at N=4 = noise | killed per pre-registration; knob-gated |
| v3 — FD/composite-key GBR | killed **in analysis** (q18 touchable ceiling 2.4 ms) | machinery shipped dark, fully validated |
| v3 — GBR volume gate | tax did not reproduce (pooled ~0 at N=4) | shipped inert |

**Campaign total (this session's stack): 15.99 → 6.677 s = −58.2%.**

### Architecture
- **Origin tracking**: generation-checked pin handles; global row addressing derived on demand from batch origin ranges (zero cost when off).
- **Two-phase materializer**: `prepare_selection` (shared per selection) + stream-ordered per-column `materialize`; raw no-sort gather for uncompressed origins, chunk-CSR sparse decode (K8/K5s/K6s JIT variants) for compressed; wave-side selection capture on fused batches.
[...]

PR #1554, opened 2026-08-13
feat(benchmark): official TPC-H power/throughput/QphH harness (scoring, clause-2.8 evolution, scratch rollback, nsys mapping)

felipeblazing:
# bench: official TPC-H power/throughput/QphH harness (RF1/RF2, per-query telemetry + nsys)

Tooling-only PR — no engine code. Branch: `pr/tpch-power-throughput-harness`, rebased onto current `dev` (2026-09-16).

## Description

Extends `test/tpch_performance/tpch_power_throughput.py` and the `bench/sf1000-repro` kit into a
harness for **official TPC-H power/throughput scoring** on a file-backed `.duckdb` dataset, with
per-query telemetry and per-query nsys profiling for analysis runs. This harness produced the
SF1000 QphH campaign on GB300 that took **QphH@Size from 5.47M to 7.52M** (reference run in the
commit message: Power@Size 7,165,616 / Throughput@Size 4,951,261 / QphH 5,956,411 on the PR #1409
stack, GPU-vs-CPU validation PASS after both refresh functions).

### Scoring (spec-conformant)

- `Power@Size = 3600·SF / geomean(22 stream-0 query times + T_RF1 + T_RF2)` — the refresh
  functions are timed as spec elements of the power geomean, not warmup. Query times below
  `slowest/1000` are floored per **clause 5.4.1.4** (the 1000:1 spread cap).
- `Throughput@Size = N·22·3600 / measurement_interval · SF` — the interval is wall-clock from
  first stream start to last stream end (**clause 5.3.6.1**), with N concurrent query streams
  (spec permutations) plus one refresh stream running N RF1/RF2 pairs.
- `QphH@Size = sqrt(Power · Throughput)`.
- `--vary-predicates` runs each stream's own qgen substitution parameters, as an official run
  requires; the default fixed parameters keep validation meaningful.
- `--update-set-offset` implements **clause 2.8 database evolution**: one evolving database is
  reused across runs (`--scratch-db <shared> --keep-scratch-db`) — the power run consumes update
  set N+1, throughput N+2..N+streams+1 — instead of restoring a 440 GB pristine copy per
  benchmark run. Validation requires offset 0 (`--update-set-offset` rejects `--validation`);
  the scratch size check accepts the evolved size and reports drift instead of failing.

### Shared-GPU probe gate (`run-power.sh`)

The config reserves ~0.95 of the device at `LOAD`. On a shared box that fails not only while
another workload runs but **for minutes after it exits** — the driver reclaims a dead process's
async-pool backing lazily, during which `nvidia-smi` already reports the memory free but large
allocations still OOM. Gating on `nvidia-smi` therefore passes and the run still dies; the only
reliable gate is the failing operation itself. `run-power.sh` retries a throwaway extension-LOAD
probe (`PROBE_TRIES`, default 20 × 60 s) before committing to the run, on the original
(non-telemetry) config so a Quent capture never sees a second same-named engine.

### Rollback-scratch: WAL-discard restore (`--rollback-scratch`, `ROLLBACK=1`)

Restores the scratch DB after a run **without re-copying** (a 15-minute, 440 GB operation at
SF1000). Recipe — three thresholds keep every refresh mutation confined to the WAL:
`SET checkpoint_threshold='1TB'`, `SET wal_autocheckpoint='1TB'`, and
`SET auto_checkpoint_skip_wal_threshold=2^40` (so bulk writes do not checkpoint-on-commit). The
run then ends via `os._exit` because DuckDB force-checkpoints the WAL into the base file on any
clean shutdown and offers no opt-out; deleting `<scratch>.wal` afterwards (run-power.sh's
`ROLLBACK=1` does it) restores content-pristine state, so every run reuses one copy at offset 0.

Qualified at small scale: 3 mutate/discard cycles returned exact pristine content each time.
**Caveat (orphan drift):** the base file keeps small orphaned-block growth from DuckDB's
optimistic bulk writes — never read, never reclaimed — so the file size drifts up across cycles;
the size check reports the drift, and the copy should be refreshed when it grows large.
Incompatible with `NSYS=1`.

### Per-query nsys (`--nsys-per-query`, `NSYS=1`)

Each sequential power-pass query is bracketed in its own cudaProfilerApi capture range
(`--capture-range=cudaProfilerApi --capture-range-end=repeat::sync`); the throughput phase gets
one whole-interval range (cudaProfilerStart is process-global — concurrent streams must not use
it). Two gotchas are baked in:

- **`repeat::sync`, and never trust report indices**: nsys silently merges adjacent capture
  ranges when profiler_stop/start pairs arrive faster than it finalizes a range, and can drop the
  last range at exit — the surviving `range.<N>.nsys-rep` files are numbered compactly, so
  index-based assignment shifts every pointer after the first merge (observed: 72 files for 89
  planned ranges). `prep_analysis_bundles.py` therefore joins reports to manifest entries by
  **absolute-timestamp overlap** (report capture window from its sqlite export, manifest window
  from the runner-recorded `start/stop_epoch_ns`, falling back to the quent query's Init..Exit
  window), and writes a full accounting to `nsys_range_map.json` — mapped / ambiguous (merged) /
  dropped / no_window / conflict; orphan reports and partial overlaps flagged, nothing silently
  reassigned. A 0-byte sqlite export is treated as failed, not cached.
- **env-clear for the nsys frontend**: the frontend's bundled `libssl.so.3` predates the pixi
  libcurl's `OPENSSL_3.2.0` requirement and dies at startup, so nsys is launched with
  `LD_PRELOAD`/`LD_LIBRARY_PATH` cleared and both restored only for the profiled python.

nsys runs are analysis runs; their metrics are never quoted as scores.

### Validation worker design

Validation cannot share a process with the pinned run: Sirius's host pool is a growing pool
allocator (unpinning returns blocks to the pool, not the OS) and DuckDB sizes its default
`memory_limit` from total system RAM with no knowledge of what Sirius holds — at SF1000 a
CPU-side q9 allocates into a machine already ~320 GB spoken for and gets OOM-killed. The GPU rows
are pickled per phase, and a **fresh child process** that never loads the extension replays
RF1/RF2 on its own copy of the pristine input to reproduce the post-refresh states, then diffs
row-by-row (sorted multisets, absolute tolerance). Related: `--duckdb-memory-limit` (default
32 GB) caps the benchmark connection itself — the unlimited default OOM-killed the pinned process
during pin materialization at SF1000.

### Telemetry and analysis prep

- Every query is labeled (`CALL sirius_set_query_label`, zero-padded `q01`..`q22`) and each phase
  bucketed into its own Quent query group via `CALL sirius_set_session_label` (`warmup`,
  `power_clean`, `power`, `power_postrf2`, `tput_s1..sN`); both calls sit outside the timed
  window. `QUENT=1` derives a telemetry-enabled config per run.
- `prep_analysis_bundles.py <run_dir> <nsys_dir> <quent_dir>` builds per-query analysis bundle
  JSONs (nsys report pointers with `nsys_status`/`nsys_shared_with`, timings, quent UUIDs) and
  pre-exports every `.nsys-rep` to sqlite in parallel.
- `--pin-layout <json>`: mixed-tier pin layouts with validated column coverage (a
  schema-qualified name creates a second entry over the same table for per-column tiering);
  SF1000 reference layout in `bench/sf1000-repro/pin-layout-sf1000.json`.
- `--warmup-pass` burns one discarded pass so JIT/first-touch costs land nowhere; crash-safe
  cleanup keeps results when unpin fails; the perf-stack env (`SIRIUS_EXP_*`, `SIRIUS_PRE_SQL`,
  `LD_PRELOAD`) is recorded into `run_info.txt`/`metrics.json` for provenance.
- `generate_tpch_refresh.sh`: dbgen `upath[128]` overflow fix (stage via `/tmp`).

## Engine dependencies (all in `dev` now)

Everything this harness leans on has landed, and the rebase commit adapts the harness to it:

- **`sirius_set_session_label`** exists in `dev`; the runner still catches `CatalogException` and
  degrades to unlabeled groups on an older engine.
- **Split pin entries** (the `main.orders` host-tier entry in the SF1000 layout) use the
  column-aware `find_pinned_entry_for_duckdb_table(requested_ids)` lookup that is in `dev`. A
  scored run must still have zero `Transparent execution fallback` lines in its `log_dir`.
- **`SIRIUS_EXP_*` gates** (#1524): `run-power.sh` exports the same names and defaults as `run.sh`
  (`SIRIUS_EXP_FUSED_SCAN_FILTER`, `SIRIUS_EXP_LATE_MAT`, `SIRIUS_EXP_LATE_MAT_PIN_UNIQUE_COLS`).
  Late-mat is inert on duckdb pins; fused scan-filter engages on the compressed GPU pins.
- **Config** (#1499): the two diagnostic YAMLs no longer set `thread_name_prefix`, which the loader
  now rejects.
- **Patched libcudf** is optional (`CUDF_SO`), matching `run.sh` after #1524; unset uses the pixi
  libcudf.
- `SIRIUS_POISON_FREES` / `SIRIUS_QUARANTINE_FREES` (used by `verify-poison-concurrent.sh`) are the
  one remaining separate piece; without them that gate degrades to a concurrent stress run with
  coredump capture.

## Reconciliation with #1470

PR #1470 (fadvise cache eviction, benchmark totals) landed in
`test/tpch_performance/performance_test.py`; this PR does not touch that file and keeps dev's
shape. The runner's own `_evict_page_cache` uses the same mechanism (`posix_fadvise(DONTNEED)`,
no sudo) but targets the `.duckdb` scratch/input files with an fsync-first dirty-page path — it
complements rather than duplicates `drop_os_cache`.

## Staged refresh (#1551)

Staged refresh landed in `dev` while this PR was open and is on by default in the runner; the
clause-2.8 offset and the rollback path both pass it through (`rf1/rf2_statements(dir, set,
staged)`), and `run-power.sh` forwards `--no-staged-refresh` for the legacy path. Staging tables
are created after the rollback thresholds are armed, so they live in the WAL and are discarded
with it.

## Rollback on every exit path

The rebase commit also closes a gap in `--rollback-scratch`: an exception mid-run used to reach
interpreter teardown, where the kept-alive connection's destructor checkpoints the WAL into the
base file and consumes the scratch — on exactly the crashed run whose scratch needs restoring.
`os._exit` now fires on every exit path once rollback is armed, and `run-power.sh` reaches its
WAL-discard step even when the runner fails (`|| RC=$?` instead of tripping `set -e`).

## Checklist

- [x] Read CONTRIBUTING.md
- [x] Cover changes with new or existing tests — `py_compile`/`bash -n`/`pre-commit` clean;
      exercised end-to-end at SF1000 (see measured context above) and at small scale for the
      rollback qualification. No unit tests: benchmark tooling, not engine code.
- [x] Document configuration changes — knob table in `bench/sf1000-repro/README.md`, flag docs
      in `test/tpch_performance/CLAUDE.md`.
- [x] Update human and agent documentation (README.md, CLAUDE.md)

## References

Refs #1470 (cache-eviction overlap, reconciled above).


felipeblazing:
Pushed two commits extending the harness with the diagnostics tooling it grew during the memory-corruption hunt:

**`f5df480e` — compute-sanitizer launcher + SF1 pressure config**
- `SANITIZER=memcheck|initcheck|racecheck|synccheck` knob in `run-power.sh`: wraps the whole workload in `compute-sanitizer` with a log file and a distinct error exit code (99). Diagnostic runs only — 10–40× kernel overhead, incompatible with `NSYS=1`.
- `sirius-sf1-memcheck.yaml`: an SF1 *pressure* config. The GPU usage cap is shrunk to 1% so the SF1000 downgrade/eviction churn regime (parked-batch conversions, RF-window eviction storms) fires at SF1 sizes — turning a minutes-scale SF1000 repro into a seconds-scale one where sanitizer overhead doesn't matter.

**`6f97024e` — concurrency verification gates**
- `verify-memcheck-sf1.sh`: one-command sanitizer sweep of the small-scale concurrent repro (SF1, `mode=both`, 3 streams, the pressure config). Every invalid device access self-identifies with a device PC + host allocation/free backtraces.
- `verify-poison-concurrent.sh`: full-scale SF1000 concurrent gate on `sirius-sf1000-gate.yaml` (0.80 usage cap, heavier-than-production downgrade churn) with CUDA exception coredumps wired (`CUDA_ENABLE_COREDUMP_ON_EXCEPTION` + friends) so the *first* kernel to touch bad memory is captured red-handed, not a downstream victim minutes later.
- Both gates refuse to start over foreign GPU work, serialize via `flock` on a lease file, and append PASS/FAIL lines to a ledger; README documents the knob and both gates.

**Why it's worth carrying in the harness PR:** these gates are how the corruption hunt converged — the SF1 memcheck sweep flagged an out-of-bounds shared-memory read inside a vendored CCCL 3.4.0 `DeviceScan` kernel *in a single pass*, after weeks of the same bug hiding behind nondeterministic downstream victims at SF1000.

**Scope note:** `verify-poison-concurrent.sh` exports `SIRIUS_POISON_FREES=1` (and the docs mention `SIRIUS_QUARANTINE_FREES`). Those are env-gated engine debug modes that ship separately in the memory-hardening PR; on an engine without them they are inert, and the script degrades gracefully to a concurrent stress gate with coredump capture. Nothing in this PR depends on engine changes.

bwyogatama (approved):

PR #1555, opened 2026-08-13
perf(scan): per-chunk MVCC gating for fused scan-filter + duckdb-native fused wiring

felipeblazing:
# perf(scan): decode-time membership pushdown for duckdb-native pins

**Rebased onto `dev`.** This PR was originally stacked on #1391 (`exp/fused-scan-filter`), which
@joosthooz reimplemented as #1474 (decompression pushdown) and #1524 (late materialization), both
now merged. The change has been re-derived against the current architecture; the branch is two
commits on top of `dev`, no stacking.

Description rewritten per @ran-yuan-rui's review — the previous one described a regression that
is not reachable in production and credited a backstop that does not cover this path. Details
under *What is live vs. future-proofing* below.

## What this actually does

**1. Native dynamic-filter channel (the live change).** `sirius_gpu_scan_operator`'s ctor and
`sirius_scan_manager::prepare_for_query` resolved the dynamic-filter channel from parquet bind
data only. Every split is stamped with that channel, and the decode-time membership snapshot in
`scan_operator_input::prepare_for_processing` reads it — so on a duckdb-native pin the snapshot
had nothing to read and decode-time membership pushdown never fired.

Ablation-verified: with the native branch removed from the ctor, a native GPU-pinned compressed
scan under a join attaches no membership source at all (`SIRIUS_EXP_FUSED_SCAN_DIAG` shows no
`member(col N)` line). With it, the probe attaches and the decode compacts 20M rows to 4K
survivors, touching 20% of blocks.

**2. Narrowed probe carriers are restored, not declined.** Pin-time compressed materialization
casts each pinned column to the narrowest carrier its values fit, while a dynamic join filter is
published at the key's *native* carrier. All three membership implementations required an exact
type match, so probing a narrowed pinned key returned null — a decline. Where membership was the
only selection source, the decode abandoned its whole filtered pass and fell back to an ordinary
full-width decode.

`compute_mask` now restores the probe first, reusing `can_restore_to` / `cast_through_rep` — the
same predicate and conversion `normalize_physical_schema` applies after decode — so a restored
probe answers exactly as the native column would. Widening only: narrowing a probe could fold
values outside the target carrier onto keys in the set, turning a decline into a false positive.
The restore is a no-op when the carriers already agree, so the matching path is unchanged.

This half is **format-agnostic** and predates this PR; it affects parquet pins too.

**3. Per-chunk MVCC gate.** The decode-side attaches in `cached_databatch_provider` were gated on
`_mvcc_masks.empty()`. The mask set is slot-sized whenever a pinned entry has any MVCC version
state, so set emptiness over-gates: a default (all-visible) slot must still serve its chunk the
pushdown. The gate is now a per-chunk slot probe.

**4. Native `ROW_FILTERED` bypass.** `duckdb_native_gpu_ingestible::post_filter_and_project`
honors the `ROW_FILTERED` contract instead of re-evaluating the filter over rows the decode
already compacted, mirroring the parquet path.

## What is live vs. future-proofing

Per @ran-yuan-rui, and verified:

- **Live:** the native channel wiring (1) and the carrier restore (2).
- **Future-proofing:** the per-chunk static gate (3) and the native `ROW_FILTERED` bypass (4).
  `filter_analysis()` is only overridden by parquet, so a native scan's `pushdown_req` is always
  empty and the static row-selection block never runs; and membership clears
  `covers_whole_filter`, so a native decode never reports `row_filtered` today. Both become
  reachable when duckdb-native grows a `filter_analysis()`. The unit test builds a synthetic
  `pushdown_request` for the same reason.
- **Correction to the previous description:** the `prepare_for_processing` backstop does *not*
  cover membership compaction — membership forces `covers_whole_filter` false, so
  `pushdown_row_filtered` is never set. Safety comes from the provider and decode-time mask
  gates, backed by the keep-mask row-count check in `materialize_table`.

## Measured — TPC-H SF1000, GB300, GPU-tier compressed **parquet** pins, best-of-3

| Arm | Total |
|---|---|
| Baseline (restore ablated) | 6.447 s |
| With the carrier restore | 6.407 s |
| | **−0.62%, within run-to-run noise** |

**22/22 query results byte-identical** between the two arms.

The mismatch is real but rare on this suite: 921 membership sources attach, of which **12 decline
(1.3%)** — 9 abandoning the filtered decode outright on 279M-row lineitem chunks, 3 partial. That
rate is why the suite delta does not clear noise. A synthetic case where a hot key does narrow
shows what one decline costs: 20M rows compacted to 80K survivors once the probe fires, versus a
full decode when it declines.

**Not claimed:** this is not an SF1000 parquet win. The suite is parquet, so it cannot measure the
native channel wiring at all — that path has no SF1000 number yet.

The `Power@Size +10.4%` figure in the previous description was measured on the #1391 stack and
does not carry over to this base.

## Testing

- New unit test (`test_dynamic_filter_publisher.cpp`): native-carrier probe and narrowed probe
  produce the same mask; an unrelated carrier still declines. Verified to fail with the restore
  ablated.
- New unit test (`test_cached_serving_hardening.cpp`): a keep-mask on chunk 0 only ⇒ chunk 1 still
  receives the row-selecting pushdown, chunk 0 does not. Verified to fail against the old
  set-emptiness gate.
- Full C++ unit suite: **2949/2949 test cases pass** (32 785 863 assertions).
- Full pre-commit hook set passes on the changed files.

## Checklist
- [x] Read CONTRIBUTING.md

## References
Supersedes the #1391-based version of this branch. Refs #1474, #1524.


ran-yuan-rui:
Non-blocking, The PR description and commit message seems to me that they were not match the code logic

1. The claimed post-refresh static-pushdown regression is not reachable in production. Static requests come from `filter_analysis()`, which only parquet overrides; duckdb-native always returns an empty analysis. MVCC masks exist only for DuckDB-native pins, and cache entries cannot serve another format. In fact, every production native pin gets a slot-sized mask set from its first cache hit, not only after a refresh. 

2. The live change is the operator-side native dynamic-filter. It lets the existing per-split snapshot attach membership immediately before decode on mask-free compressed native chunks. The provider's prepare-time snapshot is normally empty and is not what enables this path; its new per-chunk gate is defensive.

3. The stated `prepare_for_processing()` backstop does not cover membership compaction. Membership forces `covers_whole_filter` to false, so it never sets `pushdown_row_filtered`. Safety comes from the provider and decode-time mask gates, followed by the keep-mask row-count check in `materialize_table`.

Would you consider rewriting the description and commit message around the native decode-time membership path, marking the static row-selection gate, the native `ROW_FILTERED` bypass, and the synthetic-request test as future-proofing pending a native `filter_analysis()`, ?

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -1571,9 +1582,13 @@ void sirius_scan_manager::prepare_for_query(const sirius::planner::query& query,
...
-      auto const* pq = dynamic_cast<op::scan::parquet_ingestible_table_info const*>(
-        &assignment.op->get_ingestible().table_info());
-      if (pq != nullptr) { dynamic_filters = pq->sirius_dynamic_filters; }
+      auto const& info = assignment.op->get_ingestible().table_info();
+      if (auto const* pq = dynamic_cast<op::scan::parquet_ingestible_table_info const*>(&info)) {
+        dynamic_filters = pq->sirius_dynamic_filters;
+      } else if (auto const* native =
+                   dynamic_cast<op::scan::duckdb_native_ingestible_table_info const*>(&info)) {
+        dynamic_filters = native->sirius_dynamic_filters;
+      }
```
ran-yuan-rui, line 1591: A question about native membership on narrowed compressed pins.

Under the default `enable_compressed_materialization` setting an INT64 key chunk can be stored as INT32 or narrower. With `SIRIUS_EXP_FUSED_SCAN_FILTER=1`, the published filter keeps the planned key type while decode probes the stored carrier; all membership implementations decline a type mismatch. If membership is the only selection source, the batch falls back to ordinary decode. Results remain correct through the downstream join, but this probe provides no compaction.

Was this path measured with a probe key that actually narrows? Is the plan to widen before probing, or document the limitation for now?

This behavior predates the PR and also affects parquet. If deferred, please mention it in the description and point to `SIRIUS_EXP_FUSED_SCAN_DIAG`; partial declines are otherwise not reported.

ran-yuan-rui (approved):
Verified the new commit at this head. The restore targets the filter's key type, and admission guarantees that equals the probed column's native type (bare references on both join sides, cast shapes blocked), so `can_restore_to` is exactly the inverse of pin-time narrowing: signed same-family widening of range-proven values reproduces the original bits, and the refused directions (unsigned, decimal, rep-backed, wider probes) are the right set. The rebase is clean, and the rewritten first-commit message resolves my earlier description comment.

Non-blocking notes for follow-up. The bloom variant was restructured the most, but its only probing tests skip below two visible GPUs, so single-GPU CI never runs it — one direct case shaped like the new in-list test would cover it. The refused widening direction (a same-family wider probe must still decline) is untested; the FLOAT64 case exercises a different family. And the membership push lacks the `probe_storage_type` binding check the zone-map push has — a one-line defense worth adding now that the restore widens on trust.

LGTM.

PR #1557, opened 2026-08-13
perf(dynamic-filter): carrier-native membership probes, mask-aware filtered decode, and key-type generalization (INT8..UINT64, DATE/TIMESTAMP, DECIMAL, STRING, nullable keys)

felipeblazing:
> **2026-09-17: scope extended.** Six commits (`d219a326`..`b352168c`) generalize the
> membership probes to every fixed-width integer, DATE/TIMESTAMP, DECIMAL, STRING and nullable
> keys via a key-domain dispatch layer — see **section 3** below. Head moved `80801eb1` ->
> `b352168c`; full `make test` 3293/3293.
>
> **2026-09-16: rebased onto `dev` at `f54e6a10`** (new head `80801eb1`; the previous tip
> `30fe3dd9` is preserved on the fork's reflog and referenced here). Still a single commit.
> Three files conflicted:
>
> - `CMakeLists.txt` — test-source list; kept both dev's new iceberg/puffin tests and
>   `test_fused_membership_mask.cpp`.
> - `src/cuda/sirius_dynamic_bloom_filter.cu` — include block only. Dev (#1809) dropped the
>   custom `sirius_bloom_policy` in favour of `cuco::default_filter_policy`, so the
>   `<cuco/hash_functions.cuh>` include this PR carried is dead and was dropped too; the
>   `bloom_contains` Bulk kernel is unchanged.
> - `src/compression/simpatico_codegen/src/simpatico_codegen.cpp` — dev landed the T9
>   positional keep mask (#1556) as an extra AND source appended *after* wave 1, while this
>   PR moves the combine *before* the sequential membership cascade. Resolved by making the
>   keep mask a cascade prior: it is uploaded on `s0` before the combine and counts as
>   "another source", so a membership-only chunk *with* a keep mask now takes the sequential
>   pruned path. With no static range/bool8 strip nothing else writes `combined`, so the keep
>   mask lands there directly instead of a separate strip. The "every probe declined" fallback
>   is unchanged and still correct on that path (`keep_mask_applied` stays false, the caller
>   applies the mask itself). A new case in `test_fused_membership_mask.cpp` covers it.
>
> The "Deliberately not carried over" paragraph below is therefore partly stale: the T9
> keep-mask **is** now a prior source; pair (col-vs-col) filters remain out.
>
> Verified on this rebase: `[dynamic_filter]` 241/241, `[fused_scan_filter]` 22/22 (5 fused
> membership cases), `make test` 3257/3257 — 36 `gpu_execution iceberg` cases failed on the
> first pass only because the `iceberg` DuckDB extension was not provisioned on the box;
> all 36 pass after `INSTALL avro; INSTALL iceberg`. `pre-commit run -a` clean.

> **Rebased onto `dev`.** This PR previously sat on the closed fused-scan-filter
> (#1391) + late-mat (#1409) stack — the "wave-2 lineage" the old description
> referred to. Both were reimplemented and merged as #1474 and #1524, so the
> branch is now a **single commit off `dev`** with no base-branch diff
> pollution. The old wave-2 tip is preserved at `a9ac3d88`.

## 1. Carrier-native probes

`compute_mask` on all three membership filters required the probe column's type
to equal the filter set's type. Compressed materialization stores a bounded
`BIGINT`/`INTEGER` key in the narrowest fitting signed carrier, so a decode-time
probe sees `INT32` against an `INT64`-built set. The probes now accept any
signed-integer carrier (`INT8`..`INT64`) and convert per element in-kernel. A
value the key domain cannot represent is a definite non-member, so the
no-false-negative contract holds.

### This replaces `restore_probe_to`, and the two were measured against each other

#1555 fixed the same defect by materializing a widened copy of the probe column
before the lookup. The two are **independent fixes for the same problem, derived
two weeks apart** — neither the wave-2 base nor T2's own branch had
`restore_probe_to`; it appeared during #1555's own rebase. Merge order rather
than merit picked the winner, so I A/B'd them.

Pooled allocator (production hands these calls a pooled `mr`; an unpooled one
charges the restore a driver allocation the real path never pays), both arms
interleaved in one process, median of 15 warmed reps:

| filter | rows | filters/col | A: restore | B: in-kernel | B/A |
|---|---|---|---|---|---|
| hash IN-list | 1M | 1 | 23.6 µs | 19.0 µs | **0.81** |
| hash IN-list | 8M | 1 | 92.1 | 67.7 | **0.74** |
| hash IN-list | 8M | 3 | 256.4 | 200.5 | **0.78** |
| Bloom | 1M | 1 | 19.6 | 16.2 | **0.82** |
| Bloom | 8M | 1 | 67.0 | 46.1 | **0.69** |
| Bloom | 8M | 3 | 178.7 | 117.9 | **0.66** |

In-kernel is faster at **every point measured**, across two independent runs,
and both arms produce identical masks (asserted in the harness). The margin
widens with filters per column because `restore_probe_to` sits *inside*
`compute_mask` and so re-materializes the same column once per filter — which
the post-decode cascade in `dynamic_filter_merge` does routinely, on the default
path, not just under the experimental gate.

Read the pooled numbers as a **lower bound**: the benchmark pool is fresh,
dedicated and uncontended, so production's allocation costs at least this much.

Scale: ~3 µs per million rows, so ~18 ms for a full lineitem probe pass at
SF1000. Real and consistent, but modest in absolute terms — this is not a
headline win, and nothing here recovers the number the old description claimed.

`restore_probe_to` had no other callers and is removed with its helpers.

## 2. Mask-aware probing

`sirius_mask_applicable` gains a prior-keep-mask `compute_mask` overload (packed
1 bit/row); dead rows skip the set/Bloom lookup. The wave-1 orchestrator runs
membership probes sequentially on `s0` after combining the other mask sources,
handing the combined words to each probe as its prior and folding each result
back in, so later probes see earlier survivors. Membership-only chunks keep the
concurrent, prior-free path. The prior is a hint only: ignoring it is sound
because the caller ANDs with that same mask.

**This half is unmeasured and is the open question on the PR.** It trades
concurrency for pruning: a probe that previously ran round-robin on its own pool
stream now serializes on `s0` behind the combine, and only the lookup is pruned
— the full-width key decode is still paid. At high static keep-rate that may not
pay for the lost overlap. **Reasonable to split out** if you'd rather land item 1
on its own measurement and take item 2 separately.

## 3. Key-type generalization: membership probes beyond `INT32`/`INT64` (added 2026-09-17, commits `d219a326`..`b352168c`)

Item 1 fixed the *carrier* mismatch for integer keys. This section removes the *type* gate
entirely. All three membership filters (hash IN-list, small IN-list, Bloom) previously accepted
only `INT32`/`INT64` keys with `null_count()==0`; any other equality key reached the publisher and
was silently assigned no membership filter (`choose_membership_filter` -> `none`). Type gating
lived in three places, none of them the planner: the filters' `supports()`, the publisher's
build-column type-equality check, and the join-edge route gate in the publish plan.

### Dispatch design

Two orthogonal, explicitly enumerated axes instead of a `cudf::type_dispatcher` over the
(key, probe) type pair (which would instantiate hundreds of kernels):

- **Key rep** — the device element type the set / needles / Bloom are built over:
  `int32_t`, `int64_t`, `uint32_t`, `uint64_t`. New host-only header
  `src/include/op/dynamic_filter/dynamic_filter_key_domain.hpp`: `membership_key_family`
  `{signed_int, unsigned_int, date_days, timestamp, decimal, string_hash}`,
  `classify_membership_key`, `membership_key_supported`, `membership_probe_compatible`,
  `membership_build_fits_rep`. These replace every `INT32|INT64` literal in the three
  `supports()`, the publisher, `dynamic_filter_publish_plan.cpp`, and
  `direct_route_admissible`.
- **Probe adapter** — a device functor chosen on the host from `(domain, probe.type())`:
  `integral_probe_adapter<ProbeRep, KeyRep>` (generalizes item 1's `probe_key_convert`;
  signedness-aware, range-checked, sentinel-aware), and `string_hash_adapter`. Kernels
  (`set_contains`, `small_in_list_scan`, `bloom_contains`) take the adapter, a `probe_validity`
  bitmask reader and the optional prior keep-mask, and emit a **non-nullable** `BOOL8` mask.

Instantiation budget: **19 probe kernels per filter kind** (was 8 on dev, 16 for the integer
scaffold, +2 for `__int128` decimal carriers, +1 string). Compile time per `.cu` unchanged
within noise (~30 s each, measured cache-busted).

### What each family does

| Family | Rep | Build side | Probe side | Notes |
|---|---|---|---|---|
| `INT8`/`INT16`/`INT32`/`INT64`, `UINT8`..`UINT64` | i32/i64/u32/u64 | any signed/unsigned integer column; the publisher now **accepts a build column arriving at a narrower carrier** when `can_restore_to` holds (previously the key was skipped, `keys_skipped_type_mismatch`) | any carrier of the same signedness family, converted in-kernel; out-of-domain value = definite non-member | sentinel is `min()` for signed, `max()` for unsigned reps |
| `TIMESTAMP_DAYS` (DATE), `TIMESTAMP_{S,MS,US,NS}` | i32 / i64 | reads the underlying rep; a DATE build column arriving as `INT8`/`INT16` is **restored to `TIMESTAMP_DAYS`** in the publisher (a carrier-typed set would decline the native probe the post-decode cascade emits) | same unit, or `INT8`/`INT16`/`INT32` carriers for DATE; mixed units decline (unreachable anyway: DuckDB inserts a cast and a cast blocks the scan route) | zone maps already supported these and serve as the test oracle |
| `DECIMAL32`/`DECIMAL64`/`DECIMAL128` | i32 / i64 / i64-if-fits | `DECIMAL128` builds are range-probed (`compute_exact_numeric_range`); if the unscaled range fits `int64` the set is `int64`, else **all three membership filters decline** (zone map still publishes) | same scale required; `DECIMAL32`/`64`/`128` carriers all accepted, `__int128` probes range-checked into `int64` | no `__int128` Bloom (deliberately) |
| `STRING` | u64 | `cudf::hashing::xxhash_64` fingerprints materialized once per build | `XXHash_64<string_view>` computed **in-kernel** over a `column_device_view`, no materialized copy; `DICTIONARY32` declines (the fused decode reconstructs dictionary/str_split carriers to `STRING` first) | both IN-lists become **no-false-negatives** for strings (fingerprint collisions keep rows); the sentinel `UINT64_MAX` needs no remap — cuco's insert of its empty key is a no-op and the probe keeps rows whose fingerprint equals it |
| **nullable keys** (any family) | — | `supports()` no longer requires `null_count()==0`; null build keys are compacted out (`drop_nulls`, as Bloom already did); the small IN-list tier gates on *valid* rows; the publisher sizes/tiers on `valid_rows` | null probe rows emit `false` via the `probe_validity` prologue; `copy_bitmask` tails removed | exact, not approximate: null-safe keys are never admitted and the join runs `null_equality::UNEQUAL`, so a NULL never matches on either side |

Not done, on purpose: `FLOAT32/64` (no benchmark uses float keys; would need -0.0/NaN
canonicalization), `BOOL8`, `LIST`/`STRUCT`, `HUGEINT` (its lossy `INT64` mapping is a separate
bug).

### Where the expected TPC-H wins turned out not to exist

TPC-H has **only `INTEGER`/`BIGINT` join keys**, so this section is inert on TPC-H beyond the
narrow-build-carrier acceptance and nullable-key handling. In particular:

- q2's `ps_supplycost = min(...)` is planned as a `FILTER` above a `DELIM_JOIN`, not as a join
  condition — there is no DECIMAL join key to admit.
- q15's `total_revenue = (subquery)` *is* a DECIMAL128 join condition, but its build side is
  `AGGREGATE max(...)` over a `CTE_SCAN`, which `build_relation_is_opaque` /
  `build_subtree_is_filtering` do not recognize as evidence. That is the evidence policy,
  upstream of admission, and is untouched here.

The beneficiaries are TPC-DS. A survey at SF10 on the integer-only scaffold found that of the
21 queries with STRING/DATE/DECIMAL/nullable join keys, only ~7 currently run on the GPU at
all (the rest fall back on UNION / WINDOW / DISTINCT / mixed-inequality joins, or crash: q10
aborts on a MARK-join mode assertion). Those seven *do* show the key being declined today:
q83 (`d_date` -> `none`), q59 (`s_store_id` -> `none`), q30/q81/q64/q69 (nullable INT keys
downgraded IN-list -> Bloom). No TPC-DS timing is claimed in this PR.

**SF1000 TPC-H QphH, same recipe as August, no patched libcudf:** dev `f54e6a10` 7.79M
(Power 10.15M / Tput 5.98M); this branch at `80801eb1` (items 1+2 only) median of 3 clean runs
7.95M (Power 10.45M / Tput 6.04M) — Power +3.3%, inside the ±4.5% run-to-run band. Firmer than
the score: dev hit GPU-OOM retry-cap CPU fallbacks in 3 of 4 throughput phases, this branch in
0 of 3. Per-query the membership-filter consumers moved most (q17 −11%, q8 −9%, q7/q20 −7%,
q19 −6%). The type-generalization commits were not scored separately: they cannot change any
TPC-H plan.

### Tests for this section

`[dynamic_filter]` 277 cases / `[fused_scan_filter]` 27 cases; full `make test` 3293/3293 on an
idle GPU at `b352168c`. New coverage, per family and for all three filters, with and without a
prior mask, against a host oracle and (for DATE/DECIMAL) the lowered zone-map AST as an
independent oracle: cross-carrier identity `INT8..INT64` x `{INT32,INT64}` keys vs. the old
path; `INT8`/`INT16`/`UINT*` keys and unsigned sentinel semantics; `TIMESTAMP_DAYS` x
`{DAYS, INT8, INT16, INT32}` probes and unit-mismatch declines; `DECIMAL{32,64,128}` x
`DECIMAL{32,64,128}` probes, scale mismatch declines, DECIMAL128 fit -> int64 set vs. overflow
-> decline; STRING host XXH64 reference pinned against `cudf::hashing::xxhash_64` over a corpus
covering every XXH64 tail path (empty, 1 KiB, UTF-8, embedded NUL); nullable builds x nullable
probes x 4 carriers, sliced-probe offsets, all-null build/probe; publisher accepting
`INT8`/`INT16` build carriers, `DECIMAL32`-carrier for a `DECIMAL64` key, `INT16`-carrier DATE;
fused decode of `INT8`-carrier, `INT16`-carrier-DATE, `DECIMAL32`-carrier and dictionary-`STRING`
chunks against the widened sets. `docs/super-sirius/dynamic-filters.md` gains a "Key types"
section with this table.

## The original's numbers do not transfer

The old description claimed SF1000 Power **+7.2%**, q19 **−52%**. Those were
measured against a stack where a declining probe *threw* and failed the entire
filtered decode. #1524 changed that to a stand-down to the AND identity, and
#1555 removed the decline for narrowed carriers altogether. **Treat the old
numbers as void.**

## Notes

- **The multi-probe cascade is unreachable in a default build.**
  `SIRIUS_EXP_FUSED_SCAN_MAX_MEMBER` defaults to `1`, so the fold-back never
  executes; the new test arms it to 2. Worth deciding whether that default moves.
- **Small-in-list was not benchmarked** — the A/B covered hash IN-list and Bloom.
- Item 2's decode-time path sits behind `SIRIUS_EXP_FUSED_SCAN_FILTER` (off
  unless set). Item 1 is on the default path via the post-decode cascade.

## Deliberately not carried over

`dev`'s reimplementation dropped pair (col-vs-col) filters and the T9 visibility
keep-mask (still open as #1556), both of which the original also primed priors
from. Priors here are ranges and dictionary-answered equalities only. A
declining probe in the sequential cascade leaves the mask untouched, matching
`dev`'s all-ones stand-down rather than the original's throw.

## Related, not fixed here

The build side has the mirror defect with a worse failure mode:
`dynamic_filter_publisher.cpp` compares the runtime build column against the
plan's recorded logical type and **skips the key entirely** on a mismatch
(`keys_skipped_type_mismatch`), so a narrowed build key publishes no filter at
all — probe-side tolerance cannot help there. Already instrumented via
`dynamic_filter_stats`; worth checking whether that counter fires.

## Tests

- `test/cpp/operator/test_dynamic_filter_probe.cpp` (`[dynamic_filter][probe]`,
  8 cases): heterogeneous carriers, decimal/date/float refusal, sentinel
  conservation, prior-mask all-dead/all-live/patterned, null propagation.
- `test/cpp/scan/test_fused_membership_mask.cpp` (`[fused_scan_filter]`,
  4 cases): prior handed alongside a static range, prior-free membership-only,
  a two-probe cascade proving the fold-back, all-dead prior, INT32 carrier vs
  INT64 set.

Against current `dev`: `[probe]` 8/8, `[fused_scan_filter]` 4/4,
`[dynamic_filter]` 240/240, **full suite 3035/3035** (32,788,590 assertions).
`pre-commit` clean. The A/B harness was scaffolding and is not in the diff.


bwyogatama:
LGTM! I have some minor fix on code comment and code duplication that i pushed directly into your PR

bwyogatama (approved):

kevkrist (approved):

src/op/dynamic_filter/dynamic_filter_publisher.cpp
joosthooz, line 347: This now only works for DATE types, because line 186 has a type-specific check and a cast: https://github.com/sirius-db/sirius/pull/1557/changes#diff-b116a1a995667be9da8605a774a8690745728a9fe421f927e4c9742eab1b4ebdR186
We should probably do some kind of compatibility check here instead of the `==`

src/op/dynamic_filter/dynamic_filter_key_domain.cpp
```diff
@@ -0,0 +1,168 @@
...
+  // unverified column. Reuse the carrier-fit helper so decimal domain, scale, and bounds follow
+  // the same rules as compressed materialization.
+  auto const range = sirius::compute_exact_numeric_range(keys, stream, mr);
+  return range.has_value() &&
+         sirius::numeric_range_fits(cudf::data_type{cudf::type_id::DECIMAL64, keys.type().scale()},
+                                    *range);
+}
+
+// Mirrors dispatch_probe_adapter: the probe carriers each family's adapter accepts.
+bool membership_probe_compatible(membership_key_domain const& domain,
```
joosthooz, line 97: Check if this can be deduplicated with code in src/helper/numeric_narrowing.hpp

src/cuda/dynamic_filter_probe.cuh
```diff
@@ -0,0 +1,381 @@
...
+    case cudf::type_id::DECIMAL64: fn(std::int64_t{}); return true;
+    case cudf::type_id::DECIMAL128: fn(__int128_t{}); return true;
+    default: return false;
+  }
+}
+
+/// The (key domain, probe type) switch. Invokes @p fn once with the adapter that reads @p probe
+/// into KeyT, or returns false (= decline) without invoking it. KeyT must be the rep the domain
+/// was classified to; a rep/family disagreement is unreachable and also declines. Each key family
+/// owns one arm here, mirrored on the host by membership_probe_compatible.
```
joosthooz, line 239: There are now separate host and device implementations that can drift, is there a way to make them have a single source of truth?

src/cuda/dynamic_filter_probe.cuh
```diff
@@ -0,0 +1,381 @@
...
+
+//===----------------------------------------------------------------------===//
+// Element conversion
+//===----------------------------------------------------------------------===//
+
+/// Lossless conversion into the key domain. Widening always succeeds; narrowing only when @p value
+/// is representable, and a non-representable value can never equal a stored key. Both types
+/// share a signedness (the dispatchers below never mix them).
+template <class KeyT, class ProbeT>
+__device__ __forceinline__ bool probe_key_convert(ProbeT value, KeyT& out) noexcept
```
joosthooz, line 97: This might need a test to check if the behavior on device is the same as on the host (can_narrow_to/can_restore_to)

src/op/dynamic_filter/dynamic_filter_key_domain.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include "op/dynamic_filter/dynamic_filter_key_domain.hpp"
+
+#include "helper/numeric_narrowing.hpp"
+
+namespace sirius::op {
+
+// Each key family owns one arm here and a matching adapter arm in
+// cuda/dynamic_filter_probe.cuh's dispatch_probe_adapter. Keep the arms explicit; a
+// cudf::type_dispatcher here would silently widen the correctness surface.
+std::optional<membership_key_domain> classify_membership_key(cudf::data_type build_type) noexcept
```
joosthooz, line 26: See if this can be deduplicated with code in src/helper/numeric_narrowing.hpp, there is also diverging behavior for which timestamps to map to ints

src/op/dynamic_filter/sirius_dynamic_filter.hpp
```diff
@@ -164,32 +170,72 @@ class sirius_mask_applicable {
...
     rmm::device_async_resource_ref mr) const = 0;
+
+  /**
+   * @brief Prior-mask-aware variant: rows the prior keep-mask already killed skip the probe
+   *
+   * @p prior_mask_words is packed 1 bit/row over @p probe's rows (bit `row % 32` of word
+   * `row / 32`, 1 = keep), or null for no restriction. A pruning hint only: ignoring it is sound
+   * because every caller ANDs the result with that same mask.
+   */
+  [[nodiscard]] virtual std::unique_ptr<cudf::column> compute_mask(
```
joosthooz, line 200: This function is duplicated a couple of times, in multiple files, maybe they can all just use a single definition in a base class/common header?

src/cuda/sirius_dynamic_bloom_filter.cu
```diff
@@ -196,11 +229,19 @@ sirius_dynamic_bloom_filter::sirius_dynamic_bloom_filter(cudf::column_view const
...
+  if (!membership_build_fits_rep(keys, stream, mr)) {
     throw std::invalid_argument(
-      "[sirius_dynamic_bloom_filter] unsupported key type (INT32 or INT64).");
+      "[sirius_dynamic_bloom_filter] build keys do not fit the key rep (DECIMAL128 values outside "
+      "int64).");
   }
-  // Keep compacted storage alive until add_async is queued on stream.
+  _domain = *domain;
+  // Null build keys match nothing under the join's null_equality::UNEQUAL, so they are dropped
+  // exactly. Keep compacted storage alive until add_async is queued on stream.
```
joosthooz, line 244: This code (and some additional lines even) appears identically in a couple of places, maybe it can be deduplicated

joosthooz (changes requested):
1 item could be silently dropping probes, have a look and let me know what you think

joosthooz (approved):
Giving my approval to prevent being a blocker while out of office. But I do want you to have a look at the comments to see if they need to be addressed

PR #1559, opened 2026-08-13
feat(build): opt-in RAPIDS sccache distributed build farm support

bwyogatama:
## Summary

Adds an **opt-in** path that routes Sirius compilation through the [RAPIDS sccache fork](https://github.com/rapidsai/sccache), the shared `rapids-sccache-devs` S3 cache, and NVIDIA's RAPIDS distributed build cluster — so build results are shared across machines when running experiments. Mirrors the approach in [rapidsai/rmm#2101](https://github.com/rapidsai/rmm/pull/2101).

Nothing changes unless explicitly enabled: normal `pixi run make` builds (pixi's Mozilla sccache 0.15.0), CI (GHA sccache / `ci-release` ccache), and the vcpkg presets are all untouched. The mode is gated per shell on `SIRIUS_SCCACHE_DIST=1`.

## What's included

- **`scripts/sccache_dist_setup.sh`** — one-time setup: installs the RAPIDS sccache fork side-by-side under `~/.local/share/sirius/sccache-dist/` (never replacing the pixi sccache), checks `gh` auth scopes, installs `gh-nv-gha-aws`, and mints 12h AWS creds into a dedicated file (`~/.aws/credentials` and `~/.config/sccache` are never touched). `--creds-only` refreshes expired creds.
- **`scripts/sccache_dist_env.sh`** — per-shell activation: exports `SIRIUS_SCCACHE_DIST=1` plus the `SCCACHE_*` config (S3 bucket, arch-specific scheduler URL, gh-token auth, dedicated server port 4227 to avoid clashing with the pixi sccache daemon, local-compile fallback when the farm is unreachable).
- **`Makefile`** — when `SIRIUS_SCCACHE_DIST=1`, overrides the preset compiler launchers via command-line `-D` flags and sets `SCCACHE_NO_DIST_COMPILE=1` for configure-time compiler checks; a mode stamp in `build/` re-runs configure automatically when the mode toggles. With the variable unset, generated build commands are identical to before. Adds a `make sccache-dist-status` health-check target.
- **`docs/sccache-dist.md`** (+ pointer in `DEVELOPMENT.md`) — setup, usage, caveats, troubleshooting.

## Testing (aarch64 host, arm64 scheduler)

- Cold build from a fresh worktree: all 1,283 targets, **1,709 successful distributed compiles, 0 failures, 0 S3 errors**.
- Warm rebuild after `rm -rf build`: **1m42s wall, 99.76% cache hit rate** (1,239/1,242; CUDA 100%) — this is what other machines of the same arch get once the cache is populated.
- With the mode off, `make -n` output is byte-for-byte identical to `dev`; toggling the mode on/off correctly triggers exactly one reconfigure.
- Not yet exercised: the `amd64` scheduler path (same code, different `uname -m` mapping — needs an x86_64 host).

## Notes for reviewers

- Requires NVIDIA GitHub org membership + `gh` scopes `gist, repo, read:org, read:enterprise`; external contributors are unaffected (the mode is never on by default).
- First `make` after this lands re-runs the CMake configure step once (new stamp file); nothing recompiles.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,16 +48,28 @@ SCCACHE_DIST_CMAKE_FLAGS := \
 	-DCMAKE_C_COMPILER_LAUNCHER=$(SIRIUS_SCCACHE_DIST_BIN) \
 	-DCMAKE_CXX_COMPILER_LAUNCHER=$(SIRIUS_SCCACHE_DIST_BIN) \
 	-DCMAKE_CUDA_COMPILER_LAUNCHER=$(SIRIUS_SCCACHE_DIST_BIN)
+# Compile through the canonical pixi env's compilers (same absolute path on
+# every machine/checkout) so cache keys match across checkouts: conda's gcc
+# embeds its env path in the specs file, which sccache hashes into every key.
+ifneq ($(wildcard $(SIRIUS_SCCACHE_DIST_CANON_ENV)/bin/nvcc),)
+SCCACHE_DIST_ARCH_PREFIX := $(shell uname -m)-conda-linux-gnu
+SCCACHE_DIST_CMAKE_FLAGS += \
+	-DCMAKE_C_COMPILER=$(SIRIUS_SCCACHE_DIST_CANON_ENV)/bin/$(SCCACHE_DIST_ARCH_PREFIX)-cc \
+	-DCMAKE_CXX_COMPILER=$(SIRIUS_SCCACHE_DIST_CANON_ENV)/bin/$(SCCACHE_DIST_ARCH_PREFIX)-c++ \
+	-DCMAKE_CUDA_COMPILER=$(SIRIUS_SCCACHE_DIST_CANON_ENV)/bin/nvcc \
+	-DCMAKE_CUDA_HOST_COMPILER=$(SIRIUS_SCCACHE_DIST_CANON_ENV)/bin/$(SCCACHE_DIST_ARCH_PREFIX)-c++
+endif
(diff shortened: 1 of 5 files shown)


bwyogatama:
Pushed four follow-up commits that make the shared cache actually hit across checkouts, worktrees, and machines (measured on aarch64: cross-checkout C/C++ hit rate 0% → **85%**, clean build 6-7 min → 4m24s; same-checkout rebuilds hit 99.8% / 1m24s):

- **ad13cbe** canonical pixi env: conda writes the env's absolute path into gcc's specs file, which sccache hashes into every key — so checkout-local `.pixi` envs gave every checkout disjoint keys. Setup now installs one env at a fixed per-user path and dist-mode builds compile through it.
- **34eba59** `SOURCE_DATE_EPOCH=0` (`__DATE__`/`__TIME__` expansions via duckdb's `pcg_extras.hpp` baked the preprocessing wall-clock second into keys) and server-applied `SCCACHE_BASEDIRS` (sccache reads it in the server process at startup, so the value from pixi activation never took effect).
- **5582d15** pre-start the dist server (ninja's autostart stampede fails with EMFILE).
- **d502566** document measured hit rates.

Remaining known gap: whole-`nvcc` CUDA jobs and a small C/C++ tail still key checkout-specifically even though their preprocessed output is byte-identical across checkouts — that residual leak is in the rapidsai/sccache fork's key construction, to be raised upstream.

PR #1564, opened 2026-08-14
fix(io): make the local datasource selector exact

infblueocean:
## Summary

- keep `scan_manager.use_sirius_datasource` as an exact local-read backend selector
- preserve `false` as an explicit KvikIO choice on one configured GPU
- reject `false` when multiple GPUs are configured instead of silently overriding it
- retain Sirius io_uring as the default and mandatory multi-GPU path
- make no automatic fallback or performance-based routing decision
- leave independent `s3://` routing through the REST backend unchanged

## Why this shape

The earlier draft proposed automatic fallback after io_uring construction
failure. Maintainer review correctly challenged both that policy and the
reproducibility of the benchmark used to motivate it. This replacement removes
the fallback entirely and settles KvikIO's role narrowly: it is an explicit
single-GPU comparison/escape backend, not an advisory preference.

The selector now means what the configuration says. `false` either constructs
the KvikIO path on one GPU or fails; Sirius does not silently select a different
backend. Multi-GPU rejects `false` because KvikIO binds file handles to one CUDA
context and is unsafe for the multi-GPU routing contract.

## Contract

- omitted or explicit `use_sirius_datasource: true` selects Sirius io_uring
- explicit `false` selects KvikIO on exactly one configured GPU
- explicit `false` with multiple configured GPUs fails configuration loading
- no io_uring-construction failure is converted into another backend
- `s3://` continues to resolve through REST under either single-GPU local selector

## Reproducibility response for the retained AWS observation

The old `0.08836x` KvikIO/default ratio is **not a design input to this repair**.
Its exact bounded method is retained here so the limitation is explicit:

- stock Sirius `7b69a9cc63dc965c9e6d6098e02f00923b7a0433`
- TPC-H SF100/RG2M; Q1, Q6, Q12; nine Parquet files; `34,312,011,701` bytes total
- AWS `g6.4xlarge` L4, root ext4 on EBS gp3; live recheck: 2,024 GiB,
  3,000 IOPS, 125 MiB/s provisioned throughput
- one configured GPU; 22,945,254,604 GPU bytes; 8 GiB host capacity;
  prefetch cache disabled
- default arm omitted the selector; comparison arm added only
  `use_sirius_datasource: false`
- CPU oracles ran before the GPU matrix and could warm page cache; caches were
  not dropped or reset between cells
- three balanced repetitions per query/arm in fresh processes; 18/18 cells
  were oracle-correct
- statistic: geometric mean of per-query median ratios; Q1 `0.09625`, Q6
  `0.08013`, Q12 `0.08944`, overall `0.08836`; every arm/query
  `(max-min)/median` spread was below `0.05`
- the runner did not set or retain effective `KVIKIO_COMPAT_MODE`,
  `KVIKIO_GDS_THRESHOLD`, or `KVIKIO_NTHREADS` values
- every KvikIO cell reported `/proc/driver/nvidia-fs/devcount` absent and
  `cuFile running in compatible mode`; GDS was not active

The nine synthetic TPC-H file sizes were 1,246,774,692; 21,515,861,203;
219,007,402; 2,965; 6,375,017,202; 630,750,890; 4,244,285,603; 1,595; and
80,310,149 bytes. The runner verified their predeclared relative paths and
SHA-256 values before and after the matrix.

This is therefore an oracle-warmed, compatibility-mode AWS observation with
page-cache and library-default confounding. It supports retaining an exact
selector. It does not identify a causal I/O mechanism or justify automatic
fallback or a universal default.

## Identity

- exact base: `9624f06334d41f1049088185a34d4b16e2a0e29c`
- exact head: `f7485222eb5cdcf267cc32f81632b32c4ed784d4`
- base tree: `fe3bad29a859e6480c901ade52c55725ff2de41d`
- candidate tree: `ce4da3063c7e7e35b0f62b34bfaad9dde7291bb6`
- zero-context cumulative diff SHA-256: `85030ab65083f307a595dbfee2edb60f05c60e55c77b9c07722fe8aeb5bfd822`
- cumulative diff: 7 paths, 65 insertions, 19 deletions

## Validation

- complete replacement diff independently reviewed against the maintainer contract
- `git diff --check` passes
- `scripts/check_orphan_tests.py` passes
- pinned clang-format 20.1.4 dry-run passes
- codespell 2.4.1 passes
- focused configuration tests cover single-GPU selector preservation and
  multi-GPU rejection
- existing routing tests cover `true` to io_uring, `false` to KvikIO, and
  independent `s3://` to REST routing
- predecessor `9f54c7e` exposed a Catch2 v2 matcher-name error in hosted build;
  `4424cfc` repaired the API name, then hosted lint exposed its formatting;
  this head uses repository-standard `Catch::Contains` syntax and passes the
  exact clang-format and codespell hooks locally
- exact-head hosted C++/CUDA build, unit suite, TPC-H snapshot, lint,
  validation, and distribution workflows pass

## Draft dependency

Hosted workflows: Check `32922643672`, Test `32922643546`, Distribution
`32922644019`, Validate `32922728252`. The final replacement head is
author-ready and terminal-green.

This remains draft only for the named external question: whether this
exact-selector-only contract settles KvikIO's comparison-backend role as
requested in review.


infblueocean:
Exact-head validation receipt:

- base: `e885ffea7d0541f1e3493566e2db0ed6e431106f`
- head: `95d0e1871ce248b1ba2b5230616c8291243d73cc`
- tree: `cf3b93aea8d563f25b5688d3d19c304f7b44f165`
- retained mail-patch SHA-256:
  `4fab75818ef5c83add10ba8efbdf93c3fe83fee738f1990966578ba930839857`
- applying that patch to the exact base reproduces the exact candidate tree

Independent source review found one stale in-tree benchmark emission of the
legacy `true` key. The exact head removes it and corrects the two remaining
docs phrases; the reviewer cleared the selection, migration, one-warning
single-GPU fallback, multi-GPU fail-closed, loud startup failure, S3 routing,
and test contracts otherwise.

Terminal hosted jobs:

- lint `94669363276`
- GCC release `94669363262`
- Clang debug `94669363312`
- CUDA 13 build `94669416251`
- CUDA 13 full C++ unit suite + TPC-H snapshot `94669981946`
- aggregate host/CUDA checks `94670341191` / `94673226801`

All passed. This supports the construction-time policy and surface-removal
claim only; it is not a performance claim and does not claim recovery from a
later reactor-start failure.

infblueocean:
Evidence update after this PR opened: exact stock Sirius `7b69a9cc` completed 18/18 oracle-correct, backend-bound cells on both AWS L4/EBS and Brev GCP L4 persistent disk. AWS timing was admissible and measured explicit KvikIO at `0.08836x` the Sirius-io_uring default across Q1/Q6/Q12 (~11.3x faster). Brev independently confirmed backend selection and correctness, but KvikIO Q1/Q6 exceeded the predeclared 5% spread ceiling, so it neither replicates nor refutes the AWS timing result. The current Sirius-first surface removal would delete the only tested escape hatch for that material gap. I have moved this PR back to draft and withdrawn the review request while preserving its green construction-fallback work. The next revision will retain the selector as an expert escape hatch unless a separately validated automatic backend rule replaces it; no new PR or universal performance claim is being made.

ran-yuan-rui:
Two points block this change.

1. The 11.3x result needs a reproducible method.

It conflicts with our other measurements, so please post enough to rerun it: dataset and scale factor, file count and sizes, the full local scan-manager config for each arm, EBS volume type with provisioned IOPS/throughput, whether the volume was pre-warmed, the cache-reset protocol between arms, iteration count, and which statistic produces 0.08836x.

For the KvikIO setting specifically, please include the effective KVIKIO_COMPAT_MODE, KVIKIO_GDS_THRESHOLD and KVIKIO_NTHREADS, and whether GDS was actually active on that host rather than assumed.

The main thing to rule out is the page cache. Sirius device reads use O_DIRECT (backend table in docs/super-sirius/scan.md). KvikIO reads are buffered whenever it runs in POSIX compatibility mode or the request falls under the GDS threshold, so a warm-cache run can compare storage reads against page-cache hits. The cache protocol needs to be explicit before this becomes a design input.

2. KvikIO is a comparison backend, 
KvikIO support was added so the RAPIDS and Sirius IO paths could be compared apple-to-apple;

This PR makes use_sirius_datasource advisory rather than exact: true can select KvikIO after an io_uring construction failure, and false selects io_uring on multi-GPU. A comparison switch needs the opposite property — the requested backend either runs or fails visibly. A multi-GPU KvikIO request should be rejected rather than silently changed behind one warning line.

Please pause this PR until KvikIO's role is settled.

infblueocean:
Thanks — this review changed the proposal substantially.

The draft no longer adds automatic fallback. I replaced it with the narrow
contract your comment called for:

- `use_sirius_datasource: false` remains an exact KvikIO selector on one GPU;
- it now fails configuration loading with multiple configured GPUs rather than
  being silently overridden;
- io_uring construction failure never changes the configured backend; and
- `s3://` routing remains independently bound to REST.

I also rewrote the body with the reproducibility details and the limitation we
found. The retained AWS run was CPU-oracle warmed, did not reset page cache,
did not capture effective KvikIO environment defaults, and every KvikIO cell
reported cuFile compatibility mode with GDS inactive. I am therefore not using
the 11.3x observation as a rationale for fallback or a default change; it only
supports preserving an exact comparison selector.

Replacement identity: base `9624f063`, head `9f54c7e`. The complete
replacement diff passed an independent author-side review plus formatting,
spelling, orphan-test, and diff checks. I am leaving the PR in draft while the
new head's hosted C++/CUDA workflows run and while you confirm whether this
exact-selector-only role resolves the policy question you raised.

infblueocean:
Hosted build on predecessor `9f54c7e` exposed one author-side Catch2 v2 API mismatch (`ContainsSubstring` was unavailable). The replacement at `4424cfc` uses the repository-standard `Catch::Contains`; an independent diff review confirmed this is the only change and does not alter the selector contract. Fresh exact-head workflows are running.

infblueocean:
Final replacement head `f7485222eb5cdcf267cc32f81632b32c4ed784d4` is terminal-green:
Check `32922643672`, Test `32922643546`, Distribution `32922644019`,
and Validate `32922728252`.

The implementation and PR body now match the narrowed contract above: no
automatic fallback, exact single-GPU KvikIO selection, and explicit rejection
of that selector for multiple configured GPUs. I am keeping the PR in draft
only for the requested maintainer confirmation that this settles KvikIO's
comparison-backend role.

PR #1566, opened 2026-08-14
fix(scan): await async reads before freeing their source memory

9prady9:
Needs https://github.com/NVIDIA/cuCascade/pull/184

## Bug

`cudf::cast` and the view-materializing copy are **asynchronous**, but both sites freed the source immediately after enqueuing:

```cpp
columns[i] = cudf::cast(columns[i]->view(), target, stream, mr);  // assignment frees the source
```

A pinned-cache chunk's buffers are stream-ordered on the stream that produced them, not the task stream, so the free is unordered against the still-running read and the block can be handed to the next allocation mid-copy.

Silently wrong scanned values, varying run to run. Row counts and grouping stay correct. On TPC-H SF1000 Q18 this corrupted `sum(l_quantity)`, so `HAVING sum(l_quantity) > 300` emitted 63,430 / 14,879,354 / 163,210,611 / 58,233 rows across four identical iterations.

Only reachable with `enable_compressed_materialization` (default `true`).

## Fix

**`normalize_physical_schema` (the carrier cast)** — rebind the source's deallocation onto the task stream so the free is *enqueued behind* the cast rather than racing it. No copy, no kernel, no host wait; the cast's `column_view` stays valid because rebinding relocates nothing:

```cpp
auto casted = cudf::cast(columns[i]->view(), target, stream, mr);
auto source = cudf::rebind_stream(std::move(*columns[i]), stream);
columns[i]  = std::move(casted);
```

**`owning_table_view::release()` (the view-materializing copy)** — keeps `stream.synchronize()`. There the source belongs to an external owner (a pinned-cache lock held under a read-only lock), so there is no buffer of ours whose deallocation stream could be rebound; `rebind_stream` is explicitly a no-op for that case. Closing it without a host wait needs a consumer-event API in cuCascade — the mirror of `record_writer_event`, letting a reclaimer wait on an in-flight reader — which does not exist today.

The move-based materialization path enqueues no device work and is untouched.

## Checks

New `[owning_table_view][stream_order]` test, deterministic both ways:

| | `cudaStreamQuery` at reclaim | poisoned elements |
|---|---|---|
| without the fix | `cudaErrorNotReady` | 4,194,304 / 4,194,304 |
| with the fix | `cudaSuccess` | 0 |

Full C++ suite: 2,448 passed, 1 skipped.

TPC-H SF500 pinned, per-group sums re-aggregated against the ungrouped total, 9 runs per config:

| config | before | after |
|---|---|---|
| prefetcher on | 6/8 runs wrong | 9/9 correct |
| prefetcher off | 5/8 runs wrong | 9/9 correct |

Both match the `enable_compressed_materialization=false` control.

## Note for reviewers

The cast-site rebind requires the source's *previous* stream to have no in-flight work — guaranteed today by the `stream.synchronize()` after `prepare_for_processing` at `gpu_pipeline_task.cpp:547`. That sync is currently commented as existing "to ensure the timing collected by Quent and logging for preparing the task is accurate", i.e. it reads as telemetry-only. It is load-bearing for correctness; worth re-commenting separately.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -712,6 +712,7 @@ set(TEST_SOURCES
     test/cpp/scan_manager/test_pin_table_multi_gpu.cpp
     test/cpp/scan_manager/test_cached_serving_hardening.cpp
     test/cpp/scan_manager/test_insert_delta_job.cpp
+    test/cpp/scan_manager/test_memory_prefetcher_accounting.cpp
     test/cpp/scan_manager/test_mvcc_mask_job.cpp
     test/cpp/scan_manager/test_pinned_chunk_stats.cpp
     test/cpp/operator/test_physical_grouped_aggregate_merge_mgpu.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -799,6 +799,7 @@ set(TEST_SOURCES
     test/cpp/scan/test_parquet_null_count_pruning.cpp
     test/cpp/scan/test_parquet_scan_sizing.cpp
     test/cpp/scan/test_scan_schema_normalization.cpp
+    test/cpp/scan/test_stream_lineage_item5.cpp
     test/cpp/sirius_extension/test_sirius_read_parquet_cardinality.cpp
     test/cpp/sql/test_sirius_sql_rewrite.cpp
     test/cpp/telemetry/test_telemetry_context.cpp
--- a/src/include/op/scan/owning_table_view.hpp
+++ b/src/include/op/scan/owning_table_view.hpp
@@ -61,14 +61,13 @@ concept no_alloc_materializable = requires(Owner& owner) {
 };
 
 /**
- * @brief An owner that tracks consumer events on the storage behind the view
+ * @brief An owner that tracks reader events on the storage behind the view
  *        (canonically @c cucascade::read_only_data_batch, whose read lock is
  *        what keeps a cached batch alive while the view is consumed).
  */
 template <typename Owner>
-concept consumer_event_recording = requires(Owner const& owner, rmm::cuda_stream_view stream) {
-  owner.record_consumer_event(stream);
-};
+concept reader_event_recording =
+  requires(Owner const& owner, rmm::cuda_stream_view stream) { owner.record_reader_event(stream); };
 
 //===----------------------------------------------------------------------===//
 // my_view
@@ -160,12 +159,9 @@ class my_view {
     return _model->materialize(_selection, stream, mr);
   }
 
-  /// Forward a consumer-event record to the type-erased owner. No-op unless the
-  /// owner is @ref consumer_event_recording.
-  void record_consumer_event(rmm::cuda_stream_view stream)
-  {
-    _model->record_consumer_event(stream);
-  }
+  /// Forward a reader-event record to the type-erased owner. No-op unless the
+  /// owner is @ref reader_event_recording.
+  void record_reader_event(rmm::cuda_stream_view stream) { _model->record_reader_event(stream); }
 
  private:
   struct owner_concept {
@@ -178,8 +174,8 @@ class my_view {
       rmm::cuda_stream_view stream,
       rmm::device_async_resource_ref mr) = 0;
 
-    /// Default no-op for owners that do not track consumer events.
-    virtual void record_consumer_event(rmm::cuda_stream_view) {}
+    /// Default no-op for owners that do not track reader events.
+    virtual void record_reader_event(rmm::cuda_stream_view) {}
   };
 
   template <typename Owner>
@@ -225,9 +221,9 @@ class my_view {
       }
     }
 
-    void record_consumer_event([[maybe_unused]] rmm::cuda_stream_view stream) override
+    void record_reader_event([[maybe_unused]] rmm::cuda_stream_view stream) override
     {
-      if constexpr (consumer_event_recording<Owner>) { _owner.record_consumer_event(stream); }
+      if constexpr (reader_event_recording<Owner>) { _owner.record_reader_event(stream); }
     }
 
     Owner _owner;
@@ -336,13 +332,17 @@ class owning_table_view {
   void select_columns(std::span<const std::size_t> positions) const;
 
   /// Record on the data owner behind the view that reads of the exposed view
-  /// have been ENQUEUED on @p stream (see data_batch::record_consumer_event).
-  /// Call after enqueuing the reads and before this handle — whose owner may
-  /// hold the read lock keeping a cached batch alive — is dropped or
-  /// reassigned, so a later reclaim of the batch is ordered after the reads.
-  /// No-op for owned-table and empty states, and for owners that do not track
-  /// consumer events. Never allocates or syncs.
-  void record_consumer_event(rmm::cuda_stream_view stream) const;
+  /// have been ENQUEUED on @p stream (see
+  /// cucascade::read_only_data_batch::record_reader_event). Call after
+  /// enqueuing the reads and before this handle — whose owner may hold the
+  /// read lock keeping a cached batch alive — is dropped or reassigned, so a
+  /// later reclaim of the batch (which must first acquire mutable access, and
+  /// therefore waits on every recorded reader event) is ordered after the
+  /// reads. No-op for owned-table and empty states, and for owners that do
+  /// not track reader events. Never syncs on success; if event registration
+  /// fails after the reads were enqueued, the owner synchronizes @p stream
+  /// before propagating the error.
+  void record_reader_event(rmm::cuda_stream_view stream) const;
 
   /// Realize a view state into an owned table. No-op if already materialized or
   /// empty. May allocate (copying path) depending on the underlying owner.
--- a/src/include/op/scan/owning_table_view.hpp
+++ b/src/include/op/scan/owning_table_view.hpp
@@ -60,6 +60,16 @@ concept no_alloc_materializable = requires(Owner& owner) {
   { (*owner).release() } -> std::same_as<std::vector<std::unique_ptr<cudf::column>>>;
 };
 
+/**
+ * @brief An owner that tracks consumer events on the storage behind the view
+ *        (canonically @c cucascade::read_only_data_batch, whose read lock is
+ *        what keeps a cached batch alive while the view is consumed).
+ */
+template <typename Owner>
+concept consumer_event_recording = requires(Owner const& owner, rmm::cuda_stream_view stream) {
+  owner.record_consumer_event(stream);
+};
+
 //===----------------------------------------------------------------------===//
 // my_view
 //===----------------------------------------------------------------------===//
@@ -150,6 +160,13 @@ class my_view {
     return _model->materialize(_selection, stream, mr);
   }
 
+  /// Forward a consumer-event record to the type-erased owner. No-op unless the
+  /// owner is @ref consumer_event_recording.
+  void record_consumer_event(rmm::cuda_stream_view stream)
+  {
+    _model->record_consumer_event(stream);
+  }
+
  private:
   struct owner_concept {
     virtual ~owner_concept() = default;
@@ -160,6 +177,9 @@ class my_view {
       std::span<const cudf::size_type> selection,
       rmm::cuda_stream_view stream,
       rmm::device_async_resource_ref mr) = 0;
+
+    /// Default no-op for owners that do not track consumer events.
+    virtual void record_consumer_event(rmm::cuda_stream_view) {}
   };
 
   template <typename Owner>
@@ -203,6 +223,11 @@ class my_view {
         stream.synchronize();
         return out;
       }
+    }
+
+    void record_consumer_event([[maybe_unused]] rmm::cuda_stream_view stream) override
+    {
+      if constexpr (consumer_event_recording<Owner>) { _owner.record_consumer_event(stream); }
     }
 
     Owner _owner;
@@ -309,6 +334,15 @@ class owning_table_view {
   /// order (projection + reorder). Never allocates; demotes an owned table to a
   /// view first if necessary. Throws on out-of-range or duplicate positions.
   void select_columns(std::span<const std::size_t> positions) const;
+
+  /// Record on the data owner behind the view that reads of the exposed view
+  /// have been ENQUEUED on @p stream (see data_batch::record_consumer_event).
+  /// Call after enqueuing the reads and before this handle — whose owner may
+  /// hold the read lock keeping a cached batch alive — is dropped or
+  /// reassigned, so a later reclaim of the batch is ordered after the reads.
+  /// No-op for owned-table and empty states, and for owners that do not track
+  /// consumer events. Never allocates or syncs.
+  void record_consumer_event(rmm::cuda_stream_view stream) const;
 
   /// Realize a view state into an owned table. No-op if already materialized or
   /// empty. May allocate (copying path) depending on the underlying owner.
--- a/src/include/op/scan/owning_table_view.hpp
+++ b/src/include/op/scan/owning_table_view.hpp
@@ -60,11 +60,6 @@ concept no_alloc_materializable = requires(Owner& owner) {
   { (*owner).release() } -> std::same_as<std::vector<std::unique_ptr<cudf::column>>>;
 };
 
-/**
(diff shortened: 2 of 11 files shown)


9prady9:
## Detailed findings — investigation behind this PR

### 1. Where it surfaced

TPC-H **SF1000**, 4 iterations, parquet, 1× GB300, Sirius `e885ffea`. 21/22 queries passed; **Q18 failed validation**.

Sirius returned a correctly sorted, correctly valued top-100 that was *missing rows belonging in it*, back-filled with rows ranking below 100. The rank-100 cutoff dropped 561067.11 → 560441.16 — the signature of a deficient input set reaching `TOP_N`, not of a broken sort.

Note both saved `result.txt` files are truncated DuckDB displays (`100 rows (40 shown)`), so only 40 rows are comparable; within those, 11 rows differ. The true count across the full top-100 is not recoverable from the artifacts.

### 2. What was ruled out

- **`e885ffea` (TopN "gather only the top k")** — the initial prime suspect, exonerated. Telemetry shows `TOP_N` and `MERGE_TOP_N` each ran as exactly **1 task in all 4 iterations**, so there is no multi-partition top-k merge to get wrong. It emitted a correct monotone top-100 *of the deficient input it was handed*. Decisively, iterations 1 and 3 ran identical TopN code and matched DuckDB.
- **Grouping / partitioning / merge** — `MERGE_GROUP_BY (id=6)` emitted bit-identical counts on all four iterations: 750,024,705 + 749,975,295 = exactly 1,500,000,000 groups. Keys are deterministic and correct; only aggregated values are wrong.
- **Downgrades, spilling, memory pressure** — none in any Q18 iteration (min free GPU memory 54–78 GB).

### 3. Narrowing to the value path

`FILTER (id=7)` (the `HAVING`) output, summed across its two tasks per iteration:

| iter | scan source | rows out | correct? |
|---|---|---|---|
| 1 | parquet, unpinned | 63,430 | ✅ |
| 2 | pinned host | 14,879,354 | ❌ |
| 3 | pinned host | 163,210,611 | ❌ |
| 4 | pinned host | 58,233 | ❌ ← saved as `result.txt` |

`GPU_SCAN` read an identical 5,999,989,709 rows every run. Errors go in **both** directions — groups falsely admitted and falsely dropped — which rules out a deterministic logic error such as a decimal scale mistake and points at nondeterministic memory corruption. The one correct GPU iteration was the only unpinned one (`pin_after_iteration: 1`).

### 4. Reproducing it

Two false negatives are worth recording, because either could be mistaken for "fixed":

1. **SF100 cannot reproduce, structurally.** Max orderkey ≈ 6M × SF. At SF100 (~600M) it fits int32, so every chunk narrows uniformly and the int32→int64 restore never runs (0 observed). At SF1000 (~6e9) it exceeds int32, so only low-key chunks narrow — a **mixed carrier state**. Reproduction needs SF ≳ 400.
2. **Default engine tuning masks it.** SF500 under defaults was clean 8/8; the same data under the benchmark config corrupted 6/8. A synthetic 400M-row mixed-carrier dataset also stayed clean under defaults despite 48 int restores.

Because this is a race, a clean run never proves absence. Always A/B against `enable_compressed_materialization: false`.

**Minimal repro is far smaller than Q18** — no join, group-by, HAVING or TopN:

```sql
SELECT sum(l_quantity) FROM lineitem;   -- after pinning lineitem to host
```

Correct `76,500,908,501.00`; observed `1,073,525,707,087,194.30` and `1,327,320,611,408,608.28` (~14,000× inflated) and `72,980,986,485.00` (deflated). Row count stable at 3,000,028,242 every run.

### 5. Attribution

SF500, 8–12 pinned runs each:

| Condition | Outcome |
|---|---|
| CPU reference / GPU unpinned | 31,873 ✅ |
| Pinned, CM **on**, prefetcher on | 6/8 wrong |
| Pinned, CM **on**, prefetcher off | 5/8 wrong |
| Pinned, CM **off** | 10/10 correct |

Then, at the `owning_table_view` layer, a probe that awaited the copy before the owner was released gave **12/12 correct** against that 60–75% baseline failure rate. That isolated the defect this PR fixes.

### 6. Residual — why this stays draft

With the fix applied:

| Config | Pre-fix | Post-fix |
|---|---|---|
| prefetcher **on** | 6/8 counts wrong | **0/8**; 1 total off by 0.001% |
| prefetcher **off** | 5/8 counts wrong | 3/8 counts wrong; totals 2.95e15, 9.38e13 |

Net-positive but incomplete. I predicted disabling the prefetcher would clean up the residual and **the experiment showed the opposite** — prefetcher-off is markedly worse. The reading that fits: the prefetcher produces a properly-owned GPU copy, so batches reach `get_cached_batch()` and hit the path fixed here. With it disabled, splits take the `has_scan_metadata()` route through `materialize_metadata_to_table`, which returns straight to the caller with no ordering when there is no MVCC mask — while the adjacent keep-mask branch does `stream.synchronize()` for exactly this reason. That is the next place to look; not yet proven.

### 7. Recommendations

1. **Do not close the Q18 issue on this PR** — §6 shows corruption remains.
2. Consider defaulting `enable_compressed_materialization` to `false` until the second site is fixed; it is currently `true` (`sirius_config.hpp:168`) and silently yields wrong answers on stock TPC-H.
3. **Perf unmeasured**: this fix blocks the host once per copying materialization. A host destructor cannot be deferred by a device event alone, so the alternative is a deferred-reclaim queue. If the scan-path cost is material, that is the follow-up.

### Unrelated issues found along the way

- **The benchmark harness's per-query log slicer is broken.** `sirius/q18/sirius.log` contains a *different query* (timestamped 19:11:07, with `DELIM_JOIN`/`DELIM_GET` ops Q18's plan does not have); the real Q18 ran 19:12:54–19:13:02. Other dirs are shifted too (`q1/` opens with `CREATE VIEW customer`, `q11/` with `unpin_table('nation')`). Any analysis based on `qN/sirius.log` is reading the wrong query — use the combined session log.
- **Only the last iteration's output is saved**, so this bug class is under-reported: had iteration 3 been last, Q18 would have been recorded `success` while 3 of 4 runs were internally wrong. Validation should also compare iterations against each other.
- **Saved results are truncated to 40 of 100 rows** — if validation diffs these artifacts, it is blind to 60% of the result set.
- **795 warnings**: `slot_for_device: device_id 0 not found in active GPU list, falling back to slot 0` (`sirius_physical_partition.cpp:324–333`). `_active_gpu_ids` is unpopulated. Benign on 1 GPU; on multi-GPU it would mis-route every broadcast probe batch to slot 0.
- **No query labels** — the harness never calls `sirius_set_query_label`, so every query shows as `unnamed_query` in the Quent UI and must be identified by position. `run_tpch_parquet_and_generate_telemetry.sh` sets labels; `benchmark_and_validate.sh` does not.

9prady9:
## Update: second site found, Q18 now fully resolved

The earlier comment said a residual hazard remained and guessed at `materialize_metadata_to_table`. **That guess was wrong** — that is the disk-read path and is not involved. Two corrections and the actual answer:

**The residual was a second instance of the same hazard, at the cast in `normalize_physical_schema`** — the site originally suspected. The two are reached by different routes, which is why each config only responded to one fix:

| Site | Route | Exposed by |
|---|---|---|
| `owning_table_view::release()` | batch already GPU-resident → view/copy | prefetcher **on** |
| `normalize_physical_schema` cast | HOST tier → inline `convert_to` → `stolen_table` → owned table → cast | prefetcher **off** |

**Also corrected:** I claimed the scan output was "provably correct" because `sum(l_quantity)` and `sum(l_orderkey)` checksummed clean. That was over-stated — those ran on a lighter query that does not stress the allocator. The race is allocation-pressure dependent, so passing meant "did not trigger", not "cannot trigger".

### How it was isolated

Checksumming each column separately, then the grouped aggregate, against the same run:

- `sum(l_quantity)`, `sum(l_orderkey)`, `count(*)`, plain `GROUP BY` count — all exact.
- Re-aggregated per-group sums — diverged: 136,328,639,662.36 / 76,484,787,029.90 / 76,456,365,543.30 vs the correct 76,500,908,501.00, and one run lost 46M groups (703,989,440 vs 750,000,000).

`l_quantity` is `decimal(15,2)` holding whole numbers, so every legitimate sum must end in `.00`. Totals ending `.26`/`.36`/`.90`/`.30` proved the aggregate was reading garbage bytes as decimal rather than merely mis-assigning rows.

With `enable_compressed_materialization=false`, the same binary and config gave 9/9 exact — confirming one bug gated by the CM path, not an independent aggregation defect.

### Result

| config | baseline | + site 1 | + both |
|---|---|---|---|
| prefetcher on | 6/8 wrong | 0/8 counts, 1 total off | **9/9 clean** |
| prefetcher off | 5/8 wrong | 4/9 wrong | **9/9 clean** |

Both match the CM-off control. The Q18 subquery returns 31,873 on every run.

### Unrelated, worth a follow-up

`gpu_pipeline_task.cpp:547` — the `stream.synchronize()` after `prepare_for_processing` is load-bearing for correctness, but its comment says it is there "to ensure the timing collected by Quent and logging for preparing the task is accurate". Anyone optimizing away a telemetry-only sync would reintroduce a data race.

9prady9:
## The host wait is not measurable

The PR notes `stream.synchronize()` blocks the host. Measured it — there is no cost to recover.

TPC-H SF500, pinned `lineitem`, benchmark config, 10 runs each, same build with and without the two syncs (medians):

| query | with sync | without sync |
|---|---|---|
| scan-dominated (`sum(l_quantity), sum(l_orderkey)`, no join/grouping) | **0.911s** | 0.982s |
| Q18 subquery | **2.00s** | 2.08s |

The synced build is nominally *faster* on the scan-dominated query — noise, or mild backpressure from not letting allocations run ahead. Either way the wait does not register: it fires once per narrowed column per batch, at a point where the stream has almost nothing outstanding beyond the cast that just finished.

Caveats: the no-sync build returns wrong results, so its Q18 figure could reflect different work — hence weighting the scan-dominated query, which has no filter or grouping and does identical work either way. Single GPU (RTX PRO 6000), single scale factor; a machine with many more concurrent scan tasks per stream could price it differently.

## On the non-blocking alternative

For the record, since it came up: cuCascade has half the machinery for a device-side fix.

- `gpu_data_representation::record_writer_event()` / `get_writer_event()`, plus `cucascade::cuda::cuda_event::wait(stream)` (`cudaStreamWaitEvent`) — covers **reader waits on writer**.
- `cudf::rebind_stream(column&&, stream)` (`cudf/column/column_stream.hpp:40`), surfaced through `idata_representation` / `gpu_table_representation` / `data_batch` — relocates a buffer's deallocation onto your own stream, no copies, no kernels. Already used this way for the downgrade path in `convertible_data_batch.hpp:122`.

**Site 2 (the cast) could use `rebind_stream` today** to order the free behind the cast with no host wait — the precondition holds, since the upload stream is already drained by the `stream.synchronize()` at `gpu_pipeline_task.cpp:547`.

**Site 1 (the `owning_table_view` copy) cannot.** It needs the inverse of the writer event — a reclaimer waiting on an in-flight reader — and no such API exists (`reader_event` / `consumer_event` / `record_reader` are all absent). `rebind_stream` is explicitly documented as a no-op when the representation holds an `owning_table_view`, because that memory belongs to an external owner. Closing that one non-blocking would need a new consumer-event capability in cuCascade.

Given the measurement, neither is worth building now. Recorded so the option is known rather than rediscovered.

9prady9:
## Revision: the cast site alone appears to be the query-level cause

An elimination run changes the causal reading in the earlier comment, and the correction is worth recording.

Each site was tested in isolation against the SF500 aggregate check (prefetcher on, the config that originally corrupted 6/8):

| build | prefetcher-on result |
|---|---|
| no fix | 6/8 runs wrong |
| `owning_table_view` sync only | 0/8 wrong |
| cast-site rebind only | 9/9 clean |
| both | 9/9 clean |

Prefetcher-on is clean with **either** fix alone. The reading that fits: the Q18 corruption was the cast site throughout, and the `owning_table_view` sync was masking it — draining the stream immediately before `normalize_physical_schema` meant the cast completed near-instantly, closing the free race by timing rather than by correctness.

This is inference from elimination, not a direct proof; the two are hard to separate because the sync at one site changes the timing at the other.

**It does not exonerate `owning_table_view::release()`.** Dropping an owner while an async copy reads its memory is a use-after-free regardless of whether a given workload hits it, and the regression test forces it deterministically: queue a blocker ahead of the copy and every run reports `cudaErrorNotReady` at reclaim with 4,194,304/4,194,304 poisoned. Removing that sync fails the suite (1 of 2,449), even with the cast-site rebind in place.

So both changes stay, for different reasons: the cast site is what corrupted Q18 and is now fixed without a host block; the copy site is a latent use-after-free caught by construction.

## On measuring the wait

An earlier comment reported the host wait as costing nothing, based on an A/B of wall-clock timings. That measurement was weak — the synced build came out *faster* on the scan-dominated query, which should have been the tell that the signal was below the noise floor. More fundamentally, sync cost is not a fixed quantity: it is however much work happens to be queued on that stream at that moment, so any single number is an artifact of one workload. The cast site now avoids the wait entirely rather than pricing it.

bwyogatama:
Pushed 7 additional commits extending this fix, after an extensive root-cause investigation confirmed the mechanism this PR addresses (premature `cudaFreeAsync` on the idle conversion-pool stream racing task-stream reads):

**Production**
- `fix(prefetcher)`: reserve-then-gate admission (3 workers could previously triple-spend the same headroom), reservations now actually drawn down via tracker attach (the transient ~2× double-count is gone — mutation-verified regression test included), and the phantom "exclusive stream pool" removed (the converter never used it; comments now describe reality).
- cucascade submodule bump → `feat/stream-ordered-reclaim` (NVIDIA/cuCascade#186): `release_table(stream)` now rebinds ownership-transferred buffers (closes the same defect class at the API level, covering the second steal site in `grouped_aggregate_merge` and future callers), plus a consumer-event API (`record_consumer_event`/`await_consumers`) — the mirror of `record_writer_event` this PR's description asked for; `owning_table_view`'s host sync can migrate to it in a follow-up.

**Tests** (all green on GB300)
- `test/cpp/scan/test_stream_lineage_item5.cpp`: 20-iteration allocator-reuse stress on the steal path + a sensitivity check that re-binds a column back to the pre-fix foreign stream and deterministically reproduces the corruption (1,048,576/1,048,576 elements poisoned) — proving the harness detects the original bug.
- `test/cpp/scan_manager/test_memory_prefetcher_accounting.cpp`: admission floor under concurrent workers, no-double-count draw-down, bit-exact prefetch read-back, reservation-refusal back-off.

**Validation**: TPC-H SF1000 full suite, 4 sessions (3 CI-parity + 1 prefetcher-off), every iteration validated against full untruncated CPU ground truth: **352/352 clean**; Q18 `HAVING` count invariant 63,430 in all 16 iterations (pre-fix: ~39% of hot executions corrupt). Q18 timings match or beat the pre-fix build on identical hardware.

Remaining known gap (out of this PR's scope, tracked for follow-up): the filtered serving paths (`duckdb_native_gpu_ingestible.cpp`, `parquet_gpu_ingestible.cpp`) drop their view-backed owner right after enqueuing `exec.select` with no sync — same defect class, currently latent (per-query batches, destroyed post-task-sync). The new consumer-event API is the intended fix there.

bwyogatama:
Follow-up: the "remaining known gap" from my previous comment is now **closed** in the latest push.

- Both filtered serving paths (`duckdb_native_gpu_ingestible.cpp`, `parquet_gpu_ingestible.cpp`) now record a consumer event through a new `owning_table_view::record_consumer_event` pass-through (concept-gated through the type erasure; no-op for owned-table/fresh-decode splits) after enqueuing `exec.select` and before the view-backed owner drops — making these the first production users of the consumer-event API (NVIDIA/cuCascade#186).
- cuCascade side: `~data_batch` now performs an exception-safe, allocation-free `synchronize_no_throw()` on outstanding consumer events, so all three reclaim paths (tier change, `set_data`, destruction) are ordered after recorded reads — the filtered-path protection no longer depends on task-boundary sync timing.
- Regression tests added at both layers (Sirius `[stream_lineage]` filtered-serving scenario with a non-vacuity `consumers_done` check; cuCascade destructor poison-on-reclaim test). Suites green: Sirius 8 cases / 126 assertions, cuCascade 197 cases / 2,171 assertions; pre-commit clean on both repos' configs.

bwyogatama (approved):

PR #1576, opened 2026-08-14
fix(concat): enforce the batching threshold in the pull path

kevkrist:
## Summary
- CONCAT's pull path released whatever a partition held, so the task creator's drain loop (which pulls until the input is empty without re-consulting the hint) flushed under-threshold groups the hint had deliberately held back — producing runt join tasks.
- Hint and pull now share one group-forming policy (`plan_pull_for_partition`): while the source pipeline is running, a pull only releases threshold-complete groups (the overflowing batch seeds the next group); pipeline finish flushes the remainder.

## Test plan
- [x] `[physical_concat]` unit tests: 24 cases pass, incl. 3 new gating tests with a controllable source pipeline
- [x] TPC-H SF50 GPU-vs-CPU validation: 22/22 queries pass
- [x] SF50 perf A/B: neutral at the default config; join microbench with small scan batches improves ~3–8% (probe-side CONCAT outputs drop 10 → 6, runts eliminated)

### DIFF
--- a/src/op/sirius_physical_concat.cpp
+++ b/src/op/sirius_physical_concat.cpp
@@ -68,6 +68,42 @@ sirius_physical_concat::sirius_physical_concat(duckdb::vector<sirius::logical_ty
   }
 }
 
+std::optional<std::vector<uint64_t>> sirius_physical_concat::plan_pull_for_partition(
+  ::cucascade::shared_data_repository& repo,
+  std::size_t partition_idx,
+  bool pipeline_finished) const
+{
+  auto batch_ids = repo.get_batch_ids(partition_idx);
+  if (batch_ids.empty()) { return std::nullopt; }
+
+  if (_concat_all) {
+    // A concat-all group is the whole partition, so it can only form once the source is done.
+    if (!pipeline_finished) { return std::nullopt; }
+    return batch_ids;
+  }
+
+  std::vector<uint64_t> group;
+  std::size_t total_batch_size = 0;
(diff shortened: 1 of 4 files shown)


src/op/sirius_physical_concat.cpp
```diff
@@ -68,6 +68,42 @@ sirius_physical_concat::sirius_physical_concat(duckdb::vector<sirius::logical_ty
...
+    return batch_ids;
+  }
+
+  std::vector<uint64_t> group;
+  std::size_t total_batch_size = 0;
+  for (auto const batch_id : batch_ids) {
+    auto batch_idle = repo.get_data_batch_by_id(batch_id, partition_idx);
+    auto batch_ro   = batch_idle->to_read_only();
+    total_batch_size += batch_ro.get_data()->get_size_in_bytes();
+    if (total_batch_size > _concat_batch_bytes) {
```
Jedi18, line 91: Is withholding an exactly-at-threshold or lone oversized batch intentional to avoid single-batch CONCAT tasks? The documented readiness contract says `>= concat_batch_bytes`, so it may be worth clarifying the intended semantics or updating the documentation.

Jedi18 (approved):

src/op/sirius_physical_concat.cpp
kevkrist: yeah this is intentional -- both cases preserve the existing semantics on dev, and the fix moves the pull onto that same policy rather than defining a new one. Clarifying the docs.

PR #1577, opened 2026-08-15
perf(scan): see through DATE→TIMESTAMP casts in fused scan-filter range extraction

felipeblazing:
# perf(scan): see through DATE→TIMESTAMP casts in scan filter range extraction

**Branch:** `pr/cast-through-range-extraction` → base: **`dev`**. Two commits, no longer stacked: the change, and the review follow-up for @mbrobbel's ±infinity / overflow comment. Rebased onto `dev` @ `05148c78` (2026-09-15).

> [!NOTE]
> **Previously stacked on #1555**, which has since merged (`3ac080d8`). This PR was rebased onto `dev`: the ~70 never-merged commits of the #1391 fused scan-filter lineage are gone, and the one remaining change was **ported** rather than replayed — `dev` had meanwhile refactored the extraction machinery out of `scan_utils.cpp` into `op/scan/scan_filter_analysis.{hpp,cpp}` (`extract_numeric_range_pushdown` → `analyze_scan_filters`, `decoded_bound{floor,ceil}` → `sirius::numeric_range{minimum,maximum}`). The diff below is now exactly this change: **5 files, +889 / −24**.

## Problem

DuckDB constant-folds qgen-style date arithmetic (`l_shipdate <= DATE '1998-12-01' - INTERVAL '72' DAY`) into `CAST(l_shipdate AS TIMESTAMP) <= TIMESTAMP '1998-09-20 00:00:00'`, which arrives at the scan as an `EXPRESSION_FILTER`. `analyze_scan_filters` only understood `CONSTANT_COMPARISON` / `CONJUNCTION_AND`, so `fold_numeric_conjunct` refused the clause and the whole scan lost its in-decode masking: full-width decode + per-row cast AST filter + gather-compact. q1 varied 1.288 s vs 0.831 s with the equivalent DATE literal; the same shape appears in 9 predicates across q1/q4/q5/q6/q10/q12/q14/q15/q20 under spec-official (qgen) parameters.

## Change

Three additions in `src/op/scan/scan_filter_analysis.cpp`:

- **`to_decoded_bound`**: accepts finite `TIMESTAMP[_S|_MS|_NS]` constants against DATE columns, lowering ticks / ticks-per-day into the stored-day domain by floor/ceil (templated on the finer-scale DECIMAL branch). The bound is EXACT for every comparison op — midnight(d) is strictly monotonic in d — so midnight and non-midnight constants both keep full coverage and the residual filter drops; `= non-midnight` folds to a provably empty range.
- **`fold_comparison_bound`**: the comparison switch lifted out of `fold_numeric_conjunct` so both filter shapes share one implementation.
- **`fold_expression_conjunct`** + a new `EXPRESSION_FILTER` case, matching `cmp(CAST(bound_ref AS TIMESTAMP*), ts_const)` in both operand orders (comparison flipped on swap), recursing through AND conjunctions.

**Edges of the DATE domain** (review follow-up, see `lower_timestamp_to_days`): `DATE ±infinity` rows compare correctly because ±INT32_MAX days and ±INT64_MAX ticks are the extremes of their domains, and **±infinity constants now lower exactly** to the infinite date's stored days instead of being refused. A finite date whose midnight overflows the target (~±292,000 years for `TIMESTAMP/_S/_MS`, ~±292 years for `TIMESTAMP_NS`) makes DuckDB raise; a decode range cannot raise, and neither does the residual `cudf::cast` it replaces, so the range is deliberately not clipped to the castable window and such rows are kept or dropped by the instant they denote. The only divergence from DuckDB, and only on queries DuckDB refuses to answer.

**Refused (coverage cleared, residual kept):** `TRY_CAST` (NULL-on-overflow diverges from day math on >-side bounds), `TIMESTAMP_TZ` (midnight is session-timezone dependent), `INT64_MIN` ticks (below -infinity, never a real instant), and every non-comparison shape (including the `BETWEEN` DuckDB builds from two cast comparisons on one column; follow-up).

## Measured (SF1000, GB300, spec-official qgen parameters)

- **Power@Size +9.6% → 10,079,305**
- Power-run suite time **−8.3%**
- q1/q12/q14/q15 each **−36..39%** (the queries whose only selective predicate is a folded date cast)

> [!WARNING]
> These numbers were measured on the **pre-rebase fused-scan-filter stack**, not on this branch. They have **not** been re-measured against current `dev`, whose scan/decode path has since changed (#1524 late materialization, #1555, and others). Treat them as the original motivation, not as a claim about this branch's delta. Correctness on current `dev` is verified below.

## Amplification history (honest note)

Multiplying the number of fused scans exposed two classes of pre-existing bugs downstream of scan-produced geometry — neither introduced by this change, both diagnosed during its qualification:

- two pre-existing Sirius GPU use-after-free races (batch read-lock lifetime, error-path teardown);
- a pre-existing **CCCL 3.4.0** bug: the warp-specialized (TMA) `cub::DeviceScan` on Blackwell performs an out-of-bounds `__shared__` read that silently corrupts scan outputs under concurrent kernel diversity (memcheck-reproducible with a 22-line bare `ExclusiveSum`; CTK-bundled CCCL 3.2.0 is clean).

Both hardenings ship separately in **#1580** (which supersedes the now-closed #1546).

## Tests

- `test/cpp/scan/test_scan_filter_cast_ranges.cpp` (`[range_pushdown]`): exhaustive boundary matrix — all 5 ops × midnight / one-µs-either-side / noon constants, pre-epoch floor, operand flips, AND shapes, S/MS/NS flavors, extreme finite ticks, ±infinity constants (all ops, both directions, both operand orders, every flavor), `INT64_MIN` refusal, bounds not clipped to the castable window, and the refusal set.
- `test/cpp/integration/test_gpu_execution_cast_date_predicates.cpp` (`[cast_date]`): GPU-vs-CPU equivalence for the folded SQL shapes on three scan paths. **On current `dev` the fold's ranges only take effect on a GPU-pinned Parquet entry decoded through the compressed path** (the DuckDB-native ingestible does not analyze its filter), so besides the plain and DuckDB-pinned scans the fixture pins bitpacked Parquet twins, proves via census that the pin compressed, and lifts the fused-decode selectivity caps so the in-decode mask answers. New tables: `DATE ±infinity` rows (finite constants on every path, infinite constants on the fold path) and a year-300000 date that raises on CPU while the fold answers by instant.

Run on this branch, rebased onto `dev` @ `05148c78`:

| Filter | Result |
| --- | --- |
| `[range_pushdown],[cast_date],[fused_scan_filter]` | 17/17 green — 9,788 assertions |
| `[scan],[filter]` neighbor sweep | 374/374 green — 461,053 assertions |

`pixi run pre-commit run -a` clean on the changed files.

> [!NOTE]
> Found while making the tests bite, pre-existing and not touched here: the residual `cudf::cast` path multiplies `±INT32_MAX` days into int64 micros and wraps, so e.g. `d < TIMESTAMP 'infinity'` returns the +infinity row on unpinned / DuckDB-format scans today. The fold gives the right answer where it engages.


src/op/scan/scan_filter_analysis.cpp
```diff
@@ -213,6 +278,107 @@ sirius::numeric_range full_decoded_domain()
...
+    fully_covered = false;
+    return false;
+  }
+  auto const& constant = value_side->Cast<duckdb::BoundConstantExpression>().value;
+  auto const bound     = to_decoded_bound(constant, col_type);
+  if (!bound.has_value()) {
+    fully_covered = false;
+    return false;
+  }
+  return fold_comparison_bound(comparison, *bound, acc, fully_covered);
```
mbrobbel, line 379: We should check if the dates can be safely cast before we do this. `+infinity` and `-infinity` require special handling and early or late dates can overflow the timestamp type. I think we should add tests and then update the rewrite to correctly handle or skip them.

kevkrist:
lgtm aside from @mbrobbel issue

src/op/scan/scan_filter_analysis.cpp
felipeblazing: Thanks, good catch. I went through DuckDB's `DATE -> TIMESTAMP*` cast (`TryCast<date_t, timestamp_t>` plus the `_S/_MS/_NS` wrappers, which all route through the micros cast) and it has three regimes. The lowering now states each one (`lower_timestamp_to_days`) and each is tested:

1. **Finite date whose midnight fits the target.** Linear and strictly monotonic; floor/ceil of `ticks / ticks_per_day` is exact for every op. Unchanged.

2. **`DATE ±infinity`.** DuckDB maps it straight onto `TIMESTAMP ±infinity` rather than through day arithmetic, so the linear formula does not apply to the sentinels. It does not need to: `±INT32_MAX` days and `±INT64_MAX` ticks are the extremes of their respective domains, so order is preserved and a finite constant needs no special casing. An **infinite constant** was refused before; it now lowers exactly to the infinite date's stored day count (`< 'infinity'` keeps everything but +infinity, `>= 'infinity'` only it, etc.). `INT64_MIN` ticks (below -infinity, never a real instant) are refused.

3. **Finite date whose midnight overflows the target** (beyond ~±292,000 years for `TIMESTAMP/_S/_MS`, ~±292 years for `TIMESTAMP_NS`). DuckDB raises a `ConversionException` for the whole query. A decode-time range cannot raise, and neither does the residual `cudf::cast` it replaces (that one silently wraps), so I deliberately did **not** clip the range to the castable window: such a row is kept or dropped by the instant it denotes. That is the only divergence from DuckDB and it exists only on queries DuckDB refuses to answer at all. Making the GPU raise here would need a data-dependent check in the decoder that applies to the residual path too; happy to do that as a follow-up if you'd rather have error parity.

Tests added: unit matrix for ±infinity constants (all five ops, both directions, both operand orders, every flavor), the `INT64_MIN` refusal, and "bounds are not clipped to the castable window" for micros and nanos. Integration: a table with `DATE '-infinity'` / `DATE 'infinity'` rows checked GPU-vs-CPU on every scan path for finite constants and on the fold path for infinite constants, and a table with a year-300000 date checked to raise on CPU while the fold answers by instant.

Two things I found while making those tests bite, worth knowing:
- On current `dev` the fold's ranges only take effect on a GPU-pinned **Parquet** entry whose column decodes through the compressed (bitpack) path; the DuckDB-native ingestible does not analyze its filter. The original integration tests on DuckDB tables were therefore passing through the residual cast. The fixture now also pins bitpacked Parquet twins, proves via census that the pin compressed, and lifts the fused-decode selectivity caps so the in-decode mask is what answers.
- The residual `cudf::cast` path multiplies `±INT32_MAX` days into int64 micros and wraps, so `d < TIMESTAMP 'infinity'` returns the +infinity row on unpinned / DuckDB-format scans today. Pre-existing and independent of this change; the fold actually fixes it where it engages. I kept the infinite-constant predicates on the fold path only for that reason. Also, DuckDB merges two cast comparisons on one column into a single `BETWEEN` expression, which the fold does not recognize yet (residual path); separate follow-up if wanted.

mbrobbel (approved):

PR #1580, opened 2026-08-15
fix: GPU memory-lifetime hardening — order the downgrade behind producer writes and consumer reads, quiesce task teardown, and work around the CCCL 3.4.0 Blackwell DeviceScan bug

felipeblazing:
## Problem

This PR closes out the corruption class that #1546 opened: **device memory observed by one
stream while another stream frees or rebinds it**. Concurrent staged-refresh TPC-H throughput
runs at SF1000 struck near-deterministically (3-of-3, then 5-of-5 corrupted or wedged runs in
one night). Every GPU coredump was a **victim consuming scribbled string geometry**: a
`gather_chars` kernel whose int64 offsets buffer contained ASCII `l_shipmode` text written by
another actor, a batched-memcpy kernel reading a wild tuple pointer, and `LIKE`/HLL spins walking
multi-GB bogus `string_view`s. Strikes clustered at the refresh RF2 DELETE/COMMIT boundary, the
window of maximum conversion/eviction churn. Sequential runs validated 22/22 byte-identical
throughout, which is what makes this class so expensive: the freed VA must be rebound quickly by
concurrent work for the stale access to observe garbage.

Root-causing took three instruments: cuda-gdb coredump forensics, an env-gated poison-on-free
mode that made the first toucher of freed memory fault red-handed, and a **compute-sanitizer
memcheck sweep** over an SF1-sized config that reproduces the SF1000 memory-pressure regime in
minutes (the sweep ships with the harness in #1554, now on `dev`).

## Root causes

The engine hands batches off *event-ordered, not host-synced* (writer events), and batch read
locks are **host-scoped**: a consumer can enqueue kernels and drop its lock with device work
still in flight. Three seams ignored one or both directions of that contract, and one strike
signature was a vendor bug.

**A1 — the downgrade never waited the producer's writer event.** A parked batch grabbed by the
downgrade executor (`try_to_mutable` succeeds; nobody holds a lock on a parked batch) while its
producer stream still writes gets a **torn host image**: the unwritten regions D2H-copy whatever
the pool last left there. The re-upload later serves scribbled string geometry. Fix
(`convertible_data_batch::convert`): wait the batch's writer event on the conversion stream
before `convert_to` reads a byte.

**A2 — the pipeline task's SUCCESS path never quiesced its stream.** After the last operator
sync, the sink's enqueues (`publish_output`) stayed in flight while the task's read locks died and
`finalize_operator` freed **cross-task** device state (hash-table slots, the build batch's read
lock, sort partition boundaries). Fix (`gpu_pipeline_task::execute`): `stream.synchronize()`
after `publish_output`. Nearly free, since only the sink's enqueues are outstanding.

**A3/A4 — downgrade and consumer-acquire rebinds carried no reader edge.** The downgrade rebinds a
parked batch's deallocation stream onto its own idle stream and frees after conversion, and
`lock_or_prepare_batch` rebinds opportunistically on acquire, while a consumer's kernels enqueued
before its read lock dropped may still read the batch. A1 cannot cover this: the writer event was
long signaled. Fix: use cuCascade's reader-event mechanism (upstream NVIDIA/cuCascade#184, already
in dev's pin). `pipelineable_operator_data::record_reader_events` publishes every operator's
enqueued reads to its input batches on both exits of `run_one_operator`, including the throwing
one. `try_to_mutable()` then refuses, and `to_mutable()` waits, until those reads complete, so no
downgrade can rebind or free a batch under an in-flight reader.

**Vendor — CCCL 3.4.0 warp-specialized (TMA) Blackwell DeviceScan corrupts scan output under
concurrency.** The last surviving strike signature was not an ordering bug: garbage always in
scan-produced geometry, sequential runs clean, immune to every ordering fix, and it vanishes
with CCCL's own kill-switch `CCCL_DISABLE_WARPSPEED_SCAN` (routing `cub::DeviceScan` back to the
classic decoupled-lookback path). This is ablation-only evidence; there is no minimal
reproducer yet. The compute-sanitizer `Invalid __shared__ read` reported against the same kernel
turned out to be a known sanitizer ≤13.2 false positive (NVIDIA/cccl#8288, NVBug 5824899) and is
not the mechanism. The define is a defensive kill-switch until the cause is pinned down upstream.

A fifth defect found by the sweep, the scan's narrow-cast source freed mid-cast, landed on `dev`
separately as #1566.

## Verification

Measured on GB300 (SF1000, staged refresh, varied predicates) on the original fix stack, which
used cucascade-side fences for A1 and A3/A4 instead of the writer-event wait and reader events
carried here:

| Gate | Result |
|---|---|
| Sequential poison gate | Pre-fix binary wedged 4/4 at warmup q14. Post-fix: PASS. |
| SF1 memcheck sweep (mode=both, 3 streams) | Pre-fix: 15 narrow-cast UAFs (the #1566 site) plus the warpspeed-scan report later identified as a sanitizer false positive. Post-fix: **zero use-after-frees**. |
| SF1000 concurrent poison gate (mode=both, 3 streams, 0.80 cap) | PASS: 0 dumps, 0 strike lines. |
| Full-scale clean gauntlet | **6/6 clean** concurrent runs on the vehicle that struck 5-of-5 pre-fix; 22/22 byte-identical; QphH 8.22–8.39M, ±1.0%. |

On this branch (sirius-side mechanism, rebased on `dev` @ `e0080cd6`):

- Full build, the `[spill_converter_ordering]` guards, and the complete Catch2 unit suite
  (3301 cases) pass.
- **SF1000 concurrent gate, 2026-09-18** (`run-power.sh`, mode=both, 3 streams + refresh
  stream, staged refresh, production `sirius-sf1000.yaml`, fixed predicates so GPU-vs-CPU
  validation runs): all GPU phases clean, 0 coredumps, 0 strike lines, 0 CPU fallbacks;
  validation after RF1 **19/19 OK** and after RF2 **19/19 OK** (q2/q11/q16 skipped as
  refresh-invariant). Power@Size 10.05M, Throughput@Size 5.53M (42.99 s), QphH@Size 7.45M.
  Harness note: the pure-DuckDB validation worker was kernel-OOM-killed mid after-RF2 on the
  first attempt (its default 395 GiB memory_limit on a 494 GB host); the after-RF2 diff was
  completed against the same post-RF2 validation database with a 300 GB cap. The 0.80-cap
  `sirius-sf1000-gate.yaml` is no longer usable for this: the pinned layout alone is ~205 GB, so
  every MERGE_GROUP_BY OOMs and falls back to CPU.

## What ships

- **2 fix commits**: the three ordering fixes (A1, A2, A3/A4) and the CCCL workaround. No cucascade
  submodule change.
- **Regression guards** (`test/cpp/data/test_spill_converter_ordering.cpp`): a downgrade under a
  parked producer must carry the final bytes; a recorded reader event must make the non-blocking
  downgrade refuse and the blocking one wait, with a poison fill proving no VA reuse under the
  reader.
- **Docs**: `docs/super-sirius/memory-management.md` gains a short stream-ordering discipline
  section (the invariant and the three mechanisms that carry it).

## Relationship to other PRs

- **#1546** (closed): instances 1–2 of the same class (scan read-lock lifetime, error-path
  teardown). The error-path reader-event recording here covers the input-lock side of its
  teardown fix.
- **#1566** (merged): the narrow-cast source-free ordering originally carried here as B1.
- **NVIDIA/cuCascade#184** (merged, in dev's pin): the reader-event API A3/A4 builds on.
- **#1554** (merged): ships the SF1 memcheck config and gate scripts this work was verified with.

### DIFF
--- a/src/pipeline/gpu_pipeline_task.cpp
+++ b/src/pipeline/gpu_pipeline_task.cpp
@@ -640,6 +640,21 @@ void gpu_pipeline_task::execute(rmm::cuda_stream_view stream)
 
   if (output_data) { publish_output(*output_data, stream); }
 
+  // Quiesce the task stream on the SUCCESS path (freed-while-read corruption
+  // fix; the executor's abnormal exits get the symmetric quiesce in the
+  // error-path teardown fix, PR #1546). Everything that runs after this return
+  // — output_data and its read locks dying below, ~gpu_pipeline_task's
+  // unsubscribe chain, mark_task_completed -> update_pipeline_status ->
+  // finalize_operator (which frees cross-task device state: hash-table slots,
+  // build_table read locks, sort partition boundaries) — could otherwise
+  // destroy owners of device memory while sink-enqueued kernels/copies were
+  // still in flight on this stream. The freed VA rebinds to another stream's
+  // fresh allocation and the straggler scribbles it: torn string geometry,
+  // the RF2-window corruption class. Nearly free: the stream was synchronized
+  // after every operator execute; only the sink's enqueues (if any) are
+  // outstanding here.
+  stream.synchronize();
+
   // The input pipelineable_operator_data (with its _read_only_data_batches) was destroyed
   // when compute_task replaced operator_input_output_data, releasing all shared locks.
 }
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ tools/log_analyzer/metrics/__pycache__/
 test/tpch_performance/__pycache__/
 test/tpch_performance/output/
 test/tpch_performance/log/
+bench/sf1000-repro/VERIFY-LEDGER.txt
+bench/sf1000-repro/sanitizer_*.log
+bench/sf1000-repro/coredumps/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -635,6 +635,7 @@ set(TEST_SOURCES
     test/cpp/config/test_object_store_config.cpp
     test/cpp/config/test_topology_config.cpp
     test/cpp/data/test_convertible_data_batch.cpp
+    test/cpp/data/test_spill_converter_ordering.cpp
     test/cpp/data/test_data_repository_manager_registry.cpp
     test/cpp/data/test_convertible_gpu_pipeline_task.cpp
     test/cpp/downgrade/test_downgrade_disk.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,19 @@ if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
   message(STATUS "CMAKE_CUDA_ARCHITECTURES not defined. Setting to 'native'.")
 endif()
 
+# CCCL 3.4.0's new warp-specialized (TMA) Blackwell DeviceScan performs an
+# out-of-bounds __shared__ read (deterministic under compute-sanitizer memcheck
+# on GB300/sm_103: "Invalid __shared__ read of size 16", reproduced standalone
+# with this CCCL against sm_100 and sm_103a alike; clean under CCCL 3.2). Every
+# cub::DeviceScan in this codebase feeds GEOMETRY (string offsets, selection
+# offsets, memcpy tuple offsets), so a misbehaving scan scribbles exactly the
+# buffers the SF1000 concurrent-refresh corruption strikes showed as garbage.
+# CCCL left this macro in as the official kill-switch for the new
+# implementation; the classic decoupled-lookback path takes over. Full analysis
+# and a 22-line standalone reproducer: docs/cccl-3.4-warpspeed-devicescan.md.
+# TODO: drop once CCCL ships a fixed warpspeed scan (tracked upstream).
+add_compile_definitions(CCCL_DISABLE_WARPSPEED_SCAN)
+
 # Optional features
 option(ENABLE_STREAM_CHECK
        "Enable stream check library for debugging default stream usage" OFF)
--- a/bench/sf1000-repro/run.sh
+++ b/bench/sf1000-repro/run.sh
@@ -34,6 +34,21 @@ for t in LINEITEM ORDERS PART CUSTOMER SUPPLIER NATION REGION PARTSUPP; do
   export "SIRIUS_PIN_TIER_$t=gpu"
 done
 
+# SANITIZER=memcheck|initcheck|racecheck|synccheck: run the workload under
+# compute-sanitizer (DIAGNOSTIC RUNS ONLY -- 10-40x kernel overhead; never quote
+# scores from a sanitized run). Reports land in $SANITIZER_LOG.
+LAUNCHER=()
+if [ -n "${SANITIZER:-}" ]; then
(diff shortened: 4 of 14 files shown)


src/include/op/sirius_physical_operator.hpp
```diff
@@ -293,6 +293,23 @@ class pipelineable_operator_data : public operator_data {
...
+   * Read locks are host-scoped: once they drop, a downgrade may reclaim the batch even though
+   * kernels enqueued under the lock are still in flight. Recording a reader event makes
+   * try_to_mutable() refuse, and to_mutable() wait, until those reads complete. Call after the
+   * consumer has enqueued its work and before the locks are released. No-op when nothing is
+   * locked or the batch is not GPU-resident.
+   */
+  void record_reader_events(rmm::cuda_stream_view stream) const
+  {
+    if (!_read_only_data_batches) { return; }
+    for (auto const& ro : *_read_only_data_batches) {
```
bwyogatama: So right now this only covers the input batches, the intermediate batches in the pipeline is not going to be covered by `record_reader_events`. The right approach is going to be some kind of abstraction or a lease that we will use to record event, something like:

```
class execution_read_lease {
  std::vector<cucascade::read_only_data_batch> accessors;

public:
  void publish_reads(rmm::cuda_stream_view stream)
  {
    for (auto const& accessor : accessors) {
      accessor.record_reader_event(stream);
    }
  }
};
```

then you will do something like this everytime you execute an operator, instead of relying on the data batches inside the `pipelineable_operator_data`

```
auto lease = input.acquire_read_lease();
auto output = op.execute(lease.input(), stream);
lease.publish_reads(stream);
```

This is going to require some bigger refactoring though so maybe let's do this in another PR

src/pipeline/gpu_pipeline_task.cpp
```diff
@@ -792,6 +805,8 @@ void gpu_pipeline_task::execute(rmm::cuda_stream_view stream)
       local_state.get_reservation_size_info()->bytes_to_materialize_input);
   }
 
+  stream.synchronize();
```
bwyogatama: codex told me this stream synchronize can be skipped if we throw on earlier in the code, need another synchronize in the catch block

bwyogatama: perhaps this synchronize is also redundant cause we record the event right before this now

bwyogatama (approved):

src/include/op/sirius_physical_operator.hpp
bwyogatama: actually forget about this cause the intermediate batch will not be downgraded anyway, so this is not needed

ran-yuan-rui:
Cross-reference: #1845 overlaps with A1 here. It also handles GPU batches without a writer event and tests the downgrade executor directly. I found the missing producer wait through a `Free-before-alloc` report on the existing downgrade test before noticing this PR. Please take a look at #1845 and see whether it makes sense to fold it into #1580.

bwyogatama:
> Cross-reference: #1845 overlaps with A1 here. It also handles GPU batches without a writer event and tests the downgrade executor directly. I found the missing producer wait through a `Free-before-alloc` report on the existing downgrade test before noticing this PR. Please take a look at #1845 and see whether it makes sense to fold it into #1580.

This should be solved in this PR

PR #1582, opened 2026-08-15
perf(scan): serve GPU-pinned scans as views instead of per-query copies

kevkrist:
## Summary
Scans served from the GPU-tier pinned cache with no row filter currently deep-copy every split device-to-device on every query: with no owned table to move, `owning_table_view::release()` falls back to its copying materialization.

This lets the handle surrender its view together with its type-erased owner (`release_view()`, engaged exactly when `release()` would deep-copy) and has the scan operator emit the split as a view batch. Since pin-time carrier narrowing means most pinned tables still need some casts, schema-mismatched columns are cast directly from the view (preserving all normalization validation) and matched columns forward zero-copy, held together by a composite owner — the same pattern the projection operator uses. The owner keeps the pinned columns alive past `unpin_table`; non-pinned scan paths are structurally unaffected (their handles hold owned tables, for which `release_view()` disengages).

Measured on TPC-H SF50, `--pin gpu`, single GB10: the scan's redundant D2D copies are eliminated (57.8 GB → 31.8 GB total per 9-query bench; the remainder is hash-join internals) and hot medians across the 22 queries improve 11.1% (4.90 s → 4.35 s), with q5/q17 −33%, q4/q8/q11 −27%, q21 −21%.

Measured on TPC-H SF1000 with #1371 config. -1.59% cold and -2.28% warm.

## Test plan
- [x] New `[pinned_view_forward]` suite (8 cases): pointer-identity forwarding, mixed cast+view emission, ownership survival past unpin, normalization fallback
- [x] `[owning_table_view]` `[cached_serving]` `[scan_normalization]` `[convertible_data_batch]` suites green
- [x] TPC-H SF50 GPU-vs-CPU result validation: 22/22 pinned and 22/22 unpinned (disk-path regression check)
- [x] CI full suite


mbrobbel:
There are some conflicts. Marking as draft.

joosthooz:
Converted to draft because there is a refactor underway of this code by @dhruv9vats

PR #1584, opened 2026-08-15
fix(scan): empty ARRAY splits end-to-end and transactional carrier casts for cached splits

kevkrist:
Three scan changes: two correctness fixes and an extension of the transactional steal.

**Empty splits with ARRAY columns — decoder and downstream.** An empty table, or a scan whose filter prunes every row group, produces a zero-row split. Building it used `cudf::make_empty_column`, which rejects list types, so any table with an ARRAY column threw and the query silently fell back to CPU. The decoder now builds ARRAY columns with `cudf::make_empty_lists_column`, and the shared empty-table helpers are nested-safe so the resulting 0-row LIST batch also survives downstream empty rebuilds: `duckdb::make_empty_like` uses `cudf::empty_like` (top-n), and the logical-type `sirius::make_empty_table` builds empty LIST columns for ARRAY entries (hash-join orphan sides, all-pruned VALUES). The carrier-exact `make_empty_table` overload refuses nested entries loudly — an id-only `data_type` carries no element type. Known residual (follow-up): the hash-join physical-sidecar branch can still pass a LIST carrier there; the call site needs the absent side's logical types. Note: the all-pruned-scan-stays-on-GPU behavior itself landed on dev in #1065 — this PR fixes the ARRAY decode it implies and deletes the stale walker test that contradicted it.

**Transactional OOM retry for cached splits that need a cast.** A cached split may first upload or decompress into a fresh per-query table `S`, then cast selected columns into destination buffers `D`. Retaining the wrapper made the cast retryable, but the generic materialization path copied the table first, producing a `2S + D` peak while the reservation still modeled `S + D`.

The converting path now casts directly from the original `S` under an exclusive batch lock:

- Only a fresh upload/decompression result can enter the transaction; raw GPU pins never do.
- Schema/range preflight and every potentially allocating cast complete before ownership changes.
- If a cast throws `rmm::out_of_memory`, partial destinations are discarded and the complete source table remains available for retry, including on another CUDA stream.
- After every cast succeeds, commit reuses the original table, replaces converted columns in place, preserves unchanged column buffers, and empties the wrapper without another allocation.
- MVCC-masked, raw-pin, unexpected-layout, and not-yet-filtered inputs retain the existing generic path. A split whose whole row filter was already applied by a pushdown decode qualifies for the steal — but only when the ingestible reports its output assembly as a leading identity (`output_assembly_is_leading_identity()`: parquet derives it from `needs_output_assembly`; duckdb-native is identity by construction), because the steal bypasses `post_filter_and_project` and a width match alone cannot prove the projection away.
- Eligibility is one shared predicate (`converted_table_transferable`) consumed by prepare's arming gate and the steal's refusal gate; refusals precede any wrapper mutation.

This restores the affected path's peak to `S + D` without sacrificing retry safety.

**Validation**

- Built the complete `sirius_unittest` target.
- `[cached_serving]`: 22 cases / 250 assertions, including partial-cast OOM rollback, cross-stream retry, raw-pin exclusion, downgraded-wrapper refusal without mutation, and eligibility-gate cross-checks.
- `[scan]`: 279 cases / 274,677 assertions.
- `[cudf_utils]`: 11 cases pinning the empty-helper mappings (the 3 nested cases fail without the fix).
- `[array]`: 29 cases, including a new end-to-end fully pruned ARRAY scan through `ORDER BY .. LIMIT` asserting zero CPU fallbacks (fails pre-fix with the exact `make_empty_like` throw).
- TPC-H SF50 with host-pinned tables validates 22/22 against DuckDB.


src/op/scan/sirius_gpu_scan_operator_data.cpp
```diff
@@ -55,13 +55,16 @@ void scan_operator_input::prepare_for_processing(
...
+      // plain gpu_table_representation and never reach this branch), so the
+      // scan may take ownership of it here instead of deep-copying it at
+      // materialize. Three exclusions keep the view path: masked and
+      // row-filtered splits filter by copy and need the source view alive;
+      // carrier-converting splits allocate cast destinations in scan
+      // normalization after materialize, so the retained wrapper table lets an
+      // OOM retry re-enter materialize and re-run the cast. Every split that
+      // steals therefore allocates nothing after the take, and a retry can
+      // never re-enter materialize on a consumed split.
+      if (!mvcc_keep_mask.has_mask() && !row_filter_pending && !needs_carrier_conversion) {
```
ran-yuan-rui: Could we account for the extra materialization copy in the reservation?

Before this change, the converted table was stolen, so the peak was `S + D`. The new path retains decoded table `S`, materializes another `S`, then allocates cast destination `D`, but the estimate remains `S + D`. This should not return incorrect results because retry is now safe, but it can cause an avoidable OOM/retry—or repeated downgrade/retry under pressure. Could we include the extra `S` and cover the GPU-compressed conversion case?

kevkrist: I've just eliminated the S + S + D case by folding the cast case into the transactional stealing:
Casts are built directly from the retained converted source S, and the source table is released only after ll destination columns D succeed. Unchanged columns are moved into the result, so max is S + D.

ran-yuan-rui (approved):

src/include/op/scan/duckdb_native_metadata.hpp
```diff
@@ -170,13 +171,29 @@ struct duckdb_native_walk_plan {
...
+ *
+ * @param storage Table storage whose row groups will be walked.
+ * @param context Client context used to read partition statistics.
+ * @param projected_cols Projected storage columns.
+ * @param projected_types Types parallel to @p projected_cols.
+ * @param table_filters Optional pushed-down filters used for statistics pruning.
+ * @param column_ids Optional mapping from filter positions to storage columns.
+ * @return The prepared plan. Check @c viable before scheduling range walks; a
+ *         plan with every row group pruned is still viable.
+ */
```
9prady9: 👍🏾

9prady9:
Follow-up request: the downgraded-wrapper refusal currently appears to leave a query-fatal fallback path. If the converted source is downgraded between prepare and execute, `transactionally_steal_converted_table()` returns null because the representation is no longer GPU-backed. `execute()` then enters the generic materialization path, whose resident branch calls `get_cudf_table_view()` and assumes a GPU representation; with the HOST representation still installed, that path throws instead of reconverting or requesting a retry. The current downgraded-wrapper test stops after asserting the transactional refusal, so it does not exercise this subsequent execute path. Could a follow-up either reconvert under the existing reservation or propagate a retryable outcome, with an execute-level downgrade regression test?

kevkrist:
@9prady9 I don't see an actual production path in which this wrapper can get downgraded during that interval
 -- Cached serving creates a fresh batch rather than registering it in a data repository
 -- Queued-task downgrade only understande `pipelineable_operator_data` -- not `scan_operator_input`
 -- Connector-resdient scan batches are explicitly outside downgrade discovery

This of course may change in the future, so I will file an issue: #1584

PR #1585, opened 2026-08-15
perf(concat): lock-free readiness hint and taskless single-batch forwarding

kevkrist:
Two CONCAT improvements.

**The readiness check no longer touches batch locks.** Deciding whether enough input has buffered used to walk every buffered batch and briefly lock each one, on every poll, on the single thread that creates tasks for the whole engine — so a batch in the middle of being spilled to host memory could stall task creation everywhere. The operator now keeps per-partition running totals, updated as batches arrive and leave, and the check answers from those. It fires once a partition holds more than the batching threshold across at least two batches; the flush at pipeline finish is unchanged.

**Single-batch groups skip the task machinery.** When a group is exactly one batch there is nothing to concatenate, so the batch is handed directly to the downstream operator instead of running a GPU task — saving a memory reservation and two stream synchronizations per group.

No plan or result changes. TPC-H SF50 runtimes are unchanged within noise (the gain is stall immunity and fewer tasks — 42 → 33 concat tasks in a small-batch stress run); 22/22 result validation against DuckDB passes.

Related: #1576 touches the pull side of the same operator; the two are independent and compose in either order.

### DIFF
--- a/src/include/op/sirius_physical_concat.hpp
+++ b/src/include/op/sirius_physical_concat.hpp
@@ -25,6 +25,8 @@
 #include "op/sirius_physical_top_n.hpp"
 #include "sirius_config.hpp"
 
+#include <unordered_map>
+
 namespace sirius {
 namespace op {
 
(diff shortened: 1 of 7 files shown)


mbrobbel:
There are some conflicts, marking as draft.

PR #1586, opened 2026-08-15
perf(dynamic-filter): bound the exact IN-list to a fraction of probe-GPU L2

kevkrist:
## Description

When a hash join publishes a dynamic membership filter, the publisher chooses an exact cuco
IN-list whenever the set fits device L2. That rule ignores two measured facts (GB300, sweep
included below): a set competing with streaming probe traffic stops being cache-resident well
before L2 capacity, and the Bloom filter out-probes the exact set ~2.2× *even when fully
resident* (16 bits/key vs 8 B/key at load factor 0.5, single fast-range block access vs
dependent bucket probes).

This PR adds one config option and changes nothing else:

- **`dynamic_filter_inlist_max_l2_fraction`** (double, `[0,1]`, default **0.125**): the exact
  IN-list is kept only while its estimated cuco-set bytes are ≤ fraction × queried device L2;
  above that the publisher emits a Bloom filter instead. `1.0` reproduces the legacy
  fits-in-L2 rule exactly; `0` means "always Bloom where the key type supports it". The
  `small_in_list` tier and types without Bloom support are unaffected. If the device L2 size
  cannot be queried, the legacy rule applies.

The trade encoded by the default: below the threshold, exactness (zero false positives → no
leaked rows downstream) is worth ~5.7 ps/row of probe cost; above it, residency decay makes the
exact set strictly worse on both axes. The residency sweep (2 B uniform probes, GB300,
L2 = 115.5 MiB) shows the in-list flat at 10.4–10.5 ps/row through 32 MB (0.277×L2), then
climbing (12.5 at 80 MB, matching the q9 production profile) while Bloom holds 4.7 ps/row at
every size; 0.125 places the cap at ~14.4 MiB, safely inside the flat region.

## TPC-H SF1000 results (GB300, PR #1371 measurement-kit regime)

The only query whose decisions change at the default is **q9** (two L2-fitting sets demote to
Bloom: the 86.96 MB filtered-part set and the 80.0 MB supplier set). ABBA-interleaved
(I,B,B,I; best-of-3 per query per run; all runs byte-identical to reference):

| | branch | base `7b69a9cc` | delta |
|---|---|---|---|
| **q9** (firing) | **0.754 s** | **0.820 s** | **−8.1%** (arms non-overlapping) |
| suite (q3,5,9,10,19,21) | 3.064 s | 3.146 s | −2.6% |

Non-firing queries scatter within the ±3.5% run-to-run band. `fraction=0` experiment (always
Bloom): q19 −8.6% and q21 −2.1% on their newly-flipped filters, others within noise — evidence
the default is conservative; 0.125 is kept because exactness protects downstream operators on
shapes this suite doesn't exercise.

Validation: all runs byte-identical to the saved SF1000 references (which are themselves
byte-identical to a DuckDB-CPU oracle at SF1000); GPU path confirmed per run (runtimes in band,
per-key decision log at DEBUG, no fallback markers); `[dynamic_filter]` 106 cases / 549
assertions (6 new publisher cases incl. both knob endpoints and boundary bracketing),
`[config]`/`[context]` suites green; default-level log output unchanged (messages identical;
spdlog `file:line` prefixes shift with the edit).

## Notes for reviewers

- Deliberately minimal: no probe-volume terms, no selectivity priors, no coverage-gate changes —
  #1277 owns that territory and relocates these files; this diff is confined to the publisher's
  tier-2 decision expression, config registration, docs, and tests, and rebases trivially onto
  #1277 in either order.
- Sweep raw data and analysis are in the campaign records (`inlist-residency-sweep/`); happy to
  attach them to the PR thread on request.

## Checklist
- [x] Read CONTRIBUTING.md
- [x] Update human and agent documentation (docs/super-sirius/configuration.md row)


Jedi18 (approved):
LGTM thanks!

PR #1587, opened 2026-08-16
perf(expression): cascade cheap conjuncts ahead of expensive filter predicates

kevkrist:
## Description

`expression_evaluator::select()` evaluates a filter's whole conjunction in one pass — for q19's
part-table predicate that is a single 116-register JIT kernel (string OR-groups + an integer
range union) at 25% occupancy, evaluating every conjunct for all 200 M rows.

This PR adds a **filter cascade**: a top-level AND is split into structurally *cheap* conjuncts
(subtrees that are entirely elementwise fixed-width work — comparisons, BETWEEN, integer
IN-lists, NOT; anything else, strings included, is expensive **by default**) and the rest. The
cheap mask runs first; when its pass rate is ≤ `filter_cascade_max_pass_rate`, survivors are
gathered and the expensive conjuncts run on them alone; otherwise the masks are Kleene-ANDed
without a gather (never slower than baseline by more than the cheap pass). Out-of-scope gathers
(columns outside residual-referenced ∪ projected) refuse the cascade outright — no new peak-
memory point. Three-valued (NULL) semantics are preserved by construction and pinned by a 3VL
matrix test.

Cascade logic lives in its own TU (`filter_cascade.cpp` + `filter_cascade_internal.hpp`, per
the `specializations/` precedent) with the classifier directly unit-testable.

## Knobs

| knob | default | semantics |
|---|---|---|
| `filter_cascade_cheap_conjuncts` | `true` | master switch; `false` is operation-identical to pre-change code |
| `filter_cascade_min_rows` | 1048576 | engagement floor (launch-latency amortization) |
| `filter_cascade_max_pass_rate` | 0.75, validated [0,1] | gather-vs-AND break-even (~55 ps/surviving row gather vs ~180 ps/dropped row avoided JIT; s* ≈ 0.77, rounded down — not tuned to q19, whose rate is 0.300) |

## TPC-H SF1000 results (GB300, PR #1371 measurement-kit regime)

q19 is the only query where the cascade engages (measured pass rate 0.300, matching the
analyst estimate; zero activation on the other 21 queries).

- **ABBA 3v3** (A,B,B,A,A,B; full 22-query suite; best-of-3 hot): q19 **357.8 → 347.7 ms
  (−10.1 ms, −2.8%)**, above its 1.9% between-run floor on means; individual pairs
  +0.2/−9.0/−21.7 — q19's documented between-run swing makes single pairs noisy, means don't.
- **Same-build knob isolation** (adjacent runs, only the SET toggled): **350.1 → 327.5 ms
  (−22.6 ms, −6.5%)** — the confounder-free signal.
- Suite neutrality: FoM subset +22.9 ms, inside the ±24 ms floor; no other query beyond its
  floor; GPU pool peak ≤ baseline; JIT kernel cache entry count unchanged (no recompile thrash).
- Correctness: **22/22 byte-identical in every arm and run** (six campaign queries also match
  the SF1000 DuckDB-CPU oracle references); zero fallback markers; activation INFO in q19 only.
- Context for reviewers: a 2×2 evidence matrix (jit/interpret × cascade on/off, all
  byte-identical) confirms the known q19×ast_jit anomaly is real and unresolved by this PR —
  the cascade helps both strategies (−17/−18 ms), while a ~30 ms jit-vs-interpret gap persists
  (candidate future work: per-expression strategy selection).

## Honest caveats

- The pending #1474 (decompression pushdown) will pre-filter rows before this code sees them,
  which shrinks this PR's headline on q19; the mechanism remains valuable for filter shapes
  that survive pushdown (post-scan FILTER operators, non-pushdown-able conjuncts).
- Unit coverage: `[filter_cascade]` 20 cases / 321 assertions incl. classifier arms, boundary
  behavior at `max_pass_rate` (inclusive), Kleene 3VL matrix, empty/all-pass, and knobs-off
  identity. Full-suite failures on the shared measurement box were adjudicated 1:1 as
  pre-existing/environmental (base build reproduces the identical failure set).

## Checklist
- [x] Read CONTRIBUTING.md
- [x] Update human and agent documentation (docs/super-sirius/configuration.md + expression-executor.md)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -659,6 +659,7 @@ set(TEST_SOURCES
     test/cpp/expression/test_value.cpp
     test/cpp/expression_evaluator/test_expression_evaluator.cpp
     test/cpp/expression_evaluator/test_expression_evaluator_ast_equivalence.cpp
+    test/cpp/expression_evaluator/test_filter_cascade.cpp
     test/cpp/expression_evaluator/test_gpu_expression_translator.cpp
     test/cpp/expression_evaluator/test_nonduckdb_frontend_proof.cpp
     test/cpp/helper/test_cudf_utils.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -255,6 +255,7 @@ set(EXTENSION_SOURCES
     src/expression_evaluator/expression_evaluator_strategy.cpp
     src/expression_evaluator/ast_op_counter.cpp
     src/expression_evaluator/expression_evaluator.cpp
+    src/expression_evaluator/filter_cascade.cpp
     src/expression_evaluator/gpu_expression_translator.cpp
     src/expression_evaluator/regex/regex_playground.cpp
     src/expression_evaluator/specializations/between.cpp
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -529,9 +529,16 @@ These can also be set at load via the `SIRIUS_LOG_BACKEND`, `SIRIUS_LOG_DIR`, an
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `expression_evaluator_strategy` | `ast_interpret` | Expression evaluator strategy: `materialize`, `ast_interpret`, or `ast_jit` |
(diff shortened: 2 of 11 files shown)


kevkrist:
Shows only 0.2% improvement on TPC-H in general -- not worth the complexity. Closing.

PR #1591, opened 2026-08-17
fix(pin): key duckdb pin-cache identity on the table incarnation

Jedi18:
## Description

A pinned DuckDB table was identified by qualified name alone (`cache_entry_info`'s
`catalog_name` / `schema_name` / `table_name`). `DROP TABLE t; CREATE TABLE t (...)`
reuses that name, and every other check the cache applies — column layout, chunk
count, per-chunk row counts, memory-space placement — passes for any same-shaped
table. Two paths leaked the dropped table's rows:

- **Serving.** The plan-time probe found the pin, the MVCC guards passed (a new
  transaction satisfies `start_time >= v_base`; with no explicit `CHECKPOINT` the
  checkpoint generation is unchanged, since the pin suppresses auto-checkpoint), and
  the scan was served the dropped table's cached columns masked against the *new*
  table's storage. Silently wrong results, no error and no fallback.
- **Re-pinning.** `insert_pinned_entry`'s merge path keys only on row count, so a
  second `pin_table` under the same pin name after a recreate fused columns from two
  unrelated tables into a single entry.

Reproduced on `dev` — same name, same schema, same row count, different values:

| Run | `sum(a)` |
| --- | --- |
| GPU | `1,249,975,000` ← the **dropped** table's rows |
| CPU | `51,249,975,000` ← the correct rows |

### Fix

Identity now carries the table's **incarnation**: `DuckTableEntry::oid`, which DuckDB
draws from an instance-global monotonic counter (`DatabaseManager::NextOid`) and never
reuses.

- New `table_oid` on `duckdb_native_ingestible_table_info` and `cache_entry_info`,
  populated at both derivation sites — pin (`build_duckdb_pin_info`) and query
  (`build_duckdb_native_table_info`).
- `matches_duckdb_table` requires it. The qualified-name-only half is split out as
  `matches_duckdb_table_name`, used solely to *report* a superseded pin, never to
  decide that one may serve.
- New `same_source_as` gates the merge path, so a re-pin onto a different source
  replaces the entry instead of merging into it.
- `pinned_name_for_table` (the UPDATE/INSERT rejection) matches on the oid too, so
  writes to a recreated table are no longer blocked by the dead pin.

A miss alone is not enough. Under the pin's checkpoint suppression the disk-native
path is MVCC-blind and its on-disk image still belongs to the dropped table, so a
scan of a recreated table must not fall through to it. `sirius_plan_get.cpp` detects a
same-name / different-oid pin and throws `NotImplementedException` — a clean
transparent CPU fallback naming `unpin_table`, matching the existing cache-or-CPU
guards.

### Behaviour change

`ALTER TABLE` also mints a new catalog entry, so it now invalidates a pin as well.
That is the correct outcome — column storage indices shift under an `ADD`/`DROP
COLUMN`, and the previous name-only identity kept serving across it. Verified against
the pinned DuckDB revision that a fresh oid is minted only by `CREATE` and `ALTER`,
never by `INSERT`/`DELETE`/`UPDATE`/`CHECKPOINT` or reads, so the oid is stable across
the ordinary pin-then-query flow.

### Testing

- `test/cpp/scan/test_can_serve_with_columns.cpp` — matcher unit coverage: same
  qualified name at a different incarnation misses.
- `test/cpp/integration/test_pin_table_recreated_table.cpp` (new) — end-to-end for
  both leak paths: the serving case asserts the plan declines *and* that the values
  are the recreated table's; the re-pin case asserts the entry is replaced rather than
  fused. Both fail without the fix.

The serving test deliberately omits a `CHECKPOINT` after the recreate: one would bump
the checkpoint generation and trip the pre-existing staleness guard, masking this bug
behind a runtime error. The test comment records why it must not be added back.

Full run with the fix: 163 cases / 27,966 assertions pass across `[pin_table]`, all
four `[pin_table_mvcc*]` suites, `[can_serve]` and `[scan_manager]`.

No configuration options changed.

## Checklist
- [x] Read CONTRIBUTING.md

## References
Found during review of #1428; the bug predates that PR and is present on `dev`.

### DIFF
--- a/src/include/scan_manager/sirius_scan_manager.hpp
+++ b/src/include/scan_manager/sirius_scan_manager.hpp
@@ -113,11 +113,8 @@ class cache_entry_info {
   std::string catalog_name;                      ///< duckdb identity: catalog (attach alias)
   std::string schema_name;                       ///< duckdb identity: schema
   std::string table_name;                        ///< duckdb identity: table
-  /// duckdb identity: catalog object id of the pinned DuckTableEntry — the table's
-  /// *incarnation*. A qualified name is reused by a table that was dropped and
-  /// recreated (or altered); this id is not, so it is what keeps the previous
-  /// incarnation's cached rows from serving the new table. See
-  /// @c duckdb_native_ingestible_table_info::table_oid.
+  /// DuckDB catalog object id. This distinguishes tables that reuse the same
+  /// qualified name after a drop, recreate, or alter.
   duckdb::idx_t table_oid{0};
   duckdb::vector<duckdb::ColumnIndex> column_ids;  ///< cached columns, by primary index
   std::vector<std::string> names;                  ///< aligned with column_ids; gather keys
@@ -135,29 +132,20 @@ class cache_entry_info {
   [[nodiscard]] std::vector<std::size_t> can_serve_with_columns(
     const op::scan::ingestible_table_info& other) const;
 
-  /// Duckdb-identity check shared by can_serve_with_columns and the plan-time
-  /// MVCC guards — one matcher, so the probe and prepare can never drift.
-  /// Qualified name AND incarnation (@p oid, see @c table_oid). False for parquet
-  /// entries (empty table_name).
+  /// Match a DuckDB table by qualified name and catalog object id.
+  /// Returns false for parquet entries.
   [[nodiscard]] bool matches_duckdb_table(std::string_view catalog,
                                           std::string_view schema,
                                           std::string_view table,
                                           duckdb::idx_t oid) const;
 
-  /// Qualified-name-only half of @ref matches_duckdb_table, ignoring the
-  /// incarnation. Only for reporting a pin of a *superseded* incarnation (same
-  /// name, different oid) — never for deciding that an entry may serve a scan.
-  /// False for parquet entries (empty table_name).
+  /// Match only the qualified name. Used to find superseded pins, not cache hits.
   [[nodiscard]] bool matches_duckdb_table_name(std::string_view catalog,
                                                std::string_view schema,
                                                std::string_view table) const;
 
-  /// True when @p other describes the SAME source as this entry — same duckdb
-  /// table incarnation, or same parquet file set — irrespective of which columns
-  /// each caches. Read by @ref sirius_scan_manager::insert_pinned_entry to decide
-  /// whether a re-pin under an existing pin name may merge its columns into the
-  /// existing entry or must replace it. Two entries that are both identity-less
-  /// (no table name, no files) report false, so the caller replaces.
+  /// Whether both entries cache the same DuckDB table incarnation or parquet
+  /// file set. Identity-less entries never match.
   [[nodiscard]] bool same_source_as(const cache_entry_info& other) const;
 
   /// Parquet-identity check shared by can_serve_with_columns and the plan-time
@@ -235,10 +223,8 @@ struct pinned_entry {
   /// Representative memory space of a HOST-tier entry; the MVCC path expands
   /// it into a per-chunk vector. GPU-tier entries leave it null.
   cucascade::memory::memory_space* memory_space{nullptr};
-  /// Total number of rows across all pinned chunks. Used by insert_pinned_entry —
-  /// together with the cache identity — to decide whether a re-insert merges into
-  /// the existing entry (same source AND same row count → add unique columns) or
-  /// replaces it.
+  /// Total rows across pinned chunks. Re-insertion merges only when this and the
+  /// cache identity match.
   std::size_t num_rows{0};
   /// Zone-map sidecar: pin-time DuckDB types + per-chunk min/max statistics,
   /// positional with cache_info.column_ids. Absent (never prunes) when the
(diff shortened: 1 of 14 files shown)


src/include/scan_manager/sirius_scan_manager.hpp
```diff
@@ -111,10 +111,13 @@ namespace sirius::scan_manager {
...
-  std::string catalog_name;                        ///< duckdb identity: catalog (attach alias)
-  std::string schema_name;                         ///< duckdb identity: schema
-  std::string table_name;                          ///< duckdb identity: table
+  std::vector<std::string> resolved_file_paths;  ///< parquet identity (file set)
+  std::string catalog_name;                      ///< duckdb identity: catalog (attach alias)
+  std::string schema_name;                       ///< duckdb identity: schema
+  std::string table_name;                        ///< duckdb identity: table
+  /// DuckDB catalog object id. This distinguishes tables that reuse the same
+  /// qualified name after a drop, recreate, or alter.
+  duckdb::idx_t table_oid{0};
```
kevkrist: Can you be consistent with the other field documentation style? Use inline ///< instead of above-line ///, even if it wraps multiple lines. Unless it looks atrocious.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -2592,6 +2618,28 @@ pinned_entry const* sirius_scan_manager::find_pinned_entry_for_duckdb_table(
...
+  duckdb::idx_t table_oid,
+  std::optional<std::uint64_t> current_checkpoint_iteration) const
+{
+  for (auto const& [name, entry] : _pinned_entries) {
+    if (!entry.cache_info.matches_duckdb_table_name(catalog_name, schema_name, table_name) ||
+        entry.cache_info.table_oid == table_oid) {
+      continue;
+    }
+    // A later checkpoint makes the fresh disk read safe.
+    if (entry.mvcc != nullptr && current_checkpoint_iteration.has_value() &&
```
kevkrist: I don't think this is quite right. What if you pin, then checkpoint (bumping the checkpoint iteration), then drop the table? Then we will hit this `continue` and the stale disk table will be served, right?
Please add a test for this sequence.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -2010,6 +2013,25 @@ bool cache_entry_info::matches_duckdb_table(std::string_view catalog,
   return catalog_name == catalog && schema_name == schema && table_name == table;
 }
 
+bool cache_entry_info::matches_duckdb_table(std::string_view catalog,
+                                            std::string_view schema,
+                                            std::string_view table,
+                                            duckdb::idx_t oid) const
+{
+  return matches_duckdb_table_name(catalog, schema, table) && table_oid == oid;
```
kevkrist, line 2022: Not that it really matters, but this can *just* be an oid check at this point, no? Fine to leave as-is.

src/vss/vector_search.cpp
```diff
@@ -99,7 +99,7 @@ std::unique_ptr<cucascade::host_data_representation> run_vector_search(
   // The table must be GPU-pinned: the search reads its vectors and gathers the
   // output columns straight from GPU-resident chunks (same order the index built in).
   const auto* pin = ctx.get_scan_manager().find_pinned_entry_for_duckdb_table(
-    req.catalog, req.schema, req.table_name);
+    req.catalog, req.schema, req.table_name, req.table_oid);
```
kevkrist, line 102: I think you need to update this for the ANN index search too -- it will look up and use a stale index. Please add a test for this (rebuild and repin without index rebuild -> should refuse).

kevkrist (changes requested):

src/scan_manager/sirius_scan_manager.hpp
```diff
@@ -112,10 +112,13 @@ namespace sirius::scan_manager {
...
-  std::string catalog_name;                        ///< duckdb identity: catalog (attach alias)
-  std::string schema_name;                         ///< duckdb identity: schema
-  std::string table_name;                          ///< duckdb identity: table
+  std::vector<std::string> resolved_file_paths;  ///< parquet identity (file set)
+  std::string catalog_name;                      ///< duckdb identity: catalog (attach alias)
+  std::string schema_name;                       ///< duckdb identity: schema
+  std::string table_name;                        ///< duckdb identity: table
+  duckdb::idx_t table_oid{0};  ///< DuckDB catalog object id. Distinguishes tables that
+                               ///< reuse the same qualified name after a drop,
+                               ///< recreate, or alter.
```
kevkrist, line 121: It looks like ALTER TABLE preserves oid (duckdb/src/catalog/catalog_set.cpp:345), so name + oid can distinguish DROP/CREATE but not ALTER. If a table t consists of a single INTEGER column a, and it is pinned, then

ALTER TABLE t
ALTER COLUMN a TYPE INTEGER USING a + 100;
SELECT sum(a) FROM t;

will not capture the + 100. We may need to invalidate pins on relevant ALTER operations unless there is a way to sidestep the pin when detecting such operations (by, e.g., adding something to the cache key). 

Similarly,
ALTER TABLE t
ALTER COLUMN a TYPE BIGINT
will be cache-served with the narrower INTEGER type, unless I'm mistaken?

src/sirius_extension.cpp
```diff
@@ -2280,6 +2281,7 @@ static unique_ptr<FunctionData> SiriusVectorSearchBind(ClientContext& context,
   req.catalog             = entry.ParentCatalog().GetName();
   req.schema              = entry.ParentSchema().name;
   req.table_name          = entry.name;  // catalog-resolved name (matches query-side derivation)
+  req.table_oid           = entry.oid;
```
kevkrist, line 2284: Codex is telling me there's an issue here, with PREPAREd vector searches. The PREPARE caches the OID. Suppose the table is then DROPped and then reCREATEd. Later, when executing against the prepared vector search after the DROP, the stale OID will be used as lookup key in the pinned cache, resulting in an incorrect hit. Please add PREPARE -> DROP/CREATE -> EXECUTE regression test.

There may be a way to compare the cached OID with the OID in the 'current' the database catalog using the table name at execution-time (better, but potentially more invasive)? Or, at bind time, just observe whether there's a catalog version change (more conservative)? There may be other solutions.

src/vss/cuvs_index_cache.cpp
```diff
@@ -99,14 +100,34 @@ std::shared_ptr<const pinned_index_entry> cuvs_index_cache::find_by_column(
   for (auto const& kv : _entries) {
     auto const& entry = kv.second;
     if (entry->meta.catalog_name == catalog && entry->meta.schema_name == schema &&
-        entry->meta.table_name == table && entry->meta.column_name == column &&
-        canonical_metric(entry->meta.metric) == wanted) {
+        entry->meta.table_name == table && entry->meta.table_oid == table_oid &&
```
kevkrist, line 103: Again from Codex. The issue is that unpin_table removes only the scan_manager entry, not the cuvs_index_cache entry, making the following sequence unsafe:
1. Pin table and build its ANN index
2. Unpin the table
3. Insert or update vectors in the index and checkpoint.
4. Re-pin the same table withoug rebuilding the index.
5. Run a fresh ANN search.

The old index and the new pin will have the same catalog/name/OID, so the old index will be used. The existing "rebuild after re-pin reflects a new nearest row" test almost performs this sequence, but it rebuilds the index after re-pinning. Can you add a regression test for this specific sequence?

Possible solutions: invalidate the table's indexes whenever its pin is removed or replaced, or store and compare a pin identifier in the index metadata?

kevkrist (changes requested):
A few problematic sequences to consider.

Issue #1063, opened 2026-07-02
Ensure all data batches stored in repositories are idle (no read-only pins via owning_table_view owners)


bwyogatama:
## Problem

Not every `data_batch` parked in a `shared_data_repository` is `idle`. View-backed batches (`sirius::make_data_batch_from_view`, `src/include/data/data_batch_utils.hpp`) carry a type-erased owner that is typically a `read_only_data_batch` lock on a *source* batch — so for as long as the view-backed batch is alive (including while it sits in a repository), the source batch is pinned `read_only` and cannot return to `idle`. Producers of such batches include the pinned-table scan path and PROJECTION's zero-copy passthrough (`owning_table_view`, see `docs/super-sirius/data-management.md`).

**This is currently causing a real deadlock:** I'm encountering a deadlock in practice because of these read-only pins — execution ends up waiting on a batch that can never return to `idle`. So this is an active bug, not just invariant hardening.

Why this matters:

- The downgrade executor relieves memory pressure by upgrading a batch to `mutable` and converting it to HOST. A batch that is read-only-pinned by another batch's owner can't be upgraded — it either blocks or gets skipped as a downgrade candidate — so that memory is effectively unspillable.
- Tier movement and cloning (#1059) assume batches parked in repositories are in a known `idle` state; long-lived read-only pins violate that expectation and make movement policies harder to reason about.

## Proposed change

Enforce the invariant that **every batch stored in a data repository is idle**. Options:

- Materialize view-backed batches into owned tables before pushing them to a repository (`release_table()` already exists for on-demand materialization), keeping zero-copy views only within a task's lifetime.
- Rework the ownership model so views keep the underlying allocation alive without holding a `read_only` lock on the source *batch* (e.g. shared ownership of the allocation), so the source batch's state machine stays free.

Either way, add an assertion/check on `add_data_batch()` so violations surface immediately.

dhruv9vats:
This is a bit more involved that what it seemed like, and exploring the possibilities, here are some findings:

Some invariants that were established to settle on a fix:

1. Since only idle batches are spillable/movable, **every batch parked in a repository must be idle/self-owned** — Downgrade skips non-idle batches; a pinned source (inside gpu_table_repr) is unspillable and invisible to the downgrader.
2. **A batch's bytes should have exactly one physical owner** — Two owners ⇒ spilling one frees nothing and duplicates; "downgrade" must actually reduce memory footprint on the required tier.
[...]

PR #1593, opened 2026-08-17
feat: enforce task input invariants at compile-time part-1

dhruv9vats:
This PR simply updates sirius to be able to compile against updates in https://github.com/NVIDIA/cuCascade/pull/192, so we can incrementally move to a more bolted down data management strategy.

This is intentionally narrowly scoped to ease the review process, and the subsequent independent pieces will be proposed separately.

Part 1 of a few to move towards solving https://github.com/sirius-db/sirius/issues/1063 from the root of it.


dhruv9vats:
Do not merge until https://github.com/NVIDIA/cuCascade/pull/192 lands and this is rebased.

9prady9 (commented):
Change by itself looks fine, I have a related query - I understand most of the status quo existed before, I noticed it now only - hence asking.

Several forwarding operators now construct their output from get_data_batches() instead of get_read_only_batches(). The shared_ptrs preserve the batch's lifetime, but the output no longer carries the input's read locks. Once the previous input object is destroyed, a batch shared with another task could become eligible for spilling before the next operator reads it. Theoretically that can cause issues, but I wonder if we are preventing that somehow ?

dhruv9vats:
> Change by itself looks fine, I have a related query - I understand most of the status quo existed before, I noticed it now only - hence asking.

And they are most welcome :)

The inputs are already prepared and pinned before a task (of pipelines runs), so there is no need to move read_only locks through the same: https://github.com/sirius-db/sirius/blob/f35092ffd10aeeea82fb793996f1b54ce477aa15/src/pipeline/gpu_pipeline_task.cpp#L635

and in a follow, for simple passthrough batches, we will just pass a flag instead of carrying forward the shared_ptrs.

9prady9 (approved):

dhruv9vats:
Still needs to land after https://github.com/NVIDIA/cuCascade/pull/192 lands and cuCascade is bumped.

Issue #1590, opened 2026-08-17
feat(exec): non-blocking fragment execution and a push/pull FFI surface


aocsa:
Follow-up to #1481, which lands the streaming fragment as a **blocking** runner.

## What's missing

#839 asks for a session that "builds a streaming plan, **starts it on the existing task scheduler
without blocking**, and exposes `push` + `close_input` (#836), `pull` + `wait` (#837)". #1481
delivers the plan-building, binding, and execution half, but not the asynchrony:

1. **`run()` blocks.** `streaming_fragment::run()` calls `sirius_engine::execute()`, which does
   `auto future = ...start_query(); future.get(); wait_for_completion();`. The scheduler already
   exposes the non-blocking half — `start_query()` returns a future — and the current code
   discards it. No `start()`/`poll()`/`join()` split exists.

2. **The `Fragment` FFI exposes no `push`/`pull`/`wait`.** `exec::stream_session` has all of them
   and `streaming_fragment::session()` hands it out, so in-process C++ callers are fine. But the
   Rust-facing `Fragment` surface offers only `close_input()` and `relay_from()`. With no `push`,
   a wrapper cannot inject batches it received from a remote peer.

## Why it matters

Together these force **store-and-forward, one fragment at a time**. It is visible in the contracts
#1481 ships: `relay_from()` must be called after `source.run()` and before `this->run()`, and
"exactly one fragment may sit between its own `build()` and `run()`". A real multi-node plan wants
fragments running concurrently with data flowing between them, which neither constraint allows.

The design doc also concedes the transport half: "**Not yet ported:** `export_packed` /
`push_packed` / `StagingArena`".

## Scope

- [ ] Split `streaming_fragment::run()` into `start()` / `join()` (or `poll()`), keeping the
      blocking `run()` as the composition of the two for existing callers.
- [ ] Expose `push`, `pull`, `wait`, `drained` on `sirius::ffi::Fragment`.
- [ ] Revisit "exactly one fragment between build() and run()" once fragments can overlap. Note
      `streaming_fragment` teardown was recently narrowed to erase only its own bind-catalog ids
      precisely so concurrent fragments on one connection stop clobbering each other — that fix is
      in #1481 and is a prerequisite here.
- [ ] Port `export_packed` / `push_packed` / `StagingArena` for cross-process transport.

## Not in scope

Anything already landed by #1479, #1480, #1481.

aocsa:
### Where this stands after #1481

[...]

PR #1598, opened 2026-08-18
feat(ffi): Rust bindings for sirius::ffi::Fragment

aocsa:
Completes the Rust half of #1396 (closed). The C++ half shipped in #1481; this binds it.

> **Purely additive.** No existing signature changes — `SiriusContext` only gains `fragment()`.

**What.** `sirius::ffi::Fragment` — one plan fragment of a multi-fragment query — reachable from
Rust. A fragment declares its input streams (schema and senders) and its output streams, is built
from a Substrait plan, relays native GPU batches to and from its peers, and either parks results
for a downstream fragment or hands them back as Arrow.

**Design: a fragment borrows the context, it does not own it.** That single fact drives the whole
shape. Several fragments of one query are alive simultaneously — senders parked, waiting on their
receiver — so the factory has to be `SiriusContext::fragment(&self)`. A `&mut self` factory would
permit exactly one fragment at a time, which is the one thing a multi-fragment query cannot live
with.

But every call into C++ needs `Pin<&mut Context>`. Reconciling those puts the handle behind a
`RefCell<UniquePtr<Context>>`: the borrow is taken and released inside each call, never held
across one. That adds `!Sync` to the `!Send` the cxx opaque type already carried; both are
documented on the type rather than left to be discovered.

`execute_substrait` and `execute_substrait_result` keep `&mut self`. The `RefCell` does not force
them to give up the exclusive borrow, and holding it earns something: running a whole-plan query
while a fragment is alive stays a *compile* error instead of becoming a runtime rejection. What
the `&self` factory does cost is compile-time enforcement of "exactly one fragment sits between
its own `build()` and `run()`" — two can now be mid-lifecycle at once, and the engine rejects
that at runtime. Both facts are recorded on the type.

One consequence worth flagging for review: an `Arc<SiriusContext>` cannot reach a `&mut self`
method, so the `Arc`-threading sketch in the `StubExecutor` TODO
(`experimental/starrocks/src/fragment_executor.rs`) would have to adopt the shape `SiriusEngine`
already uses: an owned context pinned to its own thread. Nothing depends on that sketch today, and if we do want
`Arc` later, that is the argument for `&self` — worth saying now rather than discovering it then.

`Fragment<'ctx>` carries the context's lifetime, so a fragment cannot outlive the engine it was
built on; that is checked at compile time rather than trusted.

---

**What is bound.** The full surface `sirius_ffi.hpp` exports as of #1481:

| | |
|---|---|
| lifecycle | `make_fragment`, `build`, `run` |
| inputs | `declare_input_column`, `declare_input_sender`, `close_input` |
| outputs | `declare_output`, `declare_output_broadcast`, `declare_output_hash_key` |
| moving data | `relay_from`, `result_to_arrow`, `output_batch_count`, `output_types` |
| helper | `stream_view_name` |

Four of those — `declare_output_broadcast`, `declare_output_hash_key`, `close_input`,
`output_types` — postdate #1396 and are bound here for the first time.

**Preconditions surface as `Err`, not panics.** #1481 made three contracts explicit, and the doc
comments mirror each one rather than leaving a caller to meet a surprising exception:

- a partition mode on a single-output fragment is **rejected**, not silently ignored — every row
  would otherwise go to destination 0 while the call looked like it worked;
- `relay_from` requires the source fragment to have `run()` already, because before that an empty
  stream and a finished one are indistinguishable and the input would close early;
- `relay_from` requires the target input stream to be declared on the receiving fragment, and
  refuses a result fragment as a source;
- a plan that reads the same declared input stream id more than once (e.g. a self-join) is
  **rejected** at `build()` time instead of silently letting the second leaf overwrite the
  first's registration and hang forever with no error.

The usage order — `declare_* → build → relay_from(every sender) → run → drain` — is on the
`Fragment` type, along with the constraint that exactly one fragment may sit between its own
`build()` and `run()`.

---

**Reviewing.** Two files, no C++ changes:

1. `rust/crates/sirius-sys/src/lib.rs` — the `#[cxx::bridge]` declarations. Thin by design; every
   entry documents the C++ contract it binds, and `result_to_arrow` keeps its `# Safety` section
   because it takes a raw `ArrowArrayStream` address.
2. `rust/crates/sirius/src/lib.rs` — the safe wrapper. This is where the `RefCell` and the
   `Fragment<'ctx>` lifetime live, and where `result_to_arrow`'s raw address is made sound by
   reusing the crate's existing `FFI_ArrowArrayStream` path rather than adding a second one.

`build.rs` is unchanged — it already compiles the bridge against `src/include/sirius_ffi.hpp`.

**Tests.** CI runs `cargo test --no-run`, so all of these compile without a GPU:
`routing_modes_are_mutually_exclusive` (a hash key after broadcast and the reverse, both rejected
at declare time) and `context_makes_several_fragments_at_once` (two live fragments — the case
that forces the `&self` factory) both need a device to *run*, because they bring up a context;
`stream_view_name_matches_the_engine_convention` is pure formatting and runs anywhere. That a
`Fragment` cannot outlive its `SiriusContext` is enforced by `Fragment<'ctx>` itself, so there is
no test for it.

---

**Not here.** No `push` / `pull` / `wait` on the Rust surface, and `run()` still blocks — so a
remote peer cannot yet feed a fragment. That is the deferred half of #839, tracked in #1590. This
PR binds what #1481 exports and nothing more.

### DIFF
--- a/rust/crates/sirius-sys/src/lib.rs
+++ b/rust/crates/sirius-sys/src/lib.rs
@@ -5,9 +5,10 @@
 //! wrappers live in the [`sirius`](https://docs.rs/sirius) crate.
 //!
 //! The bridge binds Sirius's **public C++ surface** (`src/include/sirius_ffi.hpp`):
-//! an RAII [`Context`] held via [`cxx::UniquePtr`]. Constructing it brings up an
-//! initialized engine; dropping the `UniquePtr` tears it down. The header is
-//! lightweight, so the bridge compiles without any of Sirius's internal headers
+//! an RAII [`Context`] held via [`cxx::UniquePtr`], plus the [`Fragment`] it
+//! creates — one plan fragment of a multi-fragment query. Constructing a context
+//! brings up an initialized engine; dropping the `UniquePtr` tears it down. The
+//! header is lightweight, so the bridge compiles without any of Sirius's internal headers
 //! (cudf/rmm/duckdb). It is the seed of the public API `libsirius` will expose;
 //! the bindings link whichever Sirius artifact provides these symbols (the DuckDB
 //! extension today, a dedicated `libsirius` later — see `build.rs`).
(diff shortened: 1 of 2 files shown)


aocsa:
Superseded by #1702 (layer 2 of the ffi stack): the same Rust bindings re-based onto current dev on top of the transaction fix in #1697, with the fan-out tests added. Closing this draft.

Issue #1599, opened 2026-08-18
feat:  UNION ALL operator (N-ary, multi-GPU, zero-migration)


jjohnson-99:
### Describe the problem to solve or your goal to achieve

Sirius intercepts at the logical plan, so any operator it cannot lower throws `NotImplementedException` and replays the *entire* query on CPU DuckDB. Set operations are still on a blanket throw, so one `UNION ALL` in an otherwise GPU-friendly query forfeits the whole plan.

`UNION ALL` is bag concatenation — there is no de-duplication, key, or ordering guarantee. It routes rows and does nothing else. That makes it a high value-to-effort feature and opens the rest of the set-operator family (`UNION` distinct is `DISTINCT(UNION ALL)`; `EXCEPT` / `INTERSECT` are de-duplication over a semi/anti join).

### Describe the solution you would like

An N-ary `UNION` operator whose `execute` is the identity, reached by routing `LOGICAL_UNION` to a new `create_plan(LogicalSetOperation&)` guarded on `setop_all`. Topologically it is a join minus the work, so it reuses the join's fan-in skeleton, but it reports READY when *any* arm has a batch rather than all of them, since arms are independent and rarely the same length.

The key piece is a second new operator, **`PASSTHROUGH_SINK`**, terminating each arm. Arms need a pipeline sink, and the existing one is the join's `PARTITION -> CONCAT` chain — which would destroy the point of this operator. `PASSTHROUGH_SINK` is that chain collapsed to one operator, because a bag union needs neither a shuffle nor a coalesce, and it emits plain unpartitioned data.

### Describe alternatives you have considered

Reusing `PARTITION -> CONCAT` as the arm terminator (above — costs multi-GPU placement, and is more code than the new sink). `cudf::concatenate` in `execute`, which would re-materialize data that can be forwarded untouched. A binary-only operator, which would CPU-fall-back on `a UNION ALL b UNION ALL c` — DuckDB flattens that to a single three-child node.

### Sirius component

Operators & Expressions

### Additional context

Distinct `UNION`, `EXCEPT` and `INTERSECT` stay on CPU fallback behind the `setop_all` guard; each gets its own ticket.

bwyogatama:
Thanks @jjohnson-99 for posting this and looking forward to the PR!

jjohnson-99:
> Thanks [@jjohnson-99](https://github.com/jjohnson-99) for posting this and looking forward to the PR!

[...]

PR #1604, opened 2026-08-18
feat(op): Add native UNION ALL operator

jjohnson-99:
## Description

Lowers DuckDB's `LOGICAL_UNION` (`setop_all == true`) onto a new N-ary physical UNION operator, so a query containing `UNION ALL` no longer forfeits the whole plan to the CPU. Distinct `UNION`, `EXCEPT` and `INTERSECT` continue to fall back, guarded.

Bag union computes nothing, so the operator only routes batches. UNION's `execute` is the identity. Each arm is terminated by a new `PASSTHROUGH_SINK` rather than the join's

distinct UNION is out of scope but becomes available once LOGICAL_DISTINCT is lowered. `src/planner/sirius_get_operator::create_plan` currently gates on `!op.setop_all`.

Four wiring decisions worth review:

- **A distinct `union_{i}` port per arm.** `add_port` is last-writer-wins and the repository manager keys by `(operator_id, port_id)`, so a shared name would orphan an arm's repository and let the pipeline finish with rows still pending.
- **`PASSTHROUGH_SINK -> UNION` is `PARTIAL`.** `FULL` waits for every batch from every arm before UNION emits anything. Here `PARTIAL` waits for *one batch from each live arm*.
- **`UNION`** overrides both task-driver methods and the ranking inside the hint matters. A starved arm is one whose port is empty but whose source pipeline has not finished, so rows are still coming. `sirius_physical_union::get_next_task_hint()` names a starved arm first, returning `task_creation_hint(TaskCreationHint::WAITING_FOR_INPUT_DATA, starved_producer)` and falls through to READY only when nothing is starved and at least one port holds a batch. A finished arm is not starved so it is excused rather than blocking forever. Ranking `READY` first would strand a starved arm outright, since `task_creator::get_operator_for_next_task` reaches an operator only by walking `hint.producer`. A `READY` hint puts `task_creator` in a `while (!node->all_ports_empty())` loop over `get_next_task_input_data`, which pops one batch from one arm per call and rotates across arms. One batch per task, rather than the base's one-from-every-port bundle, keeps each batch on the GPU that produced it, since a task runs on one device. The hint is re-evaluated per request and every completed arm task schedules one, so these rounds are short and interleave with the arms still producing rather than running as two phases. The query ends when every arm's pipeline has finished and its port is drained.
- **`source_order()` is `NO_ORDER`.** `order_preservation_recursive` stops at the first `is_source()` operator, so UNION's answer decides the plan's, and `INSERTION_ORDER` would claim an ordering a bag union does not provide.

**Tests**. Eleven integration cases compare GPU against CPU, asserting one GPU execution and zero fallbacks, plus a plan-time fallback check for the three declined set operations. `test_physical_union_mgpu.cpp` adds four more, driven by `scan_task_batch_size` rather than hash_partition_bytes, UNION is deliberately not partition-based so the usual lever does nothing. Two run at num_gpus = 1 and are the regression test for the starvation hang; two need two GPUs. All green on a single T4. [gpu_execution] is 781/797, and the 16 failures are identical by name to a clean dev build, this branch adds none.

**Caveats**. Data locality, as in the UNION task being executed on the GPU which produced it, has not been tested and there is no mechanism to enforce it. The current implementation of UNION only assumes locality because there is no producer stamp and no partition_idx so device placement falls to the third rung, being locality.

## Checklist
- [x] Read CONTRIBUTING.md
- [x] Document configuration changes in code and su above - n/a no configuration changes.

## References
Refs #1599

### DIFF
--- a/src/op/sirius_physical_union.cpp
+++ b/src/op/sirius_physical_union.cpp
@@ -146,32 +146,36 @@ std::optional<task_creation_hint> sirius_physical_union::get_next_task_hint()
   const auto num_arms      = ports_by_arm.size();
   if (num_arms == 0) { return std::nullopt; }
 
-  sirius_physical_operator* live_producer = nullptr;
-  std::size_t live_arm                    = 0;
+  bool any_ready                             = false;
+  sirius_physical_operator* starved_producer = nullptr;
+  std::size_t starved_arm                    = 0;
   for (std::size_t offset = 0; offset < num_arms; offset++) {
-    const auto arm = (_wait_cursor + offset) % num_arms;
-    auto* p        = ports_by_arm[arm];
-    // Readiness is ANY, not ALL: one arm with a queued batch is enough to fire a task.
-    if (p->repo && p->repo->total_size() > 0) {
-      return task_creation_hint{TaskCreationHint::READY, this};
-    }
-    // Carry a still-producing arm rather than returning here: a later arm may have data, and
-    // READY outranks WAITING.
-    if (live_producer == nullptr && p->src_pipeline && !p->src_pipeline->is_pipeline_finished()) {
-      live_producer = &(p->src_pipeline->get_operators()[0].get());
-      live_arm      = arm;
+    const auto arm      = (_wait_cursor + offset) % num_arms;
+    auto* p             = ports_by_arm[arm];
+    const bool has_data = p->repo && p->repo->total_size() > 0;
+    const bool live     = p->src_pipeline && !p->src_pipeline->is_pipeline_finished();
+    any_ready           = any_ready || has_data;
+    if (!has_data && live && starved_producer == nullptr) {
+      starved_producer = &(p->src_pipeline->get_operators()[0].get());
+      starved_arm      = arm;
     }
   }
 
-  // `task_creator::get_operator_for_next_task` selects the next operator to run *only* by walking
-  // `hint.producer`, so a null producer here would leave the still-producing arm with nothing to
-  // run it. The walk is also all-or-nothing: when it yields no operator, `task_creator` abandons
-  // the whole task-creation request, so repeatedly nominating one stalled arm concentrates that
-  // loss on it. Advancing past the arm just nominated spreads it across the arms instead.
-  if (live_producer != nullptr) {
-    _wait_cursor = (live_arm + 1) % num_arms;
-    return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, live_producer};
+  // A starved arm outranks a ready one, because nothing else will start it.
+  // `task_creator::get_operator_for_next_task` reaches an operator *only* by walking
+  // `hint.producer`, and `task_scheduler::start_query` schedules `scans.front()` alone — so an arm
+  // is named here or never runs, and answering READY first spends that request on draining. The
+  // producer is likewise never null, or the still-producing arm has nothing to run it. The walk is
+  // all-or-nothing too: when it yields no operator `task_creator` abandons the whole request, so
+  // advancing past the arm just nominated spreads that loss instead of concentrating it on one.
+  if (starved_producer != nullptr) {
+    _wait_cursor = (starved_arm + 1) % num_arms;
+    return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, starved_producer};
   }
+
+  // Readiness is ANY, not ALL: one arm with a queued batch is enough to fire a task. An arm that
+  // finished without producing is not starved, which is what lets an empty arm through.
+  if (any_ready) { return task_creation_hint{TaskCreationHint::READY, this}; }
 
   return std::nullopt;
 }
--- a/src/op/sirius_physical_union.cpp
+++ b/src/op/sirius_physical_union.cpp
@@ -31,7 +31,6 @@ sirius_physical_union::sirius_physical_union(duckdb::vector<sirius::logical_type
   : sirius_physical_operator(
       SiriusPhysicalOperatorType::UNION, std::move(types), estimated_cardinality)
 {
-  // Nothing to configure: no keys, no join type, no dedup.
 }
 
 std::string sirius_physical_union::get_name() const { return "UNION"; }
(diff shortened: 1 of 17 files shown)


src/planner/sirius_plan_set_operation.cpp
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  auto union_op = duckdb::make_uniq<sirius::op::sirius_physical_union>(
+    sirius::from_duckdb_vec(op.types), op.estimated_cardinality);
+
+  for (auto& child : op.children) {
+    auto child_plan = create_plan(*child);
+    // Re-check the binder's arity invariant: a mismatch means it was misread, and falling back to
+    // the CPU beats emitting a differently-shaped batch on one arm.
+    if (child_plan->types.size() != op.types.size()) {
```
mbrobbel: This isn't always the output schema. For a materialized CTE this describes the materialized data, which can differ from what the arm returns.

src/op/sirius_physical_union.cpp
```diff
@@ -0,0 +1,217 @@
...
+  // resumes each time its splits run dry — answering READY first spends that request on draining.
+  // The producer is likewise never null, or the still-producing arm has nothing to run it. The walk
+  // is all-or-nothing too: when it yields no operator `task_creator` abandons the whole request,
+  // and an arm whose connector closed while its last tasks run is starved but yields nothing — so a
+  // fixed start would burn every request for that arm's tail, which is what `_wait_cursor` advances
+  // past.
+  if (starved_producer != nullptr) {
+    _wait_cursor = (starved_arm + 1) % num_arms;
+    return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, starved_producer};
+  }
```
mbrobbel, line 178: Maybe we can also check `any_ready` here (before returning `WAITING_FOR_INPUT_DATA`), schedule the starved producer (if it isn't already scheduled) and return `READY` to prioritize consuming queued batches?

test/cpp/operator/test_physical_union_mgpu.cpp
```diff
@@ -0,0 +1,281 @@
...
+  std::map<int, size_t> tasks_per_gpu;
+  scoped_mgpu_env env(yaml);
+  auto con = std::make_unique<duckdb::Connection>(env.make_connection());
+  require_gpu_matches_cpu(*con, query, /*force_cpu_reference=*/true);
+  auto& scheduler = env.get_task_scheduler(*con);
+  con.reset();  // flush sinks before reading metrics
+
+  scheduler.visit_executors(
+    [&](int device_id, sirius::pipeline::gpu_pipeline_executor const& exec) {
+      tasks_per_gpu[device_id] = exec.get_metrics().tasks_executed;
```
mbrobbel, line 153: We should only count union tasks here to make sure it actually runs on multiple GPUs.

mbrobbel (commented):
@jjohnson-99 thank you for your contribution!

Some questions/suggestions:

src/planner/sirius_plan_set_operation.cpp
jjohnson-99: I hadn't thought about that. I believe removing line 58, changing `child_plan->types.size()` to `child->types.size()` on line 61 and `std::move(child_plan)` to `create_plan(*child)` on line 65 should work, right? That way when `child` is the materialized CTE, we are still using the override to capture the size of the body rather than the inner query.

mbrobbel: ﻿I think that fixes the check, but [`wrap_union_child()`](https://github.com/sirius-db/sirius/pull/1604/changes#diff-f65a520fc0e9373ede880eef036979dcbfe43921959e18a84b4af499b21f868eR656) copies `child_orig->types` into the passthrough sink. For a CTE, those don't match the body output. We should fix that too I guess.

jjohnson-99: Addressed in commit c15a6556: I changed the guard to read `child->types`, this fixes the guard when `child_plan->types` does not match the actual output schema. For the `wrap_union_child()`, it now copies `union_op.types` instead of `child_orig->types`. The binder already casts the arms’ types so they match `union_op`’s declared types, so they match what the arm emits. I also added tests to `test_gpu_execution_union_all.cpp` to check the materialized CTE cases and added tests to `test_plan_tree_shape.cpp` which I had neglected to do. Tested on a single T4.

This does make me wonder whether other operators are also exposed to the `child_orig->types` misdirect. I ran tests which seem to indicate it's possible but I'm not confident enough to say for sure. It would be a follow-up in anycase.

test/cpp/operator/test_physical_union_mgpu.cpp
jjohnson-99: I’m confused about this one. Is there a mechanism to count operator specific tasks? The `executor_metrics` struct only has a general `tasks_executed` member. Other operators have this same scheduler.visit_executor(…) and REQUIRE(…) checks. Some have `partition_idx` so the device placements are necessarily distributed and the test is incidentally correct, but not all operators which have the same checks in their `*_mgpu.cpp` have partition-based placement.

kevkrist: @mbrobbel any thoughts?

src/op/sirius_physical_union.cpp
kevkrist: I think it makes the most sense to do a sequential sweep across the input sources, draining one at a time, for now. This is what DuckDB does, FWIW, and shoud reduce memory pressure. @jjohnson-99 what do you think?

test/cpp/operator/test_physical_union_mgpu.cpp
mbrobbel: Oh sorry for the late reply. This was mostly a nit, suggesting to add the machinery for that (task per operator tracking). But we don't have to do that in this PR!

mbrobbel (approved):
Thanks @jjohnson-99

mbrobbel:
It looks like we need to update this PR to handle changes from #1593

src/op/sirius_physical_union.cpp
jjohnson-99: I'll have this version committed in an hour or so.

jjohnson-99: @kevkrist wrote a sequential-drain version which I used, with small adjustments. A few variations were benchmarked but speed and memory pressure were within noise so I stayed with Kevin's as his was the simplest. I added a guard to only run the promotion if the lookahead queue is enabled, since only the initial scan and those added by the lookahead can be promoted. Promoting the initial scan is redundant and wastes a request. The order in which arms were drained vs its execution priority were backwards so I reversed the order UNION's children were built.

mbrobbel: Let's get this in! We can do follow-up PRs if needed.

PR #1606, opened 2026-08-18
perf(join): fuse COUNT-grouped-by-join-key outer joins into a dense-count operator

kevkrist:
## Summary

During physical planning, this recognizes COUNT(col) or COUNT(*) GROUP BY join_key over eligible LEFT/RIGHT single-column integer equi-joins (this can be extended in the future). It plans only one DENSE_COUNT_JOIN instead of the normal PARTITION/CONCAT, HASH_JOIN, projection, HASH_GROUP_BY, and merge operator tree.

The fused operator counts matches by key directly, avoiding join-row materialization and re-aggregation while preserving outer-join and NULL semantics. It uses a dense histogram when the observed key range fits the memory budget and an exact sparse fallback otherwise. The feature is enabled by default.

Three shapes that the COUNT GROUP BY optimization gives up:
 - SEMI-delim under LIMIT *either* side
 - ANTI-delim under LIMIT preserved side

## Performance

TPC-H SF1000 on GB300, per-query GPU pinning, HOT total (minimum, dashboard-matching 18-query intersection):

- origin/dev: 7.661 s
- DENSE on: 7.220 s
- 5.75% faster

Q13: 1.259 s to 0.838 s (33.42% faster). All 22 queries validated against DuckDB.


9prady9 (commented):
Nice fix with OOM based reties after bad estimations.

A few general observations:

- The fixed size of 2 GB could be conservative in certain situations like GB200/GB300 and  greedy in low memory size cards. It would be nice if we can adjust this dynamically. We don't need to do this now, can be done in a followup.
- Can multi-input ingress policies be expressed declaratively or shared between join operators, instead of duplicating input_port_for implementations ?

I will leave any CUDA impl related comments soon.

Thank you!, this is a great optimization.

9prady9 (approved):
CUDA impl also looks good. I want to do some experimentation of the kernel itself once it is in.

Good to merge I would say, other minor nits I have mentioned can be addressed in followups.

@kevkrist Great work once again.

src/op/sirius_physical_dense_count_join.cpp
```diff
@@ -0,0 +1,663 @@
...
+      "dense_count_join: input batch {} has {} columns; {} column index {} is out of range",
+      batch_index,
+      num_columns,
+      role,
+      index);
+  }
+  return batch.column(static_cast<cudf::size_type>(index));
+}
+
+[[nodiscard]] int64_t checked_null_count(cudf::column_view const& column, std::size_t batch_index)
```
wmalpica: In general the AI was very conservative in this PR, sometimes that is good. But this function is an example of going too far in being conservative and checking everything. I would remove this and other functions that are excessively conservative like checked_column.

src/cuda/dense_count_join_impl.cu
```diff
@@ -0,0 +1,607 @@
...
+
+  for (auto const& column : keys) {
+    if (column.size() == 0 || column.size() == column.null_count()) { continue; }
+
+    auto minmax = cudf::minmax(column, stream, mr);
+    auto launch = [&](auto key_tag) {
+      using KeyT            = decltype(key_tag);
+      auto const& batch_min = static_cast<cudf::numeric_scalar<KeyT> const&>(*minmax.first);
+      auto const& batch_max = static_cast<cudf::numeric_scalar<KeyT> const&>(*minmax.second);
+      merge_extrema_kernel<KeyT>
```
wmalpica: Do we really want to have GPU kernels do get a min max from two scalars? that seems excessive.  especially if we are just moving the results back to host anyways. Just move the scalar to host and do this simple logic on host, no?

wmalpica (commented):
still reviewing

src/cuda/dense_count_join_impl.cu
9prady9: my understanding is that it was done to avoid N syncs if say there are N columns.

I would like to handle this much differently in a followup. Hence my comment about experimenting with these a bit in a followup. I basically want to use CUB to write a custom kernel that does all of this in single launch with cross column reduction. If you are okay with that direction, I would like to create an issue and work on that as a follow up.

src/op/sirius_physical_dense_count_join.cpp
kevkrist: Agreed. I've cleaned most of this up (both functions you mentioned removed).

wmalpica (approved):

src/cuda/dense_count_join_impl.cu
kevkrist: @9prady9 happy to leave this as follow-up for you!

kevkrist:
> Nice fix with OOM based reties after bad estimations.
> 
> A few general observations:
> 
> * The fixed size of 2 GB could be conservative in certain situations like GB200/GB300 and  greedy in low memory size cards. It would be nice if we can adjust this dynamically. We don't need to do this now, can be done in a followup.
> * Can multi-input ingress policies be expressed declaratively or shared between join operators, instead of duplicating input_port_for implementations ?
> 
> I will leave any CUDA impl related comments soon.
> 
> Thank you!, this is a great optimization.

Thanks! Happy to leave these as follow-ups for you @9prady9 in the interest of getting this merged ASAP.

src/op/sirius_physical_dense_count_join.cpp
```diff
@@ -0,0 +1,569 @@
...
+    if (p->src_pipeline && !p->src_pipeline->is_pipeline_finished()) {
+      auto* producer = &(p->src_pipeline->get_operators()[0].get());
+      return task_creation_hint{TaskCreationHint::WAITING_FOR_INPUT_DATA, producer};
+    }
+  }
+  if (!all_ports_empty()) { return task_creation_hint{TaskCreationHint::READY, this}; }
+  return std::nullopt;
+}
+
+std::unique_ptr<operator_data> sirius_physical_dense_count_join::get_next_task_input_data()
```
wmalpica, line 361: Its worth pointing out that the way this operator works is that it needs to pull all the data from two sides of a join all into one task. 
In general this is a very bad pattern. In this case for now it is tolerable because this operator would only be invoked in very specific scenarios and these scenarios basically guarantee a very small number of columns.
I imagine will revisit this operator and this optimization in general to expand its applicability. When we do, we should also address this scalability issue. 
One option is to maintain an in-memory accumulator that stays in memory for the duration of the operator

wmalpica: OH wait?!?!?!?! in a multi-gpu scenario how are we co-locating the join keys? i dont think this works for multi-gpu

wmalpica (changes requested):

wmalpica (approved):

9prady9:
Here is the issue https://github.com/sirius-db/sirius/issues/1666

PR #1610, opened 2026-08-18
perf(strings): SWAR fast path for multi-literal LIKE patterns

kevkrist:
## Summary

Adds a default-on GPU fast path for ordered multi-literal LIKE and NOT LIKE patterns. The kernel scans each string eight bytes at a time for candidate matches, verifies each literal exactly, and finds the literals in pattern order. NOT LIKE is inverted in the same kernel.

The fast path applies to `%lit1%lit2%...%` patterns with 1–4 non-empty ASCII literals of at most 32 bytes each, no `_`, escape clause, NUL, or non-ASCII bytes, and an aligned cuDF character buffer with 32- or 64-bit offsets. Other patterns and layouts fall back to `cudf::strings::like`.

## Performance

TPC-H SF1000 using the sirius-ci configuration and per-query GPU pinning, with DENSE_COUNT_JOIN default-on:

- origin/dev: 9.743 s
- PR + DENSE-on: 8.807 s (9.61% faster); Q13: 1.270 s → 0.217 s
- 22/22 queries validated

These runs used NVIDIA driver 595.84. The nightly 7.6 s result used driver 610.43.03, so the same sirius-ci configuration did not reproduce that absolute number and no direct comparison to 7.6 s is claimed.


src/expression_evaluator/specializations/function.cpp
```diff
@@ -178,22 +180,36 @@ evaluate_result expression_evaluator::evaluate(sirius::ast::function_call const&
...
+    return {evaluate_result(std::move(input)), std::string_view(match_str)};
   };
   if (resolved_id == function_id::like || resolved_id == function_id::not_like) {
     auto [input, match_str] = setup_string_matching();
+    auto const invert       = resolved_id == function_id::not_like;
+
+    // `%lit1%lit2%...%litN%` patterns take a SWAR digram fast path (NOT fused in);
+    // everything else — and any ineligible column layout — takes cudf::strings::like.
+    if (_like_swar_fastpath && !input.is_scalar()) {
+      auto const parsed = _like_cache->get_or_classify(match_str);
```
Jedi18: This shared cache is checked for every batch, and each check takes a mutex even when the pattern is already cached or will fall back to cuDF. 

With concurrent batches, this creates unnecessary contention. Can we cache the result in the evaluator after the first lookup?

src/include/expression_evaluator/like_multiliteral.hpp
```diff
@@ -0,0 +1,184 @@
...
+  int32_t n;
+  int32_t total_len;
+};
+
+}  // namespace detail
+
+/// Parsed form of an eligible `%lit1%lit2%...%litN%` pattern.
+class like_multiliteral_pattern {
+ public:
+  explicit like_multiliteral_pattern(std::vector<std::string> literals);
```
Jedi18: Since this constructor is public, callers can skip the pattern classifier and create unsupported patterns such as `{"a_b"}.` The kernel would treat `_` as a normal character instead of a wildcard. Can we make this constructor private or apply the classifier’s checks here?

src/expression_evaluator/specializations/function.cpp
kevkrist: Yeah good call. Done.

src/include/expression_evaluator/like_multiliteral.hpp
kevkrist: Fixed

Jedi18 (approved):

PR #1615, opened 2026-08-19
feat: performance.py feature-parity with benchmark_and_validate.sh

joosthooz:
## Description
Added a couple of features from the legacy benchmark_and_validate.sh to performance_test.py, so that we can move away from that old bash script. Some of the features are needed for nvidia-internal nightly benchmarks, so this is needed to migrate those to performance.py:
- --pin-after-iteration N in run_grouped/run_sequential/run_isolated, so the first N iterations of a query (or round-robin pass) run unpinned before pinning kicks in for the rest — matches benchmark_and_validate.sh's --pin-after-iteration semantics.
- validate() now returns a structured {status, detail} per query (success/validation/error) instead of a bare bool, so a caller can tell a genuine mismatch apart from a missing/unparseable result file.
- New --duckdb-results <path> mode: lets a caller validate a Sirius run against a separately-captured DuckDB reference. This can be used by local developers to only have to run the (slow) duckdb validation step once and store those results somewhere, and only re-run if duckdb/the dataset/the queries changed.

## Checklist
- [x] Read CONTRIBUTING.md

### DIFF
--- a/test/tpch_performance/performance_test.py
+++ b/test/tpch_performance/performance_test.py
@@ -101,6 +101,7 @@ def setup_benchmark_dir(
     name=None,
     nsys_profile=False,
     data_source="parquet",
+    duckdb_results_source=None,
 ):
     """Create the benchmark output directory and return its paths.
 
@@ -114,13 +115,15 @@ def setup_benchmark_dir(
           <engine>/q<N>/result.txt  (one repr(row) per line)
           sirius/q<N>/sirius.log    (post-run split of combined log)
 
-    If `name` is provided, the benchmark dir is `<output_root>/<name>` (no
-    timestamp); otherwise the default `tpch_<ts>_<mode>_<engine>_iter<N>` is used.
+    If `name` is provided, the benchmark dir is `<output_root>/tpch_<ts>_<name>`
+    (timestamp kept, mode/engine/iter dropped since `name` already labels the
+    run); otherwise the default `tpch_<ts>_<mode>_<engine>_iter<N>` is used.
     """
     ts = datetime.now().strftime("%Y%m%d_%H%M%S")
-    benchmark_name = f"tpch_{ts}_{mode}_{engine}_iter{iterations}"
     if name:
-        benchmark_name = f"{benchmark_name}_{name}"
+        benchmark_name = f"tpch_{ts}_{name}"
+    else:
+        benchmark_name = f"tpch_{ts}_{mode}_{engine}_iter{iterations}"
     benchmark_dir = os.path.join(output_root, benchmark_name)
     csv_dir = os.path.join(benchmark_dir, "csv")
     log_dir = os.path.join(benchmark_dir, "log_dir")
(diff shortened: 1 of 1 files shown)


test/tpch_performance/performance_test.py
```diff
@@ -1116,9 +1172,50 @@ def parse_args():
             "(default: 90). Ignored in the other modes."
         ),
     )
+    p.add_argument(
+        "--duckdb-results",
```
9prady9, line 1178: Nit: the PR description still advertises a standalone `--validate-dir` mode, while this implements `--duckdb-results` and reruns Sirius. Could we align the description with the implemented behavior?

test/tpch_performance/performance_test.py
```diff
@@ -1202,6 +1312,10 @@ def main():
     )
     os.environ["SIRIUS_LOG_DIR"] = log_dir
 
+    if duckdb_results_dir:
+        log(f"Reusing DuckDB reference results from {duckdb_results_dir}")
+        shutil.copytree(duckdb_results_dir, os.path.join(benchmark_dir, "duckdb"))
```
9prady9: Could we persist the reused DuckDB reference identity in `metadata.json` (for example, its resolved source path and source-run metadata/commit)? The copied `duckdb/` subtree excludes the source run’s parent metadata, so the resulting artifact is not self-contained.

test/tpch_performance/performance_test.py
```diff
@@ -114,13 +114,15 @@ def setup_benchmark_dir(
...
-    timestamp); otherwise the default `tpch_<ts>_<mode>_<engine>_iter<N>` is used.
+    If `name` is provided, the benchmark dir is `<output_root>/tpch_<ts>_<name>`
+    (timestamp kept, mode/engine/iter dropped since `name` already labels the
+    run); otherwise the default `tpch_<ts>_<mode>_<engine>_iter<N>` is used.
     """
     ts = datetime.now().strftime("%Y%m%d_%H%M%S")
-    benchmark_name = f"tpch_{ts}_{mode}_{engine}_iter{iterations}"
     if name:
-        benchmark_name = f"{benchmark_name}_{name}"
+        benchmark_name = f"tpch_{ts}_{name}"
```
9prady9, line 124: Nit: this remains timestamped, so the CLI help saying same-name reruns overwrite outputs is inaccurate. Could we update the help/docs to match `tpch_<ts>_<name>`?

9prady9 (commented):
Few minor nits; otherwise this looks good.

test/tpch_performance/performance_test.py
joosthooz: I see, yes this has been updated

test/tpch_performance/performance_test.py
joosthooz: Thanks, yes, fixing that

test/tpch_performance/performance_test.py
joosthooz: I'm changing the feature to not copy the reference anymore, that is totally unnecessary. So now there is just the original metadata and this is no longer a problem. Thanks for helping simplify this.

9prady9 (approved):

PR #1622, opened 2026-08-20
fix: remove tpch_performance's separate pixi env

joosthooz:
## Description
test/tpch_performance/ had its own pixi.toml file so it could use cudf nightly for its parquet rewriting. But that is no longer used so there is no reason for it anymore. What's more, it has a separate version pin for duckdb which causes the script to fail to load the sirius extention when that doesn't match the project root env's duckdb pin. So I'm removing it here.

Also adds python=3.12.* and pyarrow>=14 to dev-libs. Drops fastavro: unused since the Iceberg tooling that needed it (wrap_parquet_as_iceberg.py, run_iceberg_benchmark.sh) was deleted in #1147.

python, duckdb, and rust were already present in the root default env, but python was pulled in transitively via pre-commit (ugly, now fixed). The only new packages are pyarrow's arrow-cpp stack (~39 MB against a 4.8 GB env).

## Checklist
- [x] Read CONTRIBUTING.md


9prady9 (approved):

pixi.toml
```diff
@@ -55,6 +55,8 @@ liburing = "*"
 openssl = "*"
 libcurl = "*"
 pnpm = "*"
+python = "3.*"
```
joosthooz, line 58: This wide pin allows duckdb-python to set the python version (it has a narrow python pin so we would otherwise have to update ours every time we update duckdb and they changed it)

PR #1624, opened 2026-08-20
fix(pipeline): signal query completion from the pipeline finish transition

### LINKED ISSUE #1486: bug(pipeline): query-completion race — mark_completed lost between task epilogue and finish cascade hangs the query forever
**Describe the bug**

A query can hang forever, silently, when its last task finishes an instant before the upstream
pipelines' finished-flags flip. Observed once in a 2528-case suite run on a single-GPU GB10:
`nation LEFT JOIN customer` (planned as HASH_JOIN type RIGHT, 6 pipelines) printed its pipeline
DAG and one telemetry line, then nothing — 18/20 threads in `futex_wait`, 0% GPU, for 42+ minutes
until killed. The same test passes standalone and in repeated randomized re-runs, consistent with
a microsecond-scale ordering race.

**Root cause**

`mark_completed()` — the only fulfiller of the future `sirius_engine::execute()` blocks on — is
called from exactly one place: the task epilogue in `gpu_pipeline_executor.cpp` (~447–469 on
`dev` @ 412a2158), gated on "this task's own pipeline sinks into RESULT_COLLECTOR **and**
`is_pipeline_finished()` is true at this instant".

`pipeline_finished` flips only in `sirius_pipeline::update_pipeline_status()`, which requires the
upstream finish signals (source exhausted, ports empty) in addition to balanced task counters. If
the last terminal-pipeline task's epilogue runs before the upstream CONCAT's finish signal
propagates, the gate reads false. The signal then arrives via the parent cascade
(`parent->update_pipeline_status(false)`) on another thread, flips the terminal pipeline to
finished — and dead-ends at `notify_downstream_pipelines()`'s deliberate RESULT_COLLECTOR
early-return (`sirius_pipeline.cpp` ~348–350, added to avoid racing engine teardown). Nothing
re-checks completion; `future.get()` parks forever with no log output.

Notes:
- `is_pipeline_finished()` already carries a `// todo (amin): potential race condition` comment.
- The task_creator manager's visited-pipelines re-evaluation (added for the RIGHT-join discard
  sweep) is a further non-epilogue driver of the same finish transition.
- Tiny single-batch queries maximize exposure: every pipeline finishes inside the same
  microsecond window.

**Expected behavior**

The terminal pipeline's finished-transition must fulfill query completion regardless of which
thread drives it — either the RESULT_COLLECTOR branch of `notify_downstream_pipelines` invokes
the completion handler, or the completion check moves under the pipeline's `_status_mutex` so
the epilogue-vs-cascade ordering cannot lose it.

**Discriminator if captured live**

Engine thread in `future.get()` with the terminal pipeline's `pipeline_finished == true` and
`tasks_created == tasks_completed` confirms this mechanism.

**Related defects found during the same audit** (candidates for follow-up issues; all
code-confirmed on `dev` @ 412a2158):

1. `terminate_query()` permanently bricks the scheduler: it `stop()`s the task queue and request
   channel, nothing restarts them, and `multi_index_priority_queue::push()` then **silently
   drops** every task of every later query (`multi_index_priority_queue.hpp` ~183); the wake-send
   failure is also swallowed.
2. `task_creator::stop()` self-deadlocks when called from its own bounded-pool worker (the
   creator lambda's catch block does exactly this): `stop()` → `wait_all()` waits for the
   caller's own still-active slot (`task_creator.cpp` ~556).
3. The executor's task-exception path orders `_task_creator->stop()` **before**
   `report_error(...)` (`gpu_pipeline_executor.cpp` ~424) — a `stop()` that blocks on a parked
   creator lambda converts a reportable error into a silent hang.
4. The partial-reservation catch (`gpu_pipeline_executor.cpp` ~234–239) `break`s out of the
   device's dispatch loop without `report_error` when `request_downgrade(...).get()` throws
   (e.g. broken promise landing in `downgrade_executor::drain()`'s per-QueryEnd interrupt
   window).
5. `management_eventloop` evaluates `*_ready_devices.begin()` unconditionally while
   `_ready_devices` can be empty (`task_scheduler.cpp` ~351) — UB, currently masked because the
   hint parameter is unused.
6. `notification_channel::notify()` collapses concurrent posts into a single boolean that
   `wait()` consumes via `std::exchange` — with ≥2 reservation waiters one strands
   (`cucascade/src/memory/notification_channel.cpp` ~76). Relatedly, plain `deallocate()`
   releases capacity without any waiter wakeup (`reservation_aware_resource_adaptor.cpp` ~513).
7. The downgrade monitor's `_monitor_request_enqueued` flag is set before a fire-and-forget push
   whose failure or cancellation leaves it latched true forever, permanently disabling
   monitor-driven downgrades (`downgrade_executor.cpp` ~473).
8. `memory_reservation_manager::request_reservation`'s slow path waits on `_wait_cv`, which no
   code path ever notifies (`memory_reservation_manager.cpp` ~163) — any thread reaching it
   parks forever (reachable from the insert-delta / mvcc-mask jobs).

_Written by Claude Code_

Jedi18:
## Summary

Move query completion signaling into the terminal pipeline's finish transition so completion no longer depends on the GPU task epilogue observing `pipeline_finished` at the right instant.

Fixes #1486.

## Problem

A query could finish all work but leave `sirius_engine::execute()` blocked forever on its completion future. `pipeline_finished` can be set by GPU workers, task-creator workers, the creator manager, streaming producers, or a parent-pipeline cascade, but only the GPU task epilogue called `completion_handler::mark_completed()`.

If another thread completed the terminal pipeline—or the epilogue sampled the flag just before the transition—the query finished without satisfying the future.

## Fix

- Install the query's completion handler on query-terminal pipelines during `prepare_for_query()`.
- Signal completion directly from `update_pipeline_status()` after releasing `_status_mutex`.
- Hold the handler weakly in pipelines and replace its strong owner per query, preventing retired pipelines from signaling later queries.
- Keep the epilogue signal as a CAS-safe fallback.

## Follow-up: completion quiescence

This change also makes streaming end-of-stream callbacks a possible signaling thread. The existing success path quiesces task-creator and GPU executor work before teardown, but it does not yet establish the same contract for externally driven producer callbacks.

The follow-up branch `fix/completion-quiescence-protocol` will add that quiescence protocol so every completion-capable context is retired before operator and stream state can be destroyed.

## Validation

A controlled test forced the old epilogue gate to miss once on an SF10 grouped aggregation:

| Build | Forced miss | Unmodified |
|---|---|---|
| `dev` | Timed out after 100 s | Passed |
| This PR | Passed in ~2 s | Passed |

Additional coverage:

- Full `sirius_unittest`: 2607 passed, 1 unrelated skip, 0 failed.
- TPC-H transparent path: 49 passed.
- Five new bounded completion tests covering non-epilogue signaling, streaming sinks, non-terminal pipelines, retired handlers, and duplicate/concurrent signals.

The completion contract is documented in `docs/super-sirius/pipeline-execution.md`, with pointers from the execution-flow and architecture overviews.

### DIFF
--- a/src/pipeline/task_scheduler.cpp
+++ b/src/pipeline/task_scheduler.cpp
@@ -224,7 +224,12 @@ std::future<void> task_scheduler::start_query()
 
 void task_scheduler::terminate_query(std::exception_ptr error)
 {
-  _completion_handler->report_error(std::move(error));
+  std::shared_ptr<completion_handler> completion;
+  {
+    std::scoped_lock lock(_query_mutex);
+    completion = _completion_handler;
+  }
+  if (completion) { completion->report_error(std::move(error)); }
   stop();
 }
 
--- a/src/pipeline/task_scheduler.cpp
+++ b/src/pipeline/task_scheduler.cpp
@@ -189,11 +189,20 @@ void task_scheduler::prepare_for_query(duckdb::shared_ptr<planner::query> query)
   std::lock_guard lock(_query_mutex);
   _query = std::move(query);
 
-  _completion_handler = std::make_unique<completion_handler>();
+  _completion_handler = std::make_shared<completion_handler>();
 
-  // Set completion handler on all executors
+  // Executors keep raw references owned by the scheduler.
   for (auto& [device_id, gpu_exec] : _gpu_executors) {
     gpu_exec->set_completion_handler(_completion_handler.get());
+  }
+
+  // Terminal pipelines keep weak references so replacing the handler retires the previous query.
+  if (_query) {
+    for (auto& pipeline : _query->get_pipelines()) {
+      if (pipeline && pipeline->is_query_terminal()) {
+        pipeline->set_completion_handler(_completion_handler);
+      }
+    }
   }
 }
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -807,6 +807,7 @@ set(TEST_SOURCES
     test/cpp/planner/test_query_index.cpp
     test/cpp/planner/test_tier_narrowing_policy.cpp
     test/cpp/pipeline/test_batch_lock_utils.cpp
+    test/cpp/pipeline/test_completion_signal.cpp
     test/cpp/pipeline/test_gpu_pipeline_executor.cpp
     test/cpp/pipeline/test_oom_reschedule.cpp
     test/cpp/pipeline/test_pipeline_memory_history.cpp
--- a/docs/super-sirius/architecture-overview.md
+++ b/docs/super-sirius/architecture-overview.md
@@ -127,13 +127,13 @@ A query through Super Sirius follows these steps:
    - Converts each TABLE_SCAN source into a unified GPU scan source with a per-table `gpu_ingestible`
    - Injects PARTITION, CONCAT, MERGE operators at pipeline boundaries
    - Wires data repositories between pipelines with barrier types
-4. **Query Preparation** — `task_scheduler::prepare_for_query()` drains leftover state and queues initial scan operators; `sirius_scan_manager::prepare_for_query()` builds each scan's split provider, installs its split connector, and matches any pinned-cache entries
-5. **Query Start** — `task_scheduler::start_query()` creates a `completion_handler`, distributes it to all sub-executors, and schedules initial work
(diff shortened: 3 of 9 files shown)


src/pipeline/task_scheduler.cpp
mbrobbel: Should we serialize this use too?

mbrobbel (approved):

src/pipeline/task_scheduler.cpp
Jedi18: Thanks yes, just the read of the `_completion_handler`, the handler itself is thread-safe.

mbrobbel (approved):

Issue #17, opened 2025-05-13
Simple queries with only count(*) or no output failed (expression does not support "No match found")


Yifei-yang7:
Query and output:

```
D call gpu_processing("select count(*) from nation");
┌───────────────────────────┐
│         PROJECTION        │
│    ────────────────────   │
│        Expressions:       │
│        count_star()       │
│                           │
│          ~1 Rows          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         AGGREGATE         │
│    ────────────────────   │
│        Expressions:       │
│        count_star()       │
│                           │
│          ~1 Rows          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│          SEQ_SCAN         │
│    ────────────────────   │
│       Table: NATION       │
│   Type: Sequential Scan   │
│                           │
│          ~25 Rows         │
└───────────────────────────┘

Error in GPUGeneratePhysicalPlan: Attempted to access index -1 within vector of size 4
=============================================
Error in GPUExecuteQuery, fallback to DuckDB
=============================================
Execute query time: 0.78 ms
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│      25      │
└──────────────┘
```
Query with not only `count(*)` can work, like `select count(*), min(n_nationkey) from nation`.

```
D call gpu_processing("select * from nation where n_nationkey = -1");
┌───────────────────────────┐
│        EMPTY_RESULT       │
│    ────────────────────   │
└───────────────────────────┘

Allocation table size 7
Setting pipeline source RESULT_COLLECTOR
Setting pipeline sink RESULT_COLLECTOR
Setting pipeline source EMPTY_RESULT
Total meta pipelines 1
total_pipelines = 2
Total meta pipelines 1



pipeline source type EMPTY_RESULT
Reading data from empty result
pipeline sink type RESULT_COLLECTOR
Segmentation fault
```
The one above may because the duckdb optimizer already checks zonemap so knows there will be no result. The one below actually needs to evaluate the predicate, and we throw "No match found".
```
D call gpu_processing("select n_nationkey, n_name from nation where n_name = 'PERV'");
┌───────────────────────────┐
│         PROJECTION        │
│    ────────────────────   │
│        Expressions:       │
│        N_NATIONKEY        │
│           N_NAME          │
│                           │
│          ~2 Rows          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│          SEQ_SCAN         │
[...]

Issue #16, opened 2025-05-13
Use a more standard way for logging


Yifei-yang7:
So far we directly print logs into the console, and ideally we should use some standard logger (e.g., spdlog and glog, glog is used by doris but spdlog is more lightweight).

PR #1626, opened 2026-08-20
fix(scan): lifetime issue in the prefetcher

joosthooz:
## Description
Borrow the prefetch worker's CUDA stream from the GPU space pool

The prefetch worker owned its stream, but the batches it converts are dealloc-bound to the stream the conversion ran on and outlive the worker: the loop exits as soon as every connector closes, while converted batches sit in connector queues until a pipeline task consumes them. Destroying the stream there left live buffers naming it, so the eventual cudaFreeAsync ran against a destroyed stream and segfaulted inside the driver. Borrow from the memory space's pool instead, which outlives every batch.

Some more agentic analysis:
A SIGSEGV inside cuMemFreeAsync when scanning a host-pinned, Simpatico-compressed table. Introduced by #1566 (50ad02fd,
  "fix(scan): await async reads before freeing their source memory").

  That PR replaced the prefetcher's member-owned stream pool with a stream owned by the worker function itself:

  -  std::unique_ptr<cucascade::memory::exclusive_stream_pool> _stream_pool;   // member
  ...
  -  auto stream = _stream_pool->acquire_stream();
  +  rmm::cuda_stream stream{rmm::cuda_stream::flags::non_blocking};           // function-local, OWNS it

  The reasoning is stated in the header comment #1566 added:

  ▎ Each worker's private CUDA stream carries no copy traffic (the converter allocates and copies on a pool stream it 
  ▎ acquires internally); it exists only as a stable per-worker key for attaching the admission reservation to the 
  ▎ allocation tracker.

  That holds for cucascade's uncompressed converter — convert_host_fast_to_gpu calls target_memory_space->acquire_stream()
  and binds the resulting buffers to that pool stream, ignoring the caller's. It does not hold for Sirius's compressed
  converter. decompress_host_to_gpu → reconstruct_and_decompress_to_gpu uses the caller's stream throughout, ending in:

  for (auto& c : cols) c = rebind_column_stream(std::move(c), stream);
  return std::make_unique<cucascade::gpu_table_representation>(std::move(decompressed), *space, stream);

  So with pin_table_compression = true, every decoded column's deallocation stream is the prefetch worker's own stream.

  Those batches outlive the worker. sweep() hands them to connector queues, where they wait for a pipeline task, while the
  worker exits as soon as the connectors close:

  if (all_closed) { break; }     // memory_prefetcher.cpp:111

  Returning from worker_loop() runs ~rmm::cuda_stream → cudaStreamDestroy, leaving queued buffers naming a destroyed
  stream. When a task later drops the batch, device_buffer::deallocate_async() calls cudaFreeAsync(ptr, <destroyed 
  stream>) and the driver dereferences a freed internal object:

  #0  0x72e564 in libcuda.so.1        ldrb w1, [x0, #112]   x0 = 0xffc7408b7503  (unaligned garbage)
  #4  cuMemFreeAsync
  #7  rmm::mr::detail::cuda_async_memory_resource_impl::deallocate
  #10 rmm::device_buffer::~device_buffer()
  #11 cucascade::gpu_table_representation::~gpu_table_representation()
  #13 cucascade::data_batch::~data_batch()
  #16 sirius::op::scan::scan_operator_input::~scan_operator_input()
  #17 sirius::pipeline::gpu_pipeline_task::compute_task(...)

  Why it stayed hidden: memory_prefetcher.enable defaults to false (config.hpp:55), and sweep() skips batches already
  resident (get_current_tier() == Tier::GPU), so GPU-tier pinning never converts anything. Only host-tier + compression +
  prefetcher-on hits it — the combination in bench/sf1000-repro/sirius-sf1000.yaml.

## Checklist
- [x] Read CONTRIBUTING.md

### DIFF
--- a/src/include/scan_manager/memory_prefetcher.hpp
+++ b/src/include/scan_manager/memory_prefetcher.hpp
@@ -102,7 +102,7 @@ class memory_prefetcher {
   }
 
  private:
-  void worker_loop();
+  void worker_loop(std::size_t worker_index);
 
   /// Attempt one sweep over all connectors; returns the number of batches converted.
   std::size_t sweep(rmm::cuda_stream_view stream);
@@ -115,6 +115,10 @@ class memory_prefetcher {
   /// short scan-bound queries). Quiet connectors allow full parallelism.
   std::unique_ptr<std::atomic<bool>[]> _drain_claims;
   cucascade::memory::memory_space* _gpu_space;
+
+  /// One stream per worker, borrowed (NOT owned): converted batches are
+  /// dealloc-bound to it and outlive the worker that made them.
+  std::vector<rmm::cuda_stream_view> _worker_streams;
 
   std::atomic<bool> _running{true};
   std::atomic<std::size_t> _batches_prefetched{0};
--- a/src/scan_manager/memory_prefetcher.cpp
+++ b/src/scan_manager/memory_prefetcher.cpp
@@ -21,7 +21,6 @@
 #include "log/logging.hpp"
 
 #include <rmm/cuda_device.hpp>
-#include <rmm/cuda_stream.hpp>
 #include <rmm/error.hpp>
 
 #include <absl/cleanup/cleanup.h>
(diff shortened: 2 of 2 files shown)


9prady9 (approved):
good catch

PR #1634, opened 2026-08-21
fix(result): convert parquet TIMESTAMP_MILLIS to DuckDB TIMESTAMP microseconds

patdevinwilson:
## Description

GPU parquet scans store `TIMESTAMP_MILLIS` as cuDF `TIMESTAMP_MILLISECONDS`. DuckDB `TIMESTAMP` is microseconds. The result chunk reader only compared physical `INT64` storage, so it memcpy'd millis into a microsecond vector and timestamps appeared ~1000× too close to the Unix epoch (e.g. `1970-01-11` instead of `1997-09-05`).

This maps cuDF timestamp types (and `BOOL8`) in `cudf_type_to_duckdb` and casts when **logical** types differ, in both `get_next_chunk` and nested `read_into`.

Verified with SpatialBench SF10 `trip.t_pickuptime` on GPU vs CPU (`1997-09-05 03:53:09` for `t_tripkey = 37529225`).

## Checklist
- [x] Read CONTRIBUTING.md

## Test plan
- [ ] GPU parquet scan of a `TIMESTAMP_MILLIS` column matches DuckDB CPU values (not epoch-offset)
- [ ] Existing `test_host_table_chunk_reader` still passes

## References
No existing issue. Related but distinct: #146 (result collector widening), #1107 (StarRocks timestamp encode).

### DIFF
--- a/src/op/result/host_table_chunk_reader.cpp
+++ b/src/op/result/host_table_chunk_reader.cpp
@@ -423,8 +423,8 @@ host_table_chunk_reader::host_table_chunk_reader(
   }
 }
 
-/// Map a cudf data_type to the DuckDB LogicalType with the same physical storage size.
-/// Used to create temp vectors for type-widening casts.
+/// Map a cudf data_type to the DuckDB LogicalType with the same physical storage.
+/// Used to create temp vectors for type-widening / unit-converting casts.
 static duckdb::LogicalType cudf_type_to_duckdb(cudf::data_type type)
 {
   switch (type.id()) {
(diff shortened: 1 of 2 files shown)


Jedi18 (approved):
Thanks for the fix!

felipeblazing:
This is great we should probably make  a follow up issue to create some tests to make sure this does not degrade.

mbrobbel (approved):

bwyogatama:
@patdevinwilson thanks for the contributions, just need to pass that lint check and this should be good to merge

patdevinwilson:
Thanks @bwyogatama — clang-format on `test_host_table_chunk_reader.cpp` is applied and pushed. Lint should be green on this run.

felipeblazing (approved):

PR #1650, opened 2026-08-25
fix(build): include variant before cuDF AST test header in one more file

Jedi18:
## Summary

Missed a spot in the last PR

### DIFF
--- a/test/cpp/scan/test_dynamic_filter_merge.cpp
+++ b/test/cpp/scan/test_dynamic_filter_merge.cpp
@@ -21,8 +21,12 @@
  *        column indices through scan_plan and skipping hive-partition columns.
  */
 
+// libcudf's AST header uses std::variant without including <variant>.
+// clang-format off
+#include <variant>
 #include <cudf/aggregation.hpp>
 #include <cudf/ast/expressions.hpp>
+// clang-format on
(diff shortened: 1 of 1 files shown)


ran-yuan-rui (approved):

Jedi18:
There are 15 direct includes of `cudf/ast/expressions.hpp`. This PR fixes the only remaining failure observed in the ARM64 nightly build, but several other sites currently work only because `<variant>` is included transitively.

Would we prefer keeping this targeted fix, or introducing a small compatibility header that includes `<variant>` before the cuDF AST header and using it at all direct include sites? The latter guarantees project-wide coverage until the upstream libcudf fix is available.

ran-yuan-rui:
> There are 15 direct includes of `cudf/ast/expressions.hpp`. This PR fixes the only remaining failure observed in the ARM64 nightly build, but several other sites currently work only because `<variant>` is included transitively.
> 
> Would we prefer keeping this targeted fix, or introducing a small compatibility header that includes `<variant>` before the cuDF AST header and using it at all direct include sites? The latter guarantees project-wide coverage until the upstream libcudf fix is available.

My take is either way is a "workaround", so let's get this PR in and wait for libcudf fix.

PR #1653, opened 2026-08-25
fix(exec): preserve scheduler during context shutdown

johanpel:
## Summary

Prevent shutdown-time use-after-free by keeping the task scheduler alive until all asynchronous borrowers have stopped.

Previously, teardown destroyed `task_scheduler_` before stopping `task_creator_` and the downgrade executors:

```cpp
task_scheduler_->stop();
task_scheduler_.reset();
task_creator_->stop_thread_pool();
executor->stop();
```

Their worker threads retain non-owning pointers to the scheduler or its task queue. A concurrent worker could therefore dereference freed memory. This rarely failed because it requires a narrow shutdown race, and freed memory can remain readable long enough to hide the invalid access.

The new order stops and joins every borrower before resetting the scheduler. Small comments document the ownership relationship without changing the existing APIs. Destructor teardown also catches exceptions to preserve its noexcept contract.

No additional synchronization or steady-state work is introduced.

## Validation

- Release build
- Scheduler tests: 9 assertions in 6 test cases
- Downgrade lifecycle tests: 53 assertions in 8 test cases
- clang-format
- git diff --check


src/sirius_context.cpp
```diff
@@ -222,7 +222,21 @@ SiriusContext::SiriusContext() = default;
...
-  if (is_initialized_) { terminate(); }
+  if (!is_initialized_) { return; }
+
+  try {
+    terminate();
+  } catch (const std::exception& e) {
+    try {
+      SIRIUS_LOG_ERROR("SiriusContext teardown failed: {}", e.what());
+    } catch (...) {
+    }
```
mbrobbel, line 233: Do we really need those `try`/`catch` blocks here for `SIRIUS_LOG_ERROR`?

johanpel: Technically yes, since SIRIUS_LOG_ERROR could throw

mbrobbel (approved):

PR #1661, opened 2026-08-25
Feature/frugal caching and dynamic io

aminaramoon:
## Checklist
- [ ] Read CONTRIBUTING.md

## References


ran-yuan-rui:
@aminaramoon Quick question first: is #1661 the branch you intend to land, with #1476 retired? The two look like the same design re-expressed on a newer dev base, and I left two review comments on #1476 that don't carry over. If #1661 is the target I'll stop watching #1476. Summaries below, with the originals linked for the detail.

1. Readahead admission is query-wide while IO capacity is backend-specific([original on #1476](https://github.com/sirius-db/sirius/pull/1476#issuecomment-5377833915)). `prepare_for_query` still takes the largest `n_max_concurrent_scans()` across the routed ioctxs and builds one `readahead_scan_manager` for the whole query (`sirius_scan_manager.cpp:683-711`), and scan.md now states that this budget is the only admission control on the read path. For a query that joins local parquet to S3, local scans inherit REST's depth and eager strategy, and a burst of cold S3 reads can borrow the shared gate into debt and hold back local readahead while the disk is idle. One more consequence I didn't have last time: `_routed_io_ctxs` is never cleared, so once an earlier query has created the REST ioctx, a later local-only query on the same scan manager also gets REST's budget and strategy (given readahead enabled, no explicit `readahead_strategy`, and REST publishing the widest budget). The suggested shape is unchanged: one manager per active ioctx, chosen per split after routing, with query-stage events broadcast to every active manager; happy to help with that if useful.

2. LIST ownership and transport routing([original on #1476](https://github.com/sirius-db/sirius/pull/1476#issuecomment-5377894616)). `rest_ioctx_for_list()` still hard-casts to the REST ioctx (`sirius_scan_manager.cpp:1027/:1042/:1092`), so any backend that also claims `s3://` gets REST's LIST endpoint, credentials, admission and failure state without saying so. With `backend: kvikio` that pairing is implicit today; with the `s3_transport` enum on dev (`object_store_config.hpp:40`) it becomes reachable to configure `backend: kvikio` together with `s3_transport: rdma`, and nothing rejects it (`sirius_config.cpp` only forces `sirius` on multi-GPU, `:872`). Two small asks: make the lister a composed member the backend hands over (cuCascade main already has `rest::object_store_lister` from #181), and reject the kvikio+rdma combination at config time until a transport actually implements it.

Neither point blocks the direction; both are about where transport-specific limits live so a future backend with registered memory and its own failure domain doesn't inherit REST's.

PR #1668, opened 2026-08-26
fix: arm64 nightly build fails

joosthooz:
## Description
This fixes a problem introduced in the late materialization work #1524: we need to include rmm/device_buffer.hpp instead of forward-declaring it. Also adds the <cstdint> that two tests rely on transitively for std::int64_t and friends.

## Checklist
- [x] Read CONTRIBUTING.md

## References

### DIFF
--- a/src/include/late_mat/column_origin.hpp
+++ b/src/include/late_mat/column_origin.hpp
@@ -43,6 +43,8 @@
 // range.start + local, at the consumer's edge —
 // codegen/selection/row_id_space.hpp is where they are narrowed back.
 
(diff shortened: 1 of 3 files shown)


9prady9 (approved):
Usually this is used to avoid full header inclusion when only pointers are involved. Did that change in this case ?

joosthooz:
> Usually this is used to avoid full header inclusion when only pointers are involved. Did that change in this case ?

No not at all, forward declaring is still the normal way to do it for pointers, but in this case it caused a namespace collision because of rmm's namespace structuring.

PR #1673, opened 2026-08-27
perf: eliminate a copy from hash_partition

joosthooz:
This eliminates a data copy during hash_partition by using views into a single large table. That reduces the peak memory usage, and the reservation required, but also keeps all data alive until the whole process is complete instead of releasing per partition. So this is mostly to reduce peak memory usage, there isn't a real performance gain measured in our sf1k workload.

## Checklist
- [ ] Read CONTRIBUTING.md

## References


wmalpica (approved):
NOTE: this PR uses the owning_table_view concept to achieve its goals, which will be evolving soon with @dhruv9vats work. 
@dhruv9vats should take a look and see if it will make his work much harder or not. But this is indeed a situation which leverages this owning view concept

joosthooz:
Converted to draft due to William's comment above

PR #1675, opened 2026-08-27
perf(dynamic-filter): carrier-native membership probes + mask-aware filtered decode (rebase of #1557)

felipeblazing:
Rebase of #1557 onto `dev`. Supersedes it — that branch sits on the closed
fused-scan-filter (#1391) + late-mat (#1409) stack, which was reimplemented and
merged as #1474 and #1524. This is a fresh branch off `dev` carrying only the
two probe-side mechanisms, ported onto those merged shapes. **#1557 should be
closed if this lands.**

## What was wrong

Compressed materialization stores a bounded `BIGINT`/`INTEGER` key in the
narrowest fitting signed carrier (`INT8`/`INT16`/`INT32`), and
`apply_stored_dtype` passes a narrow carrier through unchanged — it only re-tags
at equal width:

```cpp
if (!cudf::is_fixed_width(col->type()) || !cudf::is_fixed_width(stored) ||
    cudf::size_of(col->type()) != cudf::size_of(stored)) {
  return col;   // narrow carrier passes straight through
}
```

So a decode-time membership probe legitimately receives an `INT32` column
against an `INT64`-built set, and `compute_mask` answered `nullptr` on the type
mismatch. Since join keys are bounded integer IDs — the best narrowing
candidates in the schema — **the decode-time membership pushdown shipped in
#1474 is inert on exactly the columns it targets.** Depending on scan shape it
either fails the whole filtered decode or silently contributes no filtering at
all.

## The two changes

**1. Carrier-native probes.** All three membership filters (IN-list, small
IN-list, Bloom) now accept any signed-integer probe carrier and widen/narrow per
element in-kernel (new shared header `src/include/cuda/dynamic_filter_probe.cuh`),
so no materialized cast of the probe column is needed. A value the key domain
cannot represent is a definite non-member.

The coverage is *exactly* complete, not merely wider: `supports()` admits only
`INT32`/`INT64` build keys, those narrow only to `INT8`/`INT16`/`INT32`, and the
dispatch handles `INT8`..`INT64`. The width-based decline becomes unreachable.
Sentinel semantics are preserved — narrowing probes keep the set-sentinel
conservative keep, widening probes become exact.

**2. Mask-aware probing.** `sirius_mask_applicable` gains a prior-keep-mask
`compute_mask` overload (packed 1 bit/row); dead rows skip the set/Bloom lookup.
The wave-1 orchestrator runs membership probes sequentially on `s0` after
combining the other mask sources, handing the combined words to each probe as
its prior and folding each result back in. Membership-only chunks keep the
concurrent, prior-free path. The prior is a pruning hint only: ignoring it is
sound because the caller ANDs with that same mask.

## On the original's measured numbers

#1557 claimed SF1000 Power **+7.2%**, q19 **−52%**. Those **do not transfer
wholesale**, because #1524 changed what a declining probe costs — it split
`nullptr` from a contract violation and stands a decline down to the AND
identity. Its own comment names the narrow carrier as the case it absorbs.

| scan shape | today, on `dev` | vs. what the numbers measured |
|---|---|---|
| membership-only (join filter is the only pushdown) | `declined_members == k_total` → throw → full-width fallback | identical |
| mixed (static range/dict equality alongside) | stands down to all-ones; decode compacts on static sources, join filter contributes nothing | smaller loss, and silent |

Which shape dominates decides how much transfers. **Not measured on this stack.**

## Known caveats

- **Item 2 trades concurrency for pruning.** A probe that previously ran
  round-robin on its own pool stream now serializes on `s0` behind the combine,
  and only the set/Bloom lookup is pruned — the full-width key decode is still
  paid. At high static keep-rate that may not pay for the lost overlap.
  Unmeasured. **Item 2 is a reasonable candidate to split out** if you'd rather
  land the defect fix alone.
- **The multi-probe cascade is unreachable in a default build.**
  `SIRIUS_EXP_FUSED_SCAN_MAX_MEMBER` defaults to `1`, so the fold-back never
  executes; the new test arms it to 2. Worth deciding whether that default
  should move.
- All of this sits behind `SIRIUS_EXP_FUSED_SCAN_FILTER`, which is off unless
  set. This makes the experimental path work when enabled; it does not change a
  default build.

## Deliberately not carried over

`dev`'s reimplementation dropped two prior sources the original also primed
from — pair (col-vs-col) filters, and the T9 visibility keep-mask (still open as
#1556). Priors here are ranges and dictionary-answered equalities only. A
declining probe in the sequential cascade leaves the mask untouched, matching
`dev`'s all-ones stand-down rather than the original's throw.

## Related, not fixed here

The build side has the mirror defect with a worse failure mode:
`dynamic_filter_publisher.cpp` compares the runtime build column against the
plan's recorded logical type and **skips the key entirely** on a mismatch
(`keys_skipped_type_mismatch`), so a narrowed build key publishes no filter at
all. Probe-side tolerance cannot help there. Already instrumented via
`dynamic_filter_stats` — worth checking whether that counter fires before
deciding to fix it.

## Tests

- `test/cpp/operator/test_dynamic_filter_probe.cpp` (`[dynamic_filter][probe]`,
  8 cases): heterogeneous carriers, decimal/date/float refusal, sentinel
  conservation, prior-mask all-dead/all-live/patterned, null propagation.
- `test/cpp/scan/test_fused_membership_mask.cpp` (`[fused_scan_filter]`,
  4 cases): prior handed alongside a static range, prior-free membership-only,
  a two-probe cascade proving the fold-back, all-dead prior, INT32 carrier vs
  INT64 set.

Results on this branch: `[probe]` 8/8, `[fused_scan_filter]` 4/4,
`[dynamic_filter]` 239/239. Full suite run twice: **2959/2959 clean** on the
second; the first reported 1 failure whose identity was lost to a truncated
capture and which did not reproduce.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -806,6 +806,7 @@ set(TEST_SOURCES
     test/cpp/operator/test_physical_top_n.cpp
     test/cpp/operator/test_physical_ungrouped_aggregate.cpp
     test/cpp/operator/test_sirius_dynamic_filter.cpp
+    test/cpp/operator/test_dynamic_filter_probe.cpp
     test/cpp/operator/test_dynamic_filter_publication_claim.cpp
     test/cpp/operator/test_dynamic_filter_publisher.cpp
     test/cpp/operator/test_dynamic_filter_source_policy.cpp
@@ -856,6 +857,7 @@ set(TEST_SOURCES
     test/cpp/scan/test_duckdb_native_host_backed_decode.cpp
     test/cpp/scan/test_duckdb_native_walker.cpp
     test/cpp/scan/test_dynamic_filter_merge.cpp
+    test/cpp/scan/test_fused_membership_mask.cpp
     test/cpp/scan/test_gpu_decode_alp.cpp
     test/cpp/scan/test_gpu_decode_bitpacking.cpp
     test/cpp/scan/test_gpu_decode_rle.cpp
--- a/docs/super-sirius/dynamic-filters.md
+++ b/docs/super-sirius/dynamic-filters.md
@@ -96,6 +96,12 @@ If no probe-GPU L2 size is available, the hash IN-list is not selected; the publ
 - The domain-coverage gate skips the key before either filter is built when a proven-unique native build key covers at least `dynamic_filter_domain_coverage_threshold` of its known base-table domain.
 - The consumer keep-ratio gate disables ineffective post-decode filtering when a measured batch retains more than `dynamic_filter_keep_threshold` of its rows.
 
+### Probe-side evaluation
+
+`compute_mask` accepts any signed-integer probe carrier (`INT8`..`INT64`) regardless of the build-key width, widening or narrowing each value in-kernel; a probe value the key domain cannot represent is a definite non-member. This matters where the probe column is decoded rather than stored natively: a `BIGINT` join key can decode to a narrow `INT32` carrier, and the consumer probes that carrier directly instead of materializing a full-width cast first. Non-integer probe carriers (decimals, dates, floats) still decline with `nullptr` — a semantic mismatch, not a width mismatch.
+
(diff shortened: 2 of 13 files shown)


felipeblazing:
Closing — superseded. The intent is to update #1557 in place rather than open a parallel PR.

PR #1678, opened 2026-08-30
fix(pin_table): add compression parameter

### LINKED ISSUE #1358: Change the input pin compression config setting to a parameter for the pin command
Currently, enabling compression for pinned input tables is a parameter in the sirius config file. It might make more sense to make that a parameter to the pin command instead.

realAaronWu:
## Summary
- add a per-call `compression` named parameter to `CALL pin_table(...)`
- make explicit `compression=true/false` override the existing `pin_table_compression` setting
- keep `pin_table_compression` as the compatibility default when the argument is omitted
- update Super Sirius docs and add regression coverage for both override directions and the empty-plan-dir fallback

## Configuration/API surface
`CALL pin_table(..., compression => true/false)` is now the preferred way to request or suppress Simpatico compression for a specific pin. Existing `pin_table_compression` YAML/SET behavior remains as the default for calls that omit the new argument.

Backward compatibility is covered by the existing compression tests as well: many existing cases set `pin_table_compression = true` and then call `pin_table` without a `compression` argument. The full `[compression][pin_table]` suite passing below exercises that omitted-argument fallback alongside the new explicit-argument override test.

Closes #1358

## Validation
- `git diff --check`
- Remote GPU validation at `efda095a`, GPU `NVIDIA GeForce RTX 5060 Ti`, driver `595.84`:
  - `CUDAARCHS=120 CMAKE_BUILD_PARALLEL_LEVEL=2 pixi run make` — passed. The initial high-parallel build failed silently during CUDA compilation, and the testcontainers Go bridge needed an existing local Go module cache because external Go module downloads timed out on the node; after lowering parallelism and reusing the cache, the release build and `sirius_unittest` linked successfully.
  - `pixi run build/release/extension/sirius/test/cpp/sirius_unittest "[compression][pin_table]"` — passed, 1224 assertions / 29 test cases. This includes `pin_table compression - compression parameter overrides session setting` and `pin_table compression - empty plan dir pins uncompressed`.
  - `pixi run build/release/extension/sirius/test/cpp/sirius_unittest "[config]"` — passed, 688 assertions / 66 test cases.
- Additional remote GPU checks at `03d26110`, GPU `NVIDIA RTX PRO 4500 Blackwell`, driver `580.159.04`:
  - `CUDAARCHS=120 CMAKE_BUILD_PARALLEL_LEVEL=48 pixi run make` — passed.
  - `pixi run build/release/extension/sirius/test/cpp/sirius_unittest "[compression][plan_register]"` — passed, 16 assertions / 6 test cases.
  - `pixi run build/release/extension/sirius/test/cpp/sirius_unittest "Sirius configuration rejects invalid compression retention fractions"` — passed, 3 assertions / 1 test case.
  - `pixi run build/release/extension/sirius/test/cpp/sirius_unittest "Sirius configuration accepts intentional compression retention fraction boundaries"` — passed, 4 assertions / 1 test case.
  - SQL smoke with a single-GPU `scan_manager.use_sirius_datasource=false` config: created a parquet input and matching Simpatico plan files, then verified `compression=true` overrides `pin_table_compression=false` and `compression=false` overrides `pin_table_compression=true`. The command succeeded and Sirius logs showed compression only for `t_param_on`, not for `t_param_off`.
  - Pod limitation: full C++ `[compression][pin_table]` and broad `[config]` aborted there when test-generated/default YAMLs initialized Sirius' local `io_uring` datasource: `uring_reactor: ring init: Operation not permitted`. The pod reported `kernel.io_uring_disabled = 0`, so this appeared to be a container syscall/capability restriction. The successful remote GPU validation at `efda095a` above covers those previously blocked suites.
- Not run locally: `pixi run ...` build/tests/pre-commit, because this local machine is macOS ARM while the repo declares Linux CUDA pixi platforms.

### DIFF
--- a/src/sirius_extension.cpp
+++ b/src/sirius_extension.cpp
@@ -1176,6 +1176,11 @@ unique_ptr<FunctionData> SiriusExtension::PinTableBind(ClientContext& context,
     result->args.cols = std::move(cols);
   }
 
+  auto compression_it = input.named_parameters.find("compression");
+  if (compression_it != input.named_parameters.end() && !compression_it->second.IsNull()) {
+    result->args.compression = BooleanValue::Get(compression_it->second);
+  }
+
   // Resolve the source format: an explicit 'format' parameter, else inferred from
   // the path extension (.parquet -> parquet, .db/.duckdb -> duckdb).
   auto to_lower = [](std::string s) {
@@ -1364,15 +1369,17 @@ void SiriusExtension::PinTableFunction(ClientContext& context,
   // directory (if configured), then resolve it into a compression_pin_config. Both
   // the host and GPU pin paths compress with this when enabled.
   const auto& comp_cfg = sirius_ctx->get_config().get_compression_config();
-  if (comp_cfg.enable_pin_table_compression && comp_cfg.input_plan_dir.empty()) {
+  bool const compression_requested =
+    data.args.compression.value_or(comp_cfg.enable_pin_table_compression);
+  if (compression_requested && comp_cfg.input_plan_dir.empty()) {
     SIRIUS_LOG_WARN(
-      "[pin_table] '{}': pin_table_compression is enabled but "
+      "[pin_table] '{}': compression was requested but "
       "pin_table_input_compression_plan_dir is empty; pinning uncompressed",
       data.args.name);
   }
-  const bool comp_globally_enabled =
-    comp_cfg.enable_pin_table_compression && !comp_cfg.input_plan_dir.empty();
-  if (comp_globally_enabled) {
+  const bool compression_plan_lookup_enabled =
+    compression_requested && !comp_cfg.input_plan_dir.empty();
+  if (compression_plan_lookup_enabled) {
     namespace fs     = std::filesystem;
     const auto& name = data.args.name;
     if (!sirius::compression::plan_register::global().resolve_table_plan(name).has_value()) {
@@ -1397,7 +1404,7 @@ void SiriusExtension::PinTableFunction(ClientContext& context,
   }
 
   sirius::compression_pin_config pin_comp{};
-  if (comp_globally_enabled) {
+  if (compression_plan_lookup_enabled) {
     if (auto plan_dsl =
           sirius::compression::plan_register::global().resolve_table_plan(data.args.name);
         plan_dsl.has_value()) {
@@ -1428,7 +1435,7 @@ void SiriusExtension::PinTableFunction(ClientContext& context,
       }
     } else {
       SIRIUS_LOG_WARN(
-        "[pin_table] '{}': pin_table_compression is enabled but no plan file was found in '{}'; "
+        "[pin_table] '{}': compression was requested but no plan file was found in '{}'; "
         "pinning uncompressed",
         data.args.name,
         comp_cfg.input_plan_dir);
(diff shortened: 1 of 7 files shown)


realAaronWu:
Design question before review: I kept `pin_table_compression` as the default for calls that omit the new `compression` argument. Should it be deprecated with a removal target, or is a permanent fallback fine?

joosthooz:
Thanks for this! It was on my TODO list so this is great. Don't worry about the deprecation, I'm working on a branch that adds parameters for spill compression so we will think about it there. I think this approach is fine.

src/sirius_extension.cpp
```diff
@@ -1364,15 +1369,16 @@ void SiriusExtension::PinTableFunction(ClientContext& context,
...
     SIRIUS_LOG_WARN(
-      "[pin_table] '{}': pin_table_compression is enabled but "
+      "[pin_table] '{}': compression was requested but "
       "pin_table_input_compression_plan_dir is empty; pinning uncompressed",
       data.args.name);
   }
-  const bool comp_globally_enabled =
-    comp_cfg.enable_pin_table_compression && !comp_cfg.input_plan_dir.empty();
-  if (comp_globally_enabled) {
+  const bool compression_active = compression_requested && !comp_cfg.input_plan_dir.empty();
```
ran-yuan-rui, line 1380: Non-blocking API question: should an explicit `compression => true` still succeed when the plan directory is empty or no matching plan is available? This is documented and tested, but the call returns the same `Success = true` as a compressed pin; only the logs distinguish the fallback.

Would you prefer rejecting missing compression prerequisites for explicit requests, or keeping best-effort behavior and exposing the actual compression outcome through SQL? Calls that omit the argument can retain the existing setting-based behavior.

src/sirius_extension.cpp
```diff
@@ -3300,9 +3307,9 @@ void SiriusExtension::InitialGPUConfigs(DBConfig& config, const sirius::sirius_c
...
   config.AddExtensionOption("pin_table_compression",
-                            "Request Simpatico compression for pin_table chunks. Takes effect only "
-                            "when pin_table_input_compression_plan_dir is non-empty and contains a "
-                            "matching table plan",
+                            "Default for pin_table calls that omit the compression named "
+                            "parameter. Takes effect only when pin_table_input_compression_plan_dir "
+                            "is non-empty and contains a matching table plan",
                             LogicalType::BOOLEAN,
                             Value::BOOLEAN(compression_defaults.enable_pin_table_compression),
                             SetEnablePinTableCompression);
```
ran-yuan-rui: Could you apply clang-format's rewrapping of this `AddExtensionOption` call? It is the only change in the lint job's formatter diff. The other hooks passed or were skipped; this failure also keeps `check` red.

src/sirius_extension.cpp
joosthooz: Originally the plan files were a stopgap, until we would have plan explore functionality. Now we do, but the static plan files still are nice to have. Because we can explore offline, and make explicit choices about the codecs besides ratio and throughput. For example, we can keep some columns raw, or we can choose a codec that supports some form of pushdown/partial materialization. 
Here we can do 2 things: Either we add another parameter to the pin command, to explicitly point to a plan file, or we leave it like it is now. Both are fine for me.

joosthooz:
@realAaronWu thanks for submitting this, we just need to fix the formatting and a response to ran-yuan-rui's comments

src/sirius_extension.cpp
ran-yuan-rui: @realAaronWu I pushed the clang-format-only fix in `c17f288e`. It is a single-hunk rewrap of this call using the pinned clang-format 20.1.4, with no behavior change. The other hooks passed or were skipped in the previous run; the new CI run still needs to confirm the fix.

src/sirius_extension.cpp
ran-yuan-rui: Thanks @joosthooz. I'm fine with keeping the current behavior for this PR. I was asking whether SQL callers should be able to distinguish compression from fallback when using `compression => true`, rather than how the plan is selected. We can leave that to a separate discussion.

ran-yuan-rui (approved):

PR #1685, opened 2026-09-01
build(vss): use prebuilt static cuVS instead of compiling it from source

yayen-lin:
## Description

The `cuvs` vcpkg overlay port now downloads the prebuilt `libcuvs-static` and `libcuvs-headers` conda packages from the rapidsai channel instead of compiling libcuvs from source.

Building libcuvs from source in the distribution CI took roughly 50 to 75 minutes per matrix leg. It compiles the full cuVS kernel set across every GPU arch, even though Sirius only calls `brute_force` and `distance`, and it ran again on every cold vcpkg binary cache. The prebuilt `libcuvs-static` package already carries the same all-arch archive, so the port just downloads and installs it. Locally the cuvs port now finishes in about 7 seconds instead of about 52 minutes.

### What changed
- `vcpkg_ports/cuvs/portfile.cmake` fetches one `.conda` per (arch, cuda) variant (linux-64 and linux-aarch64, cuda 12 and 13), the same approach the nvcomp port uses. It extracts the archive, installs `libcuvs_static.a`, the headers, and the shipped target cmake files, patches their install-prefix walk for vcpkg's `share/` layout, and writes a small `cuvs-config.cmake`.

### Notes
- The prebuilt static library links the CUDA math libs (cublas, cusolver, cusparse, curand, nvJitLink, nvrtc) at link time, so those must be present in the CI docker image.

### Validation
- Local `vcpkg-release` build: the cuvs port installs in 6.9s, post-build validation passes, and Sirius links against `libcuvs_static.a` with no missing symbols.
- `[vss]` unit and integration tests pass against the prebuilt library.

## Checklist
- [x] Read CONTRIBUTING.md
- [x] Cover changes with new or existing tests (existing `[vss]` tests exercise the prebuilt library)
- [ ] Update human and agent documentation (no doc changes needed for this build-only change)

## References
Refs #1204 (added the from-source cuvs overlay port that this PR replaces)
Refs #1205

### DIFF
--- a/vcpkg_ports/cuvs/portfile.cmake
+++ b/vcpkg_ports/cuvs/portfile.cmake
@@ -1,119 +1,145 @@
 vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
 
-vcpkg_from_github(
-  OUT_SOURCE_PATH
-  SOURCE_PATH
-  REPO
-  rapidsai/cuvs
-  REF
-  v${VERSION}
-  SHA512
-  73cdf0e16e701063528c71ac48f0d3d9072dce11d1ab2ac768173c676cb8d704fcb2d3eb5b0ae8892b3b657eefc8bc03e3d83fb204befaa81cb9538328410cf1
-  HEAD_REF
-  main)
+# cuVS is consumed as a prebuilt static library from the RAPIDS conda channel
+# instead of being compiled from source. Building libcuvs from source in the
+# distribution CI took roughly 50 to 75 minutes per matrix leg, because it
+# compiles the full cuVS kernel set across every GPU arch, even though Sirius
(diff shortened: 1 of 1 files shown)


aocsa (approved):
LGTM, but I would ask @kevkrist  for a second opinion, as he is the CMake owner in this repo.

mbrobbel (commented):
I think we want our vcpkg setup to build from source when possible. We cache the artifacts so we only have to rebuild from scratch when we bump.

nvcomp is the exception because we can't built from source.

yayen-lin:
Thank you @aocsa and @mbrobbel, 
They are both very useful information. I will wait for any new update for now.

PR #1698, opened 2026-09-02
feat(pipeline): opt-in stall watchdog (SIRIUS_QUERY_WATCHDOG_SECS)

### LINKED ISSUE #1486: bug(pipeline): query-completion race — mark_completed lost between task epilogue and finish cascade hangs the query forever

aocsa:
## Description

**Motivation.** Hang-class regressions in multi-fragment queries need to fail in seconds with a named error, not hang CI to its timeout. The canonical shape: a hash-partitioned exchange gives most compute nodes zero keys for a given partition — on TPC-H Q1 the merge fragment's fan-out leaves all but one CN with an input stream that closes with zero batches. That stream closes between `build()` and `run()`, its end-of-stream hook finishes the pipeline before `prepare_for_query()` has installed the query's completion handler, and no task is ever created to carry the completion. When this class wedged, the multi-CN suite hung for the full 120 s harness timeout with nothing in the log naming the cause.

**What the investigation found.** The source-of-truth branch (`aocsa/feat/pin-table-cn`) fixed the empty-stream hang with a zero-task completion hunk in `task_creator::manager_loop()` (an exhausted-head re-check calling a new `task_scheduler::complete_query_if_finished()`). On current `dev` that hunk is redundant: #1624 already re-statuses the scheduled head's pipeline in the `visited_pipelines` loop once the query is live, and `update_pipeline_status()` now signals completion from the finish transition for a query-terminal pipeline and re-cascades to parents. The experiment on this box was direct — with `src/creator/task_creator.cpp` reverted to `origin/dev` and everything else from the source-of-truth commit kept, FRAG-11 passes (`All tests passed (10 assertions in 1 test case)`). So this PR does **not** carry the completion hunk or `complete_query_if_finished()`; it keeps the test that pins the behaviour and the watchdog that makes any regression of it loud.

**What changed** (one reviewable unit: the opt-in stall watchdog and the regression test it guards):

- `src/sirius_engine.cpp` — `query_progress_fingerprint()` (sum of every pipeline's `tasks_created + tasks_completed`, +1 per finished pipeline) and `wait_for_query_future()`, which replaces the bare `future.get()` in `execute()`. With `SIRIUS_QUERY_WATCHDOG_SECS` unset or `0` it takes an early `future.get()` return — the default path is byte-identical to today. Set, it polls the future every 250 ms and fails the query only after **zero** scheduling progress for that many seconds. The value is parsed strictly — digits only (no sign, no whitespace), `ERANGE` checked, and capped at `86400` (one day) — because `strtoull` alone accepts `-1` by wrapping it to `ULLONG_MAX`, which would narrow to `seconds(-1)` and fire on the first poll; the error message names exactly what is enforced.
- `src/include/pipeline/task_scheduler.hpp`, `src/pipeline/task_scheduler.cpp` — `fail_stalled_query(uint64_t)` reports a named `sirius query watchdog` error through the completion handler (first-call-wins: a query that completes concurrently keeps its real outcome) without stopping or draining — that belongs to `execute()`'s catch path, which observes the failed future. Called from the engine thread only; never from the task creator's manager thread.
- `test/cpp/exec/test_streaming_fragment.cpp` — a `watchdog_guard` RAII helper (sets `SIRIUS_QUERY_WATCHDOG_SECS`, restores the previous value; a non-empty `SIRIUS_TEST_WATCHDOG_SECS` in the environment overrides the test's own threshold so a slow CI GPU can raise every guarded test at once) and **FRAG-11**: a fragment reading one declared input stream whose only sender is closed without a batch before `run()`; under `watchdog_guard("20")` it must return with zero tasks created, zero tasks completed and zero rows. Its doc comment says what it is: a pin on behaviour `dev` already has since #1624, and the C++ mirror of the CN suite's `fragment_over_an_empty_input_stream_terminates`.
- `docs/super-sirius/configuration.md` — a `SIRIUS_QUERY_WATCHDOG_SECS` subsection: off by default; must sit well above any single kernel's runtime (a long kernel makes no counter progress) and below the client-side query timeout; the CN cluster env uses 60; the accepted range (`0..86400`, digits only) and the `SIRIUS_TEST_WATCHDOG_SECS` test knob are stated. `docs/super-sirius/task-creator.md` — two sentences where the manager loop is described: an input stream that closes before `run()` still finishes its pipeline through the finish transition (#1624), and FRAG-11 pins it.

**Why a watchdog, and why opt-in.** Every hang-class regression test in this stack and the next (FRAG-6..9 in the join-wedge layer, the CN suite's `under_watchdog` tests) needs a wedge to fail in seconds with a named error rather than hang CI to its timeout; that is what `SIRIUS_QUERY_WATCHDOG_SECS` provides. It is opt-in because it watches scheduling counters, not GPU activity: an in-flight long kernel is indistinguishable from a stall, so only deployments that know their per-operator runtimes should enable it. It does not cancel in-flight GPU work; it turns a silent hang into a loud failure.

Verified: GB200 aarch64 (CUDA 13.0, driver 580.105.08), d53b44b1 over dev 84ea4ab5, in a dedicated clone: `pixi run make` OK (381 steps, 0 errors); `CUDA_VISIBLE_DEVICES=3 pixi run ./build/release/extension/sirius/test/cpp/sirius_unittest "[streaming_fragment]"` → `All tests passed (85 assertions in 6 test cases)`; `CUDA_VISIBLE_DEVICES=3 pixi run ./build/release/extension/sirius/test/cpp/sirius_unittest "FRAG-11*"` → `All tests passed (10 assertions in 1 test case)`; `pixi run --locked rumdl check docs/super-sirius/configuration.md docs/super-sirius/task-creator.md` → `Success: No issues found in 2 files`; `SKIP=rumdl pre-commit run --files <the six changed files>` all hooks Passed with no rewrites; `git diff origin/dev | grep complete_query_if_finished` empty.

**Design overlap with open #1642.** @Jedi18's completion-quiescence work ledger (`fix/completion-quiescence-protocol`) rewrites the same `future.get()` try/catch in `execute()` this PR replaces with `wait_for_query_future()`, and the same `task_scheduler` completion region `fail_stalled_query()` lives in: it makes `wait_for_completion()` park the creator, close a ledger against new work and wait for a zero count before teardown. `fail_stalled_query()` reports through the completion handler first-call-wins and deliberately does not drain; if the ledger makes draining mandatory before any completion signal, the watchdog's failure path needs to be expressed in ledger terms (register the error, let the ledger reach zero, then release). This PR is opened as **Draft** to settle that with him first.

**Intentionally not handled.** A multi-source fragment where a *non-front* streaming source closes empty before `run()` and is never scheduled: FRAG-11 cannot reach that shape (its single stream is the scheduled head, so the `visited_pipelines` re-status covers it), and whether the finish-transition signal from #1624 covers a never-scheduled source is an open question raised for #1642 rather than answered with new code here. Hash-join slots whose build or probe side can never arrive (`never_buildable_slots`, the partition zero-byte strategy, FRAG-6..9) are the next layer. Input-stream cardinality (FRAG-10) is a separate PR. The CN/bench consumers of `SIRIUS_QUERY_WATCHDOG_SECS` (`experimental/starrocks` TUNABLES, cluster scripts) ship with the CN slices.

Layer 1 of the liveness stack; the next layer resolves hash-join slots that can never be built or probed, and its FRAG-6..9 run under the `watchdog_guard` introduced here.

## Checklist

## References
- Supersedes the watchdog part of #1644 (draft "Stream fragment execution", 29 commits — being closed in favour of this stack).
- Refs #1642 (open: completion quiescence protocol / work ledger — same region; design overlap noted above).
- Refs #1624 (merged: signal query completion from the pipeline finish transition; fixes #1486) — the transition-time signal that already completes the empty-stream case FRAG-11 pins.

### DIFF
--- a/docs/super-sirius/configuration.md
+++ b/docs/super-sirius/configuration.md
@@ -32,6 +32,16 @@ Set `SIRIUS_DISABLE=1` to prevent Super Sirius from initializing. This is **requ
 export SIRIUS_DISABLE=1
 ```
 
+### `SIRIUS_QUERY_WATCHDOG_SECS`
+
+Opt-in query watchdog, **off by default**. Unset or `0` leaves `sirius_engine::execute()` blocking on the query future exactly as before. Set to `N` and the engine thread instead polls the future every 250 ms and fails the query — through the scheduler's completion handler, with a named `sirius query watchdog` error the client sees — once **no pipeline has made any scheduling progress** (no task created or completed, no pipeline finished) for `N` seconds. A wedged query becomes a loud failure in `N` seconds instead of a silent hang; in-flight GPU work is not cancelled, and a query that completes concurrently keeps its real outcome (first call wins).
+
+The watchdog watches scheduling counters, not GPU activity, so a single long-running kernel looks like a stall to it. Pick a value well above the longest single kernel or operator runtime the deployment can see, and below the client-side query timeout so the client receives the engine's error rather than its own. The multi-CN cluster environment uses `60`. The value must be a plain non-negative integer no larger than `86400` (one day); anything else — a sign, whitespace, a non-digit, or a larger number — is rejected when the query starts. The C++ integration tests that run under the watchdog (`watchdog_guard` in `test/cpp/exec/test_streaming_fragment.cpp`) hard-wire a threshold sized for a fast GPU; export `SIRIUS_TEST_WATCHDOG_SECS` to raise it on a slow CI machine.
+
+```bash
+export SIRIUS_QUERY_WATCHDOG_SECS=60
+```
+
 ### Byte Suffixes
 
 Any integer config value that represents bytes supports human-readable suffixes:
--- a/docs/super-sirius/task-creator.md
+++ b/docs/super-sirius/task-creator.md
@@ -215,6 +215,8 @@ while running:
 
 The `mark_task_created()` call before data popping prevents a race condition where the pipeline could appear finished between data check and task creation.
 
+**Zero-task completion.** A pipeline whose input stream closes without ever carrying a batch finishes between `build()` and `run()` — before `prepare_for_query()` installs the query's completion handler — so no task ever runs; step 4's re-status of the scheduled head, together with `update_pipeline_status()` signalling completion for a finished query-terminal pipeline (#1624), still completes the query. `FRAG-11` in `test/cpp/exec/test_streaming_fragment.cpp` pins this.
+
 ### Look-ahead task creation
 
 **Files:** `src/include/creator/config.hpp`, `src/creator/task_creator.cpp`
--- a/src/include/pipeline/task_scheduler.hpp
+++ b/src/include/pipeline/task_scheduler.hpp
@@ -29,6 +29,7 @@
 #include <cucascade/memory/topology_discovery.hpp>
 
 #include <atomic>
+#include <cstdint>
 #include <future>
 #include <memory>
 #include <optional>
@@ -172,6 +173,19 @@ class task_scheduler {
    * @param error The error to report.
    */
   void terminate_query(std::exception_ptr error);
+
+  /**
+   * @brief Fail the running query as stalled.
+   *
+   * Reports an error through the completion handler (first-call-wins: a query that completes
+   * concurrently keeps its real outcome) WITHOUT stopping the scheduler or draining executors —
+   * draining after an error belongs to the engine's execute() catch path, which observes the
+   * failed future. Called from the engine thread by the opt-in query watchdog (see
+   * sirius_engine::execute); never from the task creator's manager thread.
+   *
+   * @param stalled_secs How long the watchdog observed no scheduling progress, for the message.
+   */
+  void fail_stalled_query(uint64_t stalled_secs);
 
   /**
    * @brief Drain all in-flight tasks after a query error.
--- a/src/pipeline/task_scheduler.cpp
+++ b/src/pipeline/task_scheduler.cpp
@@ -233,6 +233,19 @@ void task_scheduler::terminate_query(std::exception_ptr error)
   stop();
 }
 
+void task_scheduler::fail_stalled_query(uint64_t stalled_secs)
+{
+  std::scoped_lock lock(_query_mutex);
+  if (!_completion_handler) { return; }
+  SIRIUS_LOG_ERROR(
+    "task_scheduler: no scheduling progress for {}s (no task created or completed, no pipeline "
+    "finished); failing the query as stalled",
+    stalled_secs);
+  _completion_handler->report_error("sirius query watchdog: no scheduling progress for " +
+                                    std::to_string(stalled_secs) +
+                                    "s — failing the stalled query instead of wedging the engine");
+}
+
 void task_scheduler::drain_after_error()
 {
   SIRIUS_LOG_INFO("task_scheduler: draining after error");
--- a/src/sirius_engine.cpp
+++ b/src/sirius_engine.cpp
@@ -46,6 +46,12 @@
 #include <cucascade/memory/memory_space.hpp>
 
 #include <algorithm>
+#include <cctype>
+#include <cerrno>
+#include <chrono>
+#include <cstdint>
+#include <cstdlib>
+#include <cstring>
 #include <memory>
 #include <stdexcept>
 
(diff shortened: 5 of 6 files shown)


aocsa:
Closing as a duplicate of #1699. This layer stands alone today (the join-wedge follow-up lands later as its own PR), and a one-layer stack is published as a plain fork PR by our stacking rule; same commit d53b44b1, same body. The stacked/pipeline-stall-watchdog branch on sirius-db is removed.

PR #1719, opened 2026-09-03
fix(cuda): replace deprecated stream header

mbrobbel:
Replace deprecated `<cuda/stream_ref>` with `<cuda/stream>` to fix some include warnings.


ran-yuan-rui (approved):
LGTM.

PR #1725, opened 2026-09-07
perf(scan): project zero-column parquet scans to one narrow column

### LINKED ISSUE #1723: Zero-column Parquet scans are slower than expected
Sirius has a performance issue for zero-column Parquet scans such as `SELECT count(*)`.

Several comments in the codebase mention reading Parquet through cuDF with an empty column projection can hang. Sirius therefore [classifies empty and virtual-only column lists as unprojected](https://github.com/sirius-db/sirius/blob/404d1a0167982374ffcaae8e899651f16e50daf9/src/op/scan/scan_plan.cpp#L187-L210) and [clears projection for scans without physical data columns](https://github.com/sirius-db/sirius/blob/404d1a0167982374ffcaae8e899651f16e50daf9/src/op/scan/scan_plan.cpp#L305-L308).

The Parquet reader then [avoids calling set_column_names() with an empty list](https://github.com/sirius-db/sirius/blob/404d1a0167982374ffcaae8e899651f16e50daf9/src/op/scan/parquet_gpu_ingestible.cpp#L586-L595). Because the scan is unprojected, Sirius [accounts for every Parquet column](https://github.com/sirius-db/sirius/blob/404d1a0167982374ffcaae8e899651f16e50daf9/src/op/scan/parquet_gpu_ingestible.cpp#L928-L948), and the eventual [cudf::io::read_parquet call](https://github.com/sirius-db/sirius/blob/404d1a0167982374ffcaae8e899651f16e50daf9/src/op/scan/parquet_gpu_ingestible.cpp#L1085-L1089) reads and decodes them all.

With TPC-H SF100 `lineitem` as an example:

```
SELECT count(*) FROM lineitem;
-- ~ 4.5 seconds

SELECT count(*)
FROM lineitem
WHERE l_quantity > 0 AND l_quantity < 60;
-- ~ 0.1 seconds
```

Both queries return the same result because the filter has selectivity 1.0. However, the first query requires no physical columns and consequently reads all columns, while the second query projects only `l_quantity` for filter evaluation. This means extra I/O and GPU decoding for the otherwise simpler `count(*)` query.

ran-yuan-rui:
## Description

When a `count(*)` query reaches Sirius's parquet scan without requesting data columns, the scan currently reads and decodes every column just to obtain the row count. Queries that DuckDB already answers from metadata are unaffected.

This PR reads one **row-count carrier** instead: the narrowest fixed-width, non-partition column in the bind schema, ties resolved to the lowest schema position. The same choice applies to scans requesting only partition or virtual columns. cuDF cannot preserve a row count in a zero-column table, so at least one column must remain. The carrier is read as a data column with no output entry, the same shape as a pure-filter column, so output assembly and partition injection are unchanged. Selection uses decoded type width, not compressed size.

Files that lack the carrier column, or have no row groups to resolve it against, fall back to full-width reads. Fallback estimates use bind-type widths for matching fixed-width, non-partition top-level columns, and parquet metadata for the rest. The coalescer keeps projected and full-width files in separate splits. Schemas without an eligible fixed-width column keep the existing behavior.

Footer-only counting is outside this change. No configuration changes; `docs/super-sirius/scan.md` is updated.

**Validation.** Tests cover carrier selection, unchanged real projections and nested reads, missing-column fallback and memory sizing, partition-named file columns, and coalescing across reader-option changes. GPU results are checked against DuckDB CPU for flat, filtered, partitioned and multi-file counts, including empty files, all-VARCHAR schemas, schema evolution and an entirely NULL carrier.

All gates passed:
- `[carrier]`: 19 cases / 885 assertions.
- `[scan][parquet]`: 68 cases / 40,871 assertions.
- `make test`: 3,098 cases / 32,795,381 assertions.
- `make s3-test`: 92 cases.
- Pre-commit clean.

**Timing.** Local-file `SELECT count(*) FROM read_parquet(...)` on an RTX 3060. Both GPU runs disable `statistics_propagation` to keep the scan in the plan. Each binary gets one warm-up and three timed runs in the same session; results below are medians. The dev binary is `5201d9c3`, which differs from this branch's base only by #1719.

| File | dev | This PR | DuckDB CPU (metadata only) |
| --- | --- | --- | --- |
| TPC-H SF1 lineitem, 6.0M rows / 49 row groups | 0.452 s | 0.031 s | 0.004 s |
| TPC-H SF10 lineitem, 60.0M rows / 489 row groups | 4.249 s | 0.238 s | 0.014 s |

These numbers compare the local scan path, not cold S3 throughput. The CPU timings are a metadata-only reference.

## Checklist

- [x] Document configuration changes in code and summarize in the description above (none)
- [x] Update human and agent documentation (README.md, docs/, skills, CLAUDE.md): `docs/super-sirius/scan.md`

## References

Closes #1723


src/op/scan/parquet_gpu_ingestible.cpp
```diff
@@ -925,13 +958,29 @@ std::unique_ptr<scan_info> parquet_gpu_ingestible::build_file_scan_info(
       if (!is_pure_filter) { estimate.output_bytes += decoded_bytes; }
       estimate.compressed_bytes += static_cast<std::size_t>(column_metadata.total_compressed_size);
     };
-    if (_plan->is_projected()) {
+    if (file_projected) {
```
mbrobbel, line 980: Partition-only scans are now sized using just the carrier column, but they also allocate the partition columns, so we should include those in the memory estimate to avoid packing too many rows into a batch and running out of memory.

ran-yuan-rui: Good catch, thanks. The row-group estimates omitted the partition columns that `assemble_scan_output` creates. Full-width reads masked the gap; with a narrow carrier, a partition-only scan could pack too many rows into one split.

Fixed in 7a82778b: both the output and decode working-set estimates now include the partition columns, using their fixed width or the decoded string length plus a 4-byte offset per row. These estimates feed batch sizing and memory reservation. The fix also covers projected scans with partition output and files that fall back to a full-width read. Tests cover INTEGER/VARCHAR widths, the split cap, and the fallback path.

mbrobbel (approved):

PR #1727, opened 2026-09-07
fix(compression): drop the stale dictionary cap and improve logging

joosthooz:
`MAX_INDICES = 1 << 28` was a stale sanity bound predating the HyperLogLog cardinality gate that now guards the real hazard (cudf::dictionary::encode's illegal access on huge high-cardinality inputs) by distinct fraction rather than row count. The row cap only silently forced narrow, high-row-count pins (e.g. TPC-H q12's 5-column lineitem pin at ~276M rows/chunk, orders pins at ~600-900M) to an uncompressed fallback. Dictionary codes index the key set, not rows, so any column a cudf column_view can represent already encodes; replace the cap with size_type max.

Add a per-pin "N/M chunk(s) compressed" INFO line in both the host and device materialize paths. A per-chunk encode failure only WARNs today and silently pins that chunk raw — this class of degradation should never be silent.

Extracted from Felipe's #1391 (exp/fused-scan-filter), where these were bundled with unrelated explorer and scan-manager diagnostic changes.

## Checklist

## References


src/pin_table.cpp
```diff
@@ -696,6 +696,27 @@ host_pin_result materialize_pin_to_host(
...
+  {
+    std::size_t compressed_count = 0;
+    for (auto const& chunk : out.chunks) {
+      if (dynamic_cast<sirius::compressed_host_representation const*>(chunk.get()) != nullptr) {
+        ++compressed_count;
+      }
+    }
+    auto const paths = ingestible.table_info().file_paths();
+    SIRIUS_LOG_INFO("[materialize_pin_to_host] pin '{}': {}/{} chunk(s) compressed{}",
+                    paths.empty() ? "<unknown>" : paths.front(),
```
johanpel: Should this use `data.args.name` instead of `paths.front()`, so duckdb-native tables don't just all return the db path?

src/pin_table.cpp
```diff
@@ -696,6 +696,27 @@ host_pin_result materialize_pin_to_host(
...
+                    compressed_count,
+                    out.chunks.size(),
+                    compressed_count == out.chunks.size()
+                      ? ""
+                      : " — remainder pinned UNCOMPRESSED (see warnings above)");
```
johanpel: Should "see warnings above" only be emitted if there are actual warnings? Does this log also emit when compression is disabled?

johanpel (approved):

PR #1762, opened 2026-09-11
docs: refresh README branding, Python usage, and compression

bwyogatama:
## Description

Refresh the README to describe Sirius as a GPU-Native Composable Analytics Engine and show the supplied architecture and AWS TPC-H figures. Replace the old DGX performance claim with the hot-run comparison context: all 22 queries, the best Sirius g7e size versus DuckDB on m9g.16xlarge, cost per run, logarithmic scales, and lower-left results being better.

Add a short Python quick start based on the TPC-H benchmark, including the matching DuckDB Python build, extension loading, a Parquet query, and result fetching. Document how to enable Simpatico compression with both required settings before pinning, using the bundled TPC-H plans, with GPU/host support, plan requirements, fallback behavior, and Python usage.

## Validation

- Checked README local links, Python syntax, compression setting names and ordering, and the bundled plan path.
- Cross-checked the Python example against `test/tpch_performance/performance_test.py` and compression behavior against the extension implementation and compressed pinning guide.
- Confirmed both replacement image files exactly match the supplied attachments.
- `git diff --check` passes.
- GPU execution and the full Pixi pre-commit suite were not run: this host is macOS and Pixi is unavailable.

## Checklist

- [x] Read CONTRIBUTING.md and checked PR reviewability.
- [x] Validated the documentation and examples with the checks above; no runtime code changes.
- [x] Documented existing Simpatico options; no configuration behavior changes.
- [x] Updated the README and referenced supporting documentation.


mbrobbel (approved):

joosthooz (approved):
Thanks 👍

jrbourbeau (approved):
Thanks @bwyogatama -- really like the new cost vs runtime plot

A couple non-blocking comments on the plot:

- Thoughts on adding the instance type to each plot point instead of the legend? Assuming the increasing 2x, 4x, ... GPU instance correspond to the increasing scale factors but it'd be more clear if that was explicit 
- I see the graviton instance is the same for each of the scale factors. Why vary GPU instance but not CPU?

Issue #1763, opened 2026-09-11
R0: correctness prerequisites for the table-aware scan framework (#1760)


ran-yuan-rui:
Parent: #1760, child "Correctness prerequisites (R0)". No table-format adapter is enabled before this issue is done. It fixes two bugs that exist on `dev` today and documents one setting. The fixes are in PR #1811 (item 1) and PR #1764 (items 2 and 3). The reproducers below were run on the 2026-09-11 daily build (`dev` `addaa2b`) with DuckLake `d8a1881e`; scripts and logs are available on request.

### 1. Strings stored with `DICT_FSST` come back as NULL from the GPU

What happens, with no lakehouse involved: `ATTACH 'x.db' AS q (STORAGE_VERSION 'latest'); CREATE TABLE q.t AS SELECT i, 'v' || i AS s FROM range(100) r(i);`, exit so the shutdown checkpoint writes the segments, reopen, then `SELECT s FROM q.t;`. With `gpu_execution = true` every row is NULL and `min(s)` is an empty string. With `gpu_execution = false` the real values come back. `SELECT compression FROM pragma_storage_info('q.t') WHERE column_name = 's'` shows `DICT_FSST`. Segments stored as `Uncompressed` or `Dictionary` read correctly. Nothing throws; the result is simply wrong.

Why: a `DICT_FSST` segment stores its string lengths in a bit-packed block. DuckDB rounds the entry count of that block up to a multiple of 32 before placing the next block, the dictionary indices (`BitpackingPrimitives::GetRequiredSize`, `duckdb/src/storage/compression/dict_fsst/decompression.cpp` and `compression.cpp`). The Sirius decoder placed the indices block without that rounding (`src/cuda/scan/strings/dict_fsst.cu`, `prepare_dict_fsst` and the descriptor build). When the two sizes differ after 8-byte alignment, the decoder reads the indices from the wrong offset, which can interpret zero padding as the reserved NULL entry and return NULL. The synthetic segment generator of the existing decoder tests had the same mistake (`test/cpp/scan/strings_synth.hpp`), so the tests agreed with the decoder and no `DICT_FSST` segment written by DuckDB itself was decoded in those tests.

[...]

Issue #1760, opened 2026-09-11
Tracking issue: table-aware scan framework (Iceberg, DuckLake, Paimon, etc.)


ran-yuan-rui:
### Describe the problem to solve or your goal to achieve

Sirius runs a fixed list of scan functions on the GPU: DuckDB tables, `read_parquet`, and since #1425 `iceberg_scan`. Table formats such as Iceberg, DuckLake (#1721) and Paimon add what plain Parquet does not have: a snapshot, a schema version, delete files or deletion vectors, and sometimes rows stored in the format's own catalog instead of in data files. The format's DuckDB extension keeps that state in its bind data and its own transaction.

Iceberg was added by subclassing the Parquet reader and doing the metadata and delete work inside the planner hooks. Doing the same for DuckLake and Paimon would answer the same five questions three separate times:

- which snapshot the query reads;
- what to do with the SQL the extension runs on its own to find files;
- when a file or batch is ready for the GPU;
- what the GPU decoder can and cannot handle;
- what happens when something fails after execution has started.

The Iceberg implementation in #1425 is the starting point; its planner-time metadata work is not the shared execution model proposed here.

### Describe the solution you would like

One shared scan path that Iceberg, DuckLake and Paimon plug into. It is built on the pieces Sirius already has: `gpu_ingestible`, the metadata workers, the coalescer, the GPU scheduler, the cuDF Parquet decoder and the I/O context. This parent issue tracks the shared path and its dependencies; each format has its own child issue.

The rules below use one example: `SELECT count(*) FROM lake.sales WHERE day = DATE '2026-09-01'`. Here `lake` is a DuckLake catalog. Its data files are Parquet on S3, and its newest rows are still stored inside the DuckLake catalog database.

1. Remember which snapshot the query was bound to. When DuckDB binds `lake.sales`, the DuckLake extension picks a snapshot. Sirius records that snapshot, the bound schema and the transaction, and checks that the GPU plan reads the same tables and snapshots as DuckDB's original plan. Every later step reads the same snapshot: the GPU plan, the per-file checks, and a CPU fallback if one happens. Nothing re-resolves "latest" or re-parses the SQL halfway through.

[...]

PR #1764, opened 2026-09-13
feat(transparent): never intercept provider-internal SQL or hidden catalogs

ran-yuan-rui:
## Description

Exclude DuckLake metadata connections from GPU interception. Connections with a hidden default catalog and `catalog_error_max_schemas = 0` remain excluded for their lifetime. Statements reading or modifying hidden catalogs are excluded for the current planning attempt, including with optimizer hooks disabled and after a catalog detach or rename within a transaction. Add counters for provider-internal skips, hidden-catalog skips and classification failures.

Follow-up to the review: a connection carrying DuckLake's `ducklake_internal_connection` marker (duckdb/ducklake#1407) is classified as provider-internal before the settings fingerprint, which stays for connections without the marker; the three hooks now share one decision function, `should_use_duckdb`. Validated after that commit: `sirius_unittest` (3,242 cases) and `make s3-test` (92).

Document the existing SESSION default for `gpu_execution`: unqualified `SET` affects the current connection; `SET GLOBAL` affects connections in the same instance without a session value. Add the Boolean test option `sirius_test_inject_provenance_failure`, registered only under `SIRIUS_ENABLE_TEST_OPTIONS=1`.

Tests cover provider isolation, catalog detach and rename, and setting scope. Validated on this branch, based on `dev` `ef227c1a`: `sirius_unittest` (3,238 cases) and `make s3-test` (92).

The `DICT_FSST` decoding fixes that were part of this PR moved to #1811 at review request; the two PRs are independent.

## Checklist

## References

Refs #1763. Refs #1760.


mbrobbel (commented):
Hm, I'm not sure about this approach. Maybe we should suggest connection origin/owner metadata support upstream in DuckDB, so that we can query it to know if a connection is from an extension (and from what extension)?

If we go with the heuristic, I would suggest simplifying it to:

```cpp
decline_reason should_use_duckdb(
    ClientContext& context,
    StatementProperties const& properties) noexcept
  {
    if (has_ducklake_metadata_fingerprint(context)) {
      return decline_reason::provider_internal;
    }
    if (references_hidden_catalog(context, properties)) {
      return decline_reason::hidden_catalog;
    }
    return decline_reason::none;
  }
```

Which we call in the pre-, post-optimizer and final prepare hooks to skip the GPU execution.

But again, I really think we should try to upstream support for an explicit connection marker. What do you think?

ran-yuan-rui:
Agreed on the explicit marker, and DuckLake main already has one for the case this PR targets: the metadata connections created by `DuckLakeTransaction::GetConnection()` are marked with the `ducklake_internal_connection` key (duckdb/ducklake#1407), and DuckLake itself uses it to skip its eager snapshot fetch. Sirius can read it through DuckDB's existing connection-state API, with no DuckLake headers and no DuckDB change for this lookup. A common origin/owner API upstream would still help other extensions, and I would support proposing it. The DuckLake build this PR is qualified against (`d8a1881e`) predates the marker, which is why the PR reads the two settings.

I propose checking the marker first and retaining the heuristic for compatibility with the currently qualified build; the heuristic goes once the supported DuckLake versions provide verified marker coverage. The marker is set after the connection is constructed, so the first planning attempt stays the point where a connection is classified.

Yes to simplifying: I can restructure the hook calls around one decision function while retaining the runtime checks, the connection-lifetime marker and one decline count per planning attempt. The catalog check follows transaction-held references and verifies the catalog ID, so a detach or rename does not make it inspect a different database, and the plan walk stays for statements whose recorded properties are incomplete.

mbrobbel:
> Agreed on the explicit marker, and DuckLake main already has one for the case this PR targets: the metadata connections created by `DuckLakeTransaction::GetConnection()` are marked with the `ducklake_internal_connection` key ([duckdb/ducklake#1407](https://github.com/duckdb/ducklake/pull/1407)), and DuckLake itself uses it to skip its eager snapshot fetch. Sirius can read it through DuckDB's existing connection-state API, with no DuckLake headers and no DuckDB change for this lookup. A common origin/owner API upstream would still help other extensions, and I would support proposing it. The DuckLake build this PR is qualified against (`d8a1881e`) predates the marker, which is why the PR reads the two settings.
> 
> I propose checking the marker first and retaining the heuristic for compatibility with the currently qualified build; the heuristic goes once the supported DuckLake versions provide verified marker coverage. The marker is set after the connection is constructed, so the first planning attempt stays the point where a connection is classified.
> 
> Yes to simplifying: I can restructure the hook calls around one decision function while retaining the runtime checks, the connection-lifetime marker and one decline count per planning attempt. The catalog check follows transaction-held references and verifies the catalog ID, so a detach or rename does not make it inspect a different database, and the plan walk stays for statements whose recorded properties are incomplete.

Oh nice, I didn't know about `ducklake_internal_connection`.

However, this mostly renames the existing shared function. It still keeps the state, latching, and plan walk. Can we remove those and simplify this further, or am I missing why they're needed?

Can we bump ducklake to remove the fingerprint fallback?

ran-yuan-rui:
Thanks. DuckLake `main` has the marker from duckdb/ducklake#1407 on the metadata connections created by `DuckLakeTransaction::GetConnection()`. The DuckLake build distributed for our current DuckDB version does not include it.

I propose keeping the fingerprint fallback until Sirius moves to a release whose DuckLake has verified marker coverage for those connections, then removing the fallback and connection latch together. The latch preserves the classification when the default catalog or setting changes; the marker does not depend on either.

The per-attempt state preserves a decline across hooks and counts it once, even if optimization removes the scan. The plan walk checks for hidden base-table scans when recorded catalog properties are incomplete, and declines GPU interception if that inspection fails. I would simplify these in the same change, while preserving hidden-catalog checks for unmarked connections.

I'll update the branch for the header split on `dev` with the next push.

mbrobbel (approved):

Issue #1746, opened 2026-09-10
feat: Improve Partitioning Logic For Group By


wmalpica:
### Describe the problem to solve or your goal to achieve

There are a couple problems with our partitioning strategy for Group BY:
1. If we are partitioning and we have N gpus, then we always partition into N GPUs. We should instead partition only to the number of GPUs that make sense. Since Group by's will typically reduce before we get the partitioning phase, then we should only really be having 1 partition (no partition operation actually happens), and all the data goes to 1 GPU. For a "DISTINCT" case, where we are using the group by operator to just get distinct rows, then we may not be reducing much. In this case then when we do have a large number of rows, we may need to partition, but the partitioning should not automatically go to the number of GPUs. It should be contained to what is necessary or efficient. In a concurrent query context, it should be straightforward to demonstrate that it is more efficient to only divide the work between the number of GPUs that match maximum throughput per GPU. For in a single query at a time, that may not be the case. Therefore we need to test this out to see if we always reduce the number of GPUs/partitions to "what is efficient".
2. Assuming we do #1 above. We should also think about the fact that the partitioning operation has a cost. We might want to test to see at what point does performing the partitioning balance against dividing the merge operations between multiple partitions/gpus.  For a single gpu scenario, we will likely have a stronger preference to avoiding doing the partitioning, compared to the multi GPU scenario. So we should revisit this logic with this in mind. The partition_small_table_bytes in natural_num_partitions seems like it wants to do some of this, but it does not do a good job. 
3. We want to estimate the amount of data that should be arriving to the partitioning operator and make a decision earlier, to make the barrier a softer barrier, and therefore reduce the latency that comes as a consequence of having a full pipeline breaker. 

### Describe the solution you would like

sirius_physical_grouped_aggregate_merge::get_partition_strategy needs to be updated. 
Need to put the bytes estimation logic from https://github.com/sirius-db/sirius/pull/1482 to work

### Sirius component

Operators

PR #1765, opened 2026-09-13
feat(partition): size group-by partitions from projected input

Jedi18:
## Summary
Uses the runtime size-estimation API introduced in #1482 to size grouped-aggregation partitions before their full input arrives.
A grouped aggregation runs as:
`HASH_GROUP_BY → PARTITION → MERGE_GROUP_BY`
Previously, `PARTITION` used a `FULL` input barrier because its partition count depends on total input size and must remain fixed. This change allows partitioning to overlap with upstream aggregation.

## Description
- Adds `enable_runtime_size_estimation`, disabled by default.
- Uses a `PARTIAL` barrier for eligible group-by partitions when enabled.
- Waits for the producer if no estimate is available, preserving existing behavior.
- Applies `size_estimate_safety_factor` to projected totals and floors estimates at bytes already received.
- Adds sizing-basis counters and estimated-versus-actual logging.
- Adds integration coverage for engagement, correctness, on/off equivalence, and excluded `DISTINCT` paths.
- Updates configuration and architecture documentation.
The feature remains off by default because tested TPC-H workloads showed no measurable wall-clock improvement, despite accurate projections.

## Checklist

## References
#1746
Note that this only addresses point 3 of the above issue. Point 1 and 2 should be done as a separate PR after some experimentation and benchmarking


src/sirius_context.cpp
```diff
@@ -1247,6 +1247,33 @@ void SiriusContext::record_compressed_materialization_scan_narrow_targets_retrac
     count, std::memory_order_relaxed);
 }
 
+SiriusContext::size_estimation_stats SiriusContext::get_size_estimation_stats() const noexcept
```
wmalpica, line 1250: We should not add all these functions into SiriusContext which are just for testing.  For that matter, we really should remove `record_compressed_materialization_scan_narrow_targets_retracted` and all of its siblings and replace them with something that does not litter the SiriusContext class. 

I am not quite sure what it is, but there has to be another way. 
Amin's upcoming PR https://github.com/sirius-db/sirius/pull/1776  
Introduces a way of capturing events, which we could leverage. BUT it is asynchronous. So if we use that, we may need a way to synchronize, like with a flush() or wait_to_drain(event_id)

src/sirius_config.cpp
```diff
@@ -310,6 +310,8 @@ static void from_yaml(const YAML::Node& node, operator_params& opt)
     throw std::runtime_error(
       "'operator_params.avg_variable_column_bytes': must be greater than zero");
   }
+  r.optional("enable_runtime_size_estimation", opt.enable_runtime_size_estimation);
+  r.optional("size_estimate_safety_factor", opt.size_estimate_safety_factor);
```
wmalpica, line 314: I dont think we should add `size_estimate_safety_factor`. Its just another configuration option to maintain, and i dont think its necessary.

wmalpica (changes requested):

src/sirius_context.cpp
Jedi18: I will wait for Amin's PR to go in then and try that approach

Jedi18: Adding synchronization might be quite involved, so I’ve removed the new counters and added operator tests to verify the sizing directly. The integration tests still cover correctness.

We can migrate the existing compressed-materialization counters in a separate cleanup once Amin’s event infrastructure is available.

src/sirius_config.cpp
Jedi18: Ok agreed, removed it and changed it to use the estimated total directly, floored at the bytes already received.

src/sirius_context.cpp
Jedi18: Here is the PR that uses the event system for compressed materialization tests like you suggested instead of adding functions to SiriusContext https://github.com/sirius-db/sirius/pull/1847

wmalpica (approved):

PR #1766, opened 2026-09-14
fix: dense count join memory estimation/usage

joosthooz:
## Description
At scale factor 3000 on a smaller machine, the new dense join operator (https://github.com/sirius-db/sirius/pull/1606) goes off the rails and tries to reserve 800GB of device memory (10x the size of the device). This PR changes the estimation to hopefully track actual usage a bit closer (please have a look if you think there are situations where the estimate is now too low!). And it adds a fallback path for when the new operator will not fit.

Adds config parameter dense_count_join_memory_fraction to set what part of the memory to use for the dense count join. Default 10%. The other parameter, setting an explicit byte count for the threshold, still exists but is now set to 0 (auto) to use the new fraction.

## Checklist

## References


src/planner/sirius_plan_aggregate.cpp
```diff
@@ -331,6 +334,35 @@ static std::optional<dense_count_join_detection> detect_dense_count_join(
...
+/// Histogram budget in bytes, resolving the auto (0) setting to a share of GPU tier
+/// capacity. An explicit non-zero setting wins; an unreadable space falls back to a constant.
+[[nodiscard]] std::uint64_t resolve_dense_count_budget(duckdb::SiriusContext& sirius_ctx,
+                                                       sirius::operator_params const& op_params)
+{
+  if (op_params.dense_count_join_max_bytes > 0) { return op_params.dense_count_join_max_bytes; }
+  auto const& manager    = sirius_ctx.get_memory_manager();
+  std::uint64_t capacity = 0;
+  for (auto const* space : manager.get_memory_spaces_for_tier(cucascade::memory::Tier::GPU)) {
+    if (space != nullptr) { capacity = std::max<std::uint64_t>(capacity, space->get_max_memory()); }
```
kevkrist, line 346: Why max here? Should this be min, in case one gpu has significantly less memory?

src/planner/sirius_plan_aggregate.cpp
```diff
@@ -378,7 +432,9 @@ sirius_physical_plan_generator::try_plan_dense_count_join(duckdb::LogicalAggrega
     detection->preserved_key_idx,
     detection->counted_key_idx,
     detection->counted_value_idx,
-    op_params.dense_count_join_max_bytes,
+    budget,
+    histogram_bytes,
```
kevkrist, line 436: Assuming `preserved_rows` is accurate, in the worst case, when there are 8 slots per preserved row (the *least* dense case), this will underestimate the histogram by 8x. Is it better to keep `max_admitted_histogram_bytes(_max_bins_bytes, stats.bytes)` as the histogram term in `no_history_peak_memory_estimate()`? The planner estimate will still decide whether to fuse at plan-time. Of course, this can also be 8x over-conservative, but I think conservatism is the right principle here.

kevkrist (changes requested):
Thanks for this. The original dense count join PR was admittedly rushed and there are bound to be some warts.

src/op/sirius_physical_dense_count_join.cpp
```diff
@@ -440,8 +457,20 @@ std::size_t sirius_physical_dense_count_join::no_history_peak_memory_estimate(
...
+  //
+  // An estimate above the tier is not the safe direction: it cannot be granted, so it yields
+  // a clamp and a partial reservation instead of a refusal.
+  constexpr std::size_t kSparseGroupFactor = 8;
+  auto const avg_batch_bytes =
+    stats.num_batches > 0 ? stats.bytes / stats.num_batches : stats.bytes;
+  auto sparse_peak = saturating_add(allocation_floor, avg_batch_bytes);
+  sparse_peak      = saturating_add(
+    sparse_peak,
+    saturating_mul(saturating_mul(kSparseGroupFactor, key_width + sizeof(int64_t)), output_rows));
```
kevkrist, line 473: output_rows only bounds the preserved-side groups and final output rows. In sparse mode, groups for every distinct *counted*-side key are first grouped, including keys absent from the preserved side. If there are many distinct unmatched counted-side keys, this term can be much smaller than the actual hash state. Not sure what a good solution is here. Is there a safe *counted*-side group bound we could use in sparse_peak? Perhaps its worth adding a regression test for this case?

src/planner/sirius_plan_aggregate.cpp
joosthooz: Yes, fixed!

src/planner/sirius_plan_aggregate.cpp
joosthooz: This is addressed in the planner, based on the data

src/op/sirius_physical_dense_count_join.cpp
joosthooz: Now passing on a value from the planner, and added a test for it

kevkrist (approved):

kevkrist:
A couple of things to address before merging:
- Documentation at docs/super-sirius/configuration.md should be updated to reflect the new parameters.
- The old integration tests in test_gpu_execution_dense_count_join.cpp have their meaning slightly altered by these changes. Before, the query shape alone determined whether DENSE_COUNT_JOIN triggered, and the runtime cardinalities determined whether DENSE or SPARSE mode was used, but now, even with the right shape, the plantime cardinality estimates can cause JOIN -> GROUP BY, and some of these tests now (I think) compare JOIN -> GROUP BY against itself. Can you double check that?

PR #1771, opened 2026-09-14
feat(io): add frugal caching and dynamic I/O

aminaramoon:
## Description

Large and remote parquet scans need to avoid eagerly reading and decoding data that a query may never consume. This adds frugal caching and dynamic I/O so physical requests can be split, aligned, coalesced, staged, and served incrementally according to each backend's capabilities.

The layer includes:

- partial-read cache fills with reader/evictor arbitration and explicit lifecycle handling;
- reactor-owned request splitting and bulk prepared-slice APIs for REST, io_uring, and kvikIO;
- query-event-driven readahead, scheduling, and scan-manager integration;
- parquet materialization and device-copy support;
- `reset_sirius_cache()` and updated cache/backend configuration;
- `pin_table(..., tier='parquet')` for retaining undecoded parquet ranges in the I/O cache;
- updated documentation, performance scripts, and I/O benchmarks.

Tests cover cache state transitions, request planning, backend behavior, readahead lifecycle, configuration, reset behavior, and parquet scan sizing.

## Validation

- `pixi run pre-commit run -a` passes independently at all five commits in this PR.
- `pixi run cmake --build build/release -j 4` passes for the final stacked tree.

## Checklist

- [x] Cover cache, reactor, readahead, scan-manager, and parquet behavior with tests.
- [x] Document backend, cache, and pinning configuration changes in code and user documentation.
- [x] Update performance scripts, benchmarks, and agent documentation.

## Stack

Layer 6 of 6. Based on `stacked/frugal-native-05-cache-handle`.


src/include/io/cache/types.hpp
```diff
@@ -129,407 +149,841 @@ class buffer_pool {
...
     evicting  = 6,
   };
 
-  entry_state() noexcept = default;
+  static constexpr std::uint32_t MAX_PINS        = (1U << 12) - 1;
+  static constexpr std::uint32_t MAX_SUBSCRIBERS = (1U << 16) - 1;
+
+  /// Widest extent the packed field can express, in pages.  The chunk size must
+  /// stay under this many pages; @ref buffer_pool checks it at construction.
+  [[nodiscard]] static constexpr std::uint32_t max_fill_pages() noexcept { return (1U << 14) - 1; }
```
joosthooz, line 325: This causes the cache to silently not work in our gb300 optimized config, because it uses 64MiB host block_size. This value, with the 4k page size, results in a limit that is just short of that. Can we widen the field to at least accomodate 128MiB (minus 1 page)?

And there should be a loud failure when this happens.

src/io/rest/rest_reactor.cpp
```diff
@@ -367,69 +378,135 @@ std::chrono::milliseconds compute_backoff(std::size_t attempt,
...
-/// becoming a standalone single-buffer GET.  Input segments must be in file
-/// order; a null-buffer segment (bounce-staged device read) is kept as a
-/// standalone single-buffer output and never fused into a scatter group.  Each
-/// output segment's @c size is the contiguous file span its buffers cover.
-std::vector<io_object_segment> chunk_host_segments(std::span<const io_object_segment> segs,
-                                                   size_t chunk_size,
-                                                   size_t max_n_chunks,
-                                                   bool allow_split = true)
+constexpr std::size_t rest_min_segment_bytes = 4UL << 20;
+constexpr std::size_t rest_max_segment_bytes = 16UL << 20;
```
joosthooz, line 382: Here's another limit that clashes with 64MB host blocks, any downside to increasing it?

mike-wendt (approved):
Approved for `sirius-build` CODEOWNERS `Makefile` only

nit: [/test/cpp/integration/s3/fixtures/README.md](https://github.com/sirius-db/sirius/blob/stacked/frugal-native-06-dynamic-io/test/cpp/integration/s3/fixtures/README.md) still refers to `make s3-bench` and should be updated

src/io/rest/rest_reactor.cpp
aminaramoon: yes, increasing it will lieterally make the IO slow because s3 caps throughput per stream. I think cucascade block size 64MB is HACKY and should never be set that high if its goal is to avoid fragmentation

src/include/io/cache/types.hpp
ran-yuan-rui: Agree on failing loudly, and the silence has a concrete location: the cache constructor throws when the block size exceeds `max_fill_pages() * IO_BLOCK_SIZE` (`prefetching_cache.cpp:267–273`, 64 MiB − 4 KiB with 4 KiB pages), but `ioctx::initialize_cache()` (`io_context.cpp:38`) catches it, logs at ERROR and resets the cache, so the query runs with the cache silently absent. There is a second, sharper mismatch below that limit: a 32 MiB host block passes the constructor, yet when a request covers more than 16 MiB of one cache chunk the fragmented fill reaches `physical_ranges` and throws "one cache fill exceeds the REST segment maximum" (`rest_reactor.cpp:434–435`), so the cache enables and then fails at fill time. Widening the packed field alone does not close that; the cache block size and `rest_max_segment_bytes` need one validated relationship, either by splitting a fill into segments or by rejecting at load the block sizes the REST reactor cannot fill.

src/sirius_config.cpp
```diff
@@ -767,26 +805,84 @@ void sirius_config::load_from_file(const std::filesystem::path& config_path)
...
+void sirius_config::derive_uring_scan_budget()
+{
+  // The uring backend wants one concurrent scan per pipeline executor thread.
+  // io::uring::config can only spell that as the COMPILE-TIME thread count, so
+  // a config that resizes the pipeline pool would otherwise leave the readahead
+  // budget pinned to the old default and unable to keep the pool fed.
+  //
+  // Only the untouched default is replaced, so an explicit
+  // uring.n_max_concurrent_scans in the config still wins.
+  constexpr auto struct_default = static_cast<std::size_t>(exec::default_gpu_pipeline_num_threads);
```
ran-yuan-rui, line 832: `derive_uring_scan_budget` treats `uring.n_max_concurrent_scans` as "not set" whenever it equals the built-in default (:831–:832), so an explicit value that happens to equal that default is overwritten by the pipeline thread count, which changes the eager readahead budget. The REST config carries an `n_max_concurrent_scans_explicit` flag for exactly this case; the uring config could use the same, so that the comment "an explicit value still wins" holds for every value.

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -1433,20 +1452,73 @@ void sirius_scan_manager::prepare_for_query(const sirius::planner::query& query,
...
+  // Order scans ahead of demand, on the terms resolve_readahead settles: an
+  // explicit `max_readahead_scans` / `readahead_strategy`, or what the serving
+  // backend wants.  A budget of zero means "do not read ahead" — either
+  // configured off, or a backend that publishes no depth (kvikIO by default) —
+  // and start() is then a no-op.
+  //
+  // The serving backend is not necessarily the default `_io_ctx`: a path-routed
+  // backend serves its own scans (an `s3://` query reads through the REST ioctx
+  // while `_io_ctx` is the local uring one) and one readahead manager covers
+  // them all.  Take the one publishing the widest budget, since object-store
```
ran-yuan-rui, line 1476: The readahead budget is settled once per query from whichever ioctx publishes the widest `n_max_concurrent_scans` — the default ioctx plus every routed ioctx retained across queries (:1482–:1497) — and a single `readahead_scan_manager` with a single `gatekeeper` then rations every backend's prefetch tickets (one per split) against that number. A query whose splits are all served by uring is therefore prefetched to the object-store depth under the eager strategy, and a routed ioctx retained from an earlier query can set the budget for one that never touches it. The serving ioctx is known when a split is prefetched, so the budget could be settled per serving backend (one gatekeeper per ioctx, or tickets charged to the split's backend). This is the admission point I raised on the umbrella PR; re-posting it where the code is reviewed.

src/io/kvikio/kvikio_context.cpp
```diff
@@ -16,49 +16,166 @@
...
 std::shared_ptr<io_object> kvikio_context::create_io_object(std::string path)
 {
-  // cudf::io::datasource::create returns a unique_ptr; promote to shared_ptr
-  // so the kvikio_io_object can expose access without transferring
-  // ownership (the io_object outlives any single sirius_datasource we hand
-  // back from open_datasource).
-  std::shared_ptr<cudf::io::datasource> ds = cudf::io::datasource::create(path);
-  auto const file_size                     = ds->size();
-  return std::make_shared<kvikio_io_object>(std::move(path), std::move(ds), file_size);
+  if (is_s3_uri(path)) {
```
ran-yuan-rui, line 137: `kvikio_context` serves `s3://` through `kvikio::RemoteHandle` with the `object_store` credentials (:135–:140) whenever `backend: kvikio`, but `docs/super-sirius/configuration.md:324` still says kvikio "routes local files to the kvikIO fallback", which the new `scan_manager serves S3 reads from kvikio when backend=kvikio` test contradicts; the doc should say kvikio serves `s3://` too and where its credentials come from. Related: `object_store.s3_transport` is read (`sirius_config.cpp:172`) but consumed by neither backend (that was already true of the REST path on dev); since this PR is where the backend/transport matrix gets defined, it is the natural place to either honour it or reject unsupported combinations at load.

src/include/io/types.hpp
```diff
@@ -107,105 +120,162 @@ class io_object_metadata {
...
 
-  /// True iff this segment must be submitted via io_uring_prep_readv (rather
-  /// than a single io_uring_prep_read).
-  [[nodiscard]] bool is_vectored() const noexcept { return buffers.size() > 1; }
+  [[nodiscard]] bool is_contiguous() const noexcept
+  {
+    return std::holds_alternative<std::uint8_t*>(buffer);
+  }
+
+  [[nodiscard]] bool is_staged() const noexcept { return needs_staging(); }
```
wmalpica, line 194: is_staged == needs_staging
this is very confusing

src/include/io/io_request.hpp
```diff
@@ -47,209 +46,346 @@
...
+                     std::vector<prepared_io_slice> request_slices,
+                     std::shared_ptr<grouped_coordinator> group)
+    : obj(std::move(object)), slices(std::move(request_slices)), coordinator(std::move(group))
+  {
+  }
+
+  std::size_t _next{0};
+};
+
+struct device_cpy_request {
```
wmalpica, line 290: I see we have a struct with the same name and similar purpose(?) in cucascade. Will we eventually only have one? or we are just going to have two things with the same name on different repos?

src/io/cache/prefetching_cache.cpp
```diff
@@ -313,125 +375,382 @@ prefetching_cache::file_entry& prefetching_cache::get_or_create_file_entry(const
...
+  auto const& chunks = *req.chunks;
+
+  std::size_t n_chunks_needed = std::ranges::count_if(
+    chunks, [](cached_chunk* c) { return c->state.get_state() == chunk_state::empty; });
+
+  int numa_allocated = req.preferred_numa;
+  auto buffers       = _pool->allocate_bulk(n_chunks_needed, numa_allocated);
+  if (buffers.size() != n_chunks_needed) {
+    if (!buffers.empty()) { _pool->deallocate_bulk(std::move(buffers), numa_allocated); }
+    req.producer->mark_abandoned();
```
wmalpica, line 519: If we cant allocate enough here, then the producer ends up in a terminal abandoned state. It will try to retry and not be able to. @aminaramoon I will send you a detailed codex play by play of how this can happen

wmalpica (changes requested):

src/creator/task_creator.hpp
```diff
@@ -87,6 +88,22 @@ struct task_creation_request {
...
 };
 
+/// Why @ref task_creator::get_operator_for_next_task stopped.
+enum class next_task_state {
+  /// An operator is ready to have a task created for it.
+  ready,
+  /// No operator in the chain can produce right now.  Not necessarily terminal:
+  /// an operator waiting on input reports this until its input arrives.
+  depleted,
+};
```
9prady9: are we interested in any other state that is not-ready ? If not, why not just use boolean inside `next_task_result` ?

9prady9: Also, convention for enum values is all caps

9prady9: looks like we don't have any convention unfortunately, in a follow up later.

src/cuda/device_copy_batch.hpp
```diff
@@ -0,0 +1,134 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cudf/detail/utilities/cuda_memcpy.hpp>
```
9prady9: is this the right place for this header ? Given our goal for doing keeping headers clean, it might be better to move the implementation into a cpp so that the header is clean of such dependency inclusions.

src/io/cache/config.hpp
9prady9, line 1: I understand this is a simpler header with no dependency inclusions, but it is better to have headers or move it out of `include`.

But having said that, once #1806 is in, this might need to be moved perhaps.

9prady9 (commented):
I understand #1806  is not related to this, but it would help a lot for reviewing if related sources (hpp/cpp) are next to each other while reviewing when diff is extremely large.

Can you please rebase and resolve conflicts once that is in, thank you!

src/sirius_config.cpp
```diff
@@ -768,26 +806,84 @@ void sirius_config::load_from_file(const std::filesystem::path& config_path)
...
```
9prady9: What is the usual convention in sirius for these ? If sentinels make sense, I would say lets be consistent.

src/sirius_extension.hpp
```diff
@@ -129,6 +129,18 @@ class SiriusExtension : public Extension {
...
                                                  vector<string>& names);
 
+  /// reset_sirius_cache(): drop every ioctx's prefetching cache and rebuild it
+  /// empty, so the next query pays its own IO instead of reading what the last
+  /// one left resident.  A no-op where the configuration or the backend gives
+  /// no cache in the first place.
+  static void ResetSiriusCacheFunction(ClientContext& context,
+                                       TableFunctionInput& data_p,
+                                       DataChunk& output);
+  static unique_ptr<FunctionData> ResetSiriusCacheBind(ClientContext& context,
```
9prady9, line 139: nit: missing doc

src/sirius_extension.cpp
```diff
@@ -1382,6 +1388,22 @@ void SiriusExtension::PinTableFunction(ClientContext& context,
...
+    // decode, no GPU materialisation and no pinned_entry, because the residency
+    // it creates lives in the IO cache and the ordinary scan path finds it by
+    // path and offset. Everything below this point is the decode-and-place
+    // machinery the other two tiers need, so this returns rather than falls
+    // through it.
+    if (data.args.tier == "parquet") {
+      scan_mgr.pin_parquet_ranges(data.args.name, file_paths, data.args.cols);
+      window.finish();
+      output.SetCardinality(1);
+      output.SetValue(0, 0, Value::BOOLEAN(true));
```
9prady9, line 1402: Can this report success when some requested bytes are still missing from the cache? If an earlier pin populated only a prefix of a chunk, a later pin for another column in the same chunk seems to reuse it without widening the cached extent. `prefetch()` then skips that chunk and can call `on_done(true)` with no work issued. Should `pin_parquet_ranges()` verify the requested coverage before we return success here?

src/sirius_extension.cpp
```diff
@@ -1667,6 +1689,49 @@ void SiriusExtension::UnpinTableFunction(ClientContext& context,
...
+  names.emplace_back("Success");
+  return make_uniq<ResetSiriusCacheFunctionData>();
+}
+
+void SiriusExtension::ResetSiriusCacheFunction(ClientContext& context,
+                                               TableFunctionInput& data_p,
+                                               DataChunk& output)
+{
+  auto& data = data_p.bind_data->CastNoConst<ResetSiriusCacheFunctionData>();
+  if (data.finished) { return; }
```
9prady9, line 1711: Is this a valid concern for prepared statements? `finished` lives in bind data, so if DuckDB reuses the plan, could the second `EXECUTE` see the flag left by the first execution and return without resetting the cache?

```sql
SET gpu_execution = false;
PREPARE r AS SELECT * FROM reset_sirius_cache();
EXECUTE r;
EXECUTE r;
```

Should this flag live in per-execution `GlobalTableFunctionState` instead, so it starts as `false` for each execution while still preventing repeated resets within that execution?

src/scan_manager/readahead_scan_manager.cpp
```diff
@@ -0,0 +1,393 @@
...
+void readahead_scan_manager::reset() { stop(); }
+
+void readahead_scan_manager::arm_prefetching()
+{
+  // Once: reloading a live gatekeeper would forget how much the executor is
+  // currently competing and hand the readahead a budget it has already spent.
+  //
+  // This is also the only arming there is -- the gatekeeper starts with no
+  // tickets, so the worker's acquire simply times out until this runs.
+  if (_prefetching_started.exchange(true)) { return; }
```
9prady9, line 444: Is opportunistic mode intended to keep monitoring executor idleness and allow new prefetches only during those opportunities, or is the first qualifying signal meant to mark readahead ready for the rest of the query? With `_prefetching_started.exchange(true)`, the first signal permanently arms the gate and later busy/idle changes have no effect. Could we clarify which behavior is intended?

src/scan_manager/gatekeeper.hpp
```diff
@@ -0,0 +1,172 @@
...
+  explicit gatekeeper(int budget) : _budget(budget) {}
+
+  gatekeeper(gatekeeper const&)            = delete;
+  gatekeeper& operator=(gatekeeper const&) = delete;
+
+  /// Hand out the full budget, and undo any earlier @ref stop.
+  ///
+  /// Outstanding debt is cleared with it: it describes how the executor was
+  /// competing before, which says nothing about now.
+  void reload()
```
9prady9, line 75:
```suggestion
  void reset_to_full_budget()
```
something like that makes it more clear while reading the invocation sites I think

src/op/scan/gpu_ingestible.cpp
```diff
@@ -35,7 +35,7 @@ filtered_table gpu_ingestible::materialize_table(
 {
   auto* mem_space = split.gpu_memory_space;
   if (split.has_scan_metadata()) [[likely]] {
-    split.prefetch(io::cache::prefetching_stage::disposable);
+    split.update(io::cache::scan_stage::reading);
```
9prady9, line 38: Could we preserve this foreground read's ticket when opportunistic readahead is first armed? `update(reading)` reaches `acquire_or_borrow()` on the shared gatekeeper, so before activation it can take `_available` from `0` to `-1`. If an activation event arrives while this scan is still reading, `reload()` assigns `_available = _budget`, forgetting that outstanding charge. That permits the full background budget alongside this read, and its later disposal adds an extra permit. Should first activation add the budget while preserving existing demand-read debt?

src/cuda/device_copy_batch.hpp
9prady9: not needed anymore I guess

src/io/cache/types.hpp
9prady9, line 1: this file doesn't feel like a header with just types anymore, I think we should reorg this a bit more.

CMakeLists.txt
```diff
@@ -96,6 +96,24 @@ endforeach()
 find_package(cudf REQUIRED CONFIG)
 find_package(raft REQUIRED CONFIG) # raft must be found before cuvs
 find_package(cuvs REQUIRED CONFIG)
+find_package(kvikio REQUIRED CONFIG)
```
9prady9, line 99: The vcpkg port still builds KvikIO with `KvikIO_REMOTE_SUPPORT=OFF`, while this PR unconditionally calls `RemoteHandle`/`S3Endpoint`; could we enable remote support and its dependencies there, or guard this source so the vcpkg build can link?

CMakeLists.txt
```diff
@@ -96,6 +96,24 @@ endforeach()
...
+
+# kvikio::kvikio's interface pulls in BS::thread_pool, whose
+# INTERFACE_COMPILE_FEATURES include `cuda_std_17`. That requirement propagates
+# through sirius_extension into duckdb_static, whose project() scope enables
+# only C/CXX, and CMake then fails to generate with "No known features for CUDA
+# compiler". Sirius only uses kvikIO's host-side API, so drop the CUDA feature
+# from the imported target.
+if(TARGET BS::thread_pool)
```
9prady9, line 107: Could we link `kvikio::kvikio` as `PRIVATE` by converting these targets to keyword-signature `target_link_libraries`, letting CMake retain the static final-link dependency through `LINK_ONLY` without mutating the imported `BS::thread_pool` target globally?

CMakeLists.txt
```diff
@@ -483,6 +502,8 @@ set(EXTENSION_SOURCES
     src/scan_manager/insert_delta_job.cpp
     src/scan_manager/mvcc_mask_job.cpp
     src/scan_manager/pinned_chunk_stats.cpp
+    src/scan_manager/prefetching_scheduler.cpp
```
9prady9, line 506: `prefetching_scheduler` appears to be referenced only by its unit test, and `sirius_extension` is whole-archived into the shipped library; could we move this source to `TEST_SOURCES` unless production wiring is planned in this PR?

CMakeLists.txt
```diff
@@ -1192,3 +1234,277 @@ set_target_properties(
   PROPERTIES CXX_STANDARD 20
              CXX_STANDARD_REQUIRED ON
              RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/test/io")
+
+# -----------------------------------------------------------------------------
+# test/io/prefetch_benchmark — plain read_parquet vs. prefetch-then-read
+# -----------------------------------------------------------------------------
+add_executable(prefetch_benchmark test/io/prefetch_benchmark.cpp)
```
9prady9: Could we factor the seven new benchmark target blocks into a helper that applies their shared includes, libraries, link option, and C++20 properties?

src/io/cache/types.hpp
```diff
@@ -129,407 +149,841 @@ class buffer_pool {
...
-  const std::size_t expected_count  = (last_chunk_off - first_chunk_off) / chunk_size + 1;
+  auto const first_chunk_off = (offset / chunk_size) * chunk_size;
+  auto const last_chunk_off  = ((offset + size - 1) / chunk_size) * chunk_size;
+  auto const expected_count  = (last_chunk_off - first_chunk_off) / chunk_size + 1;
 
   // Find the first chunk at/after the aligned start (chunks are sorted).
-  auto first_it = std::lower_bound(first, last, first_chunk_off, [](const auto& c, std::size_t v) {
-    return std::to_address(c)->offset < v;
-  });
+  auto const first_it = std::lower_bound(
```
9prady9, line 969: Could we use `std::ranges::lower_bound(chunks, first_chunk_off, {}, &cached_chunk::offset)` to express this lookup directly with a projection?

src/op/scan/parquet_materialize.cpp
```diff
@@ -0,0 +1,234 @@
...
+  // A filter is disqualifying, not merely unsupported: materialize_all_columns
+  // ignores one rather than rejecting it, so taking this route with filtered
+  // options would hand back every row and claim it was filtered.  Enforced here
+  // rather than left to each caller to remember.
+  if (options.get_filter().has_value()) { return false; }
+  if (sources.empty()) { return false; }
+  // Every source has to qualify: the decode consumes one flattened chunk-data
+  // span covering all of them, so a single file that cannot be read this way
+  // takes the whole split to the general route.
+  return std::all_of(sources.begin(), sources.end(), [](parquet_source const& src) {
```
9prady9, line 55: `std::ranges::all_of(sources, ...)` should directly work fine I think

test/cpp/io/cache/test_chunk_state.cpp
```diff
@@ -0,0 +1,628 @@
...
+    make_allocated(s);
+    REQUIRE(s.merge_fill(needed_fill(OFF, CHUNK, OFF, OFF + filled_pages * PAGE)));
+
+    std::atomic<bool> published{false};
+
+    std::thread loader([&] {
+      chunk_fill got{};
+      if (!s.take_loading(got)) { return; }
+      auto const [lo, hi] = fill_span(got, OFF, CHUNK);
+      std::fill(buffer.begin() + static_cast<std::ptrdiff_t>(lo - OFF),
```
9prady9, line 484: Could we express this interval as `std::ranges::fill(std::span{buffer}.subspan(lo - OFF, hi - lo), WRITTEN)` and avoid the iterator-offset casts?

test/cpp/scan_manager/test_readahead_lifecycle.cpp
```diff
@@ -0,0 +1,457 @@
...
+}
+
+TEST_CASE("stop interrupts a waiting acquire", "[scan_manager][gatekeeper]")
+{
+  gatekeeper g{1};
+  g.reload();
+  REQUIRE(g.acquire_for(INSTANT));  // take the only ticket
+
+  std::atomic<bool> returned{false};
+  std::thread waiter{[&] {
```
9prady9, line 273: use `std::jthread` so a failing assertion before an explicit `join()` cannot terminate during unwinding?

src/io/kvikio/kvikio_context.cpp
```diff
@@ -16,49 +16,166 @@
...
 
 namespace {
-// -- Protected placeholders --------------------------------------------------
 
-const kvikio_io_object& as_kvikio(const io_object& obj)
+constexpr std::string_view k_s3_scheme = "s3://";
+
+[[nodiscard]] bool is_s3_uri(std::string_view path) noexcept
+{
+  return path.size() > k_s3_scheme.size() && path.compare(0, k_s3_scheme.size(), k_s3_scheme) == 0;
```
9prady9, line 45: perhaps `path.starts_with(k_s3_scheme)` clubbed with nonempty-object size check is better ?

src/scan_manager/readahead_scan_manager.cpp
```diff
@@ -0,0 +1,393 @@
...
+
+void readahead_scan_manager::start(prefetch_strategy strategy)
+{
+  // A backend that publishes a zero budget has opted out; running a worker that
+  // can never issue anything would just be a thread parked on a condvar.
+  if (_budget == 0) { return; }
+  _strategy    = strategy;
+  _stop_source = std::stop_source{};
+  _prefetch_worker =
+    std::jthread([this](const std::stop_token& st) { worker_loop(st); }, _stop_source.get_token());
```
9prady9, line 107: use `std::jthread`'s injected stop token and `_prefetch_worker.request_stop()` so `_stop_source` and the duplicate stop state can be removed?

docs/super-sirius/configuration.md
```diff
@@ -4,7 +4,7 @@ This document covers Super Sirius configuration: the `sirius_config` class, oper
 
 ## `sirius_config`
 
-**File:** `src/sirius_config.hpp`
+**File:** `src/include/sirius_config.hpp`
```
9prady9, line 7: Could we update the `src/include/...` references in this file and `optimizations.md` to their actual colocated `src/...` paths? This head has no `src/include/` directory.

docs/super-sirius/configuration.md
```diff
@@ -356,30 +370,97 @@ and transport use one trust policy; there are no separate REST YAML controls.
...
+as prepared slices without forcing a static physical layout. Device operations
+allocate as many pinned CuCascade blocks as their selected physical range needs;
+those blocks stay owned through curl retries, the asynchronous H2D copy, and its
+CUDA completion event. Staging is therefore proportional to active device work
+instead of being reserved as one fixed bounce slot per connection.
+
+### `scan_manager.kvikio` — kvikIO local-file backend (`io/kvikio/config.hpp`)
+
+Used only when `backend: kvikio` routes local files to the kvikIO
+fallback. **Every key is optional and unset means "leave kvikIO's own default
```
9prady9, line 407: Could we list `n_max_concurrent_scans` separately and scope this kvikIO-default paragraph to the remaining keys? The parser accepts it as Sirius’s readahead budget, whose default `0` disables kvikIO readahead.

docs/super-sirius/configuration.md
```diff
@@ -356,30 +370,97 @@ and transport use one trust policy; there are no separate REST YAML controls.
...
+| `mode` | enum: `none`, `os`, `sirius` | `none` | Which cache the read path goes through. Values are lowercase. |
+| `eviction` | enum: `idle`, `lru` | `lru` | What retires an idle chunk from the Sirius cache. Only meaningful under `mode: sirius`. Values are lowercase. |
+| `eviction_threshold_fraction` | double [0,1] | 0.8 | Start evicting when the cache pool fills to this fraction. |
+| `min_prefetching_budget_fraction` | double [0,1] | 0.05 | Floor of the pool reserved for prefetching. |
+
+`mode: none` bypasses every cache (`O_DIRECT`, no prefetching cache). `mode: os` reads
+through the kernel page cache instead. `mode: sirius` reads `O_DIRECT` into Sirius's own
+pinned prefetching cache.
+
+`eviction: lru` keeps idle chunks for reuse and evicts least-recently-used ones once the
```
9prady9, line 457: Could we update the LRU claims here and in the scan/optimization docs? This head removed `chunk_lifecycle` scoring: the evictor now sweeps tracked requests in queue order, and the recorded request timestamp is never read.

docs/super-sirius/configuration.md
```diff
@@ -427,8 +524,7 @@ individually.
...
-| `dense_count_join_max_bytes` | 0 (auto) | Cap on `DENSE_COUNT_JOIN`'s combined direct-address histogram footprint. Auto = `dense_count_join_memory_fraction` of the smallest visible GPU's capacity (2 GiB if capacity cannot be read at planning time). At plan time, a preserved/counted cardinality estimate whose histogram would exceed the budget declines the fusion entirely, planning the ordinary join + aggregate instead; within a fused plan, a key domain that turns out too wide at runtime still takes the operator's exact sparse (eager-aggregation) strategy. |
-| `dense_count_join_memory_fraction` | 0.10 | Fraction of the smallest visible GPU's capacity used to resolve `dense_count_join_max_bytes` when it is 0 (auto). |
+| `dense_count_join_max_bytes` | 2 GiB | Cap on `DENSE_COUNT_JOIN`'s combined direct-address histogram footprint; a key domain too wide for the budget takes the operator's exact sparse (eager-aggregation) strategy. Must be greater than zero. |
```
9prady9, line 533: Could we remove or correct this row? YAML rejects the key as engine-owned; the normal default is `0`/auto (10% of GPU tier), and `2 GiB` is only the unreadable-capacity fallback; the same-named `SET` option is test-only.

test/tpch_performance/CLAUDE.md
```diff
@@ -125,7 +125,7 @@ All commands run from the **project root** directory.
 ```bash
 export SIRIUS_CONFIG_FILE=$(pwd)/test/cpp/integration/integration.yaml
 
-# Both engines, 2 iterations, hot-cache (grouped) mode
+# Both engines, 2 iterations, hot cache (the default)
```
9prady9, line 128: Could we call this the default back-to-back ordering here and in the benchmark-skill example, or pass `--execution hot`? Omission preserves `integration.yaml`, which does not enable the Sirius cache.

test/tpch_performance/CLAUDE.md
```diff
@@ -150,29 +150,127 @@ pixi run python test/tpch_performance/performance_test.py \
...
+sirius,q1,0,0.383860
+sirius,TOTAL,0,1.062193
+```
+
+A query that failed or timed out records `nan` and is left out of the total rather than
+poisoning it; the run log says how many queries a short total covers.
+
+#### `--pin parquet`
+
+A fourth pin tier alongside `gpu` / `host`. Instead of decoding and materialising columns, it
```
9prady9, line 197: Could we also add `parquet` to the key-flags list below and the benchmark skill's `--pin` choices? Both still say only `none|gpu|host` despite `PIN_CHOICES` including `parquet`.

test/tpch_performance/CLAUDE.md
```diff
@@ -150,29 +150,127 @@ pixi run python test/tpch_performance/performance_test.py \
...
+is kept beside it as `config.yml`.
+
+| `--execution` | `cache.mode` | `cache.eviction` | ordering | between runs |
+|---|---|---|---|---|
+| *(omitted, default)* | *unchanged* | *unchanged* | back-to-back per query | nothing |
+| `cold` | `sirius` | `lru` | round-robin | drop OS cache **and** `CALL reset_sirius_cache()` |
+| `lukewarm` | `sirius` | `lru` | round-robin | nothing |
+| `hot` | `sirius` | `lru` | back-to-back per query | nothing |
+
+**The config is only rewritten when `--execution` is passed.** Omit it and your YAML is used
```
9prady9, line 230: Could we account for `--pin parquet` and external runners here? In-process parquet pinning rewrites and sanity-checks the config without `--execution`, while `nsys`/`gdb` apply neither rewrite.

test/tpch_performance/CLAUDE.md
```diff
@@ -150,29 +150,127 @@ pixi run python test/tpch_performance/performance_test.py \
...
+exactly as written — no `effective_config.yml` is produced and no sanity check runs.
+
+All three named profiles drop the OS page cache once at startup. `cold` keeps the *connection*
```
9prady9, line 233: Could we qualify the startup page-cache drop as best-effort for `hot` and `lukewarm`? Those profiles continue after a warning when passwordless sudo is unavailable.

test/tpch_performance/CLAUDE.md
```diff
@@ -150,29 +150,127 @@ pixi run python test/tpch_performance/performance_test.py \
...
+- **GPU only.** S3 has no CPU fallback (`src/sirius_context.cpp`,
+  `throw_if_s3_no_cpu_fallback`), so `--engine cpu|both`, `--validation`, and
+  `--pin` are rejected for an `s3://` input. Validate against a local copy of the
+  same data instead.
+- **Credentials must be in the Sirius YAML.** Sirius does not read the
+  environment, AWS profiles, or IMDS (`docs/super-sirius/scan.md`,
+  "Configuration") — put `endpoint` / `region` / `access_key` / `secret_key`
+  (plus `session_token` for temporary credentials) under
+  `sirius.executor.scan_manager.object_store` and pass the file via `--config`.
+  Endpoint must be the regional form `https://s3.<region>.amazonaws.com`.
```
9prady9, line 258: Could we qualify this endpoint form as AWS-specific? Sirius accepts custom `http[s]://host[:port]` S3-compatible endpoints, including MinIO.

test/tpch_performance/CLAUDE.md
```diff
@@ -150,29 +150,127 @@ pixi run python test/tpch_performance/performance_test.py \
...
 - `--validation` — byte-compare GPU vs CPU `result.txt` after timing (with `abs_tol=1e-10` on float columns). Requires `--engine both`.
-- `--mode nsys-profile` — wrap each query in `nsys profile` (one DuckDB CLI subprocess per query; the cudaProfilerApi capture range covers the cold + hot iterations). Requires `--engine gpu`; incompatible with `--validation` and `--duckdb-profiling`.
-- `--query-timeout N` — per-query subprocess timeout in nsys-profile mode (default 90s).
-- `--name <NAME>` — override the auto-timestamped benchmark subdirectory name.
+- `--precmd none|nsys|gdb` — normal in-process execution, per-query Nsight Systems capture, or per-query batch-mode GDB with an automatic all-thread crash backtrace. External runners require `--engine gpu` and are incompatible with `--execution`, `--validation`, and `--duckdb-profiling`.
+- `--query-timeout N` — per-query subprocess timeout for the `nsys` and `gdb` runners (default 90s).
+- `--name <NAME>` — label appended to the benchmark subdirectory name (`tpch_<ts>_<execution>_<engine>_iter<N>_<NAME>`); the run's parameters stay in the name.
```
9prady9, line 273: Could we use the exact `<execution-or-precmd>` template here and carry the append behavior into `.claude/skills/benchmark/SKILL.md`, which still uses `<mode>` and says `--name` replaces the directory?

9prady9 (commented):
Here are the sections that I covered so far, might have to do another pass:
- Scan manager folder and other downstream changes.
- C++ nitpicks
- CMake checks

I won't be touching REST dir as I am not familiar with that section but there are few things codex picked up, please take those comments in REST folder with a grain of salt.

src/io/cache/prefetching_cache.cpp
```diff
@@ -32,234 +33,221 @@
...
+    while (!finished.load(std::memory_order_acquire)) {
+      switch (poll.wait_for_progress()) {
+        case exec::cuda_event_completion_poll::progress::made: break;
+        case exec::cuda_event_completion_poll::progress::none:
+          // Another drainer may have popped our callback and be immediately
+          // about to invoke it outside the lane lock.
+          std::this_thread::yield();
+          break;
+        case exec::cuda_event_completion_poll::progress::undelivered:
+          std::ignore = poll.fail_all(cudaErrorUnknown);
```
wmalpica, line 97: A concrete failure sequence is:

  1. A cache hit pins a chunk’s host buffer.
  2. The cache enqueues an asynchronous host-to-device copy from that buffer.
  3. The copy’s source buffer must remain pinned and unchanged until the CUDA stream reaches and completes the copy. Returning from the enqueue call is not sufficient.
  4. A retirement callback is registered to release the pin once CUDA completion has been observed.
  5. While the copy is still pending, wait_for_progress() detects a CUDA health error on any non-idle lane in the shared completion poll and returns undelivered.
  6. wait_and_finish() responds by calling the poll’s global fail_all().
  7. fail_all() force-invokes every pending retirement callback, including callbacks belonging to healthy streams whose copies may still be running.
  8. Those callbacks call cached_copy_retirement::finish(), which calls release_pins().
  9. The evictor can now see the chunk as unpinned, reclaim its host buffer, and return that buffer to the pool.
  10. The same buffer can be assigned to another chunk and overwritten while the original H2D copy is still reading it. The destination device buffer can therefore receive bytes
     from the wrong file or offset, producing silent corruption.

Suggested direction

  Do not call fail_all() in response to progress::undelivered. The pins must remain held until the specific copy stream has been synchronized or completion has otherwise been
  proven. If completion cannot be proven, retaining or quarantining those buffers is safer than making them eligible for reuse.

  fail_all() should only be used during controlled teardown after new submissions have stopped and all relevant device work is known to have completed or stopped.

src/io/cache/prefetching_cache.cpp
```diff
@@ -273,29 +261,104 @@ prefetching_cache::prefetching_cache(
...
-  _preparation_thread = std::jthread([this](const std::stop_token& st) { prepare_loop(st); },
-                                     _preparation_stop_source.get_token());
-  _prefetch_thread    = std::jthread([this](const std::stop_token& st) { prefetch_loop(st); },
-                                  _prefetch_stop_source.get_token());
-  _evictor_thread     = std::jthread([this](const std::stop_token& st) { evict_loop(st); },
+    _armed(_io_ctx->can_use_prefetching_cache())
+{
+  _chunk_size          = _pool->chunk_size();
+  auto const max_bytes = chunk_state::max_chunk_bytes();
+  if (_chunk_size == 0) {
```
wmalpica, line 268: You should add a check to ensure that _chunk_size is a multiple of IO_BLOCK_SIZE. 

  For suffix reads, needed_fill() converts a byte count into pages using integer division:

  pages = bytes / IO_BLOCK_SIZE;

  src/io/cache/types.hpp:240

  If the chunk size is not page-aligned—for example, 2MB parsed as 2,000,000 bytes—the division rounds down. The resulting backend read can begin slightly after the requested
  range, leaving a gap at its beginning.

  The cache’s completion path nevertheless copies the entire requested range from the staging buffer:

  src/io/cache/prefetching_cache.cpp:742

  The bytes in that gap were never read from the file, so uninitialized staging-buffer contents can be returned as file data. This can cause silent decode corruption rather than
  a clear I/O error.

  A concrete example from the review:

  chunk size:        2,000,000 bytes
  requested range:   [3,000,000, 3,010,000)
  actual file read:  [3,000,576, 4,000,000)
  unread gap:        576 bytes

  The simplest safe fix is to reject cache chunk sizes that are not page-aligned:

  _chunk_size % io::IO_BLOCK_SIZE == 0

src/scan_manager/gatekeeper.hpp
```diff
@@ -0,0 +1,172 @@
...
+  {
+    {
+      std::lock_guard lock{_mutex};
+      _available = _budget;
```
wmalpica: Before opportunistic readahead is armed, ordinary scan reads can still call acquire_or_borrow(), driving the count negative.

  For example, with a budget of 2:

  Initial available count:       0
  Three scan reads start:       -3
  Readahead is armed/reloaded:   2
  Three scan reads finish:       5

  The problem is that reload() assigns _budget directly:

  _available = _budget;

A couple options:
Could make reload() do: `_available += budget;`
Or could have `release()` check against a budget upper limit when increasing _available.

wmalpica (approved):

src/scan_manager/sirius_scan_manager.cpp
ran-yuan-rui: Correction to my earlier note: the budget is now settled only from the ioctxs that serve this query's scan paths (`sirius_scan_manager.cpp:1465–1482`), so a context left over from an earlier query no longer affects it. My remaining concern is the mixed case: when a query reads both object-store and local files, the widest backend's budget is taken before `resolve_readahead()` applies the config and strategy rules, and readahead and foreground debt still share one gatekeeper.

src/io/cache/types.hpp
```diff
@@ -129,407 +149,841 @@ class buffer_pool {
...
     evicting  = 6,
   };
 
-  entry_state() noexcept = default;
+  static constexpr std::uint32_t MAX_PINS        = (1U << 12) - 1;
+  static constexpr std::uint32_t MAX_SUBSCRIBERS = (1U << 16) - 1;
+
+  /// Widest extent the packed field can express, in pages.  The chunk size must
+  /// stay under this many pages; @ref buffer_pool checks it at construction.
+  [[nodiscard]] static constexpr std::uint32_t max_fill_pages() noexcept { return (1U << 14) - 1; }
```
ran-yuan-rui: Thanks, the two sizes from this thread work now: an exact 64 MiB chunk is accepted, and a cache fill is no longer capped at 16 MiB, which covers the 32 MiB case. One thing worth knowing: a staging/chunk-size mismatch is logged and the cache is disabled rather than the config being rejected, so the maximum is still 64 MiB and an unsupported size runs without a cache after an error line.

src/sirius_config.cpp
ran-yuan-rui: This is fixed now: the uring parser records whether `n_max_concurrent_scans` was set explicitly (`sirius_config.cpp:226`) and derivation keeps that value (`:851`) even when it equals the default.

src/io/kvikio/kvikio_context.cpp
ran-yuan-rui: Still open as far as I can tell: `configuration.md:325` and `:407` describe kvikio as local-file only, while `kvikio_context.cpp:137–154` opens `s3://` through `RemoteHandle` with the `object_store` credentials. Either the doc should say S3 is supported and under which build, or this backend should refuse `s3://` routing. The vcpkg port also still builds KvikIO with `KvikIO_REMOTE_SUPPORT=OFF`, which needs enabling or a compile-time guard whichever way the doc goes.

src/io/cache/prefetching_cache.cpp
ran-yuan-rui: Looking at the current head, this still seems open: `wait_and_finish` calls `poll.fail_all()` when the poll reports `undelivered` (`prefetching_cache.cpp:97`) and `finish()` releases the pins before reporting the error (`:106`), while the poll's own comment says `fail_all` may only be called once device work has been stopped (`cuda_event_completion_poll.hpp:520–557`). After the last pin goes the chunk is `cached` again, so it can be evicted and its buffer reused while an H2D copy on a healthy stream is still reading from it. The future does fail, so I'd call it a lifetime violation rather than a proven wrong result.

src/io/cache/prefetching_cache.cpp
ran-yuan-rui: I don't think this landed: the constructor still checks only non-zero and the maximum (`prefetching_cache.cpp:266–274`), and `needed_fill` rounds the suffix down to whole pages from the chunk end (`types.hpp:240–258`). With a 2,000,000-byte chunk and a request for `[3000000, 3010000)` the suffix read starts at 3,000,576, so the first 576 bytes stay unfilled. The staging/chunk equality check doesn't catch it, since both sides can be the same unaligned size.

PR #1774, opened 2026-09-14
feat(exec): add CUDA event completion polling

aminaramoon:
## Description

Components that finish work on CUDA streams need to observe event completion without blocking an executor thread. This adds `cuda_event_completion_poll`, which polls CUDA events and dispatches their completion callbacks through the execution infrastructure.

The accompanying tests cover registration, polling, completion dispatch, shutdown, and error paths.

## Validation

## Checklist

- [x] Cover normal completion and lifecycle edge cases with tests.

## Stack

Layer 4 of 6. Based on `stacked/frugal-native-03-query-events`; followed by #1773.


src/include/exec/cuda_event_completion_poll.hpp
```diff
@@ -0,0 +1,714 @@
...
+    {
+      std::lock_guard g(pending_m_);
+      for (auto& f : staged) {
+        pending_.push_back({t, std::move(f)});
+      }
+      refresh_hint_locked();
+    }
+
+    // flags must be 0.
+    const cudaError_t e = cudaStreamAddCallback(stream_, &detail::bump_ticket, &slot_, 0);
```
kevkrist: Just an FYI, but the [docs](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1g74aa9f4b1c2f12d994bf13876a5a2498) say this function is slated for deprecation and removal.

kevkrist (dismissed):
This is nice

src/include/exec/cuda_event_completion_poll.hpp
```diff
@@ -0,0 +1,714 @@
...
+  }
+
+  enum class progress {
+    made,        // the frontier advanced; retry
+    none,        // nothing outstanding anywhere; waiting would deadlock
+    undelivered  // callbacks cannot be delivered; nothing will ever complete
+  };
+
+  // Blocks on the globally oldest outstanding ticket.
+  progress wait_for_progress() noexcept
```
kevkrist, line 520: What if lane A has ticket 1 pending, lane B has ticket 100 pending, acquire() calls drain_all(), and then wait_for_progress() sees A as the oldest ticket, and waits for A. Meanwhile, lane B's ticket 100 work has finished. acquire() will continue to wait on lane A ticket 1, no? Even though ticket 100 is finished? It's possible I'm misunderstanding but cross-stream + single ticket counter seems to break the ordering invariant on the ticket number.

src/include/exec/cuda_event_completion_poll.hpp
```diff
@@ -0,0 +1,714 @@
...
+
+    // Ordered after the acquire above, so a failure published before the
+    // frontier moved is visible here.
+    const std::uint64_t bad = slot_.first_failed.load(std::memory_order_relaxed);
+    const cudaError_t err   = slot_.first_error.load(std::memory_order_relaxed);
+
+    std::vector<detail::pending_entry> ready;
+    {
+      std::lock_guard g(pending_m_);
+      while (!pending_.empty() && pending_.front().ticket <= done) {
```
kevkrist: It seems that releasing the pending_m_ lock before invoking the completion callbacks can result in the callbacks getting called out of order. If thread A grabs the callbacks through drain(), then so does thread B, but thread B for whatever reason invokes its callbacks first. Is that even an issue? I can't tell.

src/include/exec/cuda_event_completion_poll.hpp
aminaramoon: I know, there is no good alternative since the replacement doesnt call the host callback on error. there is no deprecation date for this, so we will fix it when we get there

src/include/exec/cuda_event_completion_poll.hpp
9prady9, line 1: Could we shorten the file-level overview and document the public types and operations directly at their declarations? Each type should explain its responsibility, ownership, and lifetime requirements; methods should state their blocking, threading, and error behavior. The detailed CUDA rationale can sit beside the relevant implementation. Currently, understanding how to use the API requires piecing together the opening contract block and implementation comments.

9prady9 (changes requested):
Requesting docs update.

On another note, it would be nice to see its usage to better understand intended use case . The current design tries to give good amount of guarantees like enabling simultaneous drains etc. but are those all needed ? Which PR in stack has this being used ?

src/include/exec/cuda_event_completion_poll.hpp
```diff
@@ -0,0 +1,527 @@
...
+    const auto count    = count_.load(std::memory_order_acquire);
+    for (std::size_t i = 0; i < count; ++i) {
+      retired += lanes_[i]->fail_all(error);
+    }
+    return retired;
+  }
+  template <class TryFn>
+  auto acquire(TryFn&& try_fn) -> std::invoke_result_t<TryFn&>
+  {
+    if (auto result = try_fn()) { return result; }
```
9prady9, line 572: Please add a C++20 constraint requiring `acquire()`'s callable to return a non-reference value that supports a boolean test, preventing dangling references to the local `result`.

src/include/exec/cuda_event_completion_poll.hpp
```diff
@@ -0,0 +1,527 @@
...
+
+  [[nodiscard]] cudaStream_t stream() const noexcept;
+
+  // Stage before enqueuing device work, including construction of fn.
+  void on_retire(retire_fn fn);
+  // Idempotent. Installation failure closes the lane. A fallback sync error
+  // takes precedence in the return value; enqueue_error() retains the cause.
+  // Unconfirmed functions remain queued; successful recovery makes them
+  // drainable with the installation error (never inline under the submit lock).
+  cudaError_t commit() noexcept;
```
9prady9: mark `commit()` and `quiesce()` `[[nodiscard]]` so callers are warned when they ignore a failure ?

src/include/exec/cuda_event_completion_poll.hpp
```diff
@@ -0,0 +1,527 @@
...
+  std::uint64_t active_ticket_{no_pending_v};
+  std::atomic<std::uint64_t> oldest_hint_{no_pending_v};
+  std::atomic_flag draining_ = ATOMIC_FLAG_INIT;
+};
+
+inline cudaStream_t submission::stream() const noexcept { return lane_->stream_; }
+inline void submission::on_retire(retire_fn fn)
+{
+  if (committed_) { throw std::logic_error("on_retire after commit"); }
+  staged_.push_back({0, std::move(fn)});
```
9prady9, line 429: we should reject an empty `retire_fn` in `on_retire()` before queueing it, since invoking an empty `absl::AnyInvocable` is undefined behavior, isn't it ?

9prady9 (commented):
few c++ comments

src/include/exec/cuda_event_completion_poll.hpp
```diff
@@ -0,0 +1,527 @@
...
+  }
+  void remember_host_error(cudaError_t error) noexcept
+  {
+    auto expected = cudaSuccess;
+    host_error_.compare_exchange_strong(expected, error, std::memory_order_release);
+  }
+  void claim_drainer() noexcept
+  {
+    while (draining_.test_and_set(std::memory_order_acquire)) {
+      std::this_thread::yield();
```
9prady9: nit: why not wait ? instead of busy loop

9prady9:
```diff
--- a/src/include/exec/cuda_event_completion_poll.hpp
+++ b/src/include/exec/cuda_event_completion_poll.hpp
@@ -176,7 +176,7 @@
     if (done < oldest_hint_.load(std::memory_order_relaxed)) { return 0; }
     if (draining_.test_and_set(std::memory_order_acquire)) { return 0; }
     const auto count = drain_owned(done, false, cudaSuccess);
-    draining_.clear(std::memory_order_release);
+    release_drainer();
     return count;
   }
 
@@ -258,7 +258,7 @@
     claim_drainer();
     const auto count =
       drain_owned(no_pending_v, true, error == cudaSuccess ? cudaErrorUnknown : error);
-    draining_.clear(std::memory_order_release);
+    release_drainer();
     return count;
   }
 
@@ -279,7 +279,7 @@
     }
     claim_drainer();
     drain_owned(completed(), false, cudaSuccess);
-    draining_.clear(std::memory_order_release);
+    release_drainer();
     if (result != cudaSuccess) { return result; }
     return idle() ? cudaSuccess : cudaErrorNotReady;
   }
@@ -299,8 +299,13 @@
   void claim_drainer() noexcept
   {
     while (draining_.test_and_set(std::memory_order_acquire)) {
-      std::this_thread::yield();
-    }
+      draining_.wait(true, std::memory_order_relaxed);
+    }
+  }
+  void release_drainer() noexcept
+  {
+    draining_.clear(std::memory_order_release);
+    draining_.notify_one();
   }
   std::size_t drain_owned(std::uint64_t done, bool force, cudaError_t error) noexcept
  ```

9prady9 (commented):
last set of c++ comments

kevkrist (approved):

PR #1772, opened 2026-09-14
build: make CTrack dependency opt-in

aminaramoon:
## Description

CTrack was fetched and linked unconditionally, which made a profiling-only dependency part of every build. This adds `BUILD_WITH_CTRACK`, defaulting to `OFF`, and gates the CTrack fetch, link dependency, and compile definition behind that option so normal builds do not require CTrack.

## Validation

- `pixi run pre-commit run -a` passes at this commit.
- The final stacked tree passes the complete release build.

## Checklist

- [x] Read `CONTRIBUTING.md` and meet the PR reviewability requirements.
- [x] Preserve the existing CTrack path when `BUILD_WITH_CTRACK=ON`.
- [x] Document the new build option next to its CMake definition.

## Stack

Layer 2 of 6. Based on `stacked/frugal-native-01-rapids`; followed by #1776.

PR #1775, opened 2026-09-14
build(deps): bump RAPIDS to 26.08 and migrate off <cuda/stream_ref>

aminaramoon:
## Description

The hybrid-scan multi-file reader requires the RAPIDS 26.08 APIs, while CCCL has moved the stream reference API out of the deprecated `<cuda/stream_ref>` header. This updates libcudf, libcuvs, and libkvikio to 26.08.01, migrates the affected dynamic-filter kernels and cuco allocator to `<cuda/stream>`, and removes kvikIO's `cuda_std_17` interface feature where it would otherwise break CMake generation for the C/C++-only DuckDB target.

## Validation

## Checklist

- [x] Use existing build coverage for dependency and header migration.

## Stack

Layer 1 of 6. Based on `dev`; followed by #1772.


mike-wendt (approved):
LGTM, re-ran failing test-run job to see if it [fails](https://github.com/sirius-db/sirius/actions/runs/34894984816/job/104149860593?pr=1775) again

mbrobbel:
Manually running [Distribution](https://github.com/sirius-db/sirius/actions/runs/34944785893) to check vcpkg updates.

src/cuda/sirius_dynamic_bloom_filter.cu
```diff
@@ -63,61 +61,11 @@ std::size_t blocks_for(std::size_t num_keys)
...
- private:
-  hasher hash_{};
-};
-
 template <class KeyT>
 using sirius_bloom = cuco::bloom_filter<KeyT,
                                         cuco::extent<std::size_t>,
                                         cuda::thread_scope_device,
-                                        sirius_bloom_policy<KeyT>,
+                                        cuco::default_filter_policy<KeyT>,
```
9prady9, line 68: These are not interchangeable filters right, are we getting better results/perf using this vs the older one ?

vcpkg_ports/nvjitlink/nvjitlink-config.cmake
```diff
@@ -0,0 +1,13 @@
...
+
+if(NOT TARGET nvjitlink::nvjitlink_static)
+  add_library(nvjitlink::nvjitlink_static STATIC IMPORTED)
+  set_target_properties(
+    nvjitlink::nvjitlink_static
+    PROPERTIES IMPORTED_LOCATION
+               "${CMAKE_CURRENT_LIST_DIR}/../../lib/libnvJitLink_static.a"
+               INTERFACE_INCLUDE_DIRECTORIES
+               "${CMAKE_CURRENT_LIST_DIR}/../../include"
+               INTERFACE_LINK_LIBRARIES "${CMAKE_DL_LIBS};Threads::Threads")
```
9prady9: `CUDA::nvJitLink_static` target—also require `CUDA::nvptxcompiler_static` if I am not wrong - a hidden dependency in this case. Not an issue for sirius , but would be if this is up-streamed outside of sirius.

9prady9 (approved):
lgtm, couple of minor nits I suppose.

I understand the vcpkg port files were introduced in some earlier PR. But we should consider up-streaming them. But I do get the appeal for having them here if they aren't used by anyone else. Something we should look into later I suppose.

src/cuda/sirius_dynamic_bloom_filter.cu
mbrobbel: Yes, it seems this new policy is better (faster construction + lower false positive rate). [Temp benchmark result](https://github.com/sirius-db/sirius/actions/runs/34965633945/job/104370738010#step:8:86).

9prady9 (approved):

vcpkg_ports/nvjitlink/nvjitlink-config.cmake
mbrobbel: 43bcccf655969aecbda57adb29c871a39105093e

mbrobbel:
Re-running [distribution](https://github.com/sirius-db/sirius/actions/runs/34969532661).

vcpkg_ports/nvjitlink/nvjitlink-config.cmake
9prady9: 👍🏾 saw it already :)

mike-wendt:
> Manually running [Distribution](https://github.com/sirius-db/sirius/actions/runs/34944785893) to check vcpkg updates.

Good catch @mbrobbel

mbrobbel (approved):
> Re-running [distribution](https://github.com/sirius-db/sirius/actions/runs/34969532661).

✅

PR #1776, opened 2026-09-14
feat(event): add query event publisher

aminaramoon:
## Description

Query lifecycle notifications were coupled to their producers, making it difficult for readahead and other services to observe execution without adding direct dependencies. This introduces a query event publisher and hook-driven subscribers, stamps publications with process-wide event IDs and system-clock timestamps, and adds the thread utilities used to dispatch subscriber work safely.

Tests cover event publication, subscriber behavior, ordering metadata, and the supporting thread utilities.

## Validation

## Checklist

- [x] Cover the publisher, subscribers, and thread utilities with tests.

## Stack

Layer 3 of 6. Based on `stacked/frugal-native-02-ctrack`; followed by #1774.


src/event/query_event_subscriber.cpp
```diff
@@ -0,0 +1,216 @@
...
+  _worker = std::jthread([this] { run(); });
+  set_thread_name();
+}
+
+void query_event_subscriber::stop() noexcept
+{
+  // Latch first so a concurrent @ref start (racing with our teardown from a
+  // different thread) turns into a no-op rather than resurrecting the worker
+  // we are about to join.
+  _stopped.store(true, std::memory_order_relaxed);
```
wmalpica, line 98: You can still end up with a race condition here. You could have start go past where it checks `_stopped` and then stop can run and finish running before start ends causing a race condition.

The _stopped/_draining atomics coordinate the logical start/stop
decision but do not protect _worker/_stop_cb themselves — any concurrent access to those two
members from start() and stop() is a data race independent of what the atomics decide. Replace
the two-flag coordination with a single mutex guarding the joinable()-check-then-mutate sequence in
both functions (or a CAS-based state machine over an enum: idle / running / stopped) so only
one thread ever touches _worker/_stop_cb at a time

johanpel (commented):
Just a high-level question: could this be built by composing existing queue and channel abstractions such as `exec::channel`, `interruptible_mpmc`, and `scoped_dispatcher` or would these not cover the required behavior?

src/event/query_event_subscriber.cpp
```diff
@@ -0,0 +1,216 @@
...
+  // The sentinel wakes a worker parked on an empty mailbox; the flag stops
+  // it from working through whatever is queued behind the sentinel.
+  // A stale sentinel left in the queue would kill a restart, but subscribers
+  // are one-shot: once stopped, @ref start is a no-op, so there is no next
+  // worker to be tripped by it.
+  _draining.store(false, std::memory_order_relaxed);
+  std::ignore = _queue->try_enqueue(nullptr);  // see query_event_publisher::stop
+  _worker.join();
+  _stop_cb.reset();
+}
```
ran-yuan-rui, line 113: `stop()` permanently disables the worker but leaves its queue registered until destruction. A stopped subscriber that remains alive keeps retaining matching events with no configured bound. The replay explanation applies before the first `start()`, but there is no restart after `stop()`.

Would you consider unregistering in `stop()` to prevent further enqueues? If retaining the registration is intentional, a bounded queue with an explicit drop policy would limit the backlog.

src/include/event/query_event_publisher.hpp
```diff
@@ -0,0 +1,203 @@
...
+      if (subscribers.empty()) { return; }
+      auto payload = std::make_shared<query_events>(
+        Event{event_id, timestamp, typename Event::param_type{std::forward<Args>(args)...}});
+      for (auto* q : subscribers) {
+        q->enqueue(payload);
+      }
+    } catch (...) {  // NOLINT(bugprone-empty-catch)
+      // Telemetry is not worth failing execution over.
+    }
+  }
```
ran-yuan-rui, line 196: The header promises publication-order delivery, but concurrent reporters can deliver IDs out of order: A takes ID 1 and pauses; B takes ID 2 and delivers it before A resumes.

Moving `fetch_add` inside the shared lock would not fix this: publishers still overlap, and moodycamel does not guarantee global FIFO across producers. Could we document cross-reporter order as unspecified? If ordered delivery is required, serialized publication with a shared producer token per mailbox would preserve it. The concurrency test currently checks receipt, not order.

src/include/event/query_event_publisher.hpp
```diff
@@ -0,0 +1,203 @@
...
+ * therefore a push and never a callback: a reporter is never made to wait on
+ * what a subscriber does with the event, which matters because these are raised
+ * from the creator, scheduler and executor hot paths.  Nothing here is virtual:
+ * the publisher is the fixed relay and the subscriber is the extension point.
+ *
+ * Registration is not a public entry point of this class.  A @ref
+ * query_event_subscriber registers itself in its own constructor and drops
+ * itself in its own destructor, and nothing else needs to touch the routing
+ * table --- so those two operations are private and only that friend can reach
+ * them.  Callers hold a subscriber, not a mailbox.
```
9prady9, line 81: Could we replace the opening description through the registration paragraph with this tighter contract? This describes the class as a router, makes the asynchronous shared-payload behavior precise, and leaves the `Lifetime` and `Thread safety` sections below intact.

```suggestion
/**
 * @brief Routes execution events asynchronously to registered subscribers.
 *
 * When at least one matching subscriber exists, a @c publish_* call creates one
 * event payload and enqueues the shared payload in every matching subscriber's
 * mailbox. Subscriber hooks run on their own worker threads; publishing never
 * invokes them synchronously.
 *
 * A @ref query_event_subscriber registers its mailbox during construction and
 * unregisters it during destruction. Callers own subscribers rather than
 * interacting with mailboxes directly.
```

src/include/event/query_event_publisher.hpp
```diff
@@ -0,0 +1,203 @@
...
+  ~query_event_publisher() = default;
+
+  query_event_publisher(query_event_publisher const&)            = delete;
+  query_event_publisher& operator=(query_event_publisher const&) = delete;
+
+  /// Stop every subscriber and close the publisher to further publishing.
+  /// Requests stop on the shared token and pushes the null sentinel to each
+  /// queue, which is what wakes a subscriber parked on an empty mailbox;
+  /// subsequent @c publish_* calls are no-ops.  Does not join the subscribers:
+  /// a subscriber owns its own thread and joins it in its own destructor.
```
9prady9: Could we keep only contract level info and move impl semantics i.e. the stop-token/sentinel mechanism to the cpp file code comments ? unless you think they add to user understanding while consuming this API  

```suggestion
  /// Request all subscribers to stop and ignore subsequent @c publish_* calls.
  /// Does not join subscriber threads; each subscriber owns and joins its worker.
```

src/include/event/query_event_publisher.hpp
```diff
@@ -0,0 +1,203 @@
...
+  ///
+  /// The empty check ahead of the allocation is not just an optimisation: with
+  /// nobody listening these entry points sit on hot paths and should cost a
+  /// lock and a branch, not a heap allocation.  Publishing is @c noexcept, so a
+  /// failed allocation drops the event rather than propagating out into a
+  /// reporter that has no way to handle it.
+  ///
+  /// The template body stays in the header even though nothing else does: the
+  /// @c publish_* wrappers in the .cpp instantiate it, and moving it out would
+  /// mean spelling every instantiation there by hand.
```
9prady9, line 178:
```suggestion
```
We can drop such impl info here.

src/include/event/query_event_publisher.hpp
```diff
@@ -0,0 +1,203 @@
...
+  /// Publishing walks one subscriber list per event, so a subscriber that
+  /// names two events is skipped outright by the other six rather than being
+  /// handed a payload it will drop.  With a handful of subscribers that is a
+  /// rounding error; it is worth having because the cost grows with
+  /// subscribers x events while the useful work does not.
+  ///
+  /// Registering after events have started flowing is safe; the subscriber
+  /// simply misses what was published before it arrived.  Registering after
+  /// @ref stop yields an already-stopped token and a queue nothing will ever
+  /// be pushed to, so the subscriber never gets past a no-op @c start.
```
9prady9: I would say impl context should be as part of code comments rather than in the doc strings.

```suggestion
  /// Register a mailbox for @p events without replaying earlier publications.
  /// If the publisher is already stopped, returns an unregistered mailbox and
  /// an already-requested stop token.
  ///
  /// Called by @ref query_event_subscriber during construction.
```

src/include/event/query_event_publisher.hpp
```diff
@@ -0,0 +1,203 @@
...
+  [[nodiscard]] subscriber_registration register_subscriber(std::span<event_type const> events);
+
+  /// Drop @p queue's registration.  Blocks until any in-flight publish has
+  /// finished, so no further event reaches @p queue once this returns.  Called
+  /// by the subscriber base in its destructor.
```
9prady9: nit: rephrase makes it more clear I think ?

```suggestion
  /// Remove @p queue from routing, synchronizing with in-flight publications.
  /// After this returns, the publisher will not enqueue another event to it.
  /// Called by @ref query_event_subscriber during destruction.
```

src/include/event/query_event_publisher.hpp
9prady9: Also this isn't true, we can the declaration here in hpp and have implementation inside cpp before all concrete publish calls. That will take care of the instantiations just fine because `publish_*` implementations are all under `.cpp` file.

src/include/event/query_event_publisher.hpp
9prady9, line 196: Since every instantiation is in `query_event_publisher.cpp`, can we keep only the declaration here and move the definition above the concrete `publish_*` wrappers? The definition will be visible at every point of instantiation, so no explicit-instantiation list is needed. This also removes the inaccurate rationale below and keeps the implementation out of the public header.

```suggestion
  /// Build an event once and enqueue it for each matching subscriber.
  template <typename Event, typename... Args>
  void publish(Args&&... args) noexcept;
```

src/include/event/query_event_publisher.hpp
```diff
@@ -0,0 +1,203 @@
...
+ * subscriber alive.
+ *
+ * Thread safety: the queue set is guarded by a shared mutex.  Publishing takes
+ * it shared, so the reporting threads do not serialise against each other;
+ * registration and deregistration take it exclusively.
+ */
+class query_event_publisher : public std::enable_shared_from_this<query_event_publisher> {
+ public:
+  query_event_publisher()  = default;
+  ~query_event_publisher() = default;
```
9prady9, line 96: Should `~query_event_publisher()` call `stop()` to stop any surviving subscribers?

src/event/query_event_publisher.cpp
```diff
@@ -0,0 +1,145 @@
...
+subscriber_registration query_event_publisher::register_subscriber(
+  std::span<event_type const> events)
+{
+  auto queue = std::make_shared<event_queue>();
+  {
+    std::unique_lock g{_queues_mtx};
+    if (!_stopped) {
+      _queues.push_back(queue);
+      for (auto e : events) {
+        auto& bucket = _by_event[static_cast<std::size_t>(e)];
```
9prady9: very nit:unnamed value can index past the array.

src/event/query_event_publisher.cpp
```diff
@@ -0,0 +1,145 @@
...
+        // A repeated event would double-deliver, and a caller assembling the
+        // list from overlapping sets should not have to care.
+        if (std::ranges::find(bucket, queue.get()) == bucket.end()) {
+          bucket.push_back(queue.get());
```
9prady9: Registration is partially committed before all vector growth succeeds. If a later `bucket.push_back()` throws, subscriber construction unwinds and never unregisters, while the publisher retains the queue and any earlier bucket entries. That ghost mailbox can then accumulate events indefinitely. Please give this operation a strong exception guarantee—for example, validate/deduplicate the indexes and reserve `_queues` plus every affected bucket before performing any insertion, or add rollback.

src/event/query_event_publisher.cpp
```diff
@@ -0,0 +1,145 @@
...
+    _stopped = true;
+    queues.swap(_queues);
+    for (auto& bucket : _by_event) {
+      bucket.clear();
+    }
+  }
+  for (auto const& q : queues) {
+    // try_ so this stays genuinely noexcept: the allocating enqueue could
+    // throw, and a subscriber that misses the sentinel still comes out on the
+    // stop token within a poll interval.  The sentinel only buys promptness.
```
9prady9: The queue API reports allocation failure through `false`, and `try_enqueue` itself is not declared `noexcept`. Could we describe the actual best-effort wake-up contract instead?

```suggestion
    // Best-effort wake-up for a worker blocked on an empty queue. If this
    // fails, the worker observes the stop token at its next timed poll.
```

src/include/event/query_event_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+                                           std::size_t operator_id,
+                                           int gpu_id,
+                                           std::size_t bytes_needed) noexcept;
+
+ protected:
+  /// Invoked once when the publisher requests stop, on the thread that called
+  /// @ref query_event_publisher::stop --- not on the worker, which may still
+  /// be finishing an event.  The hook for tearing down whatever the subscriber
+  /// was driving; the worker's own exit needs no help.
+  virtual void on_stop_requested() noexcept;
```
9prady9: This callback is registered only by `start()`, so it is never called for a subscriber that was not started. It can also run on the `start()` thread if publisher shutdown occurs just before `_stop_cb.emplace()`. Could we avoid promising exactly once or a specific thread?

```suggestion
  /// Invoked when a started subscriber observes a publisher stop request.
  /// May run concurrently with event hooks.
  virtual void on_stop_requested() noexcept;
```

src/include/event/query_event_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+  /// Stop the worker and join it.  Terminal: subsequent @ref start calls are
+  /// no-ops.  Does NOT drop the registration --- that happens once, at
+  /// destruction.  Safe to call when not started, and safe to call twice.
+  /// Does not stop the publisher.
+  void stop() noexcept;
+
+  /// Whether the worker is up.  Named apart from any @c is_running a subclass
+  /// has for its own work, which is a different question.
+  [[nodiscard]] bool is_subscribed() const noexcept { return _worker.joinable(); }
```
9prady9: `std::jthread::joinable()` stays true after `run()` returns and only changes after `join()`, so this can report a stopped worker as up. Could we track worker liveness explicitly or rename this API?

src/include/event/query_event_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+  ///
+  /// Anything published between construction and this call is still in the
+  /// mailbox and gets replayed in order.
+  void start();
+
+  /// Stop the worker and join it.  Terminal: subsequent @ref start calls are
+  /// no-ops.  Does NOT drop the registration --- that happens once, at
+  /// destruction.  Safe to call when not started, and safe to call twice.
+  /// Does not stop the publisher.
+  void stop() noexcept;
```
9prady9, line 104: Can `stop()` be called from an event hook? That would join the current `_worker`; `join()` throws, and this `noexcept` function terminates.

aminaramoon: it should not be called from an event hook.

wmalpica (approved):

PR #1799, opened 2026-09-15
fix(simpatico): align CCCL headers with libcudf

9prady9:
## Description

- Use libcudf's CCCL headers and compile definitions in Simpatico.
- Include the embedded CCCL header fingerprint in JIT cache keys to prevent stale cubin reuse after header changes.

## Testing

- [x] Release build with CUDA 13.2 and libcudf 26.08.01
- [x] Standalone Simpatico configuration
- [x] `test_jit_kernel_cache` on GPU
- [x] SF50 TPC-H Q1
- [x] Pre-commit hooks for changed files


src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -73,6 +59,111 @@ else()
...
+       AND EXISTS "${_simpatico_cccl_candidate}/thrust/version.h")
+      list(APPEND _simpatico_cccl_candidates "${_simpatico_cccl_candidate}")
+    else()
+      list(APPEND _simpatico_cccl_rejected "${raw_include}")
+    endif()
+  endforeach()
+
+  list(REMOVE_DUPLICATES _simpatico_cccl_candidates)
+  list(LENGTH _simpatico_cccl_candidates _simpatico_cccl_candidate_count)
+  if(NOT _simpatico_cccl_candidate_count EQUAL 1)
```
joosthooz: This might not play nice with what gets fetched by CPM/FetchContent, and if there are multiple candidates

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -178,13 +272,35 @@ if(NOT DEFINED SIRIUS_NVTX_DOMAIN_NAME)
...
+elseif(SIMPATICO_CCCL_INCLUDE_DIR)
+  message(
+    WARNING
+      "CCCL::CCCL is unavailable; the selected CCCL headers cannot inherit "
+      "libcudf's exported compile definitions. This is only expected for a "
+      "manual SIMPATICO_CCCL_INCLUDE_DIR override without a cudf CMake package."
+  )
+else()
+  message(
+    FATAL_ERROR
```
joosthooz: Does our vcpkg build work without triggering this fatal error? We should at least try out manually

src/compression/simpatico_codegen/src/jit/kernel_cache.cpp
```diff
@@ -207,7 +239,7 @@ const CompiledKernel* KernelCache::get_or_compile_plain(const std::string& sourc
   const std::string& cdir = disk_cache_dir();
   std::string path;
   if (!cdir.empty()) {
-    path = cubin_path_for(key_material, opts.arch_cc);
+    path = cubin_path_for(cubin_key_material_from(key_material), opts.arch_cc);
```
joosthooz, line 242: Nit: this copies the whole source, maybe this can be done cheaper

joosthooz (changes requested):
Thanks Pradeep! Small change request: Can we also hash `kEmbeddedJitHeaders` so that a change to our own headers will also be detected?

And now we're working on this code, is there any way we can delete cubins that we know will no longer be needed? I don't mind if we don't tackle this here, or we can just leave it up to the user

mbrobbel (commented):
This looks like something we should fix properly in a separate PR. Simpatico should just use the imported CMake targets discovered through `find_package`, rather than reconstructing their include paths and compile definitions manually.
Then we get the `CCCL` dependency via `libcudf` and avoid this confusing resolution logic.

9prady9:
> This looks like something we should fix properly in a separate PR. Simpatico should just use the imported CMake targets discovered through `find_package`, rather than reconstructing their include paths and compile definitions manually. Then we get the `CCCL` dependency via `libcudf` and avoid this confusing resolution logic.

That would ideal @joosthooz let me try that unless you faced issues trying to use existing headers in your experiments ? can you please confirm

9prady9:
@joosthooz @mbrobbel CCCL dependency is better resolved now.

About JIT caching, I have some ideas from my previous work but lets do that in a followup as it is unrelated change.

9prady9:
> @joosthooz @mbrobbel CCCL dependency is better resolved now.
> 
> About JIT caching, I have some ideas from my previous work but lets do that in a followup as it is unrelated change.

I also found that CUDA 13.3 helps in the source accumulation logic for forwarding to nvrtc in much better way, once we bump up to CUDA 13.3 this will reduce this code cruft a lot - note that it only removes content inclusion logic from our code base but it is still required for runtime compilation.

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -165,7 +123,8 @@ add_library(
   src/simpatico_codegen.cpp
   src/api/compressed_table_io.cpp
   src/explore/compression_explorer.cpp)
-set_target_properties(simpatico PROPERTIES POSITION_INDEPENDENT_CODE ON)
+set_target_properties(simpatico PROPERTIES POSITION_INDEPENDENT_CODE ON
+                                           NO_SYSTEM_FROM_IMPORTED ON)
```
felipeblazing, line 130: **Reviewer: Build, deps, CI & repo hygiene**

**nit** NO_SYSTEM_FROM_IMPORTED added without a reason

Every other use in the root CMakeLists carries a comment explaining why the -isystem demotion is needed. Here it silently turns cudf/rmm/CCCL headers into `-I`, so their warnings now show up in simpatico TUs. If the goal is to keep libcudf's CCCL ahead of the toolkit's implicit CCCL, say so in a comment.

felipeblazing (commented):
These are all AI-generated but human reviewed comments.

---

**`src/compression/simpatico_codegen/CMakeLists.txt`**

**Reviewer: Comments, docs & AI residue**

**nit** CMake comment's first sentence doesn't say why `file(GENERATE)` is needed

"CMake evaluates transitive usage requirements and generator expressions" is generic. The real reason is that `CCCL::CCCL`'s include dirs are only known at generate time, and the `-P` script can't evaluate them. Say that in one sentence.
```suggestion
# CCCL::CCCL's (transitive) include dirs are only known at generate time, so
```

---

**`src/compression/simpatico_codegen/include/codegen/jit/cccl_embedded_headers.h`**

**Reviewer: Comments, docs & AI residue**

**minor** Fingerprint doc doesn't state its format

The test requires the fingerprint to be exactly 64 lowercase hex characters, and the README calls it SHA-256. The header only says "deterministic digest". Put that contract in the header and cut the consumer explanation down to one clause.
```suggestion
// Lowercase hex SHA-256 (64 chars) over the names and contents of the embedded
// CCCL header closure; part of the JIT cache key.
```

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -182,52 +141,40 @@ target_compile_definitions(
 add_dependencies(simpatico simpatico_embedded_headers)
 target_include_directories(
   simpatico
-  PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src ${CMAKE_CURRENT_SOURCE_DIR}/include
-          "${CMAKE_CURRENT_BINARY_DIR}/generated" "${CUDA_TOOLKIT_INCLUDE}"
-          "${CUDA_CCCL_INCLUDE}")
+  PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include
```
felipeblazing: **Reviewer: Build, deps, CI & repo hygiene**

**nit** PUBLIC include dir makes the root CMakeLists' manual simpatico include redundant

`simpatico` now exports `include/` as PUBLIC. The root `CMakeLists.txt` still adds `$<BUILD_INTERFACE:.../simpatico_codegen/include>` by hand to targets that already link `simpatico`: around line 625 for the extension targets and line 1109 for the late-mat target. Those are now dead include directories. Drop them in this PR, or in a follow-up.

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -182,52 +141,40 @@ target_compile_definitions(
...
+  PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src
+          "${CMAKE_CURRENT_BINARY_DIR}/generated")
 # In the Sirius vcpkg build the overlay nvrtc port provides a static NVRTC
 # (nvrtc::nvrtc_static) so the extension carries no libnvrtc.so runtime dep.
-# Standalone / pixi builds fall back to the conda shared 'nvrtc'.
+# Standalone / pixi builds use the toolkit's shared NVRTC target.
+if(VCPKG_BUILD AND NOT TARGET nvrtc::nvrtc_static)
```
felipeblazing, line 152: **Reviewer: Build, deps, CI & repo hygiene**

**nit** Redundant nvrtc find_package in the vcpkg branch

The root CMakeLists already runs `find_package(nvrtc CONFIG REQUIRED)` under `VCPKG_BUILD` before `add_subdirectory`, so `nvrtc::nvrtc_static` always exists by this point. This block only matters for a standalone vcpkg configure of simpatico, which nothing does. Drop it, or say in a comment that it is for that case.

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -182,52 +141,40 @@ target_compile_definitions(
...
+  find_package(nvrtc REQUIRED CONFIG)
+endif()
 if(TARGET nvrtc::nvrtc_static)
   set(SIMPATICO_NVRTC_LIB nvrtc::nvrtc_static)
 else()
-  set(SIMPATICO_NVRTC_LIB nvrtc)
+  set(SIMPATICO_NVRTC_LIB CUDA::nvrtc)
```
felipeblazing, line 159: **Reviewer: API design, layering & reuse**

**nit** Root CMakeLists NVRTC comment is now stale

Root `CMakeLists.txt:121-124` still says simpatico links the "bare 'nvrtc' library name" and that pixi uses "the conda shared libnvrtc.so". This PR changes that fallback to `CUDA::nvrtc`. Update that comment to match.

src/compression/simpatico_codegen/README.md
```diff
@@ -193,9 +193,10 @@ is compiled only once. The beam-search explorer alone enumerates hundreds of tho
 candidate shapes, so this cache is what makes runtime codegen practical.
 
 There are two levels, both keyed by the same tuple: a 64-bit FNV-1a digest of the rendered
-source (plus the kernel entry symbol), the GPU architecture (`sm_XX`), the CUDA runtime
-version, and the driver version. Keying on all four means a renderer change, a different
-GPU, or a toolchain upgrade can never produce a false hit.
+source (plus the kernel entry symbol), a SHA-256 fingerprint of the embedded CCCL header
+closure, the GPU architecture (`sm_XX`), the CUDA runtime version, and the driver version.
+This ensures a renderer or embedded CCCL change, a different GPU, or a toolchain upgrade
```
felipeblazing: **Reviewer: Comments, docs & AI residue**

**major** README says the cache cannot produce a false hit, but the embedded project headers aren't part of the key

The cache key covers the rendered source and the CCCL closure. It does not cover `kEmbeddedJitHeaders` (`rle_block.cuh`, `stdint_shim.hpp`), which NVRTC also compiles against. If one of those headers changes, the old on-disk cubin is still reused. joosthooz asked for this in their CHANGES_REQUESTED review and it is still missing at the head. Either hash those headers into the fingerprint, or narrow this sentence and the matching comments in `kernel_cache.cpp:68` and `cccl_embedded_headers.h`.

src/compression/simpatico_codegen/README.md
```diff
@@ -229,7 +236,7 @@ As a result the runtime JIT needs **no CCCL/CUDA headers on disk** — only the
 
 | Variable | Effect |
 | --- | --- |
-| `SIMPATICO_JIT_CCCL_INCLUDE` | Optional escape hatch. The CCCL headers are embedded in the binary, so this is normally unset. If a future renderer change needs a header the embedded closure lacks, set this to a CCCL dir (containing `cuda/std/cstdint`) and it is passed to NVRTC as an extra `-I`. |
+| `SIMPATICO_JIT_CCCL_INCLUDE` | Optional escape hatch. The CCCL headers are embedded in the binary, so this is normally unset. If a future renderer change needs a header the embedded closure lacks, set this to a CCCL dir (containing `cuda/std/cstdint`) and it is passed to NVRTC as an extra `-I`. The external tree is not covered by the embedded-header fingerprint; restart the process and disable or change `SIMPATICO_JIT_CACHE_DIR` after editing it. |
```
felipeblazing: **Reviewer: Error handling & fail-closed gating**

**minor** SIMPATICO_JIT_CCCL_INCLUDE cubins are cached under the embedded-closure key

When the override is set, NVRTC compiles against an external tree, but the cubin is written under the embedded fingerprint. Once the variable is unset or the tree changes, later processes load that cubin as if it came from the embedded headers. A README note asking users to clear the cache by hand is not a safeguard. Either add the override path to `cubin_key_material_from`, or skip the disk read/write while the override is set.

src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake
```diff
@@ -15,10 +16,34 @@
 
 cmake_minimum_required(VERSION 3.24)
 
-if(NOT IS_DIRECTORY "${CCCL_DIR}")
-  message(
-    FATAL_ERROR "embed_cccl_headers: CCCL_DIR '${CCCL_DIR}' is not a directory")
-endif()
+file(STRINGS "${INCLUDE_DIRS_FILE}" cccl_include_dirs)
```
felipeblazing, line 19: **Reviewer: Error handling & fail-closed gating**

**nit** Required -D inputs are no longer validated up front

The old script checked `CCCL_DIR` explicitly. The new one goes straight to `file(STRINGS "${INCLUDE_DIRS_FILE}" ...)` and never checks `OUT`, so a bad invocation gives a generic CMake error. Checking both keeps the error message descriptive, as it was before.
```suggestion
foreach(required INCLUDE_DIRS_FILE OUT)
  if(NOT DEFINED ${required} OR "${${required}}" STREQUAL "")
    message(FATAL_ERROR "embed_cccl_headers: -D${required}=... is required")
  endif()
endforeach()
if(NOT EXISTS "${INCLUDE_DIRS_FILE}")
  message(FATAL_ERROR "embed_cccl_headers: INCLUDE_DIRS_FILE '${INCLUDE_DIRS_FILE}' does not exist")
endif()
file(STRINGS "${INCLUDE_DIRS_FILE}" cccl_include_dirs)
```

src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake
```diff
@@ -15,10 +16,34 @@
...
+
+# Installed packages share one include root; source packages use separate
+# component roots. Search the directories in the order supplied by CMake.
+function(find_cccl_header name output)
```
felipeblazing, line 23: **Reviewer: Test coverage & rigor**

**major** Rewritten CCCL embedder has no test, and the layout fixture was deleted in this PR

Commit a41210e1 dropped the only test for the new `find_cccl_header` multi-root search. So the separate cub/thrust/libcudacxx roots that the README now promises, the `../` escape rejection for quoted includes and the empty-closure guard are never exercised. CI builds cover only the single-root installed layout. Add a ctest that writes a tiny split-root tree with `file(WRITE)` at test time (no committed binaries), runs `cmake -P embed_cccl_headers.cmake` on it, and checks the embedded names.

src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake
```diff
@@ -42,8 +67,8 @@ while(worklist)
   if(rel IN_LIST found)
     continue()
   endif()
-  set(abs "${CCCL_DIR}/${rel}")
-  if(NOT EXISTS "${abs}" OR IS_DIRECTORY "${abs}")
+  find_cccl_header("${rel}" abs)
+  if(NOT abs)
```
felipeblazing, line 92: **Reviewer: Error handling & fail-closed gating**

**minor** Unresolvable scan roots are skipped silently; only an empty closure fails

When a seeded root such as `cub/block/block_scan.cuh` doesn't resolve, the scan just runs `continue()`. The only guard is `n EQUAL 0`, so losing one prelude root still builds a closure that fails later in NVRTC as a missing header, far from its cause. Make the unconditionally emitted roots (`cub/block/*`, `cuda/std/*`) required, the way the marker check is, and log at least a `message(STATUS ...)` for each skipped version-dependent root (the thrust execution_policy ones).

src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake
```diff
@@ -97,13 +123,28 @@ string(APPEND body "#include \"codegen/jit/cccl_embedded_headers.h\"\n")
...
+  find_cccl_header("${rel}" header)
+  list(APPEND header_dependencies "${header}")
+  file(READ "${header}" hsrc)
+  # Hash each source independently, then hash an unambiguous ordered manifest.
+  # This avoids retaining a second full copy of the (large) closure while still
+  # binding the final fingerprint to every relative name and byte of content.
+  string(SHA256 hsrc_sha256 "${hsrc}")
+  string(LENGTH "${rel}" rel_length)
+  string(APPEND _fingerprint_manifest "${rel_length}:${rel}:${hsrc_sha256}\n")
   string(APPEND body "static const char* kCcclName${idx} = \"${rel}\";\n")
```
felipeblazing, line 152: **Reviewer: Test coverage & rigor**

**minor** Nothing checks that the fingerprint actually depends on header content

The test only checks that `kCcclEmbeddedHeadersFingerprint` is 64 lowercase hex characters. A manifest that dropped `hsrc_sha256`, or hashed only the names, would still pass that check and would reuse stale cubins after a CCCL bump with the same file set. In the embedder fixture test, run the script twice with one byte of a header changed and assert that the fingerprints differ. Also assert that they match when only the root order changes but the content is the same.

src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake
```diff
@@ -97,13 +123,28 @@ string(APPEND body "#include \"codegen/jit/cccl_embedded_headers.h\"\n")
...
   string(APPEND body
          "static const char* kCcclSrc${idx} =\nR\"${D}(${hsrc})${D}\";\n")
   math(EXPR idx "${idx}+1")
 endforeach()
+string(SHA256 cccl_fingerprint "${_fingerprint_manifest}")
+string(TOLOWER "${cccl_fingerprint}" cccl_fingerprint)
```
felipeblazing: **Reviewer: Build, deps, CI & repo hygiene**

**nit** TOLOWER on string(SHA256) output is a no-op

`string(SHA256)` already produces lowercase hex, so this line does nothing.
```suggestion
```

src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake
```diff
@@ -115,5 +156,40 @@ string(APPEND body "};\n")
 string(APPEND body "const int kCcclEmbeddedHeaderCount = ${n};\n")
 string(APPEND body "}  // namespace codegen::jit\n")
 
+get_filename_component(out_dir "${OUT}" DIRECTORY)
+file(MAKE_DIRECTORY "${out_dir}")
```
felipeblazing: **Reviewer: Build, deps, CI & repo hygiene**

**nit** Redundant MAKE_DIRECTORY before file(WRITE)

`file(WRITE)` creates missing parent directories itself. The `get_filename_component` + `file(MAKE_DIRECTORY)` pairs for `OUT` (here) and `DEPFILE` (lines 180-181) can go.
```suggestion
```

src/compression/simpatico_codegen/include/codegen/jit/cccl_embedded_headers.h
```diff
@@ -10,5 +10,9 @@ namespace codegen::jit {
 
 extern const EmbeddedJitHeader kCcclEmbeddedHeaders[];
 extern const int kCcclEmbeddedHeaderCount;
+// Deterministic digest of the names and contents of the embedded CCCL header
+// closure. This is part of every JIT cache identity so changing the headers
+// cannot reuse a cubin compiled against an older closure.
+extern const char kCcclEmbeddedHeadersFingerprint[];
```
felipeblazing: **Reviewer: API design, layering & reuse**

**major** Fingerprint ignores the other embedded JIT headers (kEmbeddedJitHeaders), which is still unaddressed

NVRTC also receives `kEmbeddedJitHeaders` (`stdint_shim.hpp`, `decode/rle_block.cuh`; see `nvrtc_compiler.cpp:173`). These headers are not part of the rendered source, so editing `rle_block.cuh` still loads a stale cubin from disk. joosthooz asked for this in their CHANGES_REQUESTED review and it is still missing at head. Compute one fingerprint over both embedded sets and name it for that (e.g. `kJitHeadersFingerprint`), instead of a field and API that are specific to CCCL.

src/compression/simpatico_codegen/include/codegen/jit/kernel_cache.hpp
```diff
@@ -16,11 +17,13 @@ struct ShapeKey {
   int arch_cc;
   uint32_t cuda_runtime;
   uint32_t driver_version;
+  std::string cccl_headers_fingerprint;
```
felipeblazing: **Reviewer: API design, layering & reuse**

**minor** Process-constant fingerprint on the in-memory key adds an allocation to every lookup

`kCcclEmbeddedHeadersFingerprint` is compiled into the binary, so every `ShapeKey` in one process has the same value. It can never make an in-memory hit miss. Even so, every `get_or_compile_plain` call copies a 64-byte `std::string` into the key and hashes it. The stale-cubin risk only exists on disk: keep `ShapeKey` as it was and apply the fingerprint only where the disk path is formed.

src/compression/simpatico_codegen/include/codegen/jit/kernel_cache.hpp
```diff
@@ -30,11 +33,25 @@ struct ShapeKeyHash {
 
 std::string source_digest(const std::string& rendered_source);
 ShapeKey shape_key_from(const std::string& rendered_source, int arch_cc);
+// Explicit-fingerprint overload used to keep cache identity construction
```
felipeblazing: **Reviewer: Comments, docs & AI residue**

**nit** "Deterministic" doesn't describe what this overload is for

Both overloads are equally deterministic. The real difference is that this one lets a test inject a fingerprint. Say that directly, as the `cubin_key_material_from` comment below already does.
```suggestion
// Test seam: same as above with an injected fingerprint instead of
```

src/compression/simpatico_codegen/include/codegen/jit/kernel_cache.hpp
```diff
@@ -30,11 +33,25 @@ struct ShapeKeyHash {
...
+// deterministic and directly testable. Production callers use the overload
+// above, which supplies kCcclEmbeddedHeadersFingerprint.
+ShapeKey shape_key_from(const std::string& rendered_source,
```
felipeblazing: **Reviewer: API design, layering & reuse**

**minor** Test-seam overloads added to the public header

Both new two-overload pairs exist only so the test can inject a fingerprint. The comments call them a "test seam", yet they are declared in `include/codegen/jit/kernel_cache.hpp`, which this PR makes a PUBLIC include dir for every Sirius target. Callers now have two ways to build a key, and one of them can build a key that production never uses. Keep a single production entry point and test the injected variant through an internal/detail header, or test the path builder directly.

src/compression/simpatico_codegen/include/codegen/jit/kernel_cache.hpp
```diff
@@ -30,11 +33,25 @@ struct ShapeKeyHash {
...
+                        int arch_cc,
+                        std::string_view cccl_headers_fingerprint);
+
+// Adds the embedded-CCCL fingerprint to the material hashed for a persistent
```
felipeblazing: **Reviewer: Comments, docs & AI residue**

**nit** Doc for `cubin_key_material_from` doesn't say what the returned string is

The return value is a length-prefixed serialization of `(kernel material, fingerprint)`. That encoding is what makes it collision-free, and callers must pass it to `cubin_path_for` rather than use it as a filename. Say this in one clause in the header. The "test seam" sentence can stay.

src/compression/simpatico_codegen/src/bridge/codegen_runtime.cpp
```diff
@@ -11,9 +11,10 @@
 //      not store: Bitpack's bp_offsets cumsum (CUB ExclusiveSum + 1-thread tail
 //      patch, see offsets_cumsum.cu) and RLE scratch, into RMM-pool transients.
 //   2. KernelCache::get_or_compile_plain resolves shape -> CUkernel (keyed
-//      on a hash of the rendered CUDA source, so a shape hits the cache
-//      across compress and decompress). First touch pays the nvrtc compile;
-//      warm hits are cheap. CUfunction is derived per-device at launch time.
+//      on the rendered CUDA source and embedded CCCL fingerprint, so a shape
+//      hits the cache across compress and decompress without crossing header
```
felipeblazing: **Reviewer: Comments, docs & AI residue**

**minor** Comment suggests the in-memory cache could match a cubin from another header version

The in-memory cache lives only as long as the process, and within one binary the fingerprint is a compile-time constant, so a lookup there can never cross header versions. Only the on-disk cubin name needs the fingerprint. The added clause invites readers to think the in-memory key needed this change. Drop the clause.

src/compression/simpatico_codegen/src/jit/kernel_cache.cpp
```diff
@@ -174,9 +184,31 @@ ShapeKey shape_key_from(const std::string& rendered_source, int arch_cc)
     arch_cc,
     static_cast<uint32_t>(CUDART_VERSION),
     static_cast<uint32_t>(driver_version),
+    std::string(cccl_headers_fingerprint),
   };
 }
 
+std::string cubin_key_material_from(std::string_view kernel_key_material)
+{
+  return cubin_key_material_from(kernel_key_material, kCcclEmbeddedHeadersFingerprint);
```
felipeblazing: **Reviewer: Error handling & fail-closed gating**

**major** Disk cubin key still ignores the embedded project JIT headers (still unaddressed)

NVRTC also compiles against `kEmbeddedJitHeaders` (`codegen/decode/rle_block.cuh`, `stdint_shim.hpp`; see `nvrtc_compiler.cpp:172`). None of that content goes into the key. If someone fixes a bug in `rle_block.cuh`, every existing `~/.cache/simpatico/jit/*.cubin` still hits, and the old kernel decodes with no error: a wrong answer, not a failure. joosthooz asked for this in the review and the head still doesn't do it. Hash `kEmbeddedJitHeaders` into the build-time fingerprint the same way as the CCCL closure, or into a sibling constant that is included in the cubin key material.

src/compression/simpatico_codegen/src/jit/kernel_cache.cpp
```diff
@@ -174,9 +184,31 @@ ShapeKey shape_key_from(const std::string& rendered_source, int arch_cc)
...
+}
+
+std::string cubin_key_material_from(std::string_view kernel_key_material,
+                                    std::string_view cccl_headers_fingerprint)
+{
+  // Length-prefix both components so arbitrary rendered source text cannot
+  // make two (kernel material, fingerprint) pairs serialize identically.
+  std::string material;
+  material.reserve(kernel_key_material.size() + cccl_headers_fingerprint.size() + 48);
```
felipeblazing: **Reviewer: Comments, docs & AI residue**

**nit** Unexplained `+ 48` in reserve

A reader has to work out that this is two decimal `size_t` lengths (up to 20 digits each) plus two `:` separators, with some slack. Add a short inline note or use a named constant.

src/compression/simpatico_codegen/src/jit/kernel_cache.cpp
felipeblazing: **Reviewer: API design, layering & reuse**

**major** Two parallel cache-identity paths; derive the cubin filename from ShapeKey

The PR adds the fingerprint twice: once to `ShapeKey`, and separately to a hand-serialized `cubin_key_material_from` string. `cubin_path_for` also queries `cuDriverGetVersion` again even though `ShapeKey` already holds the driver version. Adding the next key component needs both edits again, and the two paths can drift. Build the on-disk name from the `ShapeKey` (`source_hash` + fingerprint + arch/runtime/driver) so there is one source of truth, and delete `cubin_key_material_from`.

felipeblazing: **Reviewer: Test coverage & rigor**

**major** Disk-cache wiring of the fingerprint is untested; reverting this line still passes

The test calls `cubin_key_material_from` directly and never checks that `get_or_compile_plain` uses it. If this line goes back to `cubin_path_for(key_material, ...)`, the test still passes, and that is the stale-cubin bug the PR is meant to fix. Add a test that points `SIMPATICO_JIT_CACHE_DIR` at a temp dir before the first `disk_cache_dir()` call (it is cached in a static) and restores it afterwards. Then compile one kernel and assert that the written file's name equals the digest of `cubin_key_material_from(source|entry)`, or expose `cubin_path_for` so the test can compute it.

felipeblazing: **Reviewer: Build, deps, CI & repo hygiene**

**minor** Project JIT headers (kEmbeddedJitHeaders) still aren't part of the cache fingerprint

joosthooz's review request is still open at head. Only the CCCL closure is fingerprinted. `stdint_shim.hpp` and `rle_block.cuh`, which reach NVRTC through `kEmbeddedJitHeaders` in `nvrtc_compiler.cpp`, are not. If `rle_block.cuh` changes, the rendered source stays identical and the old on-disk cubin gets reused. Have `embed_jit_headers.cmake` emit a digest the same way and fold it into the key material.

src/compression/simpatico_codegen/tests/test_jit_kernel_cache.cpp
```diff
@@ -38,6 +39,59 @@ int main()
 {
   if (cudaSetDevice(0) != cudaSuccess) return report_fail("cudaSetDevice(0) failed");
 
+  // Cache identity must change with the exact CCCL closure supplied to NVRTC.
```
felipeblazing: **Reviewer: Comments, docs & AI residue**

**nit** Test's header comment no longer lists everything it checks

The header still lists "three properties to pin down", but the new first block checks a fourth: the CCCL fingerprint changes both the ShapeKey and the cubin key. Add it to the list. The inline comment on that block can then shrink to one line, since "remains deterministic within one build and does not need an old binary/cache artifact" only explains why the test is written this way.

src/compression/simpatico_codegen/tests/test_jit_kernel_cache.cpp
```diff
@@ -38,6 +39,59 @@ int main()
...
+  // Exercise an injected alternate fingerprint so this remains deterministic
+  // within one build and does not need an old binary/cache artifact.
+  {
+    const std::string current_fingerprint = jit::kCcclEmbeddedHeadersFingerprint;
```
felipeblazing: **Reviewer: Test coverage & rigor**

**minor** The new regression guard never runs in CI

The top-level CMakeLists adds simpatico with `EXCLUDE_FROM_ALL` and says to run the tests from the standalone build, and no workflow under `.github/` builds or runs `test_jit_kernel_cache`. The green checks therefore say nothing about the fingerprint asserts, and they, like the rest of the file, only run when someone runs them by hand on a GPU box. At minimum, the host-only fingerprint block (which needs no device) could be a separate `add_codegen_host_test` that CI can run without a GPU.

src/compression/simpatico_codegen/tests/test_jit_kernel_cache.cpp
```diff
@@ -38,6 +39,59 @@ int main()
...
+      jit::cubin_key_material_from("same rendered source|same_entry", current_fingerprint);
+    const auto alternate_cubin_material =
+      jit::cubin_key_material_from("same rendered source|same_entry", alternate_fingerprint);
+    if (current_cubin_material != explicit_current_cubin_material) {
+      return report_fail("default cubin key material does not use embedded CCCL fingerprint");
+    }
+    if (current_cubin_material == alternate_cubin_material) {
+      return report_fail("persistent cubin key ignores embedded CCCL fingerprint changes");
+    }
+    if (jit::source_digest(current_cubin_material) ==
```
felipeblazing: **Reviewer: Test coverage & rigor**

**minor** The digest-inequality check only tests FNV-1a, not this PR

Line 86 already asserts that the two materials differ. Whether `source_digest` then maps two different strings to different values is a property of FNV-1a, and the existing `hello world` block already covers it. This check cannot catch a regression in the fingerprint plumbing. Drop it, or replace it with the end-to-end filename assertion suggested on `kernel_cache.cpp:242`.

src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake
9prady9: Such comprehensive check is not necessary because we will encounter a failure at cmake CCCL target usage level itself.

src/compression/simpatico_codegen/src/jit/kernel_cache.cpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/README.md
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/README.md
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/cmake/embed_cccl_headers.cmake
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/include/codegen/jit/cccl_embedded_headers.h
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/include/codegen/jit/kernel_cache.hpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/include/codegen/jit/kernel_cache.hpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/include/codegen/jit/kernel_cache.hpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/include/codegen/jit/kernel_cache.hpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/src/bridge/codegen_runtime.cpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/src/jit/kernel_cache.cpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/src/jit/kernel_cache.cpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/src/jit/kernel_cache.cpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/tests/test_jit_kernel_cache.cpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/tests/test_jit_kernel_cache.cpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/tests/test_jit_kernel_cache.cpp
9prady9: Tracked in https://github.com/sirius-db/sirius/issues/1863. We are keeping this PR scoped to CCCL header alignment and splitting its fingerprint/cache-key additions into the follow-up. The issue captures the correctness, performance, API, documentation, and test feedback. Resolving this thread as deferred, not as a claim that cache invalidation is fixed.

src/compression/simpatico_codegen/CMakeLists.txt
9prady9: trying out dependency check, full vcpkg build will take lot of time. Will let you know.

9prady9: The dependency-only checks passed using the real repository vcpkg overlay ports and vcpkg toolchain, with Pixi supplying the CUDA toolkits:

- CUDA 12.9.86 + NVRTC 12.9.86, and CUDA 13.3.73 + NVRTC 13.3.33.
- Both select vcpkg CCCL 3.4.3. The production embedder collects 512 headers, and host/JIT assertions confirm the expected CUB and Thrust versions.
- The static NVRTC/CUDART targets link successfully. A CUB reduction kernel compiles from the embedded headers without runtime `-I` paths; the probe executables have no shared NVRTC/CUDART dependency. This was compile-only, not a GPU launch.

This verifies the dependency targets and header embedding, **not a full Simpatico/Sirius vcpkg build or the `cudf::cudf` transitive integration**. No usable dependency cache was available, so we stopped short of building libcudf from source. Leaving the full-build verification request open.

PR #1802, opened 2026-09-15
Window Function High Level Design Doc

wmalpica:
This PR is just to share a high level window function design and get feedback

[Rendered](https://github.com/wmalpica/sirius/blob/1d38d45df88aa3600a7c178778292ba4da413d51/docs/super-sirius/window-function-design.md)


Jedi18:
Thanks for the design doc, William! I compared the proposal with GQE’s implementation to see what we might learn from it. GQE’s window support is fairly narrow: it materializes the complete window input on one GPU, its SQL path currently supports only `SUM` and `RANK`, and ordered aggregate windows always use cuDF’s RANGE-based implementation without preserving the SQL `ROWS`/`RANGE` distinction.

A couple of ideas may still be useful:

### 1. Rewrite partition-only windows

For:

```sql
SUM(x) OVER (PARTITION BY key)
```

GQE rewrites the window as:

```text
GROUP BY key → compute aggregate once → join back to the original rows
```

This seems like an optimization we could reuse directly.

### 2. Experiment with skinny window inputs

GQE generates a stable row ID and joins the window result back to the original input. GQE still sends the full table through its window task, but we could extend this pattern by projecting only the columns required for window execution:

```text
row_id | partition keys | order keys | aggregate inputs
```

The operator would emit:

```text
row_id | window result
```

which would then be joined back to the original table:

```text
Original wide table ───────────────────────────────┐
                                                   │
Skinny projection                                  │
[row_id, partition, order, value]                  │
         ↓                                         │
repartition / sort                                 │
         ↓                                         │
copy skinny halos                                  │
         ↓                                         │
window calculation                                 │
         ↓                                         │
[row_id, window_result]                            │
         └──────────── join on row_id ─────────────┘
                            ↓
               [original columns, result]
```

This would make sorting cheaper and reduce halo memory from approximately:

```text
halo_rows × full_row_width
```

to:

```text
halo_rows × required_window_columns_width
```

It is not a free optimization: the original table must remain alive, and the join-back may require significant memory and distributed communication. DuckDB’s projection pruning may also mean the input is already narrower than the base-table schema. This therefore seems most useful as a planner-selected fallback when the estimated batch-plus-halo working set would otherwise exceed GPU memory, rather than as the default strategy.

wmalpica:
Excellent suggestions Akhil.
Let me explain how these would be adapted into the current design:
1. You example of this is just one of many places where we could apply a separate optimal algorithm. Another is when you dont have a PARTITION BY clause and have a window which is unbounded preceding and unbounded following. In my implementation approach, I would rather implement the general case, and then implement optimized specializations as neecessary.

2. If we have multiple window shapes, where we would have to do separate sortings, then we would need to bifurcate the query operator DAG, and in this case, then yes we would need a row identifier as a key to join things back. Different columns would go to different DAG branches. I would add this as a followup after we have a working single window shape case. 
As for the skinny halos: this is something we would do as it is already. We would only copy, concat and trim halos for the columns necessary. All other columns just ride along. If you notice, the way the tasks would be setup, we would be concatenating the halo, performing the window calculation and then trimming the halo all in one task, which means we can do so only on the necessary columns and the other columns can be "detached and reattached" to the cudf::table as needed to avoid moving or copying them.

Jedi18:
Also I ran this through Astra/Fable and it had some suggestions which I agree with. Pasting here so we can keep that in mind during implementation.

The core approach of computing within batches, exchanging halos, and combining partial aggregates makes sense. A few suggestions before implementation:
1. **Allow a SQL partition to span batches in the design**. Requiring a whole partition to fit in one batch will be restrictive for low-cardinality keys and skewed data. Even if the first implementation falls back for oversized partitions, it would help to define partition identity, batch ordering, and cross-batch state now.
2. **Chain operators for different window orderings initially**. We can calculate one set of windows, carry those result columns through the next sort, and calculate the next set. DuckDB does this already, so branching and joining results back can remain an optional optimization. Compatible ORDER BY prefixes can also share sorting.
3. **Consider explicit execution stages**. A preparation stage for batch summaries and halos, a summary-combine stage, and a final evaluation stage could reuse existing pipeline memory histories and ordinary spillable repositories. This seems worth comparing against adding per-task-type memory accounting and per-batch dependencies inside one operator. Halo ownership, cross-GPU movement, and release timing would still need to be defined.
4. **Specify exactly what each partial aggregate represents**. Summaries should cover only the batch’s original rows, excluding borrowed halo rows, to avoid double-counting. AVG needs sum/count state; ranking needs peer-boundary information. A small supported-function/frame matrix would make these contracts clearer.
5. **Make the sorting and halo limits explicit**. The current sort path needs changes to preserve range-slice identity and deliver ordered batches. Halos may extend across several batches, and a bounded RANGE frame can still contain arbitrarily many rows. We should define a memory limit and fallback or alternative strategy for those cases.

Your clarification about copying only the columns needed for halos addresses that concern well. I’d also check which whole-partition aggregate cases DuckDB’s existing optimizer already rewrites before adding equivalent logic.

For validation, deliberately tiny batches would be useful: force partitions, tied ordering values, and frames across several boundaries, then compare SUM, AVG, and ranking results against DuckDB. That should expose the important correctness issues early.

wmalpica:
> 1. **Allow a SQL partition to span batches in the design**. Requiring a whole partition to fit in one batch will be restrictive for low-cardinality keys and skewed data. Even if the first implementation falls back for oversized partitions, it would help to define partition identity, batch ordering, and cross-batch state now.

I already have in mind how we can do this and what it would entail. We would need to modify how our global sort works (which would also help with the scalability of low cardinality global sorts). It would, though, also make all the window function algorithms slower and more complex, since we cannot rely on batch boundaries also being partition boundaries. So my initial thought was to build it simpler, and faster and then make it more scalable. But i can be convinced otherwise.

> 2. **Chain operators for different window orderings initially**. We can calculate one set of windows, carry those result columns through the next sort, and calculate the next set. DuckDB does this already, so branching and joining results back can remain an optional optimization. Compatible ORDER BY prefixes can also share sorting.

Yeah, this is a good idea, although, i dont think "Compatible ORDER BY prefixes can also share sorting.". If the window functions have different ordering, then by definition we have to sort again. Or perhaps I dont understand what you mean.

> 3. **Consider explicit execution stages**. A preparation stage for batch summaries and halos, a summary-combine stage, and a final evaluation stage could reuse existing pipeline memory histories and ordinary spillable repositories. This seems worth comparing against adding per-task-type memory accounting and per-batch dependencies inside one operator. Halo ownership, cross-GPU movement, and release timing would still need to be defined.

My main reason for proposing doing it all in one operator was to avoid figuring out how we pass  halos and unbounded partial aggregates between operators. But if we can figure out a clean way to do that and to expose spilling of the halos, i am not opposed to separating things out into different operators.

Also, we should add "adding per-task-type memory accounting " to the engine anyways, since we already are doing different types of tasks in joins and putting the memory accounting into the same bucket. For example we have some tasks where we build the hash table and we also probe it (it takes two batches as inputs, left and right). But after the hash table is built, we have tasks that only probe it. THese have different memory consumption profiles, and they history is all being dumped into the same history.

> 4. **Specify exactly what each partial aggregate represents**. Summaries should cover only the batch’s original rows, excluding borrowed halo rows, to avoid double-counting. AVG needs sum/count state; ranking needs peer-boundary information. A small supported-function/frame matrix would make these contracts clearer.

Yes, we do need all that. This proposal was describing more the broad shape of the solution. I have done a cursorly survey of all aggregators and the vast majority can be solved with this general shape. 
I did forget to mention how things like AVG would be solved, and that would be with a SUM and COUNT window aggregations, followed by a projection at the end. Same idea applies to cume_dist, percent_rank and others. 

> 5. **Make the sorting and halo limits explicit**. The current sort path needs changes to preserve range-slice identity and deliver ordered batches. Halos may extend across several batches, and a bounded RANGE frame can still contain arbitrarily many rows. We should define a memory limit and fallback or alternative strategy for those cases.
> 
> Your clarification about copying only the columns needed for halos addresses that concern well. I’d also check which whole-partition aggregate cases DuckDB’s existing optimizer already rewrites before adding equivalent logic.

I have an idea of how we could support arbitrarily large halos. This is only really a challenge for RANGE frames. But it is very complex and I dont think we should worry about it for now. 

> 
> For validation, deliberately tiny batches would be useful: force partitions, tied ordering values, and frames across several boundaries, then compare SUM, AVG, and ranking results against DuckDB. That should expose the important correctness issues early.

YEs

PR #1804, opened 2026-09-16
fix(telemetry): give Sirius its own NVTX domain and fix Quent domain capture

9prady9:
## Description

Sirius emitted every NVTX range into the default domain, so profiles interleaved Sirius with `libcudf`/`CCCL` and couldn't be filtered to Sirius alone.

**NVTX `sirius` domain**

- All 80 annotation sites move into a named `sirius` domain.
- The pipeline range spans task boundaries and can't be scoped, so it switches to `nvtxDomainRangeStartEx`/`End` — semantics unchanged.
- `src/legacy/` is deliberately left on the global domain.

**The domain tag type is declared twice on purpose**

- simpatico also builds standalone, without `src/include` on its include path.
- `parquet_benchmark` gets `src/include` but not simpatico's include root.
- So neither header can include the other — collapsing them fails with `fatal error: codegen/util/nvtx.hpp: No such file or directory`.
- NVTX keys domains by name, so both tag types resolve to one `sirius` domain.
- The name comes from a single `SIRIUS_NVTX_DOMAIN_NAME` definition set from `${PROJECT_NAME}`, so it can't drift.

**quent bump**

- `381e1be` → `b21933d0`, for [quent#736](https://github.com/rapidsai/quent/pull/736), which groups explicitly named NVTX domains by name at the presentation boundary.

**Quent context ordering**

- cuCascade's ranges gained a `libcucascade` domain in [cuCascade#196](https://github.com/NVIDIA/cuCascade/pull/196), already on `dev` — but that domain never reached Quent.
- `nvtx_injection::dispatch` drops events while its `OnceLock` hook is unset, and that hook is installed by `quent::create_context`, which ran inside `telemetry_context`'s constructor — i.e. *after* the memory manager was built. cuCascade's pinned-pool setup, including the `nvtxDomainCreateA` that names the domain, landed in that dead window.
- The handle stayed valid, so later cuCascade ranges were captured but unnamed; the UI showed them under `domain0x1`.
- `telemetry_context::create` now takes an already-built `rust::Box<quent::Context>`, and `SiriusContext::initialize` builds it first via the new `make_quent_context()`. This keeps the memory manager and per-GPU device ids the context declaration needs, so no telemetry is lost.

**Verification**

- TPC-H SF100 (Q1/Q3/Q5) under nsys: one `sirius` domain alongside `libcucascade`, `libcudf` and `CCCL`; zero ranges left in the default domain.
- Under Quent: all four domains now appear by name — `libcucascade` showed as `domain0x1` before the ordering fix.
- Builds clean after rebase onto `dev` (RAPIDS nightly 26.12): `sirius_extension`, `sirius_unittest`, `parquet_benchmark`, standalone `simpatico_codegen`.
- `pre-commit run -a` green.

## Checklist
- [ ] Cover changes with new or existing tests — NVTX domain membership and Quent capture timing are observable only in a profiler, so both are verified via the nsys/Quent runs above; the two `telemetry_context::create` call sites in tests are updated for the new signature
- [ ] Update human and agent documentation — `quent-telemetry.md` doesn't enumerate domains, so nothing went stale

## References

Refs [cuCascade#196](https://github.com/NVIDIA/cuCascade/pull/196), [quent#736](https://github.com/rapidsai/quent/pull/736)


dhruv9vats (approved):
Thanks @9prady9.

Did you have a chance to try this out too? To verify if Quent indeed shows the ranges in  a separate domain now?

9prady9:
> Thanks @9prady9.
> 
> Did you have a chance to try this out too? To verify if Quent indeed shows the ranges in a separate domain now?

yes, did it. 

Will share updated screenshot after updating quent to https://github.com/rapidsai/quent/pull/736

9prady9:
> > Thanks @9prady9.
> > Did you have a chance to try this out too? To verify if Quent indeed shows the ranges in a separate domain now?
> 
> yes, did it.
> 
> Will share updated screenshot after updating quent to [rapidsai/quent#736](https://github.com/rapidsai/quent/pull/736)



Screenshot from 2026-09-16 17-32-11
`domainx01` is handle which appears so because of ordering issue with memory component w.r.t sirius/quent context. Will share fixed screenshot soon with cucascade domain too.
Screenshot from 2026-09-16 17-31-53
Screenshot from 2026-09-16 17-31-32

9prady9:
Whoop, libcucascade also in - although it is very sparse nvtx annotations because it doesn't have those calls used. We can add those gng forward now.

@dhruv9vats you can see the info on screenshots. once this PR is merged, sirius nightly runs will have all thse shown domain ranges. But for the reason I shared above, cucascade need more nvtx calls to capture more events.

image

mbrobbel (approved):

PR #1767, opened 2026-09-14
build: add standalone `sirius` shared library CMake target

mbrobbel:
This is the first step towards building Sirius as standalone library.

Closes #1732.

PR #1806, opened 2026-09-16
refactor: split public and internal headers

### LINKED ISSUE #1751: Split headers into public and internal
The supported library API headers should be separate from internal headers. I'd like to suggest the follow separation:

Public headers (installed): `include/sirius`. These have minimal other dependencies.

Internal headers are beside their implementations within each subsystem in `src`.

For example:
```
# public
include/sirius/
  engine.hpp
  session.hpp
  config.hpp
  result.hpp

# internal
src/
  core/
    engine.cpp
    engine.hpp
    pipeline/
      task_scheduler.cpp
      task_scheduler.hpp
  io/
    io_context.cpp
    io_context.hpp
  adapters/
    duckdb/
      context_adapter.cpp
      context_adapter.hpp
```

mbrobbel:
Closes #1751.


mike-wendt (approved):
Approved for `CODEOWNERS` changes only

CMakeLists.txt
```diff
@@ -762,6 +761,9 @@ install(
   LIBRARY DESTINATION "${INSTALL_LIB_DIR}"
   ARCHIVE DESTINATION "${INSTALL_LIB_DIR}")
 
+install(DIRECTORY include/sirius DESTINATION include)
+install(FILES include/sirius_extension.hpp DESTINATION include)
```
9prady9: Two problems with these two lines, one fix for both.

**1 — dropping `COMPONENT sirius_library` takes the public headers out of the standalone package.**

The rule they replace carried it, as does everything else in this block (779-781, 787, 792), and no `CMAKE_INSTALL_DEFAULT_COMPONENT_NAME` is set — so these two land in `Unspecified`:

```console
$ cmake --install build --component sirius_library --prefix /opt/sirius
-- Installing: /opt/sirius/lib/libsirius.so
-- Installing: /opt/sirius/lib/cmake/sirius/sirius-targets.cmake
-- Installing: /opt/sirius/lib/cmake/sirius/sirius-targets-noconfig.cmake
```

No `include/` at all, and since line 702 bakes that path into the export, the first consumer dies at configure time:

```
CMake Error: Imported target "sirius::sirius" includes non-existent path
  "/opt/sirius/include" in its INTERFACE_INCLUDE_DIRECTORIES.
```

**2 — `DESTINATION include` bypasses `CMAKE_INSTALL_INCLUDEDIR`, which line 702 uses for the matching export, so the two can resolve to different places under one prefix.**

Setting only the prefix is enough to split them: GNUInstallDirs applies its `usr/` insertion to the `CMAKE_INSTALL_<dir>` variables, but not to literal destinations.

```console
$ cmake -DCMAKE_INSTALL_PREFIX=/ ...
$ DESTDIR=/tmp/stage cmake --install build
-- Installing: /tmp/stage/include/sirius/ffi.hpp         # DESTINATION include
-- Installing: /tmp/stage/include/sirius_extension.hpp
-- Installing: /tmp/stage/usr/lib/libsirius.so           # ${CMAKE_INSTALL_LIBDIR}
-- Installing: /tmp/stage/usr/lib/cmake/sirius/sirius-targets.cmake
```
```cmake
# sirius-targets.cmake
INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/usr/include"
```

Headers in `/include`, consumer told `/usr/include`, library correctly in `/usr/lib`. The rest of this block already goes through `${CMAKE_INSTALL_LIBDIR}`; the headers are the only rules that opted out.

Both reproduced with these exact lines on cmake 4.4. `cmake_minimum_required(VERSION 3.30.4)` at line 17 keeps CMP0193 at `OLD`, matching the repro.

```suggestion
install(
  DIRECTORY include/sirius
  DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
  COMPONENT sirius_library)
install(
  FILES include/sirius_extension.hpp
  DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
  COMPONENT sirius_library)
```

9prady9: Two problems with these two lines, one fix for both.

**1 — dropping `COMPONENT sirius_library` takes the public headers out of the standalone package.**

The rule they replace carried it, as does everything else in this block (779-781, 787, 792), and no `CMAKE_INSTALL_DEFAULT_COMPONENT_NAME` is set — so these two land in `Unspecified`:

```console
$ cmake --install build --component sirius_library --prefix /opt/sirius
-- Installing: /opt/sirius/lib/libsirius.so
-- Installing: /opt/sirius/lib/cmake/sirius/sirius-targets.cmake
-- Installing: /opt/sirius/lib/cmake/sirius/sirius-targets-noconfig.cmake
```

No `include/` at all, and since line 702 bakes that path into the export, the first consumer dies at configure time:

```
CMake Error: Imported target "sirius::sirius" includes non-existent path
  "/opt/sirius/include" in its INTERFACE_INCLUDE_DIRECTORIES.
```

**2 — `DESTINATION include` bypasses `CMAKE_INSTALL_INCLUDEDIR`, which line 702 uses for the matching export, so the two can resolve to different places under one prefix.**

Setting only the prefix is enough to split them: GNUInstallDirs applies its `usr/` insertion to the `CMAKE_INSTALL_<dir>` variables, but not to literal destinations.

```console
$ cmake -DCMAKE_INSTALL_PREFIX=/ ...
$ DESTDIR=/tmp/stage cmake --install build
-- Installing: /tmp/stage/include/sirius/ffi.hpp         # DESTINATION include
-- Installing: /tmp/stage/include/sirius_extension.hpp
-- Installing: /tmp/stage/usr/lib/libsirius.so           # ${CMAKE_INSTALL_LIBDIR}
-- Installing: /tmp/stage/usr/lib/cmake/sirius/sirius-targets.cmake
```
```cmake
# sirius-targets.cmake
INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/usr/include"
```

Headers in `/include`, consumer told `/usr/include`, library correctly in `/usr/lib`. The rest of this block already goes through `${CMAKE_INSTALL_LIBDIR}`; the headers are the only rules that opted out.

Both reproduced on cmake 4.4, with these lines copied into a minimal standalone project rather than a full Sirius build; `cmake_minimum_required(VERSION 3.30.4)` at line 17 keeps CMP0193 at `OLD`, matching the repro.
```suggestion
install(
  DIRECTORY include/sirius
  DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
  COMPONENT sirius_library)
install(
  FILES include/sirius_extension.hpp
  DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
  COMPONENT sirius_library)
```

src/legacy/CMakeLists.txt
```diff
@@ -108,7 +108,7 @@ set(SIRIUS_LEGACY_TEST_SOURCES
     PARENT_SCOPE)
 
 set(SIRIUS_LEGACY_INCLUDE_DIR
```
Jedi18, line 110: This path appears to point one directory too high. `CMAKE_CURRENT_SOURCE_DIR` is already `src/legacy`, so adding `/..` changes the include directory to src.
The legacy source files still include headers such as "gpu_physical_plan_generator.hpp" directly, and those headers now live in `src/legacy.` This means builds with `ENABLE_LEGACY_SIRIUS=ON` will not find them.
Should this be `${CMAKE_CURRENT_SOURCE_DIR}` instead?

CMakeLists.txt
```diff
@@ -762,6 +761,9 @@ install(
   LIBRARY DESTINATION "${INSTALL_LIB_DIR}"
   ARCHIVE DESTINATION "${INSTALL_LIB_DIR}")
```
Jedi18, line 788: Could we add `COMPONENT sirius_library` to both header installation rules?
For example, this component-only installation:
```
cmake --install build --component sirius_library
```
would install the library and its CMake package files, but not the public headers required by consumers. A normal full installation still works because it installs all components; the problem only appears when `sirius_library` is installed by itself.

CMakeLists.txt
Jedi18: Could these rules use `${CMAKE_INSTALL_INCLUDEDIR}` instead of hard-coding include?
The exported Sirius target already tells consumers to look in `${CMAKE_INSTALL_INCLUDEDIR}`. If a packager customizes that directory, the headers would currently be installed in one location while consumers are told to search another.

9prady9: **This promotes `sirius_extension.hpp` to the installed public surface, but it can't be compiled from an install tree.**

It's one of the three headers this PR installs and the only one that isn't self-contained — `sirius/ffi.hpp` and `sirius/exception.hpp` include nothing but standard library headers, while this one opens with:

```cpp
#include "duckdb.hpp"                                    // line 19
#include "duckdb/function/function.hpp"
#include "duckdb/storage/statistics/node_statistics.hpp"
```

did we add find(duckdb) via find_dependency in sirius config ? otherwise this would fail.

CMakeLists.txt
```diff
@@ -1138,7 +1129,8 @@ endif()
 
 target_include_directories(
   parquet_benchmark
-  PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/include>)
+  PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
+          $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
```
9prady9: nit: consistency

```suggestion
  PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
          $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>)
```

.github/CODEOWNERS
```diff
@@ -8,13 +8,15 @@
 # Teams follow the pattern: @sirius-db/sirius-<component|language>
 
 # ── Core (Planner, Executors & Operators/Expressions) ─────────────────────────
+include/sirius/                     @sirius-db/sirius-core
+include/sirius_extension.hpp        @sirius-db/sirius-core
```
9prady9: The `src/include/*.hpp` glob these replace auto-owned every top-level public header. A literal single-file rule means the next header added under `include/` has no owner at all — with last-match-wins, nothing later in the file matches a `.hpp` there.

Leaves the tree this PR exists to declare as the public surface the only top-level source tree without directory-level review routing.

```suggestion
include/                            @sirius-db/sirius-core
```

.github/CODEOWNERS
```diff
@@ -8,13 +8,15 @@
...
+src/*.hpp                           @sirius-db/sirius-core
 src/creator/                        @sirius-db/sirius-core
-src/include/common/                 @sirius-db/sirius-core
-src/include/cudf/                   @sirius-db/sirius-core
-src/include/pipeline/               @sirius-db/sirius-core
-src/include/planner/                @sirius-db/sirius-core
+src/common/                         @sirius-db/sirius-core
+src/cudf/                           @sirius-db/sirius-core
+src/pipeline/                       @sirius-db/sirius-core
+src/planner/                        @sirius-db/sirius-core
```
9prady9, line 18: duplicates
```suggestion
```

src/compression/simpatico_codegen/CMakeLists.txt
```diff
@@ -182,8 +182,8 @@ target_compile_definitions(
 add_dependencies(simpatico simpatico_embedded_headers)
 target_include_directories(
   simpatico
-  PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src ${CMAKE_CURRENT_SOURCE_DIR}/include
-          "${CMAKE_CURRENT_BINARY_DIR}/generated" "${CUDA_TOOLKIT_INCLUDE}"
+  PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
```
9prady9: This swaps a curated 32-file `include/` surface for the merged 72-file `src/` tree, published PUBLIC — so `bridge/`, `jit/`, `plan/`, `util/` and the `.cu`/`.cpp` files now reach `sirius_extension`, `sirius_unittest` and `parquet_benchmark` (which doesn't reference simpatico at all). Every Sirius include into simpatico resolves under `api/` or `codegen/`, i.e. the tree being deleted, so the widening doesn't look required.


On another note: line 76's `include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include ...)` still points at the `include/` tree this PR deletes.

rust/README.md
```diff
@@ -45,7 +45,7 @@ dependency list:
 - **`--features static`** → `libsirius.a` (self-contained, no runtime deps — the
   fully static vcpkg build). Requires that bundled archive to exist.
 
-`build.rs` only needs `src/include` to compile the shim, because the bound
+`build.rs` only needs `include` to compile the shim, because the bound
 surface is the lightweight `sirius_ffi.h`. That header is the seed of the public
```
9prady9: the header name here is still `sirius_ffi.h` — doubly stale now (it's `ffi.hpp`, and the `.h` spelling never existed).

scripts/bootstrap-template.py
```diff
@@ -83,8 +83,8 @@ def replace_everywhere(to_find, to_replace):
     f"src/{string_to_find}_extension.cpp", f"src/{string_to_replace}_extension.cpp"
 )
 os.rename(
-    f"src/include/{string_to_find}_extension.hpp",
-    f"src/include/{string_to_replace}_extension.hpp",
+    f"src/{string_to_find}_extension.hpp",
+    f"src/{string_to_replace}_extension.hpp",
```
9prady9, line 87: Codex says the following: does it make sense ?

The mechanical `src/include/X` -> `src/X` rewrite landed here, but this PR's home for that header is `include/sirius_extension.hpp` — there's no `src/sirius_extension.hpp`, so this `os.rename` would raise after the script has already renamed the `.cpp` and rewritten CMakeLists/Makefile/README.

The script looks dead already (it opens by copying `docs/NEXT_README.md`, which doesn't exist), so deleting it may be simpler than fixing the path.

9prady9 (changes requested):

CMakeLists.txt
```diff
@@ -619,10 +619,11 @@ foreach(_target sirius_extension sirius_loadable_extension)
 
   target_include_directories(
     ${_target}
+    PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+           $<INSTALL_INTERFACE:include>
```
9prady9: Could this use `$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>` so the exported `sirius_extension` target points to where the headers are actually installed when the include directory is customized ?

Right now it goes into wrong location becoz of missing directive.

CMakeLists.txt
```diff
@@ -721,9 +722,8 @@ set_target_properties(
              CXX_STANDARD_REQUIRED ON
              CUDA_RESOLVE_DEVICE_SYMBOLS ON)
 target_include_directories(
-  sirius_shared
-  PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/include>
-         $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/sirius>)
+  sirius_shared PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+                       $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
```
9prady9, line 726: now that `sirius/exception.hpp` is installed as a public header which uses `std::format_string`, we should add `CXX_STANDARD 20` for `sirius_shared` so that it propagates to consumers.

9prady9 (commented):
couple of minor ones, rest lg

scripts/bootstrap-template.py
mbrobbel: https://github.com/sirius-db/sirius/pull/1816

9prady9 (approved):
looks good now, just have to resolve the conflicts. Unblocking from my end.

Thank you

CMakeLists.txt
```diff
@@ -785,6 +786,11 @@ install(
   LIBRARY DESTINATION "${INSTALL_LIB_DIR}"
   ARCHIVE DESTINATION "${INSTALL_LIB_DIR}")
 
+install(
```
aocsa, line 789: The `DIRECTORY include/sirius` install publishes `exception.hpp` next to `ffi.hpp`, but `ffi.hpp` never includes it. Its whole include list is five stdlib headers. `sirius::internal_exception` is engine-wide error plumbing, not the embedding API from #1751.

I installed the `sirius_library` component from a real build and confirmed `include/sirius/exception.hpp` lands in the prefix, and that a `find_package(sirius)` consumer can include it and link. Once it is in the prefix it is package ABI, `std::format_string` and all.

Keep the file on the build include path so in-tree TUs still compile, and install only `ffi.hpp` (`FILES_MATCHING PATTERN "ffi.hpp"`, or move the exceptions under `src/`).

aocsa: To reproduce: `cmake --install build/release --prefix /tmp/s1-prefix --component sirius_library
ls -1 /tmp/s1-prefix/include/sirius/`

Output:


exception.hpp
ffi.hpp

rust/crates/sirius-sys/build.rs
```diff
@@ -31,12 +31,12 @@ fn main() {
...
         .unwrap_or_else(|| repo.join("build/release"));
     let static_link = std::env::var_os("CARGO_FEATURE_STATIC").is_some();
-    let ffi_header = repo.join("src/include/sirius_ffi.hpp");
+    let ffi_header = repo.join("include/sirius/ffi.hpp");
 
     // 1. Compile the cxx glue against the public FFI header only.
     cxx_build::bridge("src/lib.rs")
         .std("c++20")
-        .include(repo.join("src/include"))
+        .include(repo.join("include"))
```
aocsa, line 39: `build.rs` still compiles against `repo/include` and links `$SIRIUS_BUILD_DIR`. That never exercises `find_package(sirius)` or the installed `include/sirius/ffi.hpp` this PR now exports.

I checked this against a real build: with the install prefix deleted, a source-tree consumer still built and passed, and only a prefix consumer failed. So a source-tree cxx build can stay green while a prefix install is unusable, and nothing in-tree would catch it.

Either teach `sirius-sys` an installed-prefix path, or add a one-file CMake consumer that does `find_package(sirius)` and `#include <sirius/ffi.hpp>`. The in-tree symlink stopgap can stay for developer builds.

Minor, same file: the comment at L17-L20 says a dedicated `libsirius` does not exist yet. After this PR the build produces `libsirius.so.0.0.0` under `extension/sirius`, so `resolve_lib_dir` takes `candidates[0]` and the symlink branch no longer runs on a normal build. Worth refreshing.

aocsa:
```
To reproduce: 
# create a real prefix
cmake --install build/release --prefix /tmp/prefix --component sirius_library

# delete it
rm -rf /tmp/prefix

# sirius-sys STILLs builds, it never looked at the prefix
cargo build --manifest-path rust/crates/sirius-sys/Cargo.toml          # green
```

aocsa (commented):
LG, I added a couple of suggestions.

Not this PR, flagging it here because this is when a packager would hit it: [sirius-config.cmake.in](https://github.com/sirius-db/sirius/blob/115329b81a338263824230a488c75e647ac33471/cmake/sirius-config.cmake.in#L1-L5) (from #1767) runs `check_required_components(sirius)` while every install rule uses `COMPONENT sirius_library` ([CMakeLists.txt L803-L819](https://github.com/sirius-db/sirius/blob/115329b81a338263824230a488c75e647ac33471/CMakeLists.txt#L803-L819)). Nothing sets `sirius_sirius_library_FOUND`, so `find_package(sirius REQUIRED COMPONENTS sirius_library)` sets `sirius_FOUND` to FALSE. 


One-line fix, in `cmake/sirius-config.cmake.in`:

    @PACKAGE_INIT@

    include("${CMAKE_CURRENT_LIST_DIR}/sirius-targets.cmake")

    set(sirius_sirius_library_FOUND TRUE)

    check_required_components(sirius)

I tested this against a real installed prefix. `find_package(sirius REQUIRED COMPONENTS sirius_library)` goes from exit 1 to `sirius_FOUND=1`, and a bogus `COMPONENTS does_not_exist` still fails correctly, so the guard keeps working.

rust/crates/sirius-sys/build.rs
mbrobbel: I'll do this in a follow-up

PR #1811, opened 2026-09-16
fix(scan): correct DICT_FSST decoding and validate string segment headers

ran-yuan-rui:
## Description

Fix `DICT_FSST` dictionary-index offsets to account for DuckDB's padded bitpacking layout, preventing string values from decoding as NULL. Preserve nullness when a dictionary contains only the NULL slot. Validate `DICTIONARY`, `FSST` and `DICT_FSST` headers on the host and reject invalid headers before decoding the affected run.

Tests cover DuckDB-written segments in all three `DICT_FSST` modes, malformed headers and NULL-only dictionaries. Validated on this branch, based on `dev` `31fb87b0`: `sirius_unittest` (3,261 cases) and `make s3-test` (92).

These commits were split out of #1764 at review request; #1764 keeps the provider-isolation change and does not depend on this PR.

## Checklist

- [ ] Update human and agent documentation (README.md, docs/, skills, CLAUDE.md) — N/A; no documentation changes required.

## References

Refs #1763. Refs #1760.


src/cuda/scan/strings/dict_fsst.cu
```diff
@@ -91,11 +91,104 @@ struct dict_fsst_pre_desc {
...
+};
+
+constexpr char const* DICT_FSST_CODEC_NAME = "DICT_FSST";
+
+//! A segment holds at most one row group of entries plus the NULL slot
+//! (DuckDB Storage::MAX_ROW_GROUP_SIZE = 2^30).
+constexpr uint64_t DICT_FSST_MAX_DICT_COUNT = (uint64_t{1} << 30) + 1u;
+
+//! Kernels form bit positions and row indices in 32-bit signed arithmetic.
+constexpr uint64_t KERNEL_INDEX_LIMIT = std::numeric_limits<int32_t>::max();
```
joosthooz: This is also defined in fsst.cu and dictionary.cu, maybe create 1 in a common header

src/cuda/scan/strings/dictionary.cu
```diff
@@ -53,6 +56,40 @@ struct dict_header_t {
...
+    fail("bitpacking_width " + std::to_string(hdr.bitpacking_width) + " > 32");
+  }
+  uint64_t const rows_end = uint64_t{seg.seg_row_start} + seg.row_count;
+  // The kernels form row and bit indices in 32-bit signed arithmetic.
+  constexpr uint64_t kernel_index_limit = std::numeric_limits<int32_t>::max();
+  if (rows_end > kernel_index_limit || rows_end * hdr.bitpacking_width > kernel_index_limit) {
+    fail("row or bit index for rows " + std::to_string(rows_end) +
+         " does not fit the kernels' 32-bit index arithmetic");
+  }
+  if (sizeof(dict_header_t) + packed_words_bytes(rows_end, hdr.bitpacking_width) >
```
joosthooz: Would this need `bitpacked_region_bytes` instead of `packed_words_bytes`?

src/cuda/scan/strings/fsst.cu
```diff
@@ -73,16 +76,53 @@ struct fsst_header_t {
...
+         " reaches past dict_end");
+  }
+  uint64_t const rows = seg.row_count;
+  // The kernels form row and bit indices in 32-bit signed arithmetic.
+  constexpr uint64_t kernel_index_limit = std::numeric_limits<int32_t>::max();
+  if (rows > kernel_index_limit || rows * hdr.bitpacking_width > kernel_index_limit) {
+    fail("row or bit index for rows " + std::to_string(rows) +
+         " does not fit the kernels' 32-bit index arithmetic");
+  }
+  if (sizeof(fsst_header_t) + packed_words_bytes(rows, hdr.bitpacking_width) >
```
joosthooz: Also here maybe `bitpacked_region_bytes` instead of `packed_words_bytes`

src/cuda/scan/strings/dictionary.cu
```diff
@@ -53,6 +56,40 @@ struct dict_header_t {
   uint32_t bitpacking_width;  ///< selection buffer
 };
 
+constexpr char const* DICTIONARY_CODEC_NAME = "DICTIONARY";
+
+//! Validate header bounds and row/bit index ranges before decoding.
+//! Payload index values remain device-side checks.
+void validate_dict_segment(size_t seg_idx,
```
joosthooz, line 63: This function is very similar to `validate_fsst_segment` and `validate_dict_fsst_segment` in other files, maybe they can be also be merged into a common header?

joosthooz (changes requested):
Thanks for this, just have a look at if we need to use packed_words_bytes

src/cuda/scan/strings/dict_fsst.cu
ran-yuan-rui: Done: `KERNEL_INDEX_LIMIT` is defined once in `common.cuh` and used by all three validators.

src/cuda/scan/strings/dictionary.cu
ran-yuan-rui: Yes. DuckDB sizes both regions with `BitpackingPrimitives::GetRequiredSize` (whole groups of 32 values), so the checks now use `bitpacked_region_bytes` like DICT_FSST does, and `packed_words_bytes` is removed. The synthetic segments now reserve at least that padding; the new tests reject regions that are too short.

src/cuda/scan/strings/fsst.cu
ran-yuan-rui: Yes. DuckDB sizes both regions with `BitpackingPrimitives::GetRequiredSize` (whole groups of 32 values), so the checks now use `bitpacked_region_bytes` like DICT_FSST does, and `packed_words_bytes` is removed. The synthetic segments now reserve at least that padding; the new tests reject regions that are too short.

src/cuda/scan/strings/dictionary.cu
ran-yuan-rui: Partly: the width check and the 32-bit index-range check are now shared helpers in `common.cuh`. The three validators stay separate because the headers and region layouts differ.

ran-yuan-rui:
Merged dev `a837c269` into the branch; the header-fetch helper and the new `prepare_dict`/`prepare_fsst` parameters now take `cuda::stream_ref`. The review points are in the last commit, `75707afa`.

joosthooz (approved):
Thank you!

PR #1812, opened 2026-09-16
fix: Update simpatico for RMM deprecations and README commands

jrbourbeau:
I was going through the simpatico README and found that (1) some of the commands were outdated and (2) I hit an RMM version bump-related error (`rmm::mr::set_current_device_resource_ref()` has been removed upstream in RMM). This PR fixes both these issues. 

Here's a concise summary of the build issue and proposed fix from Codex that I've edited myself

> ### Build issue
> The standalone Simpatico CLI and two tests failed against the upgraded RAPIDS/RMM stack because they still used the removed `rmm::mr::set_current_device_resource_ref()` API. The existing guards also saved the previous resource as a non-owning `device_async_resource_ref`, which can dangle after the current resource is replaced.
> 
> ### Fix
> Switch resource installation to `cudf::set_current_device_resource()`, retain its returned owning `cuda::mr::any_resource`, and move that resource back when restoring the previous allocator. This follows the ownership handling introduced in https://github.com/sirius-db/sirius/pull/877, the non-deprecated cuDF API migration in https://github.com/sirius-db/sirius/pull/1098, and the existing RAII pattern from https://github.com/sirius-db/sirius/pull/1205.
> The change here update the Simpatico CLI, operator-sweep test, and compression round-trip test. The build and CLI round-trip verification now pass.

cc @felipeblazing


mbrobbel (approved):
Thanks @jrbourbeau

PR #1815, opened 2026-09-17
perf(scan): Coalescing S3 GETs in Amin's stacked I/O PR when cache is disabled

joosthooz:
## Description
This improves S3 throughput when running with cache disabled (this is on top of Amin's stack PR).

## Checklist
- [ ] Read CONTRIBUTING.md and ensure PR meets "reviewability" checklist

## References


src/include/io/templated_ioctx.hpp
```diff
@@ -281,196 +194,229 @@ class templated_ioctx : public sirius_ioctx {
...
+                 slices[end].rng.offset <= slices[end - 1].rng.offset + slices[end - 1].rng.size) {
+            run_bytes += slices[end].size();
+            ++end;
+          }
+          auto const smallest = static_cast<std::size_t>(
+            std::min_element(partition_bytes.begin(), partition_bytes.end()) -
+            partition_bytes.begin());
+          partition_bytes[smallest] += run_bytes;
+          for (std::size_t i = begin; i < end; ++i) {
+            partitions[smallest].push_back(slices[i]);
```
ran-yuan-rui, line 359: What happens: with the default two reactors, a request whose slices form one contiguous run makes the query hang. `SELECT * FROM region` on the tiny TPC-H fixtures over MinIO (cache `mode: none`) prints its rows and then hangs at exit in `rest_reactor::shutdown()` → `std::thread::join()` on `a55d7444`, and passes on `44d30f47`; `SELECT r_regionkey, r_name FROM region` (two adjacent chunks) hangs the same way; with `rest_n_reactors: 1` the three-chunk case passes, and `SELECT r_regionkey, r_comment FROM region` (two non-adjacent runs) passes with two reactors. The repository's `[s3][integration][sql][tpch]` tiny case stops at Q2 on this head right after nation's three chunks are served as one GET (`bytes=4-481`), with no further request in the MinIO trace; that case is tagged `[.]` and the default unit run does not select it, which is why the checks are green.

Why: the run-level routing in `templated_ioctx.hpp` gives the whole run to the lightest partition (:343–:361) while `partition_count = min(reactors, slices)` (:332) still creates two, and :366–:374 builds and enqueues every partition, including the empty one. The REST worker never makes progress on an empty group: `expand_active` returns at :1318 without resetting `active_group`, so `next_fresh` (:1425–:1453) loops forever with nothing pending, a non-null active group, no error and no stop-token check inside the loop. The reactor that received the run completes normally, which is why the rows still come out; the other reactor spins, so the next request routed to it hangs the query, and otherwise shutdown never joins. uring resets an empty active group (`uring_reactor.cpp:1019`), so only the REST worker spins, but the empty partition is produced for every `templated_ioctx` backend and does not need a fuse to happen.

Fix and test: do not enqueue an empty partition, and keep the REST worker making progress on an empty group; the regression test needs two reactors plus a single contiguous run, and must check that a following request completes and that shutdown returns, not only the result rows.

joosthooz:
This should be revisited and retargeted to dev after merging 1771.

src/include/io/templated_ioctx.hpp
ran-yuan-rui: Thanks, this fixes it: dispatch skips empty partitions, the REST worker clears an empty group, and `next_fresh()` checks for stop. The new test with two reactors and one contiguous run checks that the next request completes and that `shutdown()` returns, which were the two things I asked for. Resolving.

PR #1817, opened 2026-09-17
refactor: migrate streams to `cuda::stream_ref`

mbrobbel:
Migrate to `cuda::stream_ref`. This fixes our nightly build and also gets rids of the deprecation warnings.


jrbourbeau (approved):
Thanks @mbrobbel. Other than the linting error this LGTM

utils/stream_check/README.md
```diff
@@ -85,7 +85,7 @@ int main() {
 │  │   if (g_throw_on_default_stream) {        │  │
 │  │     throw runtime_error(...);             │  │
 │  │   }                                        │  │
-│  │   return rmm::cuda_stream_view{};         │  │
+│  │   return ::cuda::stream_ref{cudaStream_t{}};         │  │
```
johanpel, line 88: Nit: fix dingen.txt art here

johanpel (approved):
LGTM with one nit and one general question.

Is there something that pollutes the namespaces used in Sirius with some other definition of `cuda` which requires us to prefix everything with ::? If so perhaps we can change the name of whatever that is?

src/compression/simpatico_codegen/tests/test_selection_wave.cpp
Jedi18, line 137: We should change this (and other `value()` calls at 181, 192, 240, and 313) to `.get()`

CI doesn't seem to build these tests hence why it didn't show up as an error during compilation

PR #1840, opened 2026-09-19
fix(join): keep a MARK join alive when polled before sizing; materialize inequality sides

morningman:
## Why

Two failures that only the Substrait/FFI path (`sirius_ffi.cpp` → `sirius_physical_plan_generator`) reaches. The transparent DuckDB-extension path plans the same SQL differently — DELIM_JOIN keeps the inequality in a filter, and the optimizer puts the `NOT IN` mark join on top — so it never hit them. Found by running TPC-H through the FFI path (the plans an Apache Doris FE produces, see #1842) on a T4.

## What

**`src/op/sirius_physical_hash_join.cpp` — MARK join polled before sizing (TPC-H Q16, process abort).**
A MARK join whose probe pipeline is the source of an inner join's probe side gets polled by the task creator (`WAITING_FOR_INPUT_DATA` walk) before either of its input partitions negotiated `BUILD_PROBE`. `get_next_task_hint` then takes the STANDARD/MIXED branch and `refresh_cross_schedule` throws its MARK invariant from the task-creator thread, which terminates the process. A MARK join always ends up in `BUILD_PROBE`, so when polled in that window there is nothing to schedule yet: return `WAITING_FOR_INPUT_DATA` on the build producer instead, whose tasks feed the build partition and size the join.

**`src/planner/sirius_plan_comparison_join.cpp` — DECIMAL on an inequality join side (TPC-H Q17/Q20, runtime error).**
`materialize_expression_join_keys` skipped inequality conditions entirely, leaving both sides to the mixed join's inline cuDF AST predicate. cuDF AST can neither cast to nor compute with DECIMAL, so `cast(l_quantity as decimal(38,5)) < 0.2 * avg(l_quantity)` failed with *"failed to translate mixed join inequality conditions to cuDF AST predicate"*. Inequality sides now go through the same materialization as routed null-safe keys: anything beyond a column reference or an AST-supported cast is projected into a column below the join, leaving a plain column comparison for the predicate. Equality keys are unchanged.

Net: +23/−7, no header or public-API change, no behaviour change for plans that already worked (the MARK early-return only fires in the previously-throwing window; materialization only kicks in for expressions the AST could not evaluate).

## Verification

- TPC-H SF1, all 22 queries, through the FFI path **and** through the transparent path: results match the DuckDB baseline before/after (the transparent path exercises the unchanged code paths for regression).
- Q16 no longer aborts; Q17/Q20 no longer fail at runtime on the FFI path.
- No new unit test in this PR: the failing shapes only come out of the Substrait consumer. Happy to add a `test/cpp/exec/test_sirius_ffi_fragment.cpp`-style case if a reviewer wants one — say so and I'll push it.

## Relationship to other PRs

Independent of the other two PRs from this branch split. #1842 (`experimental/doris`) needs this one at runtime for Q16/Q17/Q20 but does not build against it.


morningman:
## Reviewer walkthrough

The description above is deliberately terse. This comment spells out both failures step by step — the SQL, the plan shape that triggers each, and the exact sequence that fails — so the two hunks can be checked against a concrete scenario.

### Why only the FFI path

Sirius has two entry points, and they produce different plan shapes for the same SQL:

- **Transparent path** — SQL goes through DuckDB's optimizer first, then into Sirius.
- **FFI/Substrait path** — an external planner (here an Apache Doris FE, #1842) builds the plan and hands it to `sirius_physical_plan_generator` as Substrait.

Both code paths touched here have always existed; DuckDB's optimizer just never produces the shapes that reach them, and the Doris FE does. Details per fix below.

---

### Fix 1 — `src/op/sirius_physical_hash_join.cpp`: MARK join polled before sizing (Q16, process abort)

**What a MARK join is.** For `x NOT IN (subquery)` / `x IN (subquery)` the planner does not materialize a join result; it tags each outer row with a three-valued mark (true / false / NULL — "did the subquery contain a match?") and a Filter above consumes the mark. That tagging join is `JoinType::MARK`.

**The SQL** (TPC-H Q16, trimmed):

```sql
select p_brand, p_type, p_size, count(distinct ps_suppkey)
from partsupp, part
where p_partkey = ps_partkey
  and p_size in (49, 14, 23, 45, 19, 3, 36, 9)
  and ps_suppkey not in (
      select s_suppkey from supplier where s_comment like '%Customer%Complaints%')
group by p_brand, p_type, p_size;
```

**The plan the Doris FE produces** — the MARK join sits *below* the inner join, on its probe side:

```
Inner Join [p_partkey = ps_partkey]
  ├─ probe: Filter [NOT mark]
  │           └─ MARK Join [ps_suppkey = s_suppkey]
  │                ├─ probe: partsupp
  │                └─ build: supplier WHERE s_comment LIKE '%Customer%Complaints%'
  └─ build: part WHERE p_brand <> 'Brand#45' AND p_size IN (...)
```

**How it aborts.** The task creator polls each operator through `get_next_task_hint()`; the answer is `READY`, `WAITING_FOR_INPUT_DATA` plus a producer to walk into, or "done". A MARK join must run in `BUILD_PROBE` mode (it needs the whole build side resident for the global `build_has_null` sentinel; `get_partition_strategy` forces `build_probe = true` for MARK). But that mode is only entered when one of the join's input PARTITION operators negotiates the strategy on its first task (`sirius_physical_partition::get_next_task_input_data` → `consumer->get_partition_strategy`). Until then `_join_mode` sits at its default, `HASH_JOIN_MODE::STANDARD`.

1. The inner join is polled, has no probe data yet, answers `WAITING_FOR_INPUT_DATA` on its probe producer.
2. The task creator walks into the probe pipeline and polls the MARK join.
3. Neither of the MARK join's inputs has produced a batch, so `_join_mode == STANDARD`.
4. `get_next_task_hint` takes the `else` (STANDARD/MIXED) branch and calls `refresh_cross_schedule()`.
5. `refresh_cross_schedule` has a defensive check — "MARK must never reach this path" — and throws.
6. The throw happens on the task-creator thread, nothing catches it, and the process terminates.

**The fix.** At the top of the STANDARD/MIXED branch: if this is a MARK join, return `WAITING_FOR_INPUT_DATA` on the build producer. There is nothing to schedule for a MARK join in that window anyway. The build producer's tasks feed the build PARTITION, whose first task negotiates the strategy and flips the join into `BUILD_PROBE`; every later poll takes the existing `BUILD_PROBE` branch.

**Scope.** For a MARK join, `_join_mode` is `STANDARD` before sizing and `BUILD_PROBE` after (MARK + MIXED is rejected at construction, and `get_partition_strategy` forces BUILD_PROBE). So the new early return fires exactly in the window that used to throw, and never otherwise.

**Why the transparent path never hits it.** DuckDB's optimizer puts the `NOT IN` MARK join at the *top* of the plan (`partsupp ⋈ part` first, `NOT IN` last), so the MARK join is never the source of another join's probe pipeline and is never polled before its own inputs have run.

---

### Fix 2 — `src/planner/sirius_plan_comparison_join.cpp`: DECIMAL on an inequality join side (Q17/Q20, runtime error)

**The SQL** (TPC-H Q17):

```sql
select sum(l_extendedprice) / 7.0 as avg_yearly
from lineitem, part
where p_partkey = l_partkey
  and p_brand = 'Brand#23'
  and p_container = 'MED BOX'
  and l_quantity < (
      select 0.2 * avg(l_quantity)      -- DECIMAL arithmetic
      from lineitem
      where l_partkey = p_partkey);     -- correlated
```

**The plan the Doris FE produces.** The correlated subquery is decorrelated into a grouped aggregate joined back to the outer query, and the comparison arrives as a *join condition* next to the equality key:

```
Join [ l_partkey = agg.l_partkey                                        -- equality
       AND cast(l_quantity AS decimal(38,5)) < 0.2 * agg.avg_qty ]      -- inequality, DECIMAL
  ├─ lineitem ⋈ part (filtered)
  └─ Aggregate [l_partkey => avg(l_quantity)] over lineitem
```

**Why it failed.** A join with equality + inequality conditions runs as a cuDF mixed join: the equality keys go through the hash table, and the inequality conditions are evaluated inline, per candidate pair, as a cuDF AST predicate. cuDF AST can neither cast to DECIMAL nor multiply DECIMALs, so translating `cast(l_quantity as decimal(38,5)) < 0.2 * avg_qty` failed with

```
failed to translate mixed join inequality conditions to cuDF AST predicate
```

Before this PR, `materialize_expression_join_keys` skipped inequality conditions with `continue` and left both sides intact for the AST.

**The fix.** Inequality sides now go through the same materialization as routed null-safe keys. `is_trivial_key_side(expr, /*as_ast_predicate=*/true)` accepts only a column reference, or a cast of a column reference whose target type is in `supported_ast_cast_types`; anything else is projected into a new column below the join — an ordinary projection, the same evaluation path a SELECT-list expression takes, which does handle DECIMAL — and the condition is rewritten to reference that column. The AST then only compares two columns.

```
before:
  MixedJoin [ l_partkey = a.l_partkey AND cast(l_quantity, dec(38,5)) < 0.2 * avg ]   -- AST cannot evaluate
    ├─ lineitem ⋈ part
    └─ Aggregate

after:
  MixedJoin [ l_partkey = a.l_partkey AND $k1 < $k2 ]                                 -- column vs column
    ├─ Project [..., cast(l_quantity AS decimal(38,5)) AS k1]
    │    └─ lineitem ⋈ part
    └─ Project [..., 0.2 * avg AS k2]
         └─ Aggregate
```

In SQL terms the rewrite is:

```sql
-- before
... on a.k = b.k and cast(a.qty as decimal(38,5)) < 0.2 * b.avg_qty

-- after
with a2 as (select *, cast(qty as decimal(38,5)) as k1 from a),
     b2 as (select *, 0.2 * avg_qty as k2 from b)
... on a2.k = b2.k and a2.k1 < b2.k2
```

Q20 is the same shape with `ps_availqty > 0.5 * sum(l_quantity)`.

**Scope.** Equality keys take the unchanged path (`as_ast_predicate` is false for them unless routed null-safe, exactly as before). For inequality sides, a plain reference or an AST-supported cast is still trivial and untouched; only sides the AST could not have evaluated anyway get a projection.

**Why the transparent path never hits it.** DuckDB decorrelates this subquery with a DELIM_JOIN and keeps the `<` in a Filter *above* the join rather than as a join condition; that Filter goes through Sirius's expression evaluation, which supports DECIMAL.

---

### Summary

| | Q16 | Q17 / Q20 |
|---|---|---|
| Symptom | process abort | runtime error |
| Trigger shape | MARK join is the source of another join's probe pipeline | inequality join condition with a DECIMAL cast / arithmetic |
| Root cause | polled before sizing → STANDARD branch → `refresh_cross_schedule` MARK invariant throws on the task-creator thread | inequality side handed to cuDF AST, which has no DECIMAL support |
| Fix | answer `WAITING_FOR_INPUT_DATA` on the build producer until sized | project the side into a column below the join; the predicate becomes a column comparison |
| Effect on plans that already worked | none — only fires in the window that used to throw | none — only fires for sides the AST could not evaluate |

src/op/sirius_physical_hash_join.cpp
```diff
@@ -1082,6 +1082,17 @@ std::optional<task_creation_hint> sirius_physical_hash_join::get_next_task_hint(
...
     }
   } else {
+    // A MARK join always runs in BUILD_PROBE mode, which get_partition_strategy selects when one
+    // of its input partitions negotiates the strategy on its first task. Polled before that — a
+    // downstream join walking WAITING_FOR_INPUT_DATA into this join as the source of its probe
+    // pipeline before either input has produced a batch (e.g. an anti-join subquery on the probe
+    // side of an inner join) — there is nothing to schedule yet, and refresh_cross_schedule would
+    // report the STANDARD-mode MARK as an invariant violation. Defer to the build producer instead:
+    // its tasks feed the build partition, which sizes this join into BUILD_PROBE.
+    if (join_type == duckdb::JoinType::MARK) {
```
Jedi18, line 1092: We should also handle a build pipeline that produces no batches. If the build pipeline produces no batches (for example, when LIMIT … OFFSET skips every row), the MARK join never gets initialized. This branch keeps sending the scheduler back to the already-finished build pipeline, leaving probe rows waiting indefinitely.

I guess there are two possible fixes: explicitly handle the empty build by emitting false marks for probe rows, or synthesize a zero-row build batch and run normal join initialization.

Jedi18 (commented):
Thanks a lot for the PR and the fixes! Besides the below comment, I would also suggest making the comments more readable

morningman:
Hi @Jedi18 , review fixed. PTAL

Jedi18 (approved):
LGTM thanks!

PR #1841, opened 2026-09-19
perf(ffi): cache parquet metadata across binds, honor SIRIUS_LOG_*, log lowering/plan/execute timings

morningman:
## Why

Three gaps on the FFI path (`sirius_ffi.cpp`, the embedded engine used by `rust/crates/sirius-sys` and the `experimental/` backends), all found while profiling TPC-H SF100 through it:

1. **Substrait lowering re-parsed parquet footers on every bind.** The Substrait consumer builds the plan through the Relation API, which re-binds the whole subtree at every level, so a parquet read is bound once per operator above it. With the metadata cache off each bind re-parses the footer and re-derives the column statistics. On a 25 GB, ~5k-row-group SF100 `lineitem`, lowering took 1.3 s for Q6 and **10 s for Q21 — as long as or longer than the GPU execution (1.3 s / 4.6 s)**.
2. **No log output.** The embedded DuckDB never loads the Sirius extension, so nothing on this path read `SIRIUS_LOG_{BACKEND,DIR,LEVEL}` the way `SiriusContextExtensionCallback` does on the transparent path; the engine ran with the noop sink and the host process could not get a Sirius log.
3. **No per-phase timing.** The host only sees the total `execute_substrait` time, and the telemetry query window covers execution alone, so time spent in lowering/planning (like the 10 s above) was invisible.

## What

- `bring_up()`: set `parquet_metadata_cache = true` on the embedded DuckDB **at the database level** (`DBConfig::SetOptionByName`). It has to be database-level: the consumer's binder does not see session-level `SET` variables. This DuckDB instance is private to the engine and the cache validates file mtimes, so there is no staleness hazard for the host. Lowering drops to 55–220 ms on the cases above.
- `install_log_sink_from_env()`: at bring-up, if any of `SIRIUS_LOG_BACKEND` / `SIRIUS_LOG_DIR` / `SIRIUS_LOG_LEVEL` is set, copy them into `duckdb::Config::LOG_*` and call `install_configured_log_sink(nullptr)` (best-effort, an unknown backend is ignored rather than failing bring-up). With none of the variables set the sink is left untouched, i.e. behaviour is unchanged by default.
- `execute_substrait()`: time the lowering / physical-planning / execution phases and emit one `SIRIUS_LOG_INFO` line per query.

+52/−1, `sirius_ffi.hpp` unchanged, no API change. Applies to every FFI consumer (StarRocks backend, `sirius-sys`), all in the "faster and observable" direction.

## Verification

- TPC-H SF1 22/22 through the FFI path unchanged vs the DuckDB baseline.
- SF100 on an L40S (`g6e.8xlarge`): per-query engine time drops by the lowering delta (Q21 10 s → ~0.2 s lowering); the new log line is what exposed the split.
- Existing `test/cpp/exec/test_sirius_ffi_fragment.cpp` covers bring-up/execute on this path; no new test since the change is configuration + logging.

## Relationship to other PRs

Independent of the join fix (#1840). #1842's benchmark numbers were taken with this change in place.


morningman:
## Reviewer's guide

The description is the summary; this is the long-form walkthrough with a worked example, for anyone who has not been down the FFI path before.

### Which path this touches

Sirius has two entry points:

- **Transparent path** — `LOAD sirius` inside DuckDB; the extension intercepts normal SQL and runs it on the GPU. Not touched by this PR.
- **FFI path** (`src/sirius_ffi.cpp`) — the engine is linked into a host process (`rust/crates/sirius-sys`, the `experimental/` backends). There is no SQL text: the host hands over a Substrait plan, the engine lowers it through a *private* embedded DuckDB into a Sirius GPU plan and executes it.

Everything below is FFI-only. `sirius_ffi.hpp` and the API are unchanged.

### 1. Parquet footers were re-parsed once per operator above the scan

**What a footer is.** A parquet file ends with a footer: the list of row groups, their byte offsets, per-column min/max statistics. DuckDB has to parse it before it can plan a read. For the SF100 `lineitem` (25 GB, ~5k row groups) one parse plus statistics derivation is ~0.5 s. Parsing it once is fine; the problem is how often it happened.

**Why it happened many times.** The Substrait consumer builds the DuckDB plan through the Relation API, wrapping one relation in the next:

```
scan(lineitem)                             → bind, parse footer
filter(scan(lineitem))                     → re-bind the whole subtree, parse footer again
project(filter(scan(lineitem)))            → again
aggregate(project(filter(scan(lineitem)))) → again
...
```

Each wrap re-binds the whole subtree underneath it, so a scan with N operators above it has its footer parsed N times. Q21 has many operators above `lineitem`:

| | Q6 | Q21 |
|---|---|---|
| lowering (Substrait → optimized `LogicalOperator`) | 1.3 s | **10 s** |
| GPU execution | 1.3 s | 4.6 s |

**The fix** (the `SetOptionByName` hunk in `bring_up()`): turn on `parquet_metadata_cache` on the embedded DuckDB, so the footer is parsed once and reused across binds. Lowering drops to 55–220 ms (Q21: 10 s → ~0.2 s).

Two things worth checking:

- It is set at the **database** level (`DBConfig::GetConfig(*db->instance).SetOptionByName`), not with a session `SET`. The consumer's binder does not see session-level variables, so a `SET` would silently do nothing.
- Staleness: the DuckDB instance is private to the engine (the host never touches it), and DuckDB's metadata cache validates file mtimes, so a rewritten file is re-parsed.

### 2. The FFI path never installed a log sink

On the transparent path, `SiriusContextExtensionCallback` runs when DuckDB loads the extension; it reads `SIRIUS_LOG_BACKEND` / `SIRIUS_LOG_DIR` / `SIRIUS_LOG_LEVEL` and installs a log sink. The embedded DuckDB on the FFI path never loads the extension (the engine is linked in directly), so that callback never ran and the engine stayed on the noop sink — the host process could not get a Sirius log at all.

**The fix** (`install_log_sink_from_env()`, called first thing in `bring_up()`): if any of the three variables is set, copy them into `duckdb::Config::LOG_*` and call `install_configured_log_sink(nullptr)`, the same routine the transparent path's callback calls. With none set it returns before touching anything, so the default is unchanged.

Differences from the transparent path, on purpose:

- Best-effort: an unknown backend or level is ignored (the callback throws). Failing engine bring-up over a log setting seemed worse than running without a log.
- `install_configured_log_sink(nullptr)` can only install `spdlog` and `noop`; the `duckdb` backend needs a `DatabaseInstance` and is left alone. That matches how the callback behaves before a db exists.

Example:

```bash
SIRIUS_LOG_BACKEND=spdlog SIRIUS_LOG_DIR=/tmp/sirius-log SIRIUS_LOG_LEVEL=info <host-binary>
```

Before: no Sirius output anywhere. After: the log lands under `/tmp/sirius-log`.

### 3. No per-phase timing

`execute_substrait()` has three phases:

1. **lower** — Substrait → optimized DuckDB `LogicalOperator`
2. **plan** — `LogicalOperator` → Sirius GPU physical plan
3. **execute** — GPU execution

The host only sees the total, and the telemetry query window covers phase 3 alone. That is why the 10 s in item 1 was invisible: 14.6 s total, 4.6 s inside the window, and nothing said where the rest went.

**The fix**: time each phase and emit one line per query:

```
[sirius_ffi] execute_substrait: lower 10234.5 ms, plan 12.3 ms, execute 4600.0 ms
```

This line is what exposed item 1 (once item 2 made it visible).

### Suggested reading order

`install_log_sink_from_env()` → the two new lines in `bring_up()` (sink install, `SetOptionByName`) → the timing hunks in `execute_substrait()`. The three are independent. With no `SIRIUS_LOG_*` set, the only behavioural change is the metadata cache; the timing line goes to the noop sink.

src/sirius_ffi.cpp
```diff
@@ -59,6 +63,30 @@ constexpr duckdb::idx_t kArrowBatchSize = 1u << 20;
...
+  auto const* backend_env = std::getenv("SIRIUS_LOG_BACKEND");
+  auto const* log_dir_env = std::getenv("SIRIUS_LOG_DIR");
+  auto const* level_env   = std::getenv("SIRIUS_LOG_LEVEL");
+  if (!backend_env && !log_dir_env && !level_env) { return; }
+  if (backend_env) { duckdb::Config::LOG_BACKEND = backend_env; }
+  if (log_dir_env) { duckdb::Config::LOG_DIR = log_dir_env; }
+  if (level_env) { duckdb::Config::LOG_LEVEL = level_env; }
+  // Best-effort (null db): an unknown backend is ignored rather than failing bring-up.
+  duckdb::install_configured_log_sink(nullptr);
+}
```
ran-yuan-rui: The best-effort claim only covers an unknown backend. The spdlog path builds `daily_file_sink_mt` in its constructor, and `make_spdlog_owning_sink` documents that construction failures propagate, so an unwritable `SIRIUS_LOG_DIR` makes `Context::Context()` throw after the supplied `Config::LOG_*` values were already updated; the transparent path snapshots and restores them before rethrowing. Please pick a policy (restore the values and keep the previous sink, or restore and rethrow) and pin it with an `[isolated_context]` case in `test_sirius_ffi_fragment.cpp`, which already constructs a `Context` per case and runs in the default CI suite: point `SIRIUS_LOG_DIR` below a regular file (fails even as root), put the environment, `Config::LOG_*` and `sirius::log::get_sink()` behind a scope guard so they are restored on every exit, then assert the outcome and the restored state. A writable-directory case is optional; it can flush the sink and check `sirius.log` for the `SiriusContext: topology summary` bring-up line. No C++ test invokes `execute_substrait`, so the timing record can only be covered from the Rust GPU test.

The trigger is also wider than described: `LOG_BACKEND` defaults to `spdlog` and `LOG_DIR` to `log`, so `SIRIUS_LOG_LEVEL` alone installs a file sink under `./log` and `SIRIUS_LOG_DIR` alone installs spdlog there; an invalid level falls back to `info` with a warning rather than being ignored; `SIRIUS_LOG_BACKEND=duckdb` cannot install through the null-database call, and the FFI has no SQL or settings entry point to enable or read DuckDB logging. Please document these in `configuration.md`.

ran-yuan-rui (changes requested):
The metadata cache is safe here. `ParquetExtension` registers the option before `SetOptionByName` stores it, `MetadataCacheEnabled` reads it through the connection, cache hits are validated by version tag or mtime under the default `VALIDATE_ALL` policy (a local file modified less than ten seconds ago is re-read even when its mtime matches), and the partition-statistics callback returns nothing unless every entry is `VALID`. One qualification: `MultiFileCardinality` uses cached row counts without checking validity, so disabling `STATISTICS_PROPAGATION` does not keep cached metadata out of cardinality estimates and join ordering. That is a costing concern, not a correctness one; the GPU reader parses and caches its own cuDF footer. The repeated binds are real: each Relation wrapper the Substrait consumer builds calls `TryBindRelation`, and `Planner::CreatePlan` binds the final `RelationStatement` again.

One blocking item: define the sink-construction failure policy, restore the process-global logging state on failure, and cover that path with an isolated-context test, details inline. Two non-blocking remarks: the comment's reason for setting the option at database level does not match how settings resolve (a session `SET` on `conn` is visible to these binds, since they run on the same `ClientContext`; "every connection on the private instance inherits it" is the accurate reason), and the timing line is emitted only on the success path, so a failing query logs nothing. The documentation request is inline as well.

ran-yuan-rui (approved):
LGTM. The logging rollback path is fixed and covered by the new isolated test. Thanks for addressing it.

The PR description still has a few stale details, including the session `SET` rationale and "no new test."

Issue #137, opened 2026-01-03
[Doris] Progress on Apache Doris


dataroaring:
Hi sirius-db teams,

You're doing impressive work, and the results speak for themselves. As a member of the Apache Doris PMC, I’d be happy to offer any support or resources you might need from the project side—just let me know.

roaramburu:
Hey @dataroaring ,

We would be thrilled to partner together and invest more into SiriusDB + Apache Doris.

We can discuss further via the Sirius Slack Channel - https://join.slack.com/t/sirius-db/shared_invite/zt-3nldvrf0h-ehsdnGucSLr57Odi2lB8ZA

Please join and reference this issue, and we'll get a thread started.

roaramburu:
Hey @dataroaring, we're deprioritizing Doris support for the foreseeable future. If you would like to push this forward, please re-open the issue and we'd be happy to collaborate.

morningman:
Hi @roaramburu — picking this up from the Doris side, as you suggested when closing.

We have built a Sirius backend for Apache Doris and would like to bring it upstream: **#1842** (draft) adds `experimental/doris`, a Rust process that joins an unmodified, official Doris 4.1.4 FE as a backend node, translates the plan fragments the FE dispatches into one Substrait plan and runs it on the Sirius engine through the existing FFI path (`sirius_ffi.cpp`) — the same shape as `experimental/starrocks`. Doris itself is not modified; the integration is configuration only (an FE config file, session variables, views over `local()`).

Where it stands:

- All 22 TPC-H queries run on the GPU through the stock FE and validate row by row against a DuckDB baseline at SF1, SF10 and SF100 — 2,024 query executions across every scenario and round, zero mismatches.
- On one L40S (`g6e.8xlarge`) at SF100, Doris's native BE reading parquet vs the same FE handing that parquet to Sirius: geometric-mean speedup 3.01×, 22-query total 3.80× (154.4 s → 40.6 s), all 22 queries faster; 1.63× at SF1 and 1.77× at SF10. The full report — scenarios, method, versions under test, per-query numbers, limitations — is `experimental/doris/docs/tpch-report.md` in the PR.
- The protocol shell and the plan translator build and test without a GPU (`--no-default-features`; a `doris` job is added next to `starrocks` in the `experimental` workflow). The engine-linked build runs on a GPU box.
[...]

PR #1842, opened 2026-09-19
feat(experimental): Doris pseudo-BE — Sirius as an Apache Doris backend, TPC-H 22/22 on the GPU

morningman:
> **Draft.** Opened as a draft so the description and the review guide below can settle first; see *Before "Ready for review"* at the bottom.

Refs #137 (the Doris-support tracking issue).

## Why

Sirius as a backend for **Apache Doris**: a Rust process that joins an unmodified, official Doris FE as a backend node, receives the FE's plan fragments, translates them to Substrait and runs them on the embedded Sirius engine through the existing FFI path (`sirius_ffi.cpp`). The Doris side is configuration only — no fork of Doris, no patch to the FE. It lives under `experimental/doris`, structured like `experimental/starrocks`: a protocol shell whose every piece builds and tests without an engine or a GPU (`--no-default-features`, what CI and a laptop run), plus the engine-linked build for a GPU box.

Where it stands: **all 22 TPC-H queries run on the GPU through a stock Doris FE and validate against a DuckDB baseline** (SF1 and SF10 baselines are in the tree, SF100 generated on the box). The full report, with the version under test and the test date, is `experimental/doris/docs/tpch-report.md`. On an L40S (`g6e.8xlarge`) at SF100, Doris reading parquet with its native BE vs Doris handing the same parquet to this backend: geometric-mean speedup 3.01×, 22-query total 3.80× (154.4 s → 40.6 s), all 22 queries faster. The advantage grows with scale (1.63× at SF1, 1.77× at SF10) because at small scale the FE's planning and two-phase fragment dispatch dominate.

## What's in the tree (304 files, but most are fixtures)

| area | files | what |
|---|---|---|
| `src/` | 11 | the backend: self-registration + heartbeat (`node.rs`), the `PBackendService` gRPC shell (`backend_service.rs`), dispatch decoding (`params.rs`), analysis-time `local()` schema support (`file_schema.rs`), result store/encoder, `fragment_executor.rs`, `engine.rs` (FFI) |
| `crates/doris-plan-translator/` | 10 | `TPipelineFragmentParams` → Substrait: descriptor table, type gate, expression and plan-node translators, the single-plan stitcher (`translate_batch`), `explain` |
| `crates/doris-thrift`, `crates/doris-proto` | 6 | build-time codegen (thrift 0.22 `--gen rs`, prost/tonic) over the submodule IDL — no generated sources committed |
| `tests/corpus.rs` + `tests/fixtures/` (137) + `tests/snapshots/` (27) | | captured FE→BE dispatches for the 22 TPC-H queries and the semantic-gap probes, and the reviewed translation of each; `corpus.rs` replays and diffs them (10 tests, runs in CI) |
| `tests/expected/` | 49 | DuckDB-computed TPC-H results at SF1 and SF10 (`.tsv`, gzipped where large — Q11 at SF10 is 1.3 MB) that `run-tpch.sh` and the CPU differential validate against |
| `scripts/` | 18 | fetch the official FE/BE tarballs, start/stop, `run-tpch.sh`, the result validator, the DuckDB CPU differential of the translator (`cpu-diff.sh`), and the benchmark harness (`bench.sh`, `bench-all.sh`, `bench-report.py`, sampler, FE audit join, telemetry reader) |
| `docs/tpch-report.md` | 1 | the correctness and performance report of the SF1 / SF10 / SF100 runs: scenarios, method, per-query tables, and the exact version under test (upstream `dev` + #1840 + #1841 + this PR, submodule pins, Doris 4.1.4, driver) and test date |
| `sql/`, `conf/` | 35 | FE config, session defaults, the TPC-H views over `local()`, the 22 queries, the gap probes; `sirius.yaml` sizing for a shared box |
| `doris/` submodule | | `apache/doris` pinned to the FE release commit, **shallow, only `gensrc/thrift` and `gensrc/proto` are used** (the FE is downloaded as the official 4.1.4 binary, never built) |

Hand-written Rust is ~14.8k lines (`src/` + translator + `corpus.rs`), scripts ~2.7k.

Outside the directory: `.gitmodules` (the submodule) and `.github/workflows/experimental.yml` (a `doris` job next to `starrocks`: fmt / clippy `-D warnings` / `cargo test --workspace`, all `--no-default-features` on a CPU runner; the `experimental` summary job now needs both). `CONDA_OVERRIDE_CUDA=13` is set because the pixi platforms are CUDA-pinned even though the `be` env itself is CUDA-agnostic.

## Suggested review order

1. `README.md` — architecture diagram, layout, the three ways to run it (laptop / CPU differential / GPU), and the *Semantic gaps* table (G-01…G-19) that the refusal reasons and tests refer to.
2. `crates/doris-plan-translator/src/` — `descriptor_table.rs`, `type_mapper.rs` (the type gate), `expr_translator.rs`, `node_translator.rs`, `stitcher.rs`. This is the substance.
3. `src/backend_service.rs`, `src/params.rs`, `src/fragment_executor.rs`, `src/engine.rs` — the protocol shell and the FFI hand-off.
4. `tests/corpus.rs` and a couple of `tests/snapshots/*.txt` to see what the translator emits.
5. Skim `scripts/` and the workflow.

Fixtures, snapshots and expected results are data; they don't need line-by-line review.

## Verification

- CI: `cargo fmt --check`, `clippy -D warnings`, `cargo test --workspace --no-default-features` (corpus replay of all 22 queries + gap probes) — this is what the new workflow job runs.
- CPU differential (no GPU): every captured dispatch stitched into one plan and executed by DuckDB's Substrait consumer, 22/22 match `tests/expected/tpch-sf1`.
- GPU (Linux + NVIDIA, engine-linked build): `run-tpch.sh` executes and validates 22/22 at SF1 and SF10 (T4, L40S) and at SF100 (L40S); the benchmark harness ran the full SF1/SF10/SF100 matrix on the `g6e.8xlarge`.
- Nothing under `src/` of the engine is touched by this PR.

## Relationship to other PRs

- **Runtime dependency on #1840** (`fix(join): keep a MARK join alive when polled before sizing; materialize inequality sides`): without it Q16 aborts the process and Q17/Q20 fail on the FFI path. This PR builds and its CI passes without it (CI never links the engine).
- The benchmark numbers above were taken with #1841 (`sirius_ffi` parquet-metadata-cache) in place (it took 10 s of Substrait lowering out of Q21 at SF100).

## Before "Ready for review"

- [x] References to the planning workspace the backend grew up in (not part of this PR) — rewritten; the tree is self-contained.
- [ ] Land #1840 first, or agree to merge this one with the runtime caveat documented.


yihong0618:
wow amazing work!

.github/workflows/experimental.yml
```diff
@@ -67,15 +67,59 @@ jobs:
       - name: test
         run: pixi run -e cn cargo test --workspace --no-default-features
 
+  doris:
+    name: doris
+    runs-on: ubuntu-24.04
```
felipeblazing, line 72: **Reviewer: Build, deps, CI & repo hygiene**

**nit** Set a timeout on the new `doris` job

This job runs thrift codegen over the whole Doris IDL and builds the tonic stack from a cold cache (`cache: false`), and it has no `timeout-minutes`. A hung job would hold the merge-queue `experimental` gate for GitHub's 6-hour default.
```suggestion
    runs-on: ubuntu-24.04
    timeout-minutes: 30
```

.github/workflows/experimental.yml
```diff
@@ -67,15 +67,59 @@ jobs:
...
+
+      # The BE's `sirius-engine` feature is on by default but links a Sirius build tree
+      # (C++ toolchain + RAPIDS/CUDA) that this CI env does not have, so build the
+      # pure-Rust path here. The engine path is exercised on a GPU box via
+      # `pixi run be-build` / `be-test`.
+      - name: clippy
+        run: pixi run -e be cargo clippy --all-targets --no-default-features -- -D warnings
+
+      # Includes tests/corpus.rs, which replays the captured TPC-H dispatch fixtures.
+      - name: test
```
felipeblazing, line 111: **Reviewer: Tests + CI**

**major** CI never checks translated plans against real query results

`every_corpus_query_matches_its_snapshot` fails on any change to the translator's output, correct or not, and `UPDATE_SNAPSHOTS=1` accepts whatever the code now emits. The only check that runs the translated plans and compares the rows (`cpu-diff.sh`) is manual. So a change that swaps join sides or reorders aggregate measures passes CI as soon as someone regenerates the snapshots. Run `tpch-cpu-diff` in a CI job, even nightly or `workflow_dispatch`: an SF0.01 dataset plus a cached `.duckdb-substrait` build would make the stitched plans' correctness a regression gate.

felipeblazing, line 111: **Reviewer: Main**

**minor** CI never checks translated plans against real query results

In CI, translator correctness is only the `tests/snapshots/` diff, and `UPDATE_SNAPSHOTS=1` accepts whatever the code now emits, so a swapped join side or misaligned aggregate passes once someone regenerates the snapshots. `cpu-diff.sh` would catch it, but only runs by hand. Run `tpch-cpu-diff` (plus the gap probes) in a nightly/label/path-filtered job, using a cached `.duckdb-substrait` build and a small generated dataset with the same `<table>/part.0.parquet` layout (~SF0.1, since SF0.01 empties some queries). At minimum, make it a required pre-merge step for translator changes.

.github/workflows/experimental.yml
```diff
@@ -67,15 +67,59 @@ jobs:
...
+        run: pixi run -e be cargo test --workspace --no-default-features
```
felipeblazing, line 112: **Reviewer: Test coverage & rigor**

**minor** No CI check that the translated plans compute the right rows

In CI, translator correctness is checked only against `tests/snapshots/`, and the failure message tells the author to regenerate them with `UPDATE_SNAPSHOTS=1`. A semantic regression (wrong join side, misaligned aggregate argument) therefore passes once someone accepts the new snapshot. The DuckDB CPU differential would catch it but only runs by hand. Consider a scheduled or label-triggered job that runs `cpu-diff.sh` on a tiny generated dataset. At minimum, list it in the PR checklist as a required pre-merge step for translator changes.

experimental/doris/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+```bash
+# once: the engine (repo root; CUDA 13 + RAPIDS come from the root pixi env)
+git submodule update --init --depth=1 --jobs 3 duckdb substrait cucascade
+pixi run make TEST_BUILD_TARGET=                  # → build/release/extension/sirius/libsirius.so
+
+cd experimental/doris
+pixi run cargo build --release -p sirius-doris-be # default env carries the engine's toolchain/libs
+pixi run bash scripts/be.sh start --engine        # SIRIUS_BE_TRANSLATE_ONLY=0, --sirius-config conf/sirius.yaml
+pixi run -e fe bash scripts/run-tpch.sh --data /tmp/tpch-sf1   # execute + validate the 22 queries
```
felipeblazing, line 122: **Reviewer: Comments, docs & AI residue**

**minor** README doesn't say the GPU run needs #1840

The engine section says `run-tpch.sh` will execute and validate the 22 queries. Per the PR description, without #1840 Q16 aborts the backend process and Q17/Q20 fail. A reader who follows these steps on `dev` will hit that with no hint why. Add a one-line caveat that names #1840, or the minimum engine commit, until that PR lands.

experimental/doris/README.md
```diff
@@ -0,0 +1,219 @@
...
+FE audit numbers joined by `fe-audit.py`, `summary.csv` from the validator, `samples.csv` from
+the 0.5 s sampler: RSS, bytes read from disk, CPU ticks of the backend process, and for the
+GPU systems nvidia-smi's memory in use, `utilization.gpu` and `utilization.memory`) and
+`rounds.csv` (everything flattened by `bench-report.py rounds`: per query the peaks, deltas
+and mean GPU utilization over its window). `bench-report.py report` turns several run directories into the Markdown tables.
+`engine_ms` is the whole `execute_substrait` call (Substrait lowering, planning, execution,
+Arrow conversion); the engine's own query window is in the run's `telemetry/` directory
+(`scripts/telemetry-plans.py <run>/telemetry/<instance>/`) — a gap between the two is time
+spent outside the GPU, which is how the parquet-footer re-parse in the Substrait consumer
+was found (fixed in the engine by enabling `parquet_metadata_cache` on the FFI path).
```
felipeblazing, line 212: **Reviewer: Comments, docs & AI residue**

**minor** README claims a fix that lives in unmerged #1841

"fixed in the engine by enabling `parquet_metadata_cache` on the FFI path" is only true with #1841, which is still open. It also narrates how the problem was found, which is PR history. Drop the "which is how … was found (fixed …)" clause. If the lowering cost matters to readers, state it as a current caveat that references #1841.

felipeblazing, line 212: **Reviewer: Main**

**minor** README and headline SF100 numbers rely on unmerged #1841

"fixed in the engine by enabling `parquet_metadata_cache` on the FFI path" is only true with #1841, which is still open. The 3.01× geomean / 154.4 s → 40.6 s were also measured with it (it removes ~10 s of Substrait lowering from Q21 alone). Drop the "which is how … was found (fixed …)" history. Instead, state it as a current caveat referencing #1841, and say the benchmark numbers require it (or give them without it).

experimental/doris/conf/sirius-bench.yaml
```diff
@@ -0,0 +1,31 @@
+# Sirius engine config template for the benchmark's Sirius backend (the `sirius*` and
```
felipeblazing, line 1: **Reviewer: Configuration surface & defaults**

**nit** sirius-bench.yaml still cites the planning workspace

The header still says `plan-doc experiments/sf10-bench` and `plan §6.5`, although the last commit says every planning-workspace reference was replaced. Point these at `docs/tpch-report.md` like the other files.

experimental/doris/crates/doris-plan-translator/src/descriptor_table.rs
```diff
@@ -0,0 +1,580 @@
...
+    pub primitive: TPrimitiveType,
+    /// Substrait type mapped from the slot type and nullability.
+    pub substrait_type: Type,
+}
+
+impl SlotInfo {
+    /// Stable output name for the slot: its column name, or `col_<slot_id>` for a derived slot.
+    pub fn output_name(&self) -> String {
+        if self.col_name.is_empty() {
+            format!("col_{}", self.slot_id)
```
felipeblazing, line 65: **Reviewer: Naming, magic numbers & types**

**nit** `col_<n>` means slot id in one place and column position in another

`SlotInfo::output_name` makes `col_<slot_id>` for derived slots, while `node_translator.rs:1263` makes `col_<index>` for position-only layouts. So `col_3` can mean slot 3 or the fourth column depending on where it came from, which is confusing in the snapshots and plan dumps. Use a distinct prefix for the slot-derived name (e.g. `slot_<id>`).

experimental/doris/crates/doris-plan-translator/src/descriptor_table.rs
```diff
@@ -0,0 +1,580 @@
...
+        let slot_descriptors = desc_tbl.slot_descriptors.as_deref().unwrap_or_default();
+        let mut slots = HashMap::with_capacity(slot_descriptors.len());
+        for slot in slot_descriptors {
+            let key = SlotKey {
+                tuple_id: slot.parent,
+                slot_id: slot.id,
+            };
+            // Non-nullable slots get -1 for the bit (SlotDescriptor.toThrift); anything else is
+            // a nullable slot with a real (always 0) bit index.
+            let nullable = slot.null_indicator_bit != -1;
```
felipeblazing, line 134: **Reviewer: Naming, magic numbers & types**

**nit** Doris `-1` sentinels are bare here but named elsewhere in the PR

`node.rs:219-220` names its `-1` sentinels (`BE_RPC_PORT_DISABLED`). The translator doesn't: `null_indicator_bit != -1` here means NOT NULL, and `node_translator.rs:1420` uses `-1` for an unlimited fetch count. Name them the same way, e.g. `NOT_NULLABLE_INDICATOR_BIT: i32 = -1` and `UNLIMITED_FETCH_COUNT: i64 = -1`.

experimental/doris/crates/doris-plan-translator/src/lib.rs
```diff
@@ -0,0 +1,549 @@
...
+/// Substrait string function extension URN.
+pub const URN_STRING: &str = "extension:io.substrait:functions_string";
+/// Substrait datetime function extension URN.
+pub const URN_DATETIME: &str = "extension:io.substrait:functions_datetime";
+/// Substrait aggregate function extension URN.
+pub const URN_AGGREGATE: &str = "extension:io.substrait:functions_aggregate_generic";
+
+/// Result of translating one Doris plan fragment.
+#[derive(Clone, PartialEq)]
+pub struct TranslatedPlan {
```
felipeblazing, line 90: **Reviewer: API design, layering & reuse**

**minor** `TranslatedPlan` exposes `plan` and `output_names` as public mutable fields that must stay in sync

`output_names` duplicates `RelRoot.names` inside `plan`. Because both fields are `pub`, a caller can change one without the other, and the engine and result encoder would then disagree on the column names. Build the struct only in `translate_fragment`, make the fields private, and expose `plan()` / `output_names()` accessors. Tests can use a `#[cfg(test)]` constructor.

experimental/doris/crates/doris-plan-translator/src/node_translator.rs
```diff
@@ -0,0 +1,2379 @@
...
+            ..Default::default()
+        }))),
+    };
+    // Keys pass through; every measure is cast to its Doris slot type (G-19: DuckDB may have
+    // computed `avg` as DOUBLE or an integer `sum` as HUGEINT).
+    let mut expressions: Vec<Expression> = (0..key_count)
+        .map(expr_translator::field_reference)
+        .collect();
+    for (index, measure_type) in measure_types.into_iter().enumerate() {
+        expressions.push(expr_translator::cast(
```
felipeblazing, line 626: **Reviewer: Query semantics & logic**

**minor** Decimal avg/divide go through DOUBLE and the cast back truncates: last digit differs from Doris

DuckDB (and so Sirius) computes `avg(DECIMAL)` and `DECIMAL / DECIMAL` as DOUBLE. This measure/boundary cast then turns the result back into the Doris decimal, and on the GPU that cast truncates. Doris computes these exactly, so Q1's `avg_*` columns already come back one unit off at SF1. Above ~15 significant digits (large sums at SF1000) the error grows past the last digit. The validator's default `--ulps 1` in execution mode accepts this. Either gate these cases or document them as a known result difference, rather than widening the check.

experimental/doris/crates/doris-plan-translator/src/node_translator.rs
```diff
@@ -0,0 +1,2379 @@
...
+                    node_id: node.node_id,
+                    node_type: node.node_type,
+                    reason: "only equality (and null-safe equality) join keys are supported",
+                });
+            }
+        };
+        let left_key = ctx.expr_over(&eq.left, &join_row.row_tuples, &join_row.child_overrides)?;
+        let right_key =
+            ctx.expr_over(&eq.right, &join_row.row_tuples, &join_row.child_overrides)?;
+        if first_key.is_none() {
```
felipeblazing, line 829: **Reviewer: Main**

**major** Anti join via outer join + is_null(key) keeps matched rows when the first key is null-safe (<=>)

`first_key` is the first eq conjunct whatever its opcode, so with `EQ_FOR_NULL` a *matched* row can have a NULL build key and `is_null(other_key)` keeps it. For example, `NOT EXISTS (SELECT 1 FROM b WHERE b.x <=> a.x)` with both keys NULL returns the `a` row instead of dropping it, with no error. Only take the null-tested key from a plain `equal` conjunct, so an anti join whose keys are all null-safe is refused. `semi_and_anti_joins_keep_one_side` only covers `EQ`, so please add a `join_gates` case for this too.
```suggestion
        if first_key.is_none() && name == "equal" {
```

experimental/doris/crates/doris-plan-translator/src/node_translator.rs
```diff
@@ -0,0 +1,2379 @@
...
+        conditions.push(ctx.function(URN_COMPARISON, name, vec![left_key, right_key]));
+    }
+    for expr in join.other_join_conjuncts.as_deref().unwrap_or_default() {
+        conditions.push(ctx.expr_over(
+            expr,
+            &join_row.row_tuples,
+            &join_row.intermediate_overrides,
+        )?);
+    }
+    if conditions.is_empty()
```
felipeblazing, line 841: **Reviewer: Query semantics & logic**

**minor** `vother_join_conjunct` is silently dropped whenever the join has equality keys

The fallback runs only when `conditions` is empty, and `conditions` already holds the eq keys. So a join with `a.k = b.k` plus a non-equi predicate sent only as `vother_join_conjunct` becomes a plain equi-join and returns extra rows. The null-aware anti gate above also checks only `other_join_conjuncts`, so the same shape gets past it. Fall back when the `other_join_conjuncts` list is empty, which is what the BE does, and include `vother_join_conjunct` in the NULL_AWARE gate.
```suggestion
    if join.other_join_conjuncts.as_ref().is_none_or(|conjuncts| conjuncts.is_empty())
```

felipeblazing, line 841: **Reviewer: Error handling & fail-closed gating**

**major** `vother_join_conjunct` is silently dropped when the join has equality keys

`conditions` already holds the equality keys, so `conditions.is_empty()` is false for any equi-join. A predicate that the FE ships only in the legacy `vother_join_conjunct` (with `other_join_conjuncts` unset) is therefore discarded, and the join returns extra rows. The nested-loop path gets this right because it falls back whenever `join_conjuncts` is empty. Use the same test here.
```suggestion
    if join.other_join_conjuncts.as_deref().unwrap_or_default().is_empty()
```

experimental/doris/crates/doris-plan-translator/src/node_translator.rs
```diff
@@ -0,0 +1,2379 @@
...
+                .attach(joined, left_width)
+        }
+        JoinOutput::AntiOuter(side) => {
+            // The padded side's key is NULL exactly for the rows without a match. The key
+            // must propagate NULL (a column, or a cast of one) for that to hold.
+            let (left_key, right_key) = first_key
+                .ok_or_else(|| TranslateError::malformed("anti join has no equality key"))?;
+            let other_key = if lowering.swap { left_key } else { right_key };
+            require_column_key(node, &other_key)?;
+            let condition = ctx.function(URN_COMPARISON, "is_null", vec![other_key]);
```
felipeblazing, line 891: **Reviewer: Test coverage & rigor**

**major** Anti join with a null-safe (`<=>`) key keeps matched rows; nothing tests it

`AntiOuter` identifies unmatched rows by `is_null(other_key)`, but line 817 accepts `EQ_FOR_NULL` keys: a left row with key NULL matches a right row with key NULL via `is_not_distinct_from`, the right key is a real NULL, and the row survives the filter even though it had a match. `semi_and_anti_joins_keep_one_side` only covers `EQ`. Reject `EQ_FOR_NULL` for `AntiOuter` (or null-test a non-nullable column of the padded side), and add a `join_gates` case for it.

experimental/doris/crates/doris-plan-translator/src/node_translator.rs
```diff
@@ -0,0 +1,2379 @@
...
+    Ok(TranslatedRel::new(input.rel, row_tuples.to_vec(), width))
+}
+
+/// Applies the node's `conjuncts` as a filter over the relation's layout.
+fn apply_conjuncts(
+    input: TranslatedRel,
+    node: &TPlanNode,
+    ctx: &mut PlanContext<'_>,
+) -> Result<TranslatedRel> {
+    let conjuncts = node.conjuncts.as_deref().unwrap_or_default();
```
felipeblazing, line 1292: **Reviewer: Error handling & fail-closed gating**

**major** `TPlanNode.vconjunct` is ignored, so a filter sent that way is silently dropped

`apply_conjuncts` only reads `node.conjuncts`. The BE's `ExecNode::init` gives `vconjunct` (field 40) priority when it is set, and the join translators here already fall back to `vother_join_conjunct`/`vjoin_conjunct`. If an FE sends a scan or aggregate predicate in `vconjunct`, it translates to an unfiltered relation, so the query returns too many rows instead of an error. Refuse the node when `vconjunct` is set (or translate it as one more conjunct):
```suggestion
    if node.vconjunct.is_some() {
        return Err(TranslateError::UnsupportedPlanNode {
            node_id: node.node_id,
            node_type: node.node_type,
            reason: "legacy vconjunct filters are not supported",
        });
    }
    let conjuncts = node.conjuncts.as_deref().unwrap_or_default();
```

experimental/doris/crates/doris-plan-translator/src/node_translator.rs
```diff
@@ -0,0 +1,2379 @@
...
+            Some(expression::RexType::Cast(cast)) => {
+                cast.input.as_deref().is_some_and(is_column_reference)
+            }
+            _ => false,
+        }
+    }
+    if is_column_reference(key) {
+        return Ok(());
+    }
+    Err(TranslateError::UnsupportedPlanNode {
```
felipeblazing, line 1616: **Reviewer: Test coverage & rigor**

**minor** `require_column_key` rejection is never exercised

No test sends an anti join whose padded-side key is an expression (for example `a + 1`), so removing this gate would leave the suite green. Without it, the `is_null` filter would test an expression that might not propagate NULL. Add a `join_gates` case with a non-slot key that asserts `reason.contains("plain column")`.

experimental/doris/crates/doris-plan-translator/src/stitcher.rs
```diff
@@ -0,0 +1,1460 @@
+//! The single-plan stitcher (MVP-A0): the fragments the FE dispatched to this backend, joined
```
felipeblazing, line 1: **Reviewer: Comments, docs & AI residue**

**nit** Planning milestone ids (`MVP-A0`, `P1.x`) in code comments and test docs

`MVP-A0` appears in doc comments in `stitcher.rs`, `lib.rs`, `backend_service.rs` (lines 161 and 208), `dump-fragments.rs`, `node_translator.rs:2060` and `conf/fe.conf`. The `tests/corpus.rs` test docs start with `P1.1:`…`P1.5:`. These are milestone names from the planning workspace and mean nothing to a reader who only has the code. Replace `MVP-A0` with what it means ("single-plan execution: the whole dispatch runs as one plan on this backend") and drop the `P1.x:` prefixes.

experimental/doris/crates/doris-plan-translator/src/stitcher.rs
```diff
@@ -0,0 +1,1460 @@
...
+            .ok_or(TranslateError::MissingField {
+                context: "TPipelineFragmentParams",
+                field: "fragment",
+            })?;
+        let plan = fragment.plan.as_ref().ok_or(TranslateError::MissingField {
+            context: "TPipelineFragmentParams.fragment",
+            field: "plan",
+        })?;
+        let instances = params.local_params.as_deref().unwrap_or_default();
+        if instances.len() != 1 {
```
felipeblazing, line 145: **Reviewer: Error handling & fail-closed gating**

**major** Stitcher does not check that this backend is the only sender, so a split query returns a partial result

The stitcher only checks that each fragment has one *local* instance. If the FE splits a scan fragment across two live backends (for example the native BE in `be-native.sh` left registered and Alive next to this one), this backend receives one instance holding only its own share of the files. The stitched plan then reads only those files and returns undercounted aggregates without any error. Before stitching, reject the dispatch unless every exchange's sender count is 1 (`per_exch_num_senders[dest_node_id]`) and every stream sink's `destinations` point at this backend.

experimental/doris/crates/doris-plan-translator/src/stitcher.rs
```diff
@@ -0,0 +1,1460 @@
...
+            }
+            TDataSinkType::DATA_STREAM_SINK => {
+                let stream = sink
+                    .stream_sink
+                    .as_ref()
+                    .ok_or(TranslateError::MissingField {
+                        context: "TDataSink(DATA_STREAM_SINK)",
+                        field: "stream_sink",
+                    })?;
+                if stream
```
felipeblazing, line 176: **Reviewer: Test coverage & rigor**

**minor** Stream-sink projection/conjunct gate has no test

If this check is deleted, the stitcher drops the sender's output exprs/conjuncts and returns wrong rows, and every test still passes. `stitching_gates` also skips the duplicate-sender, exchange-with-conjuncts (`has_node_work`) and "scan node has ranges in more than one fragment" errors. Add one mutated `two_phase_dispatch()` case for each, asserting the reason text.

experimental/doris/crates/doris-plan-translator/src/type_mapper.rs
```diff
@@ -0,0 +1,653 @@
+//! Doris `TTypeDesc` → Substrait type mapping and the type gate.
+//!
+//! Sirius consumes the plan through DuckDB's Substrait consumer, so the emitted
+//! type has to be one that consumer accepts *and* one Sirius can carry on the GPU.
+//! The allowlist is deliberately small: the TPC-H corpus over `local()` parquet needs
+//! `BOOLEAN`, the four signed integers, `DECIMAL32/64/128I`, `DATEV2`, `VARCHAR` and
+//! `STRING`; `FLOAT`/`DOUBLE`/`CHAR`/`DATETIMEV2` are mapped because they are
+//! unambiguous. Everything else is rejected with a reason that names the gap (`G-nn`, the
+//! numbered list under *Semantic gaps* in the crate's `README.md`):
```
felipeblazing, line 9: **Reviewer: Design**

**minor** G-xx gap codes in ~80 error messages point at a doc that isn't in the tree

The type gate, expression gate and node translator put `G-01`…`G-29` into user-visible `TranslateError` reasons, which reach the FE client. Their definitions live in `plan-doc/reference/semantics-gaps.md`, which is not part of this PR. The Draft checklist says it will rewrite or drop the `plan-doc/` references, but these codes would then mean nothing to anyone reading an error. Commit `semantics-gaps.md` under `experimental/doris/docs/` and point the references there.

felipeblazing, line 9: **Reviewer: Comments, docs & AI residue**

**major** 78 `G-NN` gap ids point at a `semantics-gaps.md` that is not in the tree

`type_mapper.rs`, `expr_translator.rs`, `node_translator.rs`, the scripts, and the `sql/gaps/gNN-*` / `tests/fixtures/gaps/gNN-*` names all key off ids like G-06 and G-19 from `plan-doc/reference/semantics-gaps.md`. The "Before Ready for review" checklist only covers the literal `plan-doc/` paths, and removing those paths still leaves the ids pointing at nothing. The ids name the fixture directories and are the vocabulary of the rejection reasons, so I'd commit `semantics-gaps.md` under `experimental/doris/docs/` and link it from here. Otherwise, drop the ids and keep the inline reasons, which the table below already mostly states.

experimental/doris/crates/doris-plan-translator/src/type_mapper.rs
```diff
@@ -0,0 +1,653 @@
...
+use substrait::proto::Type;
+use substrait::proto::r#type;
+
+use crate::error::{Result, TranslateError};
+
+/// Highest decimal precision the 128-bit carrier holds (Doris `DECIMAL128I` tops out here too).
+pub const MAX_DECIMAL_PRECISION: i32 = 38;
+/// Decimal precisions at or below this are stored as INT16 by DuckDB and have no cuDF carrier
+/// (`sirius::get_cudf_type` throws; G-06).
+pub const MIN_DECIMAL_PRECISION_EXCLUSIVE: i32 = 4;
```
felipeblazing, line 44: **Reviewer: Naming, magic numbers & types**

**minor** `MIN_DECIMAL_PRECISION_EXCLUSIVE = 4` forces a `+ 1` at the call site

The constant holds the largest *rejected* precision, but the name reads as a minimum. So `expr_translator.rs:422` has to write `MIN_DECIMAL_PRECISION_EXCLUSIVE + 1` to get the real floor, and the doc says "widened to 5" as a separate literal. Store the smallest accepted precision instead (`pub const MIN_DECIMAL_PRECISION: i32 = 5;`), test `precision < MIN_DECIMAL_PRECISION` here, and use `precision.max(MIN_DECIMAL_PRECISION)` at the widening site.

experimental/doris/pixi.toml
```diff
@@ -0,0 +1,105 @@
...
+
+[feature.client.tasks]
+client = {cmd = "mysql --host 127.0.0.1 --port 9030 --user root",description = "Open a MySQL client against the local FE (start it with `fe-run` in another terminal first)"}
+
+[feature.be.dependencies]
+# protoc for prost/tonic codegen of the Doris proto IDL, and the thrift compiler for the
+# thrift IDL (both under the `doris/` submodule).
+libprotobuf = ">=7.35,<8"
+rust = "*"
+thrift-compiler = "*"
```
felipeblazing, line 37: **Reviewer: Build, deps, CI & repo hygiene**

**minor** Pin the thrift compiler version the codegen post-processing depends on

`build.rs` rewrites the generated Rust by line shape (derives, enum impls, inner attributes). It also runs the output against the `thrift = "0.24"` crate, while the lock currently resolves compiler 0.22.0. With `"*"`, any `pixi update` can swap in a compiler that emits a different shape and silently break the build. The starrocks manifest already pins its compiler (`thrift-compiler = "0.20.*"`).
```suggestion
thrift-compiler = "0.22.*"
```

experimental/doris/scripts/be.sh
```diff
@@ -0,0 +1,109 @@
...
+        # device) is only released when the CUDA context is torn down; a start right behind a
+        # stop then fails its bring-up with cudaErrorMemoryAllocation. Wait for the process,
+        # then for the device to be free again (nvidia-smi is absent on the no-engine path).
+        for _ in $(seq 1 60); do
+            [ -n "${pid}" ] && kill -0 "${pid}" 2>/dev/null || break
+            sleep 0.5
+        done
+        if command -v nvidia-smi >/dev/null 2>&1; then
+            for _ in $(seq 1 60); do
+                used=$(nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits 2>/dev/null | head -1 | tr -d ' ')
```
felipeblazing, line 95: **Reviewer: CUDA streams & device affinity**

**minor** be.sh stop waits on GPU 0's memory, not the device the backend used

`nvidia-smi ... | head -1` always reads physical GPU 0 and ignores `CUDA_VISIBLE_DEVICES`. If the backend ran on another GPU, this wait passes right away and the next start can still hit `cudaErrorMemoryAllocation`. On a shared box where something else holds more than 2 GiB on GPU 0, every stop stalls for the full 30 s. Waiting for the pid is enough, since the pool is released when the process exits; if you keep the check, pass `-i "${CUDA_VISIBLE_DEVICES:-0}"` and compare against the used memory recorded before start rather than a fixed 2 GiB. Also, when the backend is stopped through `pkill`, `pid` is empty, so the process wait is skipped entirely.

felipeblazing, line 95: **Reviewer: CUDA streams & device affinity**

**minor** `be.sh stop` always checks GPU 0's memory, whichever device the engine used

`nvidia-smi --query-gpu=memory.used | head -1` reads the first device in PCI order. The engine's device is whatever CUDA picks after `CUDA_VISIBLE_DEVICES`, which `nvidia-smi` does not honour. With `CUDA_VISIBLE_DEVICES=1`, stop checks the wrong GPU and the next `start` can still hit `cudaErrorMemoryAllocation`. If some other tenant holds more than 2 GiB on GPU 0, every stop waits the full 30 s. Instead, wait for the backend's pid to disappear from `nvidia-smi --query-compute-apps=pid --format=csv,noheader`, which ties the check to this process. The bench.sh sampler (`head -1`, line 187) has the same GPU-0 assumption.

experimental/doris/scripts/be.sh
```diff
@@ -0,0 +1,109 @@
...
+                [ -n "${used}" ] && [ "${used}" -gt 2048 ] || break
```
felipeblazing, line 96: **Reviewer: Main**

**minor** `be.sh stop` waits on GPU 0's total memory.used against a bare 2048 MiB threshold

`head -1` always reads the first GPU in PCI order (not what `CUDA_VISIBLE_DEVICES` selected), and the check is device-wide, not this process. On a shared box another tenant above 2 GiB makes every stop wait the full 30 s and then continue quietly, and the next `start` can still hit `cudaErrorMemoryAllocation`. Poll `nvidia-smi --query-compute-apps=pid --format=csv,noheader` until the BE's pid is gone, and warn on timeout. If a threshold stays, name it. The bench.sh sampler (line 187) has the same GPU-0 assumption.

experimental/doris/scripts/bench.sh
```diff
@@ -0,0 +1,319 @@
...
+        bash scripts/be-native.sh stop
+    fi
+}
+
+# The effective Sirius config for this run, from the template.
+write_sirius_config() { # use_odirect telemetry_dir
+    local ram_kib host spill
+    if [ -z "${HOST_CAPACITY}" ]; then
+        ram_kib=$(awk '/MemTotal/ { print $2 }' /proc/meminfo)
+        host=$(awk -v kib="${ram_kib}" 'BEGIN { g = kib / 1048576; a = g * 0.9; b = g - 14; printf "%dGi", (a < b ? a : b) }')
```
felipeblazing, line 139: **Reviewer: CUDA streams & device affinity**

**minor** bench.sh sizes a per-NUMA-node host tier from total system RAM

`memory.host.capacity_bytes` is applied per NUMA node (`sirius_config.cpp` host_mem_config), but this computes `min(0.9·MemTotal, MemTotal − 14 GiB)` over the whole box. On a two-socket host with 512 GiB, each node would try to pin ~460 GiB, which is about 1.8× the RAM the box has, and bring-up fails or the FE gets starved. Divide by the node count (`ls -d /sys/devices/system/node/node[0-9]* | wc -l`) before substituting, or switch the template to `capacity_fraction`.

felipeblazing, line 139: **Reviewer: Main**

**major** bench.sh sizes the per-NUMA pinned host tier from the whole machine's RAM

`memory.host.capacity_bytes` applies per NUMA node, but `min(0.9·MemTotal, MemTotal−14G)` is computed for the whole box. On a 2-socket host every node tries to pin ~90% of all RAM, so bring-up fails or the box OOMs. With <14 GiB of RAM the formula emits a negative value. Divide by the node count (`ls -d /sys/devices/system/node/node[0-9]* | wc -l`) and clamp to a positive floor. The same caveat applies to `--host-capacity`.

experimental/doris/scripts/run-tpch.sh
```diff
@@ -0,0 +1,253 @@
...
+
+DATA=""
+SQL_DIR="sql/tpch"
+QUERIES=""
+TRANSLATE_ONLY=false
+OUT=""
+DUMP_DIR="${SIRIUS_BE_DUMP_FRAGMENTS:-${ROOT}/log/dump}"
+EXPECTED=""
+VALIDATE=true
+ULPS=1
```
felipeblazing, line 63: **Reviewer: Test coverage & rigor**

**minor** `--ulps 1` slack applies to every numeric column, not just Q1's avg

The one-ulp slack exists because of the truncated last digit in Q1's `avg_*` columns, but `ULPS=1` is the execution-mode default for all 22 queries and every column. A last-digit decimal error in any `sum`/revenue column would pass as OK with only a note. Scope the slack to the known (query, column) pairs, or keep the default at 0.5 and pass `--ulps 1` only for q01.

experimental/doris/scripts/validate_tpch_results.py
```diff
@@ -0,0 +1,570 @@
...
+        if csv_path:
+            csv_path.parent.mkdir(parents=True, exist_ok=True)
+            with csv_path.open("w", newline="") as handle:
+                writer = csv.writer(handle)
+                writer.writerow(["query", "status", "detail"])
+                for v in self.verdicts:
+                    writer.writerow([v.query, v.status, v.detail])
+        ok, mismatch, error, skipped = (self.count(s) for s in ("OK", "MISMATCH", "ERROR", "SKIPPED"))
+        print(f"==> {ok} ok, {mismatch} mismatch, {error} error, {skipped} skipped", flush=True)
+        return 1 if mismatch or error else 0
```
felipeblazing, line 105: **Reviewer: Main**

**major** Validator exits 0 when plans or baselines are missing (everything SKIPPED)

`finish` fails only on MISMATCH/ERROR, so a missing `.substrait` plan or expected `.tsv` becomes SKIPPED and `cpu-diff.sh`/`run-tpch.sh` still pass. This already happened in this PR: the root `.gitignore` dropped 20 of the 22 SF1 baselines and the differential skipped them silently. Fail on skips unless the caller opts in explicitly.
```suggestion
        return 1 if mismatch or error or skipped or not ok else 0
```

experimental/doris/scripts/validate_tpch_results.py
```diff
@@ -0,0 +1,570 @@
...
+    beyond_half_ulp: int = 0
+
+
+def scalars_match(lhs: str, rhs: str, tolerance: Tolerance) -> bool:
+    if lhs == rhs:
+        return True
+    lnum, rnum = to_decimal(lhs), to_decimal(rhs)
+    if lnum is None or rnum is None:
+        return canonical(lhs) == canonical(rhs)
+    coarse_scale = min(scale_of(lhs), scale_of(rhs))
```
felipeblazing, line 224: **Reviewer: Tests + CI**

**major** Validator uses the coarser scale, so a result that lost its fractional digits still passes

`coarse_scale = min(scale_of(lhs), scale_of(rhs))` takes the ulp from whichever side prints fewer digits. For example, an engine that returns `26` where the baseline has `25.52` gets ulp 1, and the 0.48 difference passes; with `--ulps 1` (the execution-mode default), anything up to 1.0 passes. So a DECIMAL→integer truncation or a wrong cast in the translator would show up as OK. Take the scale from the expected value, and add a small pytest for `scalars_match`/`compare`, since the 22/22 claim rests on this script and it has no tests.

experimental/doris/src/backend_service.rs
```diff
@@ -0,0 +1,1025 @@
...
+    }
+
+    /// Whether survey (translate-only) mode is on: unset or anything but `0`/`false`/`off`.
+    fn translate_only() -> bool {
+        match std::env::var(TRANSLATE_ONLY_ENV) {
+            Ok(value) => !matches!(
+                value.trim().to_ascii_lowercase().as_str(),
+                "0" | "false" | "off" | "no"
+            ),
+            Err(_) => true,
```
felipeblazing, line 116: **Reviewer: Configuration surface & defaults**

**major** Translate-only defaults to on even in the engine-linked build

With the var unset, `translate_only()` returns true regardless of the `sirius-engine` feature. So running `target/release/sirius-doris-be --sirius-config ...` directly (not through `be.sh`) brings up the GPU, takes 90% of VRAM and then refuses every query. The same parse also treats a typo like `fasle` as "on" without saying so. Make this a clap flag (`--translate-only`, `env = "SIRIUS_BE_TRANSLATE_ONLY"`) that defaults to `!cfg!(feature = "sirius-engine")`, is parsed once at startup and rejects unknown values. That also puts it next to every other knob in `main.rs`.

experimental/doris/src/backend_service.rs
```diff
@@ -0,0 +1,1025 @@
...
+            info!(%query_id, index = fragment.index, shape = %fragment.shape(), "received fragment");
+        }
+        let result_fragment = batch
+            .fragments
+            .iter()
+            .find(|fragment| fragment.is_result_fragment());
+        // Register before doing anything slow so an early `fetch_data` parks instead of
+        // failing as unknown. A batch without a RESULT_SINK (a multi-node query whose result
+        // fragment runs elsewhere) still gets a slot keyed by query id so cancel/evict work.
+        let slot = self.results.register(
```
felipeblazing, line 138: **Reviewer: Concurrency, lifetime & hangs**

**minor** A cancel that arrives before `register` is lost, and the slot then leaks with the whole materialized result

`cancel_plan_fragment` calls `evict`, which is a no-op when no slot exists yet. Interleaving: the FE cancels (kill or timeout) while `dispatch` is still decoding the payload on the blocking worker; `evict` finds nothing; then `register` inserts a slot. Nothing ever removes that slot, because eviction only happens on cancel, and it holds the full encoded result once execution finishes. Record cancelled query ids, as tombstones checked under the map lock in `register` (and before execution), or evict the slot on EOS in `fetch_data` so a lost cancel cannot pin results.

experimental/doris/src/backend_service.rs
```diff
@@ -0,0 +1,1025 @@
...
+        // fetches). A translation error is reported through the result slot rather than
+        // failing dispatch: the FE has already committed the query to us at this point.
+        if result_fragment.is_none() {
+            slot.fail(format!(
+                "query {query_id}: no RESULT_SINK fragment was dispatched to this backend; \
+                 multi-node execution is not implemented"
+            ));
+            return Ok(());
+        }
+        match translated {
```
felipeblazing, line 172: **Reviewer: Test coverage & rigor**

**major** Execute-mode success path of `dispatch` is untested

`execute_mode_reports_translator_errors_through_the_slot` only covers a translation failure. Nothing in CI exercises translate → `executor.execute` → `MysqlResultEncoder::encode` → `slot.push`/`close`, or checks that rows arrive under the RESULT_SINK instance id the FE fetches. You could delete lines 172–184 and every test would still pass. `StubExecutor` is already there: dispatch the q06 fixture with translate-only off, fetch it by its instance id, and assert one encoded row followed by eos. Add one more case for the `result_fragment.is_none()` branch.

experimental/doris/src/backend_service.rs
```diff
@@ -0,0 +1,1025 @@
...
+            Ok(plan) => match self.execute_timed(query_id, &plan) {
```
felipeblazing, line 173: **Reviewer: Concurrency, lifetime & hangs**

**major** Cancel/timeout never reaches the engine; cancelled queries still occupy the single engine thread

Here is the interleaving. FE thread: query A's `exec_plan_fragment` hits `remote_fragment_exec_timeout_ms` (or the user runs KILL), so tonic drops the handler future and `cancel_plan_fragment` evicts A's slot. The `spawn_blocking` worker is detached, though, and still calls `executor.execute`. Queries B and C, each already waiting in the engine's mpsc queue, run in turn even if they were cancelled too. So one slow query turns into a pile-up where every later query times out. At minimum, pass the slot (or an `Arc<AtomicBool>` set by `evict`) through `execute`. Check it before `send` and again on the engine thread before `execute_substrait`, so dead requests are skipped.

experimental/doris/src/backend_service.rs
```diff
@@ -0,0 +1,1025 @@
...
+                Ok(result) => match MysqlResultEncoder::encode(result.batches(), 0) {
```
felipeblazing, line 174: **Reviewer: Main**

**major** Whole result is encoded into one TResultBatch / fetch_data packet: unbounded memory and it can exceed the FE message cap

`encode(result.batches(), 0)` turns the whole result into one packet, so the Arrow batches, the per-row `Vec<u8>`s and the `to_binary()` blob coexist at the peak (~3× the result size, none of it accounted for). A large result (for example `SELECT * FROM lineitem` at SF100) can OOM-kill the BE, or fail on the FE's `max_msg_size_of_result_receiver` (100 MiB) after the query succeeded. The slot already supports consecutive `packet_seq`, so encode and `push` one bounded packet per Arrow batch (or per row count) and drop each batch once it is encoded.

experimental/doris/src/backend_service.rs
```diff
@@ -0,0 +1,1025 @@
...
+        request: Request<PCancelPlanFragmentRequest>,
+    ) -> Result<Response<PCancelPlanFragmentResult>, Status> {
+        let request = request.into_inner();
+        let id = request
+            .query_id
+            .as_ref()
+            .map(UniqueId::from)
+            .unwrap_or_else(|| UniqueId::from(&request.finst_id));
+        info!(%id, reason = ?request.cancel_reason, "cancel_plan_fragment");
+        self.results.evict(id);
```
felipeblazing, line 384: **Reviewer: Correctness**

**minor** cancel_plan_fragment evicts the slot but doesn't stop the running query

Cancellation only fails the parked `fetch_data`. The plan keeps running on the single `sirius-engine` thread, and every later dispatch queues behind it, including queries resubmitted after a user cancel or an FE timeout. Its result is then pushed into an orphaned slot. At minimum, record the cancel so a query still waiting in the engine queue is skipped. Document that a query already running can't be interrupted.

experimental/doris/src/engine.rs
```diff
@@ -0,0 +1,397 @@
+//! GPU-backed fragment executor: owns the Sirius engine on a dedicated thread.
```
felipeblazing, line 1: **Reviewer: Design**

**major** Engine actor, executor seam, result encoder and ExtensionRegistry are copies of experimental/starrocks

`src/engine.rs` and `src/fragment_executor.rs` match `experimental/starrocks/src/{engine,fragment_executor}.rs` line for line apart from names. `result_encoder.rs` has the same structure, and the translator's `ExtensionRegistry`, `unique_names` and `URN_*` constants are copied from `starrocks-plan-translator/src/lib.rs`. A fix to the engine-thread teardown, the Arrow→MySQL text rendering or anchor allocation now has to land twice, and the copies will drift. Move the FE-agnostic parts (engine actor, `FragmentExecutor`/`FragmentResult`, MySQL row encoder, extension registry) into a shared crate next to `rust/crates/sirius` that both backends depend on, or at least record in the README that they are forked on purpose.

experimental/doris/src/engine.rs
```diff
@@ -0,0 +1,397 @@
...
+    respond: Sender<Result<Vec<RecordBatch>, String>>,
+}
+
+/// GPU-backed [`FragmentExecutor`] running plans on an embedded Sirius engine.
+///
+/// The engine context lives on a dedicated thread; this handle forwards plans to it and waits for
+/// the result. Dropping it closes the request channel, which ends the thread and tears the context
+/// down (joined for an ordered teardown).
+#[derive(Debug)]
+pub struct SiriusEngine {
```
felipeblazing, line 42: **Reviewer: API design, layering & reuse**

**major** `engine.rs` and `fragment_executor.rs` are verbatim copies of the StarRocks ones

`SiriusEngine`, `ExecuteRequest`, `engine_thread`, `build_context`, `Drop`, `FragmentResult`, `FragmentExecutor`, `StubExecutor` and all of their tests are line-for-line identical to `experimental/starrocks/src/{engine,fragment_executor}.rs`. The only difference is which `TranslatedPlan` is imported, and that type is used only for `to_substrait_bytes()`. Move the dedicated-thread actor into `rust/crates/sirius` (or a sibling crate) as a `Send + Sync` handle whose `execute` takes the plan bytes. That is also where the one-context-per-process rule (today's `GPU_CONTEXT_LOCK` in the `sirius` tests) belongs. Each backend then keeps only a thin adapter, and a fix to teardown or error handling lands once instead of twice.

experimental/doris/src/engine.rs
```diff
@@ -0,0 +1,397 @@
...
+        }
+    };
+    info!("sirius-engine thread ready");
+    // One query at a time until the handle (and its sender) is dropped.
+    while let Ok(request) = requests.recv() {
+        // `execute_substrait` drains the Arrow stream and drops the context-referencing wrapper
+        // here, on the engine thread, returning owned batches whose buffers are released via their
+        // own Arrow C release callbacks — independent of the context. So the batches are safe to
+        // send to, and drop on, the caller's thread.
+        let result = context
```
felipeblazing, line 101: **Reviewer: Concurrency, lifetime & hangs**

**minor** A panic on the engine thread kills the engine permanently while the node keeps heartbeating Alive

`engine_thread` has no `catch_unwind` around `execute_substrait`. A single panic unwinds the thread and drops the context. Every later `execute` then fails with "sirius-engine thread is not running", yet the heartbeat server keeps the backend Alive, so the FE keeps routing queries to a dead engine. Wrap the per-request body in `catch_unwind` and reply with an error, or exit the process when the engine thread dies so the node goes down visibly.

experimental/doris/src/fragment_executor.rs
```diff
@@ -0,0 +1,110 @@
+//! Execution of a translated fragment into Arrow result batches.
+//!
+//! The engine→BE result interchange is the Arrow C Data Interface (see the `SiriusExecutor`
```
felipeblazing, line 3: **Reviewer: Comments, docs & AI residue**

**minor** Stale: the `SiriusExecutor` TODO is already implemented as `SiriusEngine`

`engine.rs` already ships the engine-backed executor (`SiriusEngine`), and `main.rs` wires it in. The module doc ("see the `SiriusExecutor` TODO", "Today a StubExecutor stands in for the GPU engine") and the TODO at line 54 describe a type that will never exist. Reword the module doc to say the stub is the engine-less (`--no-default-features`) executor, and delete the TODO at line 54. The streaming TODO at line 38, and "lifted by the streaming evolution" in `engine.rs`, should reference a tracking issue instead of `doris-execute`.

experimental/doris/src/fragment_executor.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn execute(&self, translated: &TranslatedPlan) -> Result<FragmentResult, String>;
+}
+
+/// Placeholder executor that fabricates one row so the result path works without a GPU.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct StubExecutor;
+
+impl FragmentExecutor for StubExecutor {
+    fn execute(&self, translated: &TranslatedPlan) -> Result<FragmentResult, String> {
+        // TODO(doris-execute): replace with a SiriusExecutor that hands
```
felipeblazing, line 54: **Reviewer: Main**

**minor** Stale `SiriusExecutor` TODO and module doc: `SiriusEngine` already exists, with a different design

The module doc says a stub stands in for the GPU engine "today", and the TODO says to hold an `Arc<sirius::SiriusContext>` threaded in from `main`. `SiriusEngine` in `engine.rs` already ships, and it keeps the context on a dedicated thread precisely because `SiriusContext` is `!Send`/`!Sync`. Following the TODO would reintroduce the design `engine.rs` avoids. Reword the doc to say `StubExecutor` is the engine-less (`--no-default-features`) executor, point to `SiriusEngine`, and delete the TODO.

experimental/doris/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+pub struct BackendConfig {
+    /// Address the thrift/gRPC listeners bind to.
+    #[arg(long, default_value = "0.0.0.0")]
+    pub bind_host: Host,
+    /// Host the FE registers and heartbeats this backend under (`ALTER SYSTEM ADD BACKEND
+    /// "<advertise_host>:<heartbeat_port>"`); must be reachable from the FE.
+    #[arg(long, default_value = "127.0.0.1")]
+    pub advertise_host: Host,
+    /// Thrift `HeartbeatService` port (Doris default 9050).
+    #[arg(long, default_value_t = 9050)]
```
felipeblazing, line 129: **Reviewer: Naming, magic numbers & types**

**minor** Default ports are written twice (clap attrs and `Default` impls)

9050/9060/8040/8060 appear in both the `#[arg(default_value_t = …)]` attributes and `impl Default for BackendConfig`. 9030 is likewise repeated in `FeConfig`. If someone changes one copy, the CLI default and the `Default` used by tests quietly disagree. Add named constants (`DORIS_DEFAULT_HEARTBEAT_PORT: u16 = 9050`, …) and use them in both places.

experimental/doris/src/main.rs
```diff
@@ -0,0 +1,327 @@
...
+use tracing::{debug, error, info, instrument, warn};
+use tracing_subscriber::fmt::format::FmtSpan;
+
+// Initial delay for the startup registration backoff; doubles up to the cap on each retry.
+const REGISTRATION_RETRY_INTERVAL: Duration = Duration::from_secs(1);
+// Upper bound on the exponential backoff delay so a large attempt count stays bounded.
+const REGISTRATION_MAX_RETRY_INTERVAL: Duration = Duration::from_secs(30);
+const REGISTRATION_REFRESH_INTERVAL: Duration = Duration::from_secs(10);
+// The FE heartbeats every 5 s (`heartbeat_interval_second`); three misses is stale.
+const HEARTBEAT_STALE_AFTER: Duration = Duration::from_secs(30);
```
felipeblazing, line 26: **Reviewer: Naming, magic numbers & types**

**minor** HEARTBEAT_STALE_AFTER comment says three 5 s misses, value is 30 s

The comment says the FE heartbeats every 5 s and that three misses count as stale, which is 15 s. The constant is 30 s, which is six misses. The Doris copy added this comment; the StarRocks original has none. Name the FE interval (`FE_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5)`) and derive the threshold with an explicit miss count, so the comment can't drift from the value again.

experimental/doris/src/main.rs
```diff
@@ -0,0 +1,327 @@
...
+        let result = RunningBackend {
+            heartbeat_server,
+            backend_server,
+            grpc_server,
+            registration_task,
+        }
+        .wait_until_shutdown()
+        .await;
+
+        // The servers have stopped by the time `wait_until_shutdown` returns, so no in-flight RPC
```
felipeblazing, line 106: **Reviewer: Concurrency, lifetime & hangs**

**minor** SIGTERM waits for the running query; `be.sh stop` then reports stopped while the GPU is still held

`serve_with_incoming_shutdown` drains in-flight RPCs gracefully, and `exec_plan_fragment` runs the whole query inside the RPC. So `wait_until_shutdown` does not return until the current engine query, and anything queued behind it, finishes. That can take up to 10 minutes. `be.sh stop` gives up after 30 s + 30 s, prints "backend stopped" anyway, and the next `start` fails with `cudaErrorMemoryAllocation`, the failure b57abffc was meant to fix. Either cancel pending engine requests on shutdown (see the cancel finding), or put a timeout on the gRPC drain and have `be.sh stop` escalate to SIGKILL and fail loudly.

experimental/doris/src/node.rs
```diff
@@ -0,0 +1,1319 @@
...
+    /// Latest accepted FE epoch (bumped on FE leader change); older epochs are rejected.
+    pub epoch: Option<types::TEpoch>,
+    /// FE-assigned backend id.
+    pub backend_id: Option<i64>,
+    /// Host the FE registered this backend under (`TMasterInfo.backend_ip`).
+    pub registered_host: Option<String>,
+    /// FE leader's thrift address (`TMasterInfo.network_address`).
+    pub frontend_address: Option<types::TNetworkAddress>,
+    /// Wall-clock timestamp of the latest accepted heartbeat.
+    pub last_heartbeat_ms: Option<u128>,
```
felipeblazing, line 76: **Reviewer: Naming, magic numbers & types**

**nit** Heartbeat recency stored as wall-clock `u128` millis, then clamped

`last_heartbeat_ms: Option<u128>` is only used to compute elapsed time. That forces `min(u64::MAX as u128) as u64` clamping, and it can go wrong if the wall clock steps. Store `Option<Instant>` and use `.elapsed()`. This is the monotonic type for a staleness check, and the casts go away.

experimental/doris/src/node.rs
```diff
@@ -0,0 +1,1319 @@
...
+                // poisoned mutex) must not tear down the whole server — log and keep accepting.
+                let active_connection = match shutdown.track_connection(&stream) {
+                    Ok(active_connection) => active_connection,
+                    Err(err) => {
+                        warn!(service = name, error = %err, "failed to track thrift connection; continuing");
+                        continue;
+                    }
+                };
+                if let Err(err) =
+                    handle_thrift_connection(name, processor.clone(), stream, active_connection)
```
felipeblazing, line 766: **Reviewer: Concurrency, lifetime & hangs**

**major** Thrift servers handle one connection at a time, so an idle pooled FE connection starves every other connection

`handle_thrift_connection` runs inline on the accept thread and returns only on EOF, so an open connection blocks every later one. The FE keeps its thrift clients in a per-backend object pool and leaves them open while idle. If two FE threads ever borrow a client at the same moment (for example, a heartbeat round overlapping a slow one, or two `BackendService` probes), the second connection completes its TCP handshake but is never served while the first sits idle. The heartbeat then times out and `SHOW BACKENDS` flips the node to dead. The processor is already `Arc<P: Sync>`, so spawn a thread per accepted connection and keep the tracked-connection set for shutdown.

experimental/doris/src/result_encoder.rs
```diff
@@ -0,0 +1,302 @@
...
+    Int16Array, Int32Array, Int64Array, LargeStringArray, RecordBatch, StringArray,
+    StringViewArray, TimestampMicrosecondArray,
+};
+use arrow_schema::{DataType, TimeUnit};
+use doris_thrift::data::TResultBatch;
+use thrift::protocol::{TBinaryOutputProtocol, TSerializable};
+
+/// Encodes Arrow result batches into a Doris `TResultBatch` of MySQL text rows.
+#[derive(Default)]
+pub(crate) struct MysqlResultEncoder {
```
felipeblazing, line 21: **Reviewer: API design, layering & reuse**

**major** Result encoder, thrift server, heartbeat and registration code duplicated from StarRocks

`result_encoder.rs` is about 98% identical to the StarRocks copy. `node.rs` repeats StarRocks `lib.rs` almost exactly: `ThriftServer`/`ThriftServerShutdown`, `run_thrift_server`, `SharedHeartbeatState`, `register_node`, and most of the shutdown tests. `main.rs`'s `RegistrationMonitor`/`register_node_with_retries` and the translator's `ExtensionRegistry`, `URN_*` constants and `unique_names` are copies too. Put the engine-agnostic pieces in a shared `experimental/common` crate before this lands. The protocol-specific thrift types can stay behind a small trait or be wrapped by each caller (for example, the encoder returns row bytes and each backend builds its own `TResultBatch`). Doris's parked multi-batch `ResultStore` is what StarRocks' own TODO asks for, which is one more reason to share.

experimental/doris/src/result_encoder.rs
```diff
@@ -0,0 +1,302 @@
...
+        Ok(encoder.into_result_batch(packet_seq))
+    }
+
+    /// Encodes every row of `batch` as a MySQL text row.
+    fn add_batch(&mut self, batch: &RecordBatch) -> Result<(), String> {
+        let columns = batch.columns();
+        for row in 0..batch.num_rows() {
+            let mut encoded = MysqlTextRow::default();
+            for column in columns {
+                encoded.push_cell(Self::render_cell(column.as_ref(), row)?.as_deref());
```
felipeblazing, line 41: **Reviewer: Performance**

**minor** Result encoder downcasts, dispatches on type and heap-allocates once per cell

`render_cell` runs `data_type()` matching and `downcast_ref` for every (row, column) pair. It allocates a fresh `Vec<u8>` for each value, and numerics, dates and decimals allocate an intermediate `String` too, before copying into the row buffer. That means two allocations per cell on the only per-row path in the backend. Downcast each column once per batch into a small enum or closure, then write straight into `MysqlTextRow::buf`: use `write!`/`itoa` for numbers and `extend_from_slice` for strings. Also reserve the row buffer.

experimental/doris/src/result_encoder.rs
```diff
@@ -0,0 +1,302 @@
...
+            }
+            DataType::Date32 => {
+                let typed = Self::downcast::<Date32Array>(array)?;
+                let date = as_date::<Date32Type>(i64::from(typed.value(row)))
+                    .ok_or_else(|| format!("date32 value {} out of range", typed.value(row)))?;
+                Ok(Some(date.format("%Y-%m-%d").to_string().into_bytes()))
+            }
+            // Doris DATETIME is timezone-naive wall-clock; a tz-aware timestamp would need
+            // offset handling we don't implement, so let `Some(tz)` fall through to the error arm.
+            DataType::Timestamp(TimeUnit::Microsecond, None) => {
```
felipeblazing, line 105: **Reviewer: Correctness**

**major** DATETIME results with scale 0–3 can't be encoded; only microsecond timestamps are handled

`type_mapper` maps `DATETIMEV2(0)` (plain `DATETIME`) to `PrecisionTimestamp{0}` and scales 1–3 to precision 3. G-19 casts outputs to that type, so DuckDB returns Arrow `Timestamp(Second|Millisecond, None)`, which falls into the "not implemented yet" arm here. Any query that projects a DATETIME column translates, runs on the GPU, and then fails at encode time. Handle the Second and Millisecond units too, and format the fraction to the Doris scale rather than always `%.6f`.

felipeblazing, line 105: **Reviewer: Correctness**

**minor** Result encoder only handles microsecond timestamps, but the type gate emits second and millisecond ones

`type_mapper` maps `DATETIMEV2(0)` (plain Doris `DATETIME`) to `PrecisionTimestamp{0}` and scales 1–3 to `{3}`. DuckDB turns those into `TIMESTAMP_S`/`TIMESTAMP_MS`, which reach Arrow as `Timestamp(Second|Millisecond)`. So any query that outputs a `DATETIME` column, or a millisecond parquet timestamp, fails with "result encoding … not implemented" after it has already run on the GPU. Handle the other `TimeUnit`s here, and print the Doris scale's number of fractional digits instead of always 6.

experimental/doris/src/result_encoder.rs
```diff
@@ -0,0 +1,302 @@
...
+                self.buf.extend_from_slice(bytes);
+            }
+            None => self.buf.push(0xFB),
+        }
+    }
+
+    /// Writes a MySQL `length-encoded integer` prefix. `0xFB` is reserved for NULL, so a one-byte
+    /// length never reaches 251.
+    fn push_length(&mut self, len: usize) {
+        if len < 251 {
```
felipeblazing, line 156: **Reviewer: Naming, magic numbers & types**

**nit** MySQL length-encoding markers and thresholds are bare literals

`0xFB`, `0xFC`, `0xFD`, `0xFE`, `251`, `1 << 16` and `1 << 24` are MySQL protocol constants, and `0xFB` is repeated in the tests. `value % 1_000_000` on line 111 is microseconds per second. Name them (`MYSQL_NULL_MARKER: u8 = 0xFB`, `LENENC_2_BYTE_PREFIX`, …, `MICROS_PER_SECOND`), ideally in one module shared with `experimental/starrocks`, which has the same copy.

experimental/doris/src/result_encoder.rs
```diff
@@ -0,0 +1,302 @@
...
+            vec!["NULL", "1970-01-01", "1970-01-01 00:00:01.500000"]
+        );
+        assert_eq!(
+            cells(&result.rows[2]),
+            vec!["NULL", "1970-01-01", "1970-01-01 00:00:00.001500"]
+        );
+    }
+
+    #[test]
+    fn long_value_uses_two_byte_length_prefix() {
```
felipeblazing, line 276: **Reviewer: Test coverage & rigor**

**nit** Length-prefix boundary and most type arms of the encoder are untested

The test uses 300 bytes, not the 250/251 boundary where the one-byte length would collide with the `0xFB` NULL marker. The `0xFD`/`0xFE` branches are untested, and so are the Boolean, Int8/16/32, Float and Utf8View arms and the unsupported-type error. Add cases at 250 and 251 bytes and one row covering the remaining types.

experimental/doris/src/result_store.rs
```diff
@@ -0,0 +1,368 @@
...
+#[derive(Debug, Default)]
+pub(crate) struct ResultStore {
+    inner: Mutex<HashMap<UniqueId, Arc<ResultSlot>>>,
+}
+
+impl ResultStore {
+    /// Registers a slot for `query_id` (and the result fragment's `instance_ids`) before
+    /// execution starts, so a poll arriving early parks instead of failing as unknown.
+    /// Re-registering a query replaces its previous slot.
+    pub(crate) fn register(
```
felipeblazing, line 170: **Reviewer: Memory reservations & accounting**

**minor** A result nobody fetches stays in the ResultStore until cancel arrives, with no cap or TTL

A slot is only freed by `cancel_plan_fragment`. If the FE restarts, times out the RPC (`remote_fragment_exec_timeout_ms`) or never polls, the fully encoded result stays in the process-wide map indefinitely. Combined with the unbounded encoding above, a few abandoned large queries can hold gigabytes. Evict the slot once EOS or an error has been delivered, and add a TTL or a byte cap for undrained slots.

experimental/doris/src/result_store.rs
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    /// Looks up the slot a `fetch_data` poll refers to (query id or instance id).
+    pub(crate) fn get(&self, id: UniqueId) -> Option<Arc<ResultSlot>> {
+        self.lock().get(&id).cloned()
+    }
+
+    /// Drops every key pointing at the slot registered for `id` (query cancelled/finished).
+    /// A poll parked on the slot is failed so the FE stops waiting.
+    pub(crate) fn evict(&self, id: UniqueId) {
```
felipeblazing, line 191: **Reviewer: Memory reservations & accounting**

**minor** A result nobody fetches stays in memory for the life of the process

A result slot is removed only by `cancel_plan_fragment`. EOS does not free it, and nothing expires it. If the FE times out, restarts or loses the connection before it sends the cancel, the fully encoded result stays in the `ResultStore` for the life of the process. A few large abandoned queries are enough to exhaust the host headroom the configs reserve. Evict on the EOS packet as well (the FE's extra post-EOS poll can be answered from a small tombstone), or expire slots after a TTL.

experimental/doris/tests/corpus.rs
```diff
@@ -0,0 +1,647 @@
...
+        .chain(
+            captured_batches_in(&gaps_dir())
+                .into_iter()
+                .map(|batch| ("gaps", batch)),
+        );
+    for (corpus, (query, payload, _)) in all {
+        let batch = decode(&payload);
+        let actual = snapshot_text(corpus, &query, &batch, &translator);
+        assert_renders_cleanly(&actual, &query);
+        let path = snapshot_dir().join(format!("{query}.txt"));
```
felipeblazing, line 577: **Reviewer: Tests + CI**

**minor** Snapshot test misses orphaned snapshots and name clashes between the two corpora

The test only walks the captured batches, so a `tests/snapshots/*.txt` whose fixture was removed or renamed stays in the tree unchecked. Also, snapshots are keyed by bare `{query}.txt` across both the `tpch` and `gaps` corpora, so a gap probe named like a TPC-H query would silently share one file. Assert that the set of snapshot files equals the set of replayed queries.

felipeblazing (commented):
This is a good pass but I think we want to share as much as possible with starrocks. We can find a way to merge this in experimentally but I think we should get together and come up with how this will tie in with our efforts  to modularize Sirius called libSirius. We could then make rust bindings for libSirius and share as much as possible between doris and starrocks

morningman:
Review update for head `58e77f09`: rebased onto upstream `main` (both #1840 and #1841 are merged), fixed the join/stitcher fail-closed cases, result lifetime/packet limits, Thrift concurrency, timestamp encoding, CLI defaults and validator gaps, and moved the benchmark harness plus SF10 baselines out of this PR. The PR description now maps the changed areas and remaining single-plan/interrupt limits.

Verification on this head: Doris engine-linked Rust tests passed (57 library + 1 CLI + 10 corpus); the stock Doris FE ran TPC-H 22/22 and three NULL-bearing probes 3/3 on the GPU against DuckDB baselines; the CPU Substrait differential passed 22/22 plus 5/5 gap probes. The C++ suite's 3409 other cases passed; its 39 Iceberg cases needed the local DuckDB `avro`/`iceberg` extensions, and all 39 passed on rerun after provisioning them.

GitHub currently marks the fork PR workflows `action_required` pending repository-admin approval. I attempted the workflow approval API and received 403 (admin rights required), so the new-head CI remains external to my local verification.

PR #1843, opened 2026-09-19
refactor(dynamic-filter): make publication lifecycle explicit

kevkrist:
## Description

Refactors the existing dynamic-filter implementation without adding multi-batch build support.

- Moves filter publication out of the hash join into a session that handles completion, cancellation, and cleanup.
- Tracks which producers can still publish filters, so consumers can tell when no more filters will arrive.
- Gives each filtering pass a consistent snapshot for choosing filters, applying them, and measuring their usefulness.
- Keeps filter storage alive while GPU work uses it and adds tests for these lifetimes.

Build data is still delivered before filters are constructed and copied to other GPUs. There are no cuCascade changes. No performance comparison is included.

## Checklist
- [ ] Read CONTRIBUTING.md and ensure PR meets "reviewability" checklist

## References


src/op/sirius_physical_hash_join.cpp
```diff
@@ -2183,213 +2180,49 @@ std::unique_ptr<operator_data> sirius_physical_hash_join::execute(const operator
...
-        "only for a build-side sizing decision that lands in a single partition and finds a "
-        "build-side CONCAT to fold. Probe-driven sizing (right-family joins), a hash-partitioned "
-        "multi-partition build, and a missing build-side CONCAT each fail that. See this join's "
-        "partition strategy log line.",
-        get_operator_id(),
-        mode == HASH_JOIN_MODE::BUILD_PROBE  ? "BUILD_PROBE"
-        : mode == HASH_JOIN_MODE::MIXED_JOIN ? "MIXED_JOIN"
-                                             : "STANDARD");
-      if (_dynamic_filter_stats != nullptr) {
+        "[sirius_physical_hash_join] dynamic filter skipped (id={}): build is not one whole "
```
Jedi18, line 2204: `nit:` Please update `test_physical_hash_join_mgpu.cpp:791` too. It still checks for `"dynamic filter NOT published"`, so the assertion always passes after this rename.

src/op/scan/parquet_gpu_ingestible.cpp
```diff
@@ -1127,30 +1129,42 @@ filtered_table parquet_gpu_ingestible::materialize_metadata_to_table(
...
 
   rmm::device_async_resource_ref mr_ref(mem_space.get_default_allocator());
-  auto [table, _] =
-    cudf::io::read_parquet(std::move(sources), std::move(metadatas), opts, stream, mr_ref);
+  auto [table, _] = [&] {
+    try {
+      auto result =
+        cudf::io::read_parquet(std::move(sources), std::move(metadatas), opts, stream, mr_ref);
+      // The reader AST borrows scalars owned by this checkpoint's snapshot.
+      if (dynamic_reader_filter) { stream.sync(); }
```
Jedi18, line 1161: nit: Do we need this sync? The ingestible still owns the channel and its filters after the snapshot goes away

Jedi18 (approved):
LGTM, just two minor things

src/op/sirius_physical_hash_join.cpp
```diff
@@ -353,7 +353,7 @@ sirius_physical_hash_join::sirius_physical_hash_join(
     conditions(std::move(cond)),
     join_type(join_type),
     delim_types(std::move(delim_types)),
-    _dynamic_filter_plan(std::move(dynamic_filter_plan))
+    _dynamic_filter_session(std::move(dynamic_filter_plan), dynamic_filter_stats_sink)
```
joosthooz: `dynamic_filter_publication_session` is now built in this initializer list, before some checks run in the constructor body. But it already increments `stats->producers_enabled` and registered producers, so if one of those checks fails things run out of sync

src/op/scan/sirius_gpu_scan_operator_data.cpp
```diff
@@ -168,27 +163,27 @@ void scan_operator_input::prepare_for_processing(
       // drain-time one. The mapping invariant lives in
       // snapshot_membership_probes. A masked split takes probes only if its rep carries
       // the visibility mask too, so the two compose into one selection.
-      if (sirius::decompression_pushdown_enabled() && dynamic_filters &&
-          dynamic_filters->has_filters()) {
```
joosthooz: The `has_filters()` was a quick check to prevent having to take a lock and doing some work for every batch, are you sure you wish to remove it?

src/scan_manager/sirius_scan_manager.cpp
```diff
@@ -350,34 +350,14 @@ struct cached_databatch_provider : public databatch_provider {
...
-        // This drain runs on the metadata thread at query PREPARE, before any
-        // join build has published, so this snapshot is almost always EMPTY. It
-        // is kept as a free early base; the authoritative snapshot is taken at
-        // decode time by scan_operator_input::prepare_for_processing (same
-        // builder, same mapping invariant), which replaces this one.
+        // The drain usually precedes join publication. Decode-time refresh replaces this early
+        // snapshot; snapshot_membership_probes owns the shared output-to-slot mapping contract.
+        // Masked chunks compose these probes with the visibility mask attached above.
         if (sirius::decompression_pushdown_enabled() && _dynamic_filters) {
-          if (_dynamic_filters->has_filters()) {
```
joosthooz: Here also the quick check is removed

src/op/scan/parquet_gpu_ingestible.cpp
```diff
@@ -1127,30 +1129,42 @@ filtered_table parquet_gpu_ingestible::materialize_metadata_to_table(
     if (ast_expression) { reader_filter_root = &ast_expression->back(); }
   }
 
-  if (!split.disable_filter_pushdown && _sirius_dynamic_filters &&
-      _sirius_dynamic_filters->has_filters()) {
```
joosthooz: This also removes the quick check

src/op/scan/sirius_physical_dynamic_filter.cpp
```diff
@@ -66,11 +67,12 @@ std::unique_ptr<operator_data> sirius_physical_dynamic_filter::execute(
   output_batches.reserve(ro_batches.size());
 
   for (std::size_t i = 0; i < ro_batches.size(); ++i) {
+    if (i != 0) { snapshot = _filters->snapshot(); }
```
joosthooz: Do we really need to take a new snapshot instead of reusing the snapshot already taken above at line 58? Could it be a bit slow for many small batches?

src/op/dynamic_filter/sirius_dynamic_filter.cpp
```diff
@@ -466,95 +466,206 @@ cudf::ast::expression const& sirius_dynamic_zone_map_filter::to_ast(
...
-  std::scoped_lock lk(_mu);
-  auto it = _filters.find(col_idx);
-  if (it == _filters.end()) { return {}; }
+  std::scoped_lock lock(_state->mutex);
+  auto it = _state->filters.find(col_idx);
+  if (it == _state->filters.end()) { return {}; }
   return it->second;
 }
 
 std::vector<std::size_t> sirius_dynamic_filter_set::filtered_columns() const
```
joosthooz: This looks like it could be dead code now, please verify

src/op/dynamic_filter/sirius_dynamic_filter.hpp
```diff
@@ -334,24 +334,120 @@ class sirius_dynamic_bloom_filter final : public sirius_dynamic_filter,
...
+   *         have been added so far.
+   * @note The snapshot is valid even if new filters are added or the filter set is destroyed after
+   *       the snapshot is taken (the snapshot owns its entries).
+   */
+  [[nodiscard]] dynamic_filter_snapshot snapshot() const;
 
   /**
    * @brief Returns an insertion-order owning snapshot valid after later pushes or destruction
    */
   [[nodiscard]] std::vector<std::shared_ptr<sirius_dynamic_filter const>> filters_for_column(
```
joosthooz: This looks like it could be dead code now, please verify

src/op/dynamic_filter/dynamic_filter_publisher.hpp
```diff
@@ -17,15 +17,102 @@
...
 
 namespace sirius::op {
 
+class sirius_physical_hash_join;
+struct dynamic_filter_stats;
+
+/**
+ * @brief A single-batch whole-build delivery
+ */
+class complete_build_delivery final {
```
joosthooz: Is there a way to check if this is indeed the whole build side? Like throw when it is being created twice from the same place

test/cpp/operator/test_dynamic_filter_publisher.cpp
```diff
@@ -83,6 +83,38 @@ constexpr std::size_t kProbeColumnIndex = 7;
...
+    auto result = sirius::op::publish_dynamic_filters(plan, table, stream, producers);
+    stream.sync();
+    for (auto const& producer : producers) {
+      producer.finish(result.filters_pushed != 0
+                        ? sirius::op::sirius_dynamic_filter_set::completion::published
+                        : sirius::op::sirius_dynamic_filter_set::completion::skipped);
+    }
+    return result;
+  } catch (...) {
+    stream.sync();
```
joosthooz: We sync here, but also in `publish_dynamic_filters` if something in it throws, so it can probably be removed there

src/op/dynamic_filter/dynamic_filter_publisher.cpp
```diff
@@ -127,184 +385,201 @@ dynamic_filter_publication_outcome publish_dynamic_filters(dynamic_filter_publis
...
+      probe_targets.size(),
+      build_view.num_rows(),
+      outcome.keys_considered,
+      admitted_keys.size());
+    outcome.filters_pushed = total_pushed;
+    return outcome;
+  } catch (...) {
+    // Retire accepted work before the outer filter owners are destroyed during unwinding.
+    auto error = std::current_exception();
+    stream.sync();
```
joosthooz: Check if this sync can be eliminated because we are also syncing in the catch blocks of the callers.

joosthooz (approved):
Only some small comments, have a look if you want to address them before merging

src/op/scan/sirius_physical_dynamic_filter.cpp
kevkrist: Yes, because dynamic filter publication and dynamic filter execution can race, and it is not uncommon for new filters to be published after the first batch. The per-batch refresh is how this task participates in progressive, opportunistic attach. This PR does introduce an additional lock when creating the snapshot -- I will keep an eye on it but I have not observed any perf regression on, e.g., TPC-H. 

BUT -- I have added an additional check before taking a fresh snapshot that checks whether any more filters will be added to the channel -- this should save all unnecessary locking.

src/op/dynamic_filter/sirius_dynamic_filter.cpp
kevkrist: Yes, dead, removed and tests updated.

src/op/dynamic_filter/dynamic_filter_publisher.cpp
kevkrist: I don't think so -- the per_key_zone_map / per_key_membership are locals outside the try, so on an exception, if we don't sync, the destructors will free the memory here before returning to the caller, and without a sync there could be ongoing kernel work using these locals -> segfault. I will keep for now -- I think these caller/callee syncs are serving different ownership purposes.

test/cpp/operator/test_dynamic_filter_publisher.cpp
kevkrist: See below

src/op/sirius_physical_hash_join.cpp
kevkrist: The only thing that will be out of sync is the producers_enabled stats -- the destructro will cancel and finish live producers registered on the channel. I think this is OK for now.

src/op/dynamic_filter/dynamic_filter_publisher.hpp
kevkrist: This is private to hash join, and relies on the CONCAT logic setting _build_arrives_whole, so this is more an observation that that logic has been trigerred, and it doesn't seem like the job of this dynamic filter token to validate that that logic has been honored.

